PR #76, opened 2026-02-15
Register as .ipynb file handler via Tauri file associations

rgbkrk:
- Add fileAssociations to tauri.conf.json bundle config for .ipynb files
- Handle RunEvent::Opened to load notebooks opened via OS file association
- If current window is empty, load the file in-place and emit reload event
- If a notebook is already open, spawn a new process for the file
- Frontend listens for notebook:file-opened to reload cells


rgbkrk:
Closing in favor of #89.

PR #149, opened 2026-02-18
Fix crash when creating multiple notebooks rapidly (Cmd-N)

rgbkrk:
## Summary

Fixes the crash that occurs when rapidly creating multiple notebook windows using Cmd-N. Multiple windows are separate processes that share a prewarmed environment pool on disk, causing race conditions when they try to claim or use the same environments simultaneously.

## Changes

- **Path validation in take()**: Validates that environment paths still exist before returning them from the pool. If another process claimed/renamed an env, gracefully skip to the next one.
- **Warming locks**: Serialize environment creation across processes using atomic lock files (.warming.lock and .warming-conda.lock) to prevent resource contention.
- **Stale lock cleanup**: Automatically remove lock files older than 5 minutes to prevent stuck prewarming from crashed processes.
- **Shared pool model**: Multiple windows can have the same prewarmed environments in their pools. When one claims an env (renames it), others find it missing and skip it.

### DIFF
--- a/crates/notebook/src/conda_env.rs
+++ b/crates/notebook/src/conda_env.rs
@@ -22,7 +22,7 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::RwLock;
-use std::time::Instant;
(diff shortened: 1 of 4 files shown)


rgbkrk:
Closing in favor of #150

PR #163, opened 2026-02-20
Add notebook sync protocol types and fractional indexing

rgbkrk:
## Motivation

The `runtimed` daemon currently handles environment pool management. We're extending it with a notebook sync protocol — a persistent bidirectional connection where the daemon is the single source of truth for notebook state, broadcasting versioned deltas to connected clients (Tauri UI, agents, MCP servers, TUI).

This replaces LiveStore (used in intheloop/Santiago). The frustrations driving this change:
- LiveStore forces all clients to JS
- CRDTs add complexity we don't need (single daemon serializes all writes)
- Presence is bolted on
- No permission system to control who can emit outputs vs edit cells

The daemon serializes all writes — no CRDTs. Clients send edits, daemon validates (including permission checks based on `client_kind`), applies to state, increments version, broadcasts deltas.

## What's in this PR

This PR introduces the protocol types and supporting modules. No server logic yet — that comes next.

- **`notebook_protocol.rs`** — Full client/server message types with two-level JSON tagging (`type` + `op`). Covers edit operations, delta broadcasting, execution lifecycle, comm/widget relay, presence, and snapshot types. Includes a permission matrix on `ClientKind` controlling what each client type can do (e.g., only `kernel` and `agent` can emit outputs, only `agent` can manage runtime sessions).
- **`fractional_index.rs`** — Base-36 fractional indexing for cell ordering, allowing insertions between any two adjacent keys without reindexing. Returns `Result` for safe handling of invalid input.
- **`notebook_server.rs`** — Stub for per-notebook server instances with validated notebook IDs and platform-specific socket paths (`#[cfg(unix)]` / `#[cfg(windows)]`).

### Key design decisions

1. **`after: Option<cell_id>` instead of client-computed fractional index** — Clients say "put it after cell X" (or `null` for beginning). Daemon computes the fractional index server-side.
2. **Separate module from pool protocol** — Pool daemon is request-response NDJSON; notebook sync is persistent bidirectional streaming. Different enough for separate modules, same crate.
3. **`serde_json::Value` for output data, representations, and comm state** — Keeps protocol independent of `jupyter-protocol`/`nbformat` internals.
4. **`MediaRepresentation::Inline` / `::Blob`** — Small data inline, large outputs served via HTTP from local blob storage.

## Test plan

- [x] `cargo test -p runtimed --lib` — 140+ unit tests covering roundtrip serde for every message variant, wire format shape verification, snapshot serialization, permission matrix, and error cases
- [x] `cargo clippy --all-targets -- -D warnings` — clean

### DIFF
--- a/crates/runtimed/src/fractional_index.rs
+++ b/crates/runtimed/src/fractional_index.rs
@@ -4,19 +4,56 @@
 //! between any two adjacent keys without reindexing. This is the Rust
 //! equivalent of `generateKeyBetween` used in intheloop's schema.
 
+use std::fmt;
+
 const BASE: u32 = 36;
 const DIGITS: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyz";
 
-fn digit_value(c: u8) -> u32 {
+/// Errors that can occur during fractional index operations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FractionalIndexError {
+    /// A key contains a character that is not a valid base-36 digit.
+    InvalidCharacter(char),
+    /// `before` is not strictly less than `after`.
+    InvalidOrder { before: String, after: String },
+    /// No string can be generated between the given bounds.
+    NoSpace { before: String, after: String },
+}
+
+impl fmt::Display for FractionalIndexError {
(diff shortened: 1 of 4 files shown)


rgbkrk:
Closing in favor of automerge experimentation.

Issue #188, opened 2026-02-22
Unified settings read path: frontend should read settings.json when daemon is unavailable


rgbkrk:
## Context

PRs #185 and #199 added Automerge-based settings sync with nested `uv`/`conda` sub-maps and list-typed packages. The **write** path has a solid fallback: `set_synced_setting` always writes to `settings.json` first, then best-effort syncs to the daemon. But the **read** path has a gap.

## Current behavior

When the daemon is unavailable on startup:

| Layer | What happens |
|-------|-------------|
| **Frontend** (`useSyncedSettings`) | `get_synced_settings` invoke fails → catch silently → keeps hardcoded defaults (python, uv, `[]` packages) |
| **Backend** (`start_default_python_kernel_impl`) | Reads `settings.json` via `load_settings()` for `default_python_env` and `default_uv_packages`/`default_conda_packages` |

This means the settings panel can show "uv" with no packages while the backend creates a conda env with `["numpy", "pandas"]` because that's what `settings.json` says. The user's configured preferences are invisible in the UI until the daemon comes up.

## What needs to happen

1. **`get_synced_settings` should fall back to `settings.json`** — when the daemon is unreachable, the Tauri command should read `load_settings()` and return a `SyncedSettings` from it rather than returning an error. This way the frontend always shows the user's persisted preferences.

2. **Theme stays special** — theme still uses localStorage for FOUC prevention, so the read fallback for theme can remain `"system"` (or whatever localStorage has). The `settings.json` file doesn't store theme.

3. **Test the fallback read path** — unit test that `get_synced_settings` returns values from `settings.json` when the daemon socket doesn't exist.

## Current architecture reference

```
Write path (working well):
  UI setter → invoke("set_synced_setting")
    → save_setting_locally(settings.json)  [always]
    → SyncClient::put_value(daemon)        [best-effort]

Read path (the gap):
  UI mount → invoke("get_synced_settings")
    → try_get_synced_settings(daemon)      [fails if daemon down]
    → Err → frontend keeps hardcoded defaults  ← should read settings.json instead
```

The fix is small — in the `get_synced_settings` Tauri command, catch the daemon error and construct a `SyncedSettings` from `load_settings()`.

PR #209, opened 2026-02-22
Settings schema unification and file watching

### LINKED ISSUE #190: Watch settings.json for external changes and propagate via Automerge
## Context

The daemon writes a JSON mirror of the Automerge settings doc to `settings.json` (via `save_json_mirror()` in `persist_settings()`) after every sync round-trip. The Tauri app also writes to `settings.json` directly as a local fallback in `save_setting_locally()`. But the reverse direction isn't wired up — if a user (or another tool) edits `settings.json` directly, those changes aren't picked up by the daemon.

Currently, `settings.json` is only read at daemon startup in `SettingsDoc::load_or_create()` as a migration source when no `settings.automerge` file exists.

## Current flow

```
settings.json is written by:
  1. Daemon's persist_settings() after each sync message  → nested format
  2. Tauri's save_setting_locally() on every set_synced_setting call → flat format (AppSettings)

settings.json is read by:
  1. Daemon startup (one-time migration if no .automerge exists)
  2. Tauri's load_settings() for fallback env creation
  3. Tauri's load_settings() for Cmd+N menu default runtime

NOT read by:
  - Running daemon (never re-reads after startup)
```

## What needs to happen

1. **Add a file watcher** in the daemon for `settings.json` (using `notify` crate or similar)
2. **On detected change:** read the JSON, diff against current Automerge doc state, apply changed keys via the existing `put()`/`put_list()` methods
3. **Propagate** to all connected sync clients as usual (generate sync message, broadcast)
4. **Handle the feedback loop:** daemon writes JSON mirror → triggers watcher → detect it's a self-write and skip. Could use a timestamp/hash guard or a "suppress next notification" flag.

## Format consideration

The daemon writes nested JSON (`{ "uv": { "default_packages": [...] } }`) but the Tauri app writes flat JSON (`{ "default_uv_packages": [...] }`). The watcher should handle both formats — `SettingsDoc::from_json()` already does this with its backward-compat parsing.

## Why this matters

This makes `settings.json` a true human-editable interface while Automerge handles real-time distribution. Users can `vim ~/.config/runt-notebook/settings.json`, save, and see the change propagate to all open windows immediately. It's also useful for scripting and external tool integration.

rgbkrk:
## Summary

- Unified `AppSettings` to serialize in the same nested format as `SyncedSettings`, fixing a schema mismatch where the daemon and notebook wrote different JSON formats to the same file
- Added JSON Schema support via schemars crate for machine-readable settings documentation
- Implemented file watching for `settings.json` via notify crate with automatic change propagation to all connected notebook windows
- Added `cargo xtask install-daemon` for convenient daemon upgrades during development
- Improved `get_synced_settings` fallback to use local settings file when daemon is unavailable (issue #188)

## Verification

To test the file watching workflow:

1. Edit `~/Library/Application Support/runt-notebook/settings.json` manually while the daemon is running
2. Check that changes appear in all open notebook windows within 500ms
3. Verify backward compatibility by starting with an old flat-format `settings.json`
4. Test `cargo xtask install-daemon` to verify daemon reinstallation and version detection

Closes #190

### DIFF
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -1849,7 +1849,7 @@ async fn start_default_python_kernel_impl(
         // Include default uv packages from settings
         let default_deps: Vec<String> = {
             let s = settings::load_settings();
-            s.default_uv_packages
+            s.uv.default_packages
         };
         if !default_deps.is_empty() {
             info!(
@@ -2050,7 +2050,7 @@ async fn start_default_python_kernel_impl(
         let mut conda_deps_list = vec!["ipykernel".to_string()];
         {
             let s = settings::load_settings();
-            let extra = s.default_conda_packages;
+            let extra = s.conda.default_packages;
             if !extra.is_empty() {
                 info!("[prewarm:conda] Including default packages: {:?}", extra);
                 conda_deps_list.extend(extra);
@@ -2914,18 +2914,38 @@ async fn set_default_python_env(env_type: String) -> Result<(), String> {
 }
 
 /// Get synced settings from the Automerge settings document via runtimed.
-/// Returns an error if the daemon is unavailable — the frontend should
-/// keep its local state (localStorage) rather than overwriting with defaults.
+/// Falls back to reading settings.json when the daemon is unavailable,
+/// so the frontend always gets real settings instead of hardcoded defaults.
 #[tauri::command]
 async fn get_synced_settings() -> Result<runtimed::settings_doc::SyncedSettings, String> {
-    runtimed::sync_client::try_get_synced_settings()
-        .await
-        .map_err(|e| e.to_string())
+    match runtimed::sync_client::try_get_synced_settings().await {
+        Ok(settings) => Ok(settings),
+        Err(e) => {
+            log::warn!(
+                "[settings] Daemon unavailable ({}), falling back to settings.json",
+                e
+            );
+            let local = settings::load_settings();
+            Ok(runtimed::settings_doc::SyncedSettings {
+                theme: local.theme,
+                default_runtime: local.default_runtime.to_string(),
+                default_python_env: local.default_python_env.to_string(),
+                uv: local.uv,
+                conda: local.conda,
+            })
+        }
+    }
 }
 
 /// Persist a setting to local settings.json (for keys that have local representation).
 fn save_setting_locally(key: &str, value: &serde_json::Value) -> Result<(), String> {
     match key {
+        "theme" => {
+            let value_str = value.as_str().ok_or("expected string")?;
+            let mut s = settings::load_settings();
+            s.theme = value_str.to_string();
+            settings::save_settings(&s).map_err(|e| e.to_string())
+        }
         "default_runtime" => {
             let value_str = value.as_str().ok_or("expected string")?;
             let runtime: Runtime =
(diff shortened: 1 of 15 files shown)


crates/notebook/src/lib.rs
```diff
@@ -2948,16 +2968,15 @@ fn save_setting_locally(key: &str, value: &serde_json::Value) -> Result<(), Stri
...
             settings::save_settings(&s).map_err(|e| e.to_string())
         }
         "conda.default_packages" => {
             let packages = json_value_to_string_vec(value);
             let mut s = settings::load_settings();
-            s.default_conda_packages = packages;
+            s.conda.default_packages = packages;
             settings::save_settings(&s).map_err(|e| e.to_string())
         }
-        // Theme has no local fallback in settings.json (handled by localStorage)
```
rgbkrk, line 2960: Actually theme should have a local fallback in settings.json. We can do local storage first but there is a theme key.

crates/notebook/src/settings.rs
```diff
@@ -112,16 +94,72 @@ fn settings_path() -> PathBuf {
...
+    if settings.conda.default_packages.is_empty() {
+        if let Some(pkgs) = extract_flat_packages(&json, "default_conda_packages") {
+            settings.conda.default_packages = pkgs;
+        }
+    }
+
+    settings
+}
+
+/// Extract packages from an old flat key (comma-separated string or array).
```
rgbkrk: Considering we haven't released this yet with the flat key, it's going to be ok to rip this out before mondays pre release.

crates/runtimed/src/daemon.rs
```diff
@@ -365,6 +372,130 @@ impl Daemon {
...
+                _ = self.shutdown_notify.notified() => {
+                    if *self.shutdown.lock().await {
+                        info!("[settings-watch] Shutting down");
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
```
rgbkrk, line 503: The only funny thing I've noticed is that if I change the file in zed then it also reformats the file. Then the daemon gets the update and writes the file with the new setting back but formatted in its own json way. Just to cause less churn I feel like we should adopt json formatting here.

Separately I think we should use the json with comments format so we can leave comments in the document for the user.

rgbkrk, line 503: The only funny thing I've noticed is that if I change the file in zed then it also reformats the file. Then the daemon gets the update and writes the file with the new setting back but formatted in its own json way. Just to cause less churn I feel like we should adopt json formatting here.

Separately I think we should use the json with comments format so we can leave comments in the document for the user.

docs/settings.md
```diff
@@ -41,35 +43,50 @@ Environment-specific settings (packages, future: channels) live under `uv/` and
...
 ```
 
-For backward compatibility, the old comma-separated string format is also accepted when reading:
+### JSON Schema
+
+The settings structs derive `schemars::JsonSchema`. Both `SyncedSettings` (in runtimed) and `AppSettings` (in the notebook crate) serialize to the same JSON schema.
+
+### Backward Compatibility
+
+For backward compatibility, the old flat format is accepted when reading:
```
rgbkrk: Let's evaluate if we really need this.

PR #220, opened 2026-02-23
Add blob store, HTTP server, and unify daemon sockets

rgbkrk:
## Summary

- **Blob store** (`blob_store.rs`): Content-addressed on-disk storage using SHA-256, with two-level hex sharding, metadata sidecars, atomic writes, and concurrent-put safety.
- **Blob HTTP server** (`blob_server.rs`): Read-only hyper 1.x server on `127.0.0.1:0` serving `GET /blob/{hash}` with correct `Content-Type`, immutable cache headers, and CORS.
- **Socket consolidation**: Unifies the two separate daemon sockets (`runtimed.sock` for pool IPC, `runtimed-sync.sock` for settings sync) into a single multiplexed socket. Connections identify their channel via a JSON handshake as the first frame. The shared framing module (`connection.rs`) replaces duplicated length-prefixed code.
- **CLI tooling**: `runt pool info` (with liveness verification) and `runt pool flush` commands for daemon health checks.

This is the foundation for moving notebook outputs (images, HTML, rich data) out of the automerge CRDT document into content-addressed storage.

## What changed

The NDJSON line protocol between pool client and daemon is replaced with length-prefixed binary framing. The separate settings sync socket is removed — all IPC goes through the unified `runtimed.sock`. Both sockets were introduced earlier this week so there's no backward-compatibility concern.

The blob write path goes through the IPC socket (`BlobRequest::Store` + binary data frame), while the read path is HTTP (`GET /blob/{hash}`). The daemon writes `blob_port` into `daemon.json` so clients can discover it.

## Verification

- [x] `runt pool info` reports daemon version, PID, blob port, and uptime; shows "STALE" when daemon is down
- [x] `curl http://127.0.0.1:{blob_port}/health` returns `OK`
- [x] Store a blob via IPC, fetch via HTTP — correct bytes and `Content-Type`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5330,6 +5330,7 @@ name = "runt-cli"
 version = "1.3.0-dev.1"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap",
  "dirs 5.0.1",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5327,7 +5327,7 @@ dependencies = [
 
 [[package]]
 name = "runt-cli"
-version = "1.2.0"
+version = "1.3.0-dev.1"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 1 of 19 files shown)


rgbkrk:
@codex review

PR #254, opened 2026-02-24
Lazy load isolated renderer bundle to reduce initial bundle size

rgbkrk:
## Summary

Reduced main bundle size from 6.5 MB to 1.52 MB (77% smaller) by lazy loading the isolated renderer bundle. The ~5 MB renderer is now dynamically imported as a separate chunk instead of embedded in the main entry point.

The bundle is preloaded at app startup via `preloadRendererBundle()` so it's ready when cells render, maintaining the smooth UX of the existing `preloadIframe` optimization.

## Changes

- Convert `renderer-bundle.ts` to lazy loader with singleton caching
- Add `useRendererBundle()` hook for components to access the loaded bundle
- Update CodeCell and MarkdownCell to use the hook
- Preload bundle at App.tsx startup

## Verification

- [ ] Open notebook with code cells containing outputs
- [ ] Verify outputs render correctly (HTML, Markdown, images, widgets)
- [ ] Check Network tab: renderer chunk should load once, lazily
- [ ] Verify dark/light mode toggling still works
- [ ] Test Markdown cell with content on initial page load

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -2,6 +2,11 @@ import { invoke } from "@tauri-apps/api/core";
 import { listen } from "@tauri-apps/api/event";
 import { getCurrentWebview } from "@tauri-apps/api/webview";
 import { useCallback, useEffect, useRef, useState } from "react";
+import { preloadRendererBundle } from "./renderer-bundle";
+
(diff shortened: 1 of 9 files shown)


rgbkrk:
Coming back to this later.

PR #343, opened 2026-02-28
fix(xtask): build sidecar UI before runt-cli to enable build-dmg

rgbkrk:
## Summary

Fixes `cargo xtask build-dmg` failing with "UI assets not found" error.

The sidecar crate embeds UI assets at compile time via rust_embed. Its build.rs checks that `apps/sidecar/dist/index.html` exists before compiling. Previously, `build_runtimed_daemon()` would build runt-cli immediately after runtimed, but before pnpm had built the sidecar assets, causing the build to fail.

This adds `ensure_sidecar_ui()` which builds the sidecar UI if the dist folder doesn't exist, called from `build_runtimed_daemon()` before building runt-cli.

## Verification

- Run `rm -rf apps/sidecar/dist && cargo xtask build-dmg` to verify it now succeeds


rgbkrk:

PR #358, opened 2026-03-01
feat(notebook): run notebook windows in a single tauri process

### LINKED ISSUE #354: Each notebook is spawned into its own window/process
Right now, each notebook is spawned as a distinct process which results in some weird interactions that don't feel super native.

It seems like we can add a clearer window concept and map each instance of `NotebookState` to its associated window.

Nothing in the current implementation tells me we _need_ to set it up as one notebook-one process.

captainsafia:
Video isn't super interesting but all this is happening within multiple windows under the same Tauri process.

image

https://github.com/user-attachments/assets/30caeece-3d7e-466a-a65c-b93595ba33f3

Closes https://github.com/nteract/desktop/issues/354

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -408,20 +408,21 @@ function AppContent() {
 
   const handleExecuteCell = useCallback(
     async (cellId: string) => {
+      // Resolve source up front before awaiting sync operations.
+      const cell = cells.find((c) => c.id === cellId);
+      if (!cell || cell.cell_type !== "code") return;
+      const source = cell.source;
+
       // Clear outputs immediately so user sees feedback
       clearCellOutputs(cellId);
 
       // Broadcast clear to other windows
       await clearOutputs(cellId);
 
-      // Check if cell is code (early exit before daemon call)
-      const cell = cells.find((c) => c.id === cellId);
-      if (!cell || cell.cell_type !== "code") return;
-
-      // Start kernel via daemon if not running, then execute cell
-      // ExecuteCell reads source from the synced document (document-first execution)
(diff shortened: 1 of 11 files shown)


rgbkrk:
Nice! I'm assuming cmd-backtick flips between them now?

captainsafia:
> Nice! I'm assuming `cmd-`` flips between them now?

Yep!

apps/notebook/src/App.tsx
```diff
@@ -408,10 +408,9 @@ function AppContent() {
 
   const handleExecuteCell = useCallback(
     async (cellId: string) => {
-      // Resolve source up front before awaiting sync operations.
+      // Resolve cell up front before awaiting sync operations.
       const cell = cells.find((c) => c.id === cellId);
       if (!cell || cell.cell_type !== "code") return;
```
rgbkrk, line 413: Thank you. Good cleanup.

crates/notebook/src/deno_env.rs
```diff
@@ -88,9 +88,22 @@ struct RawDenoConfig {
     tasks: Option<serde_json::Value>,
 }
 
-/// Check if Deno is available (either on PATH or bootstrappable via rattler)
+/// Check if Deno is available on PATH or already cached.
+///
+/// This intentionally avoids triggering a full bootstrap during UI initialization,
+/// because daemon kernel launch handles on-demand bootstrap when needed.
 pub async fn check_deno_available() -> bool {
-    tools::get_deno_path().await.is_ok()
```
rgbkrk, line 93: I wonder if `get_deno_path()` is now dead code.

PR #404, opened 2026-03-01
fix(runtimed): clear execution queue on interrupt

### LINKED ISSUE #398: Interrupt does not stop later queued cells
If you have a long running cell that needs to be interrupted, like say `!vim` and you ran a cell after, if you interrupt that `!vim` cell it will run the other cell right after. The queue needs to be cleared.

rgbkrk:
When a user interrupts a running cell, queued cells that were pending execution are now properly cleared. Previously, these cells would still execute after the interrupt.

The fix adds queue clearing to the `interrupt()` method in kernel_manager, ensuring that all pending work stops when interrupt is called.

Closes #398

## Verification
- Queue multiple cells and run them while the first is executing
- Click interrupt on the running cell
- Verify that remaining queued cells do not execute

### DIFF
--- a/crates/runtimed/src/kernel_manager.rs
+++ b/crates/runtimed/src/kernel_manager.rs
@@ -1333,8 +1333,8 @@ impl RoomKernel {
         Ok(())
     }
 
-    /// Interrupt the currently executing cell.
-    pub async fn interrupt(&self) -> Result<()> {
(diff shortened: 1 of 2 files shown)


rgbkrk:
Manual QA verified. Yay!

PR #476, opened 2026-03-03
feat(daemon): use Tauri sidecar for daemon installation

rgbkrk:
## Summary

- Replace direct `ServiceManager` calls with Tauri's sidecar API to spawn `runtimed install`
- Add diagnostic logging for daemon startup flow
- Update `runtimed install` to also start the daemon after installing

## Why

The previous approach called `ServiceManager::install()` directly from Rust code, which:
1. Duplicated logic that already exists in `runtimed install`
2. Made it harder to debug (no CLI to reproduce the install flow)
3. Didn't use Tauri's sidecar API which handles binary spawning properly

## Changes

| File | Change |
|------|--------|
| `crates/notebook/Cargo.toml` | Add `tauri-plugin-shell` dependency |
| `crates/notebook/src/lib.rs` | Add `ensure_daemon_via_sidecar()`, register shell plugin |
| `crates/notebook/capabilities/default.json` | Add sidecar spawn permission |
| `crates/runtimed/src/main.rs` | `install_service()` now also starts daemon |

## How it works

1. On app startup, `ensure_daemon_via_sidecar()` pings the daemon
2. If not running (and not in dev mode), spawns `runtimed install` via sidecar
3. `runtimed install` self-installs and starts the daemon
4. App waits up to 10 seconds for daemon to become ready
5. Progress events emitted throughout for UI feedback

## Test plan

- [ ] Build with `cargo xtask build`
- [ ] Clean install test:
  ```bash
  rm -rf ~/Library/Application\ Support/nteract
  rm -rf ~/Library/Caches/runt
  launchctl bootout gui/$(id -u)/io.nteract.runtimed 2>/dev/null
  rm -f ~/Library/LaunchAgents/io.nteract.runtimed.plist
  ```
- [ ] Run app and verify logs show:
  - `[startup] Checking if daemon is running...`
  - `[startup] Daemon not responding, spawning runtimed install via sidecar...`
  - `[runtimed install] Installing runtimed service...`
  - `[runtimed install] Starting daemon...`
  - `[startup] Daemon ready at ...`
- [ ] Verify service installed: `launchctl list | grep nteract`

## Follow-up work

**Onboarding window architecture** - The current onboarding screen is implemented as a hidden `Untitled.ipynb` notebook window. This causes issues:
1. When onboarding completes, it reveals a broken notebook with `cwd` set to a macOS temp directory
2. Starting a kernel on this notebook uses the wrong working directory

The proper fix (separate PR):
- Detect first launch in the app startup flow
- Show a dedicated onboarding window (not a notebook)
- After onboarding completes, open a fresh new notebook with proper working directory
- The onboarding window should be a standalone UI, not a notebook in disguise


crates/runtimed/src/main.rs
```diff
@@ -324,11 +324,14 @@ fn install_service(binary: Option<PathBuf>) -> anyhow::Result<()> {
 
     manager.install(&source_binary)?;
 
+    // Start the daemon immediately (don't wait for login)
+    println!("Starting daemon...");
+    manager.start()?;
```
rgbkrk, line 329: Wait... should we be forcing the system service to kick off the service here instead?

PR #547, opened 2026-03-05
Phase 2: Add frontend Automerge document with local-first mutations

rgbkrk:
## Summary

Implements Phase 2 of the notebook architecture redesign, giving the frontend its own Automerge document replica. Cell mutations (add, delete, edit) now apply locally via `Automerge.change()` and sync to the daemon via binary relay, eliminating RPC round-trip latency. This is a feature-flagged implementation that coexists with the existing `useNotebook` hook.

## Key Changes

**Frontend Automerge Integration**
- Added `@automerge/automerge` WASM library with Vite plugin configuration (`vite-plugin-wasm`, `vite-plugin-top-level-await`)
- Created `automerge-schema.ts` defining TypeScript schema mirroring the Rust `NotebookDoc` structure
- Implemented `useAutomergeNotebook` hook (535 lines) with local-first cell mutations and sync state management
- Extracted shared materialization logic into `automerge-utils.ts` for reuse by both old and new hooks

**Binary Sync Relay Infrastructure**
- Added `get_automerge_doc_bytes` Tauri command to export the daemon's Automerge doc for frontend initialization
- Added `send_automerge_sync` Tauri command to receive raw sync messages from the frontend
- Extended `NotebookSyncClient` with `connect_split_with_raw_sync()` and `receive_and_relay_sync_message()` to handle frontend sync messages
- Spawned raw sync relay task in `initialize_notebook_sync` to forward incoming daemon sync messages to frontend via `automerge:from-daemon` event

**Feature Flag & Dispatch**
- Created `feature-flags.ts` with `USE_AUTOMERGE_FRONTEND` flag (togglable via `localStorage`)
- Added `useNotebookDispatch.ts` wrapper that selects between `useNotebook` and `useAutomergeNotebook` based on flag
- Updated `App.tsx` to use dispatch hook instead of direct `useNotebook` call

**Documentation**
- Added comprehensive `HANDOFF.md` detailing the Phase 2 architecture, implementation plan, and verification results

## Implementation Details

- **Local mutations are instant**: `updateCellSource`, `addCell`, `deleteCell` apply to local doc and React state immediately, then sync asynchronously
- **Dual output flow**: Outputs still stream via `daemon:broadcast` for real-time feedback; Automerge sync provides eventual consistency for cross-window state
- **Transitional architecture**: Tauri maintains its own Automerge replica during migration; frontend bootstraps from it via `get_automerge_doc_bytes`
[...]

PR #553, opened 2026-03-06
feat(notebook): add useAutomergeNotebook hook via runtimed-wasm

rgbkrk:
Phase 2 of the local-first Automerge migration. The frontend owns a local Automerge document via the `runtimed-wasm` `NotebookHandle`. All cell mutations (add, delete, edit source) execute instantly in WASM — no RPC round-trip. React state is derived from the doc. Sync messages flow through the Tauri relay to the daemon.

## What changed

- **`useAutomergeNotebook`** — WASM-backed hook with identical interface to `useNotebook`. Bootstrap from `get_automerge_doc_bytes`, mutations via `NotebookHandle`, sync via `automerge:from-daemon` events.
- **`useNotebookDispatch`** — Feature flag toggle. `VITE_USE_AUTOMERGE=true cargo xtask dev`, or `localStorage.USE_AUTOMERGE_FRONTEND`, or `?automerge=true`.
- **`materialize-cells`** — Extracted shared output resolution (blob manifests, stream merging) from `useNotebook` so both hooks use it.
- **`App.tsx`** — Imports from `useNotebookDispatch` instead of `useNotebook` directly.

## Key constraint

The JS frontend never creates Automerge objects directly. All doc mutations go through `NotebookHandle` methods which execute the same Rust code as the daemon — no string→Text CRDT type mismatch.

## QA

```
VITE_USE_AUTOMERGE=true cargo xtask dev
```

Tested: cell CRUD, source sync, execution with outputs, ipywidgets (IntSlider with bidirectional comms), save-as with room reconnect, restart & run all (4 cells). Feature flag off: existing `useNotebook` works unchanged.

## Known limitation

Initial load shows empty notebook briefly while WASM initializes and doc bytes load. Optimization deferred — the daemon status banner covers the gap in practice.

Depends on #547's relay infrastructure (cherry-picked to `540/relay-cherry-pick`).


rgbkrk:
It's working and it's _flying_.

**One minor thing ~~I~~ Opus noticed:** `addCell` does the WASM mutation first, then the optimistic React update. That's the right order — WASM is the source of truth. But `updateCellSource` does React first, then WASM. Both are fine since they're synchronous, but worth noting the inconsistency. Not blocking.

PR #566, opened 2026-03-06
refactor: move notebook metadata types to shared notebook-doc crate

rgbkrk:
Moves notebook metadata types (`NotebookMetadataSnapshot`, `RuntMetadata`, `KernelspecSnapshot`, etc.) from `runtimed` to the shared `notebook-doc` crate. All peers now share one definition.

## Why

Every peer (daemon, Tauri relay, WASM frontend, Python bindings) needs to read and write notebook metadata — kernelspec, dependencies, trust signatures. Previously these types lived in `runtimed`, forcing all consumers to depend on the full daemon crate. Now they're in `notebook-doc` alongside the Automerge doc they describe.

## What changed

- `git mv runtimed/src/notebook_metadata.rs` → `notebook-doc/src/metadata.rs` (history preserved)
- Re-export from `runtimed` as `notebook_metadata` — zero breakage for existing callers
- Add typed helpers on `NotebookDoc`:
  - `get_metadata_snapshot()` — deserialize `NotebookMetadataSnapshot` from doc
  - `set_metadata_snapshot()` — serialize and write back
  - `detect_runtime()` — kernelspec + language_info detection ("python", "deno", or None)

## What this enables

Any consumer with a `NotebookDoc` (or WASM `NotebookHandle`) can do:
```rust
let snapshot = doc.get_metadata_snapshot();
let runtime = doc.detect_runtime(); // Some("python")
```

Without manual JSON parsing or depending on `runtimed`. This is the foundation for the frontend reading metadata directly from its WASM doc instead of via Tauri commands.

## QA

- [ ] Existing notebook operations work (kernel launch, save, dependency management)
- [ ] Runtime detection produces correct results

31 notebook-doc tests, 225 runtimed tests, all passing.


rgbkrk:
## Follow-on work enabled by this PR

With metadata types in `notebook-doc`, two paths open up for further relay simplification:

### 1. Frontend reads metadata from its own WASM doc

The Tauri metadata commands (`get_notebook_dependencies`, `set_notebook_dependencies`, `get_notebook_runtime`, `verify_notebook_trust`, etc.) currently go through `handle.get_metadata()` — an async hop to the relay's sync task to read from its local Automerge doc. 

The frontend's WASM `NotebookHandle` has the same data. Once we expose `get_metadata_snapshot()` and `detect_runtime()` from the WASM bindings, these Tauri commands can be replaced by direct frontend reads. No IPC, no relay involvement.

**Scope:** Rebuild `runtimed-wasm` with the new `notebook-doc` metadata API, then migrate frontend hooks (`useDependencies`, `useCondaDependencies`, `useDenoDependencies`, `useTrust`) to read from WASM instead of `invoke()`.

### 2. Daemon-owned initial population

`initialize_notebook_sync` in `lib.rs` populates the relay's Automerge doc with cells from `NotebookState` via `handle.add_cell()` + `handle.update_source()`. This is the last path that writes cells through the handle.

Alternative: the daemon reads the ipynb file directly (it already does for saved notebooks via `notebook-watch`) and populates its own doc. The relay connects with an empty doc and syncs from the daemon. This eliminates the `NotebookState` → relay → daemon bootstrap path entirely.

**Scope:** Larger change — touches the connection handshake protocol and the untitled-notebook creation flow. Good candidate for a dedicated spike.

PR #685, opened 2026-03-10
feat(runtimed): create new notebook when opening non-existent path

rgbkrk:
## Summary

When `runt notebook newfile.ipynb` is called with a path that doesn't exist, the daemon now creates a new empty notebook at that path instead of failing. This matches the expected behavior of CLI tools like vim.

## Changes

- Check file existence before canonicalizing path
- For non-existent files: use absolute path for notebook_id, create empty notebook in doc
- For permission/IO errors: return error to client instead of silently failing

## Verification

* [x] Start dev daemon: `cargo xtask dev-daemon`
* [x] Run `./target/debug/runt notebook /tmp/test-new.ipynb`
* [x] Verify notebook opens with one empty cell
* [x] Save the notebook and verify `/tmp/test-new.ipynb` is created on disk
* [x] Close and reopen the notebook to verify saved content loads
* [x] Test error case: `./target/debug/runt notebook /root/nopermission.ipynb` returns error

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -953,6 +953,7 @@ impl Daemon {
                     initial_metadata,
                     false, // Send ProtocolCapabilities for legacy NotebookSync handshake
                     None,  // No streaming load for legacy handshake
+                    false, // Not a newly-created notebook at path
                 )
                 .await
             }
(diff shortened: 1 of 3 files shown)


rgbkrk:
## Addressed Copilot Review Comments

1. **Include path in error message** - Error messages now include the notebook path for better debugging: `Cannot access notebook '{}': {}`

2. **Handle canonicalize failures explicitly** - Canonicalize errors for existing files now return proper error responses instead of silently falling back

3. **Remove drain that could hang** - Removed the `tokio::io::copy` drain that could hang indefinitely if the client waits for server to close first

4. **Send error response on create_empty_notebook failure** - Now sends proper `NotebookConnectionInfo { error: ... }` response instead of returning early

5. **Fix auto-launch for new notebooks at path** - Added `created_new_at_path` parameter to `handle_notebook_sync_connection` so auto-launch works for notebooks created via `runt notebook newfile.ipynb`

PR #688, opened 2026-03-10
feat(markdown): resolve relative image paths via daemon and blob store

### LINKED ISSUE #683: feat(notebook): resolve relative image paths in markdown cells
## Summary

Markdown cells with relative image paths like `` don't render because the iframe sandbox uses blob: URLs with opaque origins.

## Security Context

The app uses iframe isolation for markdown/HTML rendering:
- Blob URLs create opaque origins (intentional security)
- `allow-same-origin` explicitly NOT included in sandbox
- CSP allows `img-src * data: blob:` but relative paths can't resolve

## Recommended Approach

Use the blob store (daemon-side) to create entries for relative images:
1. Create blob store entries for relative images, done in the daemon
2. Pump those into the rendered markdown somehow

## Security Considerations

- Path validation to prevent traversal (`../../../etc/passwd`)
- Restrict to notebook directory + configured asset paths
- All file access happens in trusted parent context, not iframe

rgbkrk:
Adds support for markdown cells with relative image paths like ``. The daemon resolves these paths against the notebook directory, stores resolved images in the blob store, and provides path → hash mappings to the frontend. Images are rendered by rewriting their paths to blob URLs before iframe display.

## Changes

**Backend (Rust):**
- `crates/notebook-doc`: Added `attachments` field to cells with getter/setter methods
- `crates/runtimed/src/markdown_images.rs` (new): Extracts relative image paths from markdown using pulldown-cmark
- `crates/runtimed/src/notebook_sync_server.rs`: `process_markdown_attachments()` validates paths (canonicalization, directory containment), stores in blob store, updates cell attachments
- `crates/runtimed-wasm`: Expose `attachments` in WASM bindings

**Frontend (TypeScript):**
- `apps/notebook/src/types.ts`: Added `attachments` to `MarkdownCell` interface
- `apps/notebook/src/lib/materialize-cells.ts`: Pass through attachments from sync
- `apps/notebook/src/components/MarkdownCell.tsx`: Apply attachments during render by rewriting paths to blob URLs

## Verification

* [ ] Create notebook with image directory (e.g., `assets/`)
* [ ] Add markdown cell with relative path: `test`
* [ ] Verify image renders in preview mode
* [ ] Verify path traversal is blocked (`../../../etc/passwd`)
* [ ] Verify attachments persist in notebook document
* [ ] Test on different machine (images resolve from blob store)

Closes #683

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1921,6 +1921,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "getopts"
+version = "0.2.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfe4fbac503b8d1f88e6676011885f34b7174f46e59956bba534ba83abded4df"
+dependencies = [
+ "unicode-width 0.2.2",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


rgbkrk:
Closing in favor of #717

PR #751, opened 2026-03-13
fix(runtimed-py): widen shell/iopub drain window from 50ms to 500ms

rgbkrk:
**Superseded by #755** — which reads outputs from the Automerge doc instead of the broadcast stream, eliminating the drain window entirely.

Keeping this as a reference. The drain window fix is correct but addresses a symptom. #755 addresses the root cause: `collect_outputs` shouldn't accumulate data from an ephemeral broadcast channel when the Automerge doc is the source of truth.

---

<details>
<summary>Original description</summary>

The Jupyter protocol has a well-known race condition where shell replies (`ExecutionDone`) can arrive before the last iopub message (error output). `collect_outputs` used a single 50ms timeout after receiving `ExecutionDone` to drain straggling messages. If the error output hadn't arrived yet, the result silently reported `status=ok, outputs=0`.

Replace the single-shot drain with a 500ms deadline: after `ExecutionDone`, keep polling at 50ms intervals until the deadline expires. Worst case adds 500ms to completed executions (typically much less since outputs arrive quickly). Applied to both `Session` and `AsyncSession`.

Found via CI probing in #746 — hit `test_error_traceback_captured` returning `ExecutionResult(status=ok, outputs=0)` on 1 of 4 pre-fix runs.
</details>

### DIFF
--- a/crates/runtimed-py/src/async_session.rs
+++ b/crates/runtimed-py/src/async_session.rs
@@ -2169,9 +2169,23 @@ async fn collect_outputs_async(
             .ok_or_else(|| to_py_err("Not connected"))?;
 
         // Before ExecutionDone: 100ms poll interval.
-        // After ExecutionDone: 50ms polls but keep going until the
-        // drain deadline (500ms after done) expires.
-        let timeout_ms = if drain_deadline.is_some() { 50 } else { 100 };
+        // After ExecutionDone: poll for the lesser of 50ms or time remaining
+        // until drain deadline, so we exit promptly when the deadline arrives
+        // even if messages keep flowing.
+        let timeout_ms = match drain_deadline {
+            Some(deadline) => {
+                let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
+                if remaining.is_zero() {
+                    log::debug!(
+                        "[async_session] Drain deadline reached (top of loop), finishing with {} outputs",
+                        outputs.len()
+                    );
+                    break;
+                }
+                remaining.as_millis().min(50) as u64
+            }
+            None => 100,
+        };
         let broadcast = tokio::time::timeout(
             std::time::Duration::from_millis(timeout_ms),
             broadcast_rx.recv(),
@@ -2240,11 +2254,12 @@ async fn collect_outputs_async(
                         }
                     }
                     NotebookBroadcast::KernelError { error } => {
+                        // KernelError is a hard failure (kernel/daemon died), not
+                        // the shell/iopub ordering race. Return immediately.
                         success = false;
                         outputs.push(Output::error("KernelError", &error, vec![]));
-                        drain_deadline = Some(
-                            tokio::time::Instant::now() + std::time::Duration::from_millis(500),
-                        );
+                        log::debug!("[async_session] KernelError received, returning immediately");
+                        break;
                     }
                     _ => {}
                 }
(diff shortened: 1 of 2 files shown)


crates/runtimed-py/src/async_session.rs
```diff
@@ -2221,16 +2227,24 @@ async fn collect_outputs_async(
...
-                            done_received = true;
                         }
                     }
                     NotebookBroadcast::KernelError { error } => {
                         success = false;
                         outputs.push(Output::error("KernelError", &error, vec![]));
-                        done_received = true;
+                        drain_deadline = Some(
+                            tokio::time::Instant::now() + std::time::Duration::from_millis(500),
+                        );
```
rgbkrk: Fixed — `KernelError` now breaks immediately. It's a hard failure, not the shell/iopub ordering race.

crates/runtimed-py/src/session.rs
```diff
@@ -1839,15 +1849,18 @@ impl Session {
...
-                        log::debug!(
-                            "[session] Drain timeout, finishing with {} outputs",
-                            outputs.len()
-                        );
-                        break;
+                    // Timeout - if drain deadline has passed, we're done
+                    if let Some(deadline) = drain_deadline {
+                        if tokio::time::Instant::now() >= deadline {
+                            log::debug!(
+                                "[session] Drain deadline reached, finishing with {} outputs",
```
rgbkrk, line 1871: Fixed — deadline is now checked at the top of each iteration, and the timeout is bounded to `min(50ms, remaining_time)`. If messages keep arriving, we still exit promptly at the deadline.

crates/runtimed-py/src/session.rs
```diff
@@ -1816,18 +1821,23 @@ impl Session {
...
-                                done_received = true;
                             }
                         }
                         NotebookBroadcast::KernelError { error } => {
                             success = false;
                             outputs.push(Output::error("KernelError", &error, vec![]));
-                            done_received = true;
+                            drain_deadline = Some(
+                                tokio::time::Instant::now() + std::time::Duration::from_millis(500),
+                            );
```
rgbkrk: Fixed — `KernelError` now breaks immediately.

crates/runtimed-py/src/session.rs
```diff
@@ -1747,8 +1750,10 @@ impl Session {
                 .as_mut()
```
rgbkrk, line 1750: Valid concern. The mutex is held during `recv()` which is at most 50-100ms per iteration (same pattern as before, just more iterations during drain). For the sync `Session`, `block_on` serializes all Python calls so contention is only with the sync task — which doesn't need this lock. For `AsyncSession` it's more relevant but the 500ms total window is bounded now. I'll note this for a follow-up refactor — extracting the `broadcast_rx` or using `resubscribe()` to decouple from the state lock.

crates/runtimed-py/src/async_session.rs
```diff
@@ -2239,13 +2253,18 @@ async fn collect_outputs_async(
...
-                    log::debug!(
-                        "[async_session] Drain timeout, finishing with {} outputs",
-                        outputs.len()
-                    );
-                    break;
+                // Timeout - if drain deadline has passed, we're done
+                if let Some(deadline) = drain_deadline {
+                    if tokio::time::Instant::now() >= deadline {
+                        log::debug!(
+                            "[async_session] Drain deadline reached, finishing with {} outputs",
```
rgbkrk, line 2275: Fixed — same approach as the sync version: deadline checked at loop top, timeout bounded to remaining time.

rgbkrk (commented):
Addressed all five review comments. Pushing the fixes now.

PR #755, opened 2026-03-13
fix(runtimed-py): read outputs from Automerge doc instead of broadcast stream

rgbkrk:
`collect_outputs` was accumulating cell outputs from the broadcast stream — a `tokio::broadcast` channel with capacity 64 that silently drops messages when the receiver lags. This caused:

- **Shell/iopub race**: error outputs arriving after `ExecutionDone` were missed (50ms drain too short)
- **Broadcast lag**: any dropped `Output` message meant silent data loss with no recovery
- **Complex drain logic** needed to catch straggling messages

The Automerge document is the source of truth. The daemon writes all outputs to the doc as they arrive from the kernel. By the time `ExecutionDone` fires, the doc has the complete, canonical output list.

**Now `collect_outputs` uses broadcasts purely as signals** (`ExecutionDone`, `KernelError`) and reads the final cell state from the doc snapshot via `handle.get_cells()` → `output_resolver::resolve_cell_outputs()`. This eliminates the drain window, the output accumulation logic, and the entire class of broadcast-ordering bugs.

This supersedes the drain window fix in #751 — if this lands, #751 can be closed.

### DIFF
--- a/crates/runtimed-py/src/async_session.rs
+++ b/crates/runtimed-py/src/async_session.rs
@@ -2157,18 +2157,22 @@ impl AsyncSession {
 // Helper functions (outside impl block for async use)
 // =========================================================================
 
-/// Collect outputs for a cell until ExecutionDone is received.
+/// Wait for execution to complete, then read outputs from the Automerge doc.
+///
+/// Uses the broadcast stream only as a signal for when execution is done.
+/// The Automerge document is the source of truth for cell outputs — this
+/// sidesteps the Jupyter shell/iopub race condition entirely since the
+/// daemon writes all outputs to the doc before signaling ExecutionDone.
 async fn collect_outputs_async(
     state: &Arc<Mutex<AsyncSessionState>>,
     cell_id: &str,
     blob_base_url: Option<String>,
     blob_store_path: Option<PathBuf>,
 ) -> PyResult<ExecutionResult> {
-    let mut outputs = Vec::new();
-    let mut execution_count = None;
-    let mut success = true;
-    let mut done_received = false;
+    let mut kernel_error: Option<String> = None;
 
+    // Phase 1: Wait for ExecutionDone or KernelError signal via broadcast.
+    // We don't accumulate outputs here — the Automerge doc is the source of truth.
     loop {
         let mut state_guard = state.lock().await;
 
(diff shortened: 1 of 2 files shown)


crates/runtimed-py/src/async_session.rs
```diff
@@ -2177,91 +2181,77 @@ async fn collect_outputs_async(
...
+        .as_ref()
+        .ok_or_else(|| to_py_err("Not connected"))?;
+
+    let cells = handle.get_cells();
+    let snapshot = cells.into_iter().find(|c| c.id == cell_id).ok_or_else(|| {
+        to_py_err(format!(
+            "Cell not found in doc after execution: {}",
+            cell_id
+        ))
+    })?;
```
rgbkrk: Fixed — added `handle.confirm_sync().await` after the break, before reading the doc.

crates/runtimed-py/src/async_session.rs
```diff
@@ -2177,91 +2181,77 @@ async fn collect_outputs_async(
...
+            outputs: vec![Output::error("KernelError", &error, vec![])],
+            success: false,
+            execution_count,
+        });
+    }
+
+    // Resolve outputs from doc snapshot (blob hashes → Output objects)
+    let outputs =
+        output_resolver::resolve_cell_outputs(&snapshot.outputs, &blob_base_url, &blob_store_path)
+            .await;
```
rgbkrk, line 2260: Fixed — snapshot is now captured while holding the lock, then lock is dropped before the `resolve_cell_outputs` await.

crates/runtimed-py/src/session.rs
```diff
@@ -1758,111 +1758,81 @@ impl Session {
...
 
+        // Phase 2: Read canonical cell state from the Automerge doc.
+        // The daemon writes outputs to the doc as they arrive from the kernel,
+        // so by the time ExecutionDone fires, all outputs are persisted.
+        let state = self.state.lock().await;
+        let handle = state
+            .handle
+            .as_ref()
+            .ok_or_else(|| to_py_err("Not connected"))?;
+
```
rgbkrk: Fixed — KernelError now returns immediately before any doc access. `execution_count` is `None` since we can't reliably read the doc if the kernel died.

crates/runtimed-py/src/session.rs
```diff
@@ -1758,111 +1758,81 @@ impl Session {
...
+        let cells = handle.get_cells();
+        let snapshot = cells.into_iter().find(|c| c.id == cell_id).ok_or_else(|| {
+            to_py_err(format!(
+                "Cell not found in doc after execution: {}",
+                cell_id
+            ))
+        })?;
```
rgbkrk: Fixed — added `handle.confirm_sync().await` after ExecutionDone, before reading `get_cells()`. This ensures the local Automerge replica has all output sync frames.

crates/runtimed-py/src/session.rs
```diff
@@ -1758,111 +1758,81 @@ impl Session {
...
+                execution_count,
+            });
+        }
+
+        // Resolve outputs from doc snapshot (blob hashes → Output objects)
+        let outputs = output_resolver::resolve_cell_outputs(
+            &snapshot.outputs,
+            &blob_base_url,
+            &blob_store_path,
+        )
```
rgbkrk, line 1840: Fixed — state lock is now scoped to capture the snapshot + config, then dropped before the `resolve_cell_outputs` await.

crates/runtimed-py/src/async_session.rs
```diff
@@ -2177,91 +2181,77 @@ async fn collect_outputs_async(
...
 
+    // Phase 2: Read canonical cell state from the Automerge doc.
+    // The daemon writes outputs to the doc as they arrive from the kernel,
+    // so by the time ExecutionDone fires, all outputs are persisted.
+    let state_guard = state.lock().await;
+    let handle = state_guard
+        .handle
+        .as_ref()
+        .ok_or_else(|| to_py_err("Not connected"))?;
+
```
rgbkrk: Fixed — KernelError returns immediately before any doc access.

rgbkrk (commented):
All 6 comments addressed. The CI failure (`status=ok, outputs=0`) confirmed Copilot's concern — `ExecutionDone` arrived before the final Automerge sync frame. Adding `confirm_sync()` after the signal and before reading the doc fixes this.

PR #762, opened 2026-03-13
fix(daemon): use channel-aware paths in xtask and recover daemon on retry

rgbkrk:
## Problem

The nightly daemon frequently dies mid-session, leaving users with "Runtime unavailable" errors. Investigation revealed two issues:

1. **xtask install-daemon bug**: Used hardcoded stable-only paths (`io.nteract.runtimed`, `runt`), so when nightly dev agents ran it, they accidentally stopped the stable system daemon instead of the nightly one
2. **App lacks recovery**: When the daemon dies mid-session, clicking Retry only attempted socket reconnection and couldn't recover if the daemon was dead

## Solution

**Part 1: Fix xtask to use channel-aware paths**
- Replace hardcoded `io.nteract.runtimed` with `daemon_launchd_label()` 
- Replace hardcoded `runtimed.service` with `daemon_service_basename()`
- Replace hardcoded `runt` cache dir with `cache_namespace()`

**Part 2: Recover daemon when user clicks Retry**
- Add `DaemonRestartInProgress` state to prevent multiple windows racing to restart
- Add `is_daemon_dead_error()` helper to detect when socket is missing
- When reconnect fails with daemon-dead error, call `ensure_daemon_via_sidecar()` to restart
- Poll daemon readiness before retrying connection
- In dev mode, show helpful guidance instead of attempting recovery

## Verification

* [ ] Verify `cargo xtask install-daemon` from nightly build targets `io.nteract.runtimed.nightly` service
* [ ] Kill daemon mid-session with `runt-nightly daemon stop`, click Retry, verify app recovers
* [ ] Kill daemon with multiple notebook windows open, click Retry in each, verify only one restart attempt occurs
* [ ] Verify dev mode shows guidance message instead of attempting recovery

### DIFF
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -86,6 +86,10 @@ impl WindowNotebookRegistry {
 /// Newtype wrapper for reconnect-in-progress flag (distinguishes from other AtomicBool states).
 struct ReconnectInProgress(Arc<AtomicBool>);
 
+/// Newtype wrapper for daemon-restart-in-progress flag.
+/// Prevents multiple windows from attempting to restart the daemon simultaneously.
+struct DaemonRestartInProgress(Arc<AtomicBool>);
+
 /// Tracks the last daemon progress status for UI queries.
 /// This allows the frontend to check status on mount (in case events were missed).
 struct DaemonStatusState(Arc<Mutex<Option<runtimed::client::DaemonProgress>>>);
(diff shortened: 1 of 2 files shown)


rgbkrk:
Addressed Copilot feedback in d3143275:

1. **Increased wait timeout**: Changed from 10s (20×500ms) to 30s (60×500ms) for the "another window is restarting" path
2. **Added ping readiness check**: Before retrying connection, poll `PoolClient::ping()` for up to 10s to verify daemon is actually ready

Re: the per-window reconnect concern - looking at the code flow, each window that calls `reconnect_to_daemon` will:
- Either perform the restart (if first)
- Or wait for restart to complete

Then **all windows** fall through to the retry section and attempt their own `initialize_notebook_sync_open/create` call. So each window does reconnect itself after the daemon is back. The ping check ensures we don't retry before the daemon is ready.

PR #802, opened 2026-03-14
fix(notebook-doc): address unresolved Copilot review findings from recent PRs

rgbkrk:
## Summary

Audit of 12 PRs merged in the last 12 hours found ~30 unaddressed Copilot review comments. Most were already resolved by subsequent refactors, but three substantive data-integrity concerns remain in the current codebase. This PR fixes all three.

### 1. Preserve output array length on unreadable elements (PR #791 feedback)

**File:** `crates/notebook-doc/src/lib.rs`

`read_cell` and `get_cells_from_doc` used `filter_map` when reading Automerge list elements for cell outputs. If `read_str` returned `None` for any index, that element was **silently dropped**, shrinking the output array and shifting all subsequent indices.

Changed to `.map(|i| read_str(...).unwrap_or_default())` so unreadable elements become empty strings, preserving the expected array length.

### 2. Propagate serialization errors in `merge_into_metadata_value` (PR #791 feedback)

**File:** `crates/notebook-doc/src/metadata.rs`

All three `serde_json::to_value()` calls for `kernelspec`, `language_info`, and `runt` used `if let Ok(v) = ...` with no else branch. On failure, the native key was silently not written — potentially leaving the on-disk `.ipynb` metadata incomplete.

Changed the return type to `Result&lt;(), serde_json::Error&gt;` and replaced silent swallowing with `?` propagation. Callers in `notebook_sync_server.rs` use `.ok()` (acceptable — metadata is pre-validated), tests use `.unwrap()`.

### 3. Preserve unknown `runt` keys through round-trips (PR #800 Copilot feedback)

**File:** `crates/notebook-doc/src/metadata.rs` + all `RuntMetadata` construction sites

Added `#[serde(flatten)] pub extra: HashMap&lt;String, serde_json::Value&gt;` to `RuntMetadata`. Without this, any `runt` keys not modeled in the Rust struct (e.g., from newer schema versions, third-party extensions, or future additions) would be silently stripped during deserialization → serialization round-trips.

Updated all `RuntMetadata` struct literals across `notebook-doc`, `runtimed`, and `notebook` crates.

## Audit context

| Severity | Finding | Source PR |
|---|---|---|
| 🔴 Data integrity | `filter_map` drops outputs, shifting indices | #791 |
| 🔴 Data integrity | Silent error swallowing in metadata serialization | #791 |
| 🔴 Data integrity | Unknown `runt` keys lost on round-trip | #800 |

## Testing

- **164 notebook-doc tests** pass (including metadata round-trip, merge, and serialization tests)
- **277 runtimed unit tests** + **22 integration tests** pass
- `cargo clippy -p notebook-doc -p runtimed -- -D warnings` clean

### DIFF
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -1634,7 +1634,7 @@ impl NotebookDoc {
             Some(list_id) => {
                 let len = self.doc.length(&list_id);
                 (0..len)
-                    .filter_map(|i| read_str(&self.doc, &list_id, i))
+                    .map(|i| read_str(&self.doc, &list_id, i).unwrap_or_default())
                     .collect()
             }
             None => vec![],
@@ -1985,7 +1985,7 @@ pub fn get_cells_from_doc(doc: &AutoCommit) -> Vec<CellSnapshot> {
                 Some((automerge::Value::Object(ObjType::List), list_id)) => {
                     let len = doc.length(&list_id);
                     (0..len)
-                        .filter_map(|j| read_str(doc, &list_id, j))
+                        .map(|j| read_str(doc, &list_id, j).unwrap_or_default())
                         .collect()
                 }
                 _ => vec![],
--- a/crates/notebook-doc/src/metadata.rs
+++ b/crates/notebook-doc/src/metadata.rs
@@ -53,6 +53,12 @@ pub struct RuntMetadata {
     /// Timestamp when the notebook was last trusted (RFC 3339 format).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub trust_timestamp: Option<String>,
+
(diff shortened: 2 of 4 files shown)


rgbkrk (commented):
Good audit. All three fixes address real data-integrity concerns.

The `#[serde(flatten)] extra` field on `RuntMetadata` is the right pattern for preserving unknown keys. The verbosity of `extra: HashMap::new()` at every construction site is the main cost — could be mitigated by a builder pattern or constructor function later, but not blocking.

One note: the `merge_into_metadata_value` return type change to `Result` is a signature change that could affect callers we haven't seen in this diff. The `.ok()` usage at the save sites is fine since metadata is pre-validated, but worth a grep for any other callers that might now need updating.

PR #805, opened 2026-03-14
fix(toolbar): move environment errors into tooltip

### LINKED ISSUE #578: bug: show full error in tooltip instead of inline in the menu
Noticed this while triggering an error state in a different PR. We should use a tooltip here instead of showing the detailed error inline in the toolbar.

Image

captainsafia:
## Summary
- keep the inline environment status concise when setup fails
- preserve the full error detail in the toolbar tooltip
- cover the concise error label with a Vitest assertion

Closes #578


captainsafia:
I think this turned out pretty nice! Only appears when hovering over environment error.

image

PR #806, opened 2026-03-14
fix(editor): route select all through codemirror

### LINKED ISSUE #771: bug(editor): Select All followed by a click leaves CodeMirror in a broken state
Reported during the March 13, 2026 nteract bug bash.

## Current behavior
After using `Select All` and then clicking back into a CodeMirror editor, the editor goes "weird" and appears to end up in a broken selection / cursor state.

## Expected behavior
A click after `Select All` should return the editor to a normal cursor / selection state without odd behavior.

## Notes
The original report was brief, so follow-up repro details may be needed during triage, but this is a user-visible editor defect worth tracking.

captainsafia:
## Summary
- replace the native Select All menu item with an app event
- select the active CodeMirror document directly instead of the whole page
- fall back to standard input selection behavior and cover the helper with tests

Closes #771

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -40,6 +40,7 @@ import { startCursorDispatch } from "./lib/cursor-registry";
 import { KERNEL_STATUS } from "./lib/kernel-status";
 import { logger } from "./lib/logger";
 import { useDetectRuntime } from "./lib/notebook-metadata";
+import { selectAllInActiveElement } from "./lib/select-all";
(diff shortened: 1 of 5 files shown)


rgbkrk:
This is what I meant for #771:

image

Basically, clicking inside codemirror when selecting on all doesn't clear the selection. I actually don't know if that's us or codemirror.

rgbkrk:
I'll close this since we couldn't reproduce it separately.

PR #820, opened 2026-03-14
feat(uv): add prerelease strategy support for inline notebook dependencies

rgbkrk:
## Summary

Add support for UV prerelease strategies in inline notebook dependencies. Users can now select from five prerelease strategies ("disallow", "allow", "if-necessary", "explicit", "if-necessary-or-explicit") via a new dropdown in the dependency panel.

The prerelease setting is stored in notebook metadata and passed to `uv pip install --prerelease <strategy>` when creating environments, enabling users to install prerelease versions of packages like `runtimed-py` without manually specifying exact versions.

## Changes

- **Metadata Schema**: Added `prerelease: Option<String>` to `UvInlineMetadata` struct
- **Environment Management**: Updated UV environment creation to pass `--prerelease` flag and include it in cache key hash
- **Daemon**: Extended environment preparation to accept and use prerelease strategy
- **WASM Bindings**: Added `set_uv_prerelease()` method for frontend communication
- **Frontend UI**: Added prerelease dropdown selector in dependency panel, integrated with hooks and metadata sync
- **Tests**: Added tests verifying hash differentiation by prerelease strategy

## Verification

* [ ] Open a notebook and select "Dependencies" (click icon in toolbar)
* [ ] Verify "Prereleases:" dropdown appears below Python version field
* [ ] Select "Allow all" and add a prerelease-only dependency (e.g., a prerelease version of nteract)
* [ ] Restart kernel and verify prerelease package installs without explicit version pinning
* [ ] Switch dropdown to different strategy and restart kernel to verify different environment is created


rgbkrk:
Addressed Copilot review feedback in bfab350a:

1. **Hash stability**: Now only adding a leading newline before `prerelease` when it's set, so existing environment hashes remain stable.

2. **Progress event packages**: Now building the packages list separately and passing it directly to the progress event, so only actual package names are reported (not `--prerelease` flags).

3. **Env drift tracking**: Valid point — for now this is acceptable since the primary use case is setting prerelease before starting the kernel. Adding prerelease to `LaunchedEnvConfig` for drift detection would be a nice follow-up.

PR #821, opened 2026-03-14
feat(notebook): show "Cell hidden" chip and group consecutive hidden cells

rgbkrk:
When both source and outputs are hidden on a code cell, instead of showing two separate chips (one for source, one for outputs), show a single **"Cell hidden"** chip. When multiple consecutive cells are fully hidden, they collapse into a single **"N cells hidden"** chip that expands all cells in the group on click.

This also keeps the left-alignment for chips when only source or only outputs are hidden.

## Verification

* [ ] Hide both source and outputs on a single cell — shows "Cell hidden" chip
* [ ] Click "Cell hidden" — both source and outputs are restored
* [ ] Hide both source and outputs on 2+ consecutive cells — first cell shows "N cells hidden", others are collapsed
* [ ] Click "N cells hidden" — all cells in the group are expanded
* [ ] Non-consecutive fully-hidden cells each show their own "Cell hidden" chip
* [ ] Hiding only source or only outputs still works as before (individual chips)

### DIFF
--- a/apps/notebook/src/components/CodeCell.tsx
+++ b/apps/notebook/src/components/CodeCell.tsx
@@ -375,7 +375,7 @@ export function CodeCell({
             {/* Source visibility toggle + Editor */}
             {bothHidden ? (
               /* Compact layout: both badges side by side when both hidden */
-              <div className="flex justify-end gap-2">
+              <div className="flex justify-start gap-2">
                 <button
                   type="button"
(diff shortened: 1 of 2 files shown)


rgbkrk:
Feedback from @captainsafia: 

> It's nice although I wish they were above and below each other instead of side by side so the heirarchy of inputs and outputs was still clear

PR #873, opened 2026-03-16
fix(trust): retry trust check on daemon:ready to fix session restore race

rgbkrk:
## Problem

On nightly session restore, untitled notebooks displayed console errors:
- `Failed to check trust: "Not connected to daemon"`
- `sync to relay failed: "Not connected to daemon"`

This prevented kernel auto-start, leaving the notebook in a broken state.

## Root Cause

`useTrust` was the only frontend hook without `daemon:ready` recovery. On startup, it called `checkTrust()` on mount before the daemon relay handle was stored, failed permanently, and blocked kernel initialization.

## Solution

- Add `daemon:ready` event listener that retries `checkTrust()` when the daemon connects (mirrors existing pattern in `useDaemonKernel` and `useAutomergeNotebook`)
- Downgrade "Not connected to daemon" error to debug level—this is an expected transient condition during startup

## Verification

- [x] `cargo test --verbose` passes (118 tests)
- Session restore should now proceed cleanly without console errors
- Kernel should auto-start after daemon connects

## Follow-up

A deeper fix is planned to defer secondary window sync until after daemon availability is confirmed (Rust-side change in `setup()`).

---

### DIFF
--- a/apps/notebook/src/hooks/useTrust.ts
+++ b/apps/notebook/src/hooks/useTrust.ts
@@ -1,4 +1,5 @@
 import { invoke } from "@tauri-apps/api/core";
+import { getCurrentWebview } from "@tauri-apps/api/webview";
 import { useCallback, useEffect, useState } from "react";
 import { logger } from "../lib/logger";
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
@copilot review

rgbkrk:
Superseded by #874 which includes this fix plus the backend change.

PR #903, opened 2026-03-17
fix(runtimed-py): make create_cell(index=None) append instead of prepend

rgbkrk:
## Summary

Fixed a footgun where `create_cell(index=None)` prepended cells at the top instead of appending at the bottom. API consumers expect `None` to mean "just add it at the end," but it was inserting before everything, producing reverse-chronological notebooks.

The underlying issue was in `crates/runtimed-py/src/session_core.rs`: when `index=None`, it was converted to `after_cell_id=None`, which in the Rust layer means "insert before everything."

Now `index=None` correctly maps to appending after the last cell.

## Verification

- [x] All lint checks pass
- [x] All tests pass

### DIFF
--- a/crates/runtimed-py/src/session_core.rs
+++ b/crates/runtimed-py/src/session_core.rs
@@ -472,15 +472,14 @@ pub(crate) async fn create_cell(
             .as_ref()
             .ok_or_else(|| to_py_err("Not connected"))?;
 
-        // Determine after_cell_id from index
-        let after_cell_id = index.and_then(|i| {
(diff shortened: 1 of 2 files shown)


rgbkrk:
**Note on optimization:** Append is the most common operation, and currently it clones the full cell list just to read the last cell's ID. A materialized structure mapping cell positions to fractional indices would make this O(1), but the path to that within Automerge's CRDT model needs more thought. Leaving this as a future optimization opportunity.

PR #911, opened 2026-03-17
fix(session): upgrade session save, migrate main geometry, correct scale factor

rgbkrk:
## Summary

Fix three session restore bugs:

1. **Upgrade session save**: Add a second session save at the top of `run_upgrade()`, right before any windows are closed. This captures the final registry state including any windows that opened between `begin_upgrade()` (which also saves) and the actual upgrade execution. The `write_session()` helper returns early on an empty registry, so the saved file is never overwritten by the `RunEvent::Exit` fallback.

2. **Stale "main" geometry migration**: Users upgrading from pre-97b0422f have window geometry orphaned under the `"main"` key in `.window-state.json` (now denylisted). A one-time migration renames the entry to the new `notebook-{hash}` label before the window-state plugin loads, so the first launch after upgrade preserves window position/size.

3. **Scale factor correction**: `tauri-plugin-window-state` stores physical pixels without scale factor context. A 1100×750 window on a 1x external monitor becomes 550×375 logical on a 2x Retina display. We now save `scale_factor` in the session state and apply `new_physical = old_physical × (current_scale / saved_scale)` after the plugin restores.

## Verification

* [ ] Open several notebooks across multiple monitors, trigger upgrade flow, verify all notebooks are restored after relaunch
* [ ] Open notebooks on an external monitor at different DPI, close app, relaunch on laptop — verify windows are correctly sized
* [ ] Verify normal Cmd+Q session restore still works

### DIFF
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -3690,8 +3690,18 @@ pub fn run(
     // Pre-97b0422f versions used a hardcoded "main" label for the first window.
     // The plugin now denylists "main", orphaning its saved geometry. This renames
     // the entry so the new deterministic label picks up the old geometry.
-    if let Some(ref session) = restored_session {
-        migrate_main_window_state(session);
+    //
+    // Runs unconditionally: the session file may exist even when restored_session
+    // is None (e.g., Finder launch, expired session, CLI open). Uses the
+    // age-ignoring loader so stale sessions still trigger the one-time migration.
+    {
+        let session_for_migration = restored_session
+            .as_ref()
+            .cloned()
+            .or_else(session::load_session_ignoring_age);
+        if let Some(ref session) = session_for_migration {
+            migrate_main_window_state(session);
+        }
     }
 
     let app = tauri::Builder::default()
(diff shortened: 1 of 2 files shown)


crates/notebook/src/lib.rs
```diff
@@ -4171,7 +4303,7 @@ pub fn run(
         // Use Exit (not ExitRequested) as it fires reliably on all platforms
         if let RunEvent::Exit = &event {
```
rgbkrk: Because `WindowEvent::Destroyed` removes each notebook from the registry before `RunEvent::Exit` runs ([lib.rs#L4276-L4300](https://github.com/nteract/desktop/blob/9937ace418e6c261cabacbd9c5df14e57ad2d6e5/crates/notebook/src/lib.rs#L4276-L4300)), this `save_session` call sees an empty registry on a normal quit. `write_session()` then returns early without writing anything ([session.rs#L129-L133](https://github.com/nteract/desktop/blob/9937ace418e6c261cabacbd9c5df14e57ad2d6e5/crates/notebook/src/session.rs#L129-L133)), so Cmd+Q/session restore is still broken unless the upgrade flow wrote the session earlier.

PR #913, opened 2026-03-17
fix(daemon): eliminate keepalive race and bind socket earlier during upgrade

rgbkrk:
## Summary

- **KeepAlive race fix**: `upgrade()` now uses `launchd_start()` (bootout+bootstrap) instead of `start()` which calls `ensure_loaded` — a no-op if `KeepAlive.Crashed` already auto-restarted the old binary during the stop→copy window
- **Early socket bind**: Unix socket is bound right after blob server starts, before `daemon.json` write, orphan reaping, env scanning, and warmup spawning — so `ping()` succeeds much sooner
- **Timeout increase**: Upgrade readiness timeout bumped from 10s to 20s as safety margin

## Test plan

- [ ] `cargo xtask lint` passes
- [ ] `cargo test -p runtimed` passes (22 tests)
- [ ] Test nightly upgrade: stop daemon, install new version, verify single-start and fast readiness in logs

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -458,6 +458,24 @@ impl Daemon {
             }
         };
 
+        // Bind the Unix socket early so clients can connect (and ping) while
+        // the rest of initialisation finishes.  The accept loop runs later.
+        #[cfg(unix)]
+        let unix_listener = {
+            let listener = UnixListener::bind(&self.config.socket_path)?;
+            // Restrict socket permissions to owner-only (0600) so other users
+            // cannot connect to the daemon.
+            {
+                use std::os::unix::fs::PermissionsExt;
+                std::fs::set_permissions(
+                    &self.config.socket_path,
+                    std::fs::Permissions::from_mode(0o600),
+                )?;
+            }
+            info!("[runtimed] Listening on {:?}", self.config.socket_path);
+            listener
+        };
+
         // Write daemon info so clients can discover us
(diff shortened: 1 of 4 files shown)


crates/runtimed/src/daemon.rs
```diff
@@ -458,6 +458,24 @@ impl Daemon {
             }
         };
 
+        // Bind the Unix socket early so clients can connect (and ping) while
```
rgbkrk, line 461: Binding the Unix socket before `run_unix_server()` starts accepting creates a worse failure mode for the upgrade check. On Unix, `connect()` succeeds as soon as the socket is listening, so the readiness probes in [`crates/notebook/src/lib.rs#L804-L816`](https://github.com/nteract/desktop/blob/901ad9b5ce92a35bb918f29ea7ec8cb77261241a/crates/notebook/src/lib.rs#L804-L816) and [`crates/runtimed/src/client.rs#L251-L324`](https://github.com/nteract/desktop/blob/901ad9b5ce92a35bb918f29ea7ec8cb77261241a/crates/runtimed/src/client.rs#L251-L324) can now connect, write the ping request, and then block forever waiting for a response because nothing is calling `accept()` yet. I reproduced this locally with a bound-but-unaccepted AF_UNIX socket: `connect()` and `send()` succeeded, `recv()` timed out. Before this change the probe failed fast and retried; after this change the first probe can hang until startup finishes, which defeats the new timeout logic and can stall app startup entirely.

PR #918, opened 2026-03-17
perf(notebook): incremental WASM materialization and perf benchmarks

rgbkrk:
## Summary

Two additions to the typing performance story:

**Incremental materialization via per-cell WASM accessors** — When incoming sync frames include source attributions (which tell us exactly which cells changed), uses the new per-cell WASM accessors (`get_cell_source`, `get_cell_type`, `get_cell_outputs`, etc.) to update only those cells instead of serializing the entire document via `get_cells_json()` → JSON.parse.

- Source edits with attributions → incremental (per-cell WASM reads + `updateCellById`)
- Structural changes (add/delete/move) → detected via `get_cell_ids()` comparison, falls back to full materialization
- Non-attributed changes (outputs, exec counts) → full materialization

This means when an agent edits cell 5 in a 100-cell notebook, we read 1 cell from WASM instead of serializing all 100.

**Vitest bench harness** for cell store and materialization hot paths, run in CI alongside unit tests.

### Key benchmark results (100-cell notebook)

| Operation | Throughput |
|-----------|-----------|
| `updateCellById` (typing hot path) | ~15M ops/sec |
| `updateNotebookCells` (old full-array path) | ~222K ops/sec |
| `getCellById` (Map lookup) | ~37M ops/sec |
| Single-cell materialization (warm cache) | ~13M ops/sec |
| Full 100-cell materialization (warm cache) | ~409K ops/sec |

`updateCellById` is **69x faster** than `updateNotebookCells` for a single source edit. Single-cell materialization is **31.5x faster** than full materialization.

## Verification

- [ ] Open a notebook with 20+ cells, type rapidly — should feel responsive
- [ ] While an agent streams output, type in a different cell — should remain smooth
- [ ] Execute a cell (Shift+Enter) after typing — should run latest code
- [ ] Add/delete/move cells — ordering stays correct
- [ ] `pnpm vitest bench apps/notebook/src/lib/__tests__/` runs and produces results
- [ ] No test regressions: `pnpm vitest run`

### DIFF
--- a/apps/notebook/src/lib/materialize-cells.ts
+++ b/apps/notebook/src/lib/materialize-cells.ts
@@ -1,4 +1,5 @@
 import type { JupyterOutput, NotebookCell } from "../types";
+import type { NotebookHandle } from "../wasm/runtimed-wasm/runtimed_wasm.js";
 import { logger } from "./logger";
 import type { OutputManifest } from "./manifest-resolution";
 import { isManifestHash, resolveManifest } from "./manifest-resolution";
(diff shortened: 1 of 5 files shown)


apps/notebook/src/lib/materialize-cells.ts
```diff
@@ -255,3 +256,71 @@ export async function cellSnapshotsToNotebookCells(
...
+      metadata,
+    };
+  }
+
+  if (cellType === "markdown") {
+    return {
+      id: cellId,
+      cell_type: "markdown",
+      source,
+      metadata,
```
rgbkrk, line 323: This incremental markdown path drops `resolvedAssets`, unlike the existing full materializers in [`materialize-cells.ts`](https://github.com/nteract/desktop/blob/f3d70e7298ae4617ef04940a64feeaf1d909284b/apps/notebook/src/lib/materialize-cells.ts#L177-L184) and [`materialize-cells.ts`](https://github.com/nteract/desktop/blob/f3d70e7298ae4617ef04940a64feeaf1d909284b/apps/notebook/src/lib/materialize-cells.ts#L240-L247). [`MarkdownCell.tsx`](https://github.com/nteract/desktop/blob/f3d70e7298ae4617ef04940a64feeaf1d909284b/apps/notebook/src/components/MarkdownCell.tsx#L226-L245) relies on `cell.resolvedAssets` to rewrite attachment and relative image refs, so any attributed markdown edit will replace the store entry with a cell that no longer has those mappings and embedded images stop resolving until the next full materialization.

PR #961, opened 2026-03-19
refactor(notebook): direct CodeMirror CRDT bridge to eliminate text corruption

rgbkrk:
## Direct CodeMirror ↔ CRDT Bridge

Replaces `@uiw/react-codemirror`'s controlled `value` prop pattern with a direct `ViewPlugin` that splices edits into Automerge at character granularity. **Eliminates the text corruption bug** — no more typing latch, no stale closure captures, no full-document replacement transactions.

### What changed

**Rust + WASM:**
- `splice_source(cell_id, index, delete_count, text)` on `NotebookDoc` and WASM `NotebookHandle` — calls Automerge `splice_text` directly, no Myers diff

**CRDT bridge ViewPlugin** (`crdt-editor-bridge.ts`):
- Outbound: CM `iterChanges` → `splice_source` per change (character-level)
- Inbound: `applyRemoteChanges()` dispatches to CM with `externalChangeAnnotation`
- `replaceSource()` for imperative full-source replacements (history search, MCP set_cell)
- Echo avoidance: annotation filtering + `isProcessingOutbound` flag

**Editor rewrite** (`codemirror-editor.tsx`):
- `EditorView` managed directly via refs + Compartments for dynamic reconfiguration
- `initialValue` prop (mount only) — not controlled `value`
- Custom keymaps placed before `baseExtensions` for correct precedence (Shift-Enter, etc.)
- IPython magic detection via dynamic `updateListener` (not static `useMemo`)
- `onValueChange` still available for non-CRDT consumers but no longer used by notebook cells

**React wiring:**
- `CrdtBridgeProvider` context + `useCrdtBridge(cellId)` hook
- Bridge extension added to CodeCell, MarkdownCell, RawCell
- Subscribes to frame bus for inbound text attributions per cell
- `onUpdateSource` / `onUpdateCellSource` prop chain removed entirely

**Cleanup:**
- `@uiw/react-codemirror` removed from both `package.json` files
- Daemon pool heartbeat logs downgraded from `info!` to `debug!`

### QA results

| Test | Result |
|------|--------|
| Bridge outbound (typing → CRDT) | ✅ splice_source, no Myers diff |
| Bridge inbound (MCP edits → CM) | ✅ text attributions reach editor |
| Concurrent editing (human + MCP agent) | ✅ CRDT merge, no corruption |
| Shift-Enter / Ctrl-Enter execute | ✅ keymap precedence fixed |
| History search (`replaceSource`) | ✅ Codex verified |
| IPython magic detection (%%bash, etc.) | ✅ dynamic updateListener |
| Output rendering (normal) | ✅ works |
| Text corruption (`import time` → `uim`) | ✅ **eliminated** |
| Output rendering (rapid execution) | ❌ pre-existing, not caused by this PR |

### Tests — 66 Deno WASM tests

Comprehensive `splice_source` test suite covering:
- Basic operations, boundary conditions, error handling
- Rapid typing simulation (character-by-character, backspace corrections)
- Sync between two handles, concurrent edits (splice vs splice, splice vs update_source)
- Concurrent edits at same position (conflict resolution)
- Interleaved splice + sync (real bridge pattern)
- Text attributions, unicode/CJK/emoji, newlines, CRLF
- Delete-all and retype, save/load persistence, changeset correctness
- Regression tests for the exact `import time\ntime.sleep(0)` corruption scenario

### Follow-up PRs

1. **Eliminate optimistic store mutations** — `clearCellOutputs` and `setExecutionCount` write directly to the cell store, bypassing the CRDT. This races with the materialization pipeline during rapid execution (outputs disappear). Move these into WASM so the cell store is a read-only projection of the CRDT.

2. **Vitest regression coverage** — history search recall via `bridge.replaceSource()`, IPython magic switching, bridge echo avoidance (flagged by Codex review as test gaps).

3. **Lockfile prune** — run `pnpm install` to remove `@uiw/react-codemirror` from `pnpm-lock.yaml`.


.zed/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "context_servers": {
     "supervisor": {
-      "command": "./target/debug/mcp-supervisor",
+      "command": "./target/release/mcp-supervisor",
```
rgbkrk: This looks like a local editor-environment change rather than part of the feature. Pointing Zed at `./target/release/mcp-supervisor` means anyone pulling the branch now needs a release build on disk, while our dev docs/examples use the debug binary. I think this should either stay on `./target/debug/mcp-supervisor` or be left out of the PR entirely.

rgbkrk:
One follow-up from review: the cleanup removed `@uiw/react-codemirror` from the checked-in manifests and from `pnpm-lock.yaml`, but `package-lock.json` still lists it as a direct dependency at the repo root. That leaves the npm lockfile out of sync with `package.json`, so I think we should regenerate/update `package-lock.json` as part of this PR too.

PR #964, opened 2026-03-19
fix(notebook): route clearCellOutputs and setExecutionCount through CRDT

### LINKED ISSUE #963: fix(notebook): eliminate optimistic store mutations that race with CRDT materialization
## Problem

`clearCellOutputs` and `setExecutionCount` write directly to the cell store, bypassing the CRDT. During rapid execution (Ctrl-Enter spam), these optimistic mutations race with the materialization pipeline:

1. `clearCellOutputs` writes `outputs: []` to the store
2. Daemon executes, writes output blobs to the CRDT
3. Sync frame arrives → materialization resolves outputs → writes to store
4. But another Ctrl-Enter fires `clearCellOutputs` again, overwriting the store with `[]`
5. After enough cycles, the pipeline gets stuck and outputs stop appearing

Discovered during QA of #961. The output drop is pre-existing (not caused by the CodeMirror refactor) but became more visible with the direct CRDT bridge.

## Approach

The cell store should be a **read-only projection of the CRDT**. All mutations go through the CRDT → sync → materialize → store pipeline.

### Option A: Remove optimistic clears, rely on daemon `clearOutputs`

The codebase already calls `clearOutputs(cellId)` (daemon RPC) alongside `clearCellOutputs(cellId)` (local store). The daemon RPC clears outputs in the CRDT and syncs back. Removing the local mutation eliminates the race. Tradeoff: slight visual delay before outputs disappear (~20-50ms on localhost).

### Option B: Add WASM-side `clear_outputs`

Add `clear_outputs(cellId)` to the WASM NotebookHandle that clears outputs in the local CRDT. The store is updated via the normal materialization path. The WASM change syncs to the daemon. This preserves instant feedback without the race.

Option B is cleaner but requires more work (Rust + WASM + rebuild). Option A is a quick fix to validate the architecture.

## Call sites

```
handleExecuteCell → clearCellOutputs + clearOutputs
handleRunAllCells → clearCellOutputs + clearOutputs  
restartAndRunAll → clearCellOutputs + clearOutputs
menu:clear-outputs → clearCellOutputs + clearOutputs
menu:clear-all-outputs → clearCellOutputs + clearOutputs
onClearOutputs (broadcast from other windows) → clearCellOutputs
```

## Follow-up from

- #961 (direct CodeMirror CRDT bridge)
- #958 (stabilization plan)

rgbkrk:
## Fix output drops during rapid execution

Follow-up to #961. Eliminates the race between optimistic store mutations and CRDT materialization that caused outputs to disappear when spamming Ctrl-Enter.

### Root cause

`clearCellOutputs` and `setExecutionCount` wrote directly to the cell store, bypassing the CRDT. During rapid execution:

1. `clearCellOutputs` → store: `outputs: []`
2. Materialization pipeline resolves outputs from CRDT → writes to store
3. Another `clearCellOutputs` overwrites store with `[]` again
4. After enough cycles, the pipeline gets stuck

### Fix

Both functions now go through the WASM CRDT handle first:

```
clearCellOutputs:
  1. handle.clear_outputs(cellId)      ← CRDT is source of truth
  2. handle.set_execution_count(cellId, "null")
  3. sourceSync$.next()                ← trigger daemon sync
  4. updateCellById(...)               ← instant visual feedback (safe — agrees with CRDT)

setExecutionCount:
  1. handle.set_execution_count(cellId, String(count))
  2. sourceSync$.next()
  3. updateCellById(...)               ← instant visual feedback
```

Since the CRDT and store now agree, materialization writes the same values — no more two-path conflict.

### Changes

- Expose `clear_outputs()` and `set_execution_count()` on WASM `NotebookHandle`
- Update `clearCellOutputs` and `setExecutionCount` in `useAutomergeNotebook` to write to CRDT first
- Rebuild WASM artifacts

Closes #963


apps/notebook/src/hooks/useAutomergeNotebook.ts
```diff
@@ -289,6 +289,24 @@ export function useAutomergeNotebook() {
...
+
+    // Clear outputs in the CRDT (not the store directly).
+    // The store is updated via the normal materialization pipeline
+    // when the CRDT change is processed, eliminating the race between
+    // optimistic store writes and materialization.
+    handle.clear_outputs(cellId);
+    handle.set_execution_count(cellId, "null");
+
+    // Trigger sync so the daemon sees the clear immediately.
+    sourceSync$.current.next();
```
rgbkrk: `useDaemonKernel` invokes these callbacks from daemon broadcasts (`execution_started` / `outputs_cleared`), but this change now mutates the local CRDT and schedules a sync back to the daemon in response. The daemon already writes `execution_count` / cleared outputs into the canonical doc *before* broadcasting those events, so every open window will now re-author the same state as `human:<session>`, mark the notebook dirty, and emit redundant sync traffic. I think we still want the fast store update here, but the CRDT mutation/sync should probably stay reserved for truly local UI actions.

PR #985, opened 2026-03-20
fix(notebook): guard Run All against concurrent invocations

### LINKED ISSUE #982: bug: rapid Run All clicks flood relay, causing 30s request timeouts and stuck outputs
Rapidly clicking "Run All" floods the relay with sequential requests. Each click fires `clear_outputs_via_daemon` for every cell plus `run_all_cells_via_daemon`. The relay processes requests one at a time, so 5 rapid clicks with 11 cells = 55 ClearOutputs + 5 RunAllCells queued sequentially.

The daemon's sync loop can't keep up — it's generating sync frames for each mutation while also trying to read the next request from the socket. Eventually `ClearOutputs` times out at 30s, and the connection is effectively stuck. Outputs stop materializing until reload.

```
[relay] Request timed out after 30s: ClearOutputs { cell_id: "983f3602-..." }
```

Reproduces on `main` — not related to #977 (RuntimeStateDoc).

**Fix options:**
- Debounce Run All on the frontend (ignore clicks within Xms of last)
- Coalesce ClearOutputs into a single bulk request
- Cancel in-flight Run All when a new one is requested

rgbkrk:
## Summary

Rapidly clicking "Run All" flooded the relay with duplicate `ClearOutputs` and `RunAllCells` requests (e.g. 5 clicks × 11 cells = 55 ClearOutputs + 5 RunAllCells queued sequentially), causing 30s request timeouts and stuck outputs.

Two changes:

1. **In-flight guard** — a shared `useRef` prevents concurrent `handleRunAllCells` / `restartAndRunAll` invocations. The second click is silently skipped while the first is in progress.

2. **Remove redundant daemon ClearOutputs** — `clearOutputsLocal` already writes to the WASM CRDT which syncs to the daemon via Automerge, and the kernel clears terminal state on `execute_input`. The per-cell `ClearOutputs` relay requests were redundant and the primary source of flooding.

Together these reduce a rapid 5-click Run All from ~60 relay requests to exactly 1 (`RunAllCells`).

Closes #982

See #988 for a follow-up to add a bulk `ClearAllOutputs` request for the menu action.

## Verification

- [ ] Open a notebook with several code cells
- [ ] Rapidly click "Run All" multiple times — only one execution cycle should occur
- [ ] Rapidly click "Restart & Run All" — same single-execution behavior
- [ ] Click "Run All" then immediately "Restart & Run All" — the second should be blocked
- [ ] Confirm outputs render correctly after Run All completes
- [ ] Use "Clear All Outputs" from the menu — should still work (uses separate code path)

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -169,6 +169,9 @@ function AppContent() {
   // Track pending kernel start that was blocked by trust dialog
   const pendingKernelStartRef = useRef(false);
 
+  // Guard against concurrent Run All / Restart & Run All operations (#982)
+  const runAllInFlightRef = useRef(false);
+
   // Notebook runtime type — reactive read from WASM Automerge doc.
   // Re-renders automatically when metadata changes (bootstrap, sync, writes).
   const detectedRuntime = useDetectRuntime();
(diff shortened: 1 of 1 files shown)


rgbkrk (commented):
Clean fix. The shared `useRef` guard is the right approach — simpler than debouncing and correctly handles both Run All and Restart & Run All sharing the same lock.

Two small observations:

1. The `finally` block correctly resets the guard on both success and error. The early `return` when `codeCells.length === 0` is inside the try block, so it does hit finally — guard gets properly released. ✅

2. The `Promise.all(codeCells.map(clearOutputs))` pattern fires N parallel ClearOutputs requests. Even with the guard preventing duplicate Run All clicks, a single click on a 50-cell notebook still sends 50 parallel requests. A follow-up could batch these into a single `ClearAllOutputs` daemon request — but that's separate from this fix.

Looks good to merge.

rgbkrk:
The in-flight guard is a good safety net, but the root cause is the N×`ClearOutputs` request pattern. Each cell gets its own request/response round-trip through the relay before execution starts. On a 50-cell notebook that's 50 sequential round-trips.

The daemon's IOPub `ExecuteInput` handler already clears outputs when execution truly starts — the frontend `clearOutputs` calls via daemon are redundant cross-window sync. The frontend only needs `clearOutputsLocal` for instant visual feedback.

A better fix would be to either:
1. Drop the daemon `clearOutputs` calls entirely from Run All (the daemon clears authoritatively on `ExecuteInput`)
2. Add a bulk `ClearAllOutputs` daemon request that clears in a single Automerge transaction

Either way the guard is still useful for the shutdown+relaunch race in `restartAndRunAll`.

PR #986, opened 2026-03-20
chore: track runtimed-wasm binary with git lfs

### LINKED ISSUE #975: chore: track WASM binary with Git LFS
## Problem

`runtimed_wasm_bg.wasm` (1.5MB) has been committed 57 times without Git LFS. That's ~85MB of git history from a single binary blob that changes on every WASM rebuild.

```
$ du -h apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm
1.5M

$ git log --oneline -- apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm | wc -l
57
```

No `.gitattributes` exists in the repo.

## Fix

1. Install Git LFS: `git lfs install`
2. Track the WASM binary: `git lfs track "*.wasm"`
3. Add `.gitattributes` to the repo
4. Migrate existing history with `git lfs migrate import` (optional — rewrites history)

The migration step is destructive (force-push required). May be worth doing on a quiet day or alongside a major version boundary.

## Also consider tracking

- Any other binary artifacts that change frequently
- The WASM `.js` glue file is text and fine in regular git

### LINKED ISSUE #978: chore: track runtimed-wasm binary with Git LFS
`apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm` is a 1.5MB binary committed directly to git. Every WASM rebuild adds another 1.5MB blob to the pack — ~10 commits have touched it so far.

Should add a `.gitattributes` entry:
```
apps/notebook/src/wasm/runtimed-wasm/*.wasm filter=lfs diff=lfs merge=lfs -text
```

Retroactive migration (rewriting history to move existing blobs into LFS) is optional — the main win is stopping future growth. CI will need `git lfs install` if it doesn't already have it.

rgbkrk:
## Summary

The `runtimed_wasm_bg.wasm` binary (1.5MB) has been committed ~57 times without Git LFS, adding ~85MB of bloat to git history. This PR adds LFS tracking to stop future growth and includes a developer-facing check for clones without LFS.

- Add `.gitattributes` to track `*.wasm` via Git LFS
- Add `ensure_wasm_resolved()` in xtask that checks the WASM magic bytes before frontend builds — if a dev clones without LFS, they get a clear error with fix instructions instead of a cryptic Vite failure
- Add `lfs: true` to CI checkout steps that run `pnpm build`

Closes #978
Closes #975

## Verification

- [ ] `git lfs ls-files` lists the `.wasm` file after checkout
- [ ] Clone the repo without LFS installed, run `cargo xtask build` — should show the `ensure_wasm_resolved` error with install/fix instructions
- [ ] CI build passes (confirms `lfs: true` on the right checkout steps)


rgbkrk (commented):
Good approach. The `ensure_wasm_resolved()` check in xtask is a nice touch — developers who clone without LFS get a clear error with install instructions instead of a cryptic Vite failure.

A few things to verify:

1. **CI workflows covered**: `build.yml`, `pr-binary-generation.yml`, and `release-common.yml` all get `lfs: true`. The PR check workflow (`ci.yml` or similar) that runs lint/test should also have it if it runs `pnpm build`. Check that the "Build shared UI artifacts" job has `lfs: true` — that's the one that failed earlier on #977.

2. **The `.wasm` file in this PR**: The diff shows the file was modified — this is the LFS migration converting the blob to a pointer. After merge, `git lfs ls-files` should list it. Anyone who already has the repo cloned will need `git lfs pull` once.

3. **`cargo xtask wasm` output**: When wasm-pack rebuilds, it writes the real binary to the output dir. The `ensure_wasm_resolved` check runs before `pnpm build`, not before `cargo xtask wasm` — so developers who build WASM themselves won't hit the pointer issue. Good.

4. **The glob pattern** `apps/notebook/src/wasm/runtimed-wasm/*.wasm` is specific enough. If we add more WASM crates later, we'd extend the pattern, but this is fine for now.

Solid PR. The xtask guard is the best part — LFS issues are always confusing without a good error message.

PR #987, opened 2026-03-20
fix(runtimed): resolve duplicate rooms after saving untitled notebook

### LINKED ISSUE #942: bug(daemon): duplicate rooms for same notebook path after save
Observed during stress testing: saving an untitled notebook to `notebooks/letsgo.ipynb` resulted in two separate rooms for the same file path — one with the UI peer, one orphaned. A Python client connecting by path joined the wrong room and edits were invisible in the UI.

`runt notebooks` showed:
```
│ ~/…/notebooks/letsgo.ipynb │ python │ uv:prewarmed │ idle │ 1 │
│ ~/…/notebooks/letsgo.ipynb │ python │ uv:prewarmed │ idle │ 0 │
```

The room re-keying path (`rekey_ephemeral_room`) works for the `create_notebook` → `save_notebook` flow (Python sessions). But opening an existing `.ipynb` file may create a room keyed by path while a stale room from a previous session still exists, or the re-key races with new connections.

rgbkrk:
## Summary

- When `rekey_ephemeral_room` found a different room already at the canonical path (created by a racing `get_or_create_room` call), it silently returned `None` — leaving the UUID-keyed room orphaned in the rooms map. This produced two rooms for the same file path, causing edits from Python clients to be invisible in the UI.
- Now `rekey_ephemeral_room` evicts the interloper room (which is empty/just-created) and proceeds with the normal rekey, so only one room exists for a given path.
- Adds an `Arc::ptr_eq` guard to the eviction task to prevent a stale eviction timer from accidentally removing a rekeyed room that now lives at the same key.

Closes #942

## Verification

- [ ] Create an untitled notebook, add some cells/content, then save it to a file path. Run `runt notebooks` and confirm only one room exists for that path.
- [ ] Open the saved notebook from a Python MCP client — edits should be visible in the UI and vice versa.
- [ ] Stress test: rapidly save an untitled notebook while simultaneously opening the same path from another connection. Confirm no duplicate rooms appear.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -2430,7 +2430,12 @@ async fn rekey_ephemeral_room(
                 if let Some(interloper) = interloper {
                     tokio::spawn(async move {
                         if let Some(mut kernel) = interloper.kernel.lock().await.take() {
-                            let _ = kernel.shutdown().await;
+                            if let Err(e) = kernel.shutdown().await {
+                                warn!(
+                                    "[notebook-sync] Failed to shut down interloper kernel: {}",
+                                    e
+                                );
+                            }
                         }
                         if let Some(tx) = interloper.watcher_shutdown_tx.lock().await.take() {
                             let _ = tx.send(());
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -1098,11 +1098,23 @@ where
                     );
                 }
 
-                rooms_guard.remove(&notebook_id_for_eviction);
-                info!(
-                    "[notebook-sync] Evicted room {} (idle timeout)",
-                    notebook_id_for_eviction
-                );
+                // Only remove if the room in the map is still the one we're evicting.
+                // A rekey may have replaced the room at this key with a different one.
+                if rooms_guard
+                    .get(&notebook_id_for_eviction)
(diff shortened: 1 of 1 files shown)


rgbkrk (commented):
This fixes a real race condition — the `Arc::ptr_eq` guard on eviction is the key safety addition. Without it, a stale eviction timer could remove a rekeyed room that happens to be at the same key.

The interloper eviction logic is sound: the ephemeral room has the user's content (cells, outputs, kernel), while the interloper is an empty/just-created room from a racing `get_or_create_room`. Removing the interloper and proceeding with rekey is the right call.

One concern on the interloper cleanup:

The `tokio::spawn` for interloper shutdown runs detached. If the interloper had a kernel (unlikely but possible if a second client connected and triggered auto-launch during the race window), the shutdown runs asynchronously. The interloper's `watcher_shutdown_tx` and kernel shutdown are fire-and-forget — if they fail, the resources leak silently. This is acceptable for a rare race condition, but worth a `warn!` if the kernel shutdown fails:

```rust
if let Err(e) = kernel.shutdown().await {
    warn!("[notebook-sync] Failed to shut down interloper kernel: {}", e);
}
```

Also worth noting: the interloper's `persist_tx` and `changed_tx` channels will be dropped when the Arc is dropped, which is correct — the debounced persistence task will exit when the watch channel closes.

The `Arc::ptr_eq` guard on eviction is the more important fix here and is clean. Good PR.

PR #1028, opened 2026-03-21
fix(runtimed): handle directory paths in --attach without autosave spin loop

### LINKED ISSUE #1027: fix(runtimed): `--attach <directory>` causes autosave spin loop instead of opening untitled notebook
`cargo xtask notebook --attach ~/notebooks/` passes a directory path to the daemon as a notebook ID. The daemon tries to open/autosave it as a file, hits `Is a directory (os error 21)`, and retries every second indefinitely.

Expected: if the path is a directory, create an untitled notebook with that directory as the working directory (for project file detection, kernel CWD, etc.).

Current behavior: infinite warn loop in the daemon log:
```
[autosave] Failed to save: Failed to write notebook: Is a directory (os error 21)
```

The daemon should check `path.is_dir()` early in the open/create path and handle it as a working directory hint rather than a notebook file.

rgbkrk:
## Summary

- When `--attach <directory>` is used, the daemon now detects the directory early in `handle_open_notebook`, verifies it's writable, and delegates to `handle_create_notebook` with the directory as `working_dir` — creating an untitled notebook instead of entering a spin loop
- Introduces a typed `SaveError` enum (`Retryable` / `Unrecoverable`) for `save_notebook_to_disk` so the autosave loop can distinguish permanent failures (`IsADirectory`, `PermissionDenied`) from transient ones and break instead of retrying forever

Closes #1027

## Verification

- [ ] `cargo xtask notebook --attach /tmp/some-dir/` opens an untitled notebook (no spin loop in daemon logs)
- [ ] `--attach /path/to/existing.ipynb` still opens the existing notebook
- [ ] `--attach /path/to/new.ipynb` still creates a new notebook at that path
- [ ] `--attach /tmp/readonly-dir/` (no write perms) returns an error to the client
- [ ] Daemon logs show no "Is a directory" warnings

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -1114,11 +1114,12 @@ impl Daemon {
                     Ok(p) => p.to_string_lossy().to_string(),
                     Err(_) => path_buf.to_string_lossy().to_string(),
                 };
-                // Verify directory is writable
-                let test_file = path_buf.join(".runtimed_write_test");
-                match tokio::fs::write(&test_file, b"").await {
+                // Verify directory is writable using a uniquely-named temp file.
+                // create_new uses O_EXCL so it can never clobber an existing file.
+                let probe = path_buf.join(format!(".runtimed_probe_{}", uuid::Uuid::new_v4()));
+                match std::fs::File::create_new(&probe) {
                     Ok(_) => {
-                        let _ = tokio::fs::remove_file(&test_file).await;
+                        let _ = std::fs::remove_file(&probe);
                     }
                     Err(e) => {
                         let (_reader, mut writer) = tokio::io::split(stream);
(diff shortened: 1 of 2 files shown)


crates/runtimed/src/daemon.rs
```diff
@@ -1104,6 +1104,42 @@ impl Daemon {
...
+                info!(
+                    "[runtimed] Path {} is a directory, creating untitled notebook with working_dir",
+                    path
+                );
+                let dir_path = match path_buf.canonicalize() {
+                    Ok(p) => p.to_string_lossy().to_string(),
+                    Err(_) => path_buf.to_string_lossy().to_string(),
+                };
+                // Verify directory is writable
+                let test_file = path_buf.join(".runtimed_write_test");
```
rgbkrk: This writability probe is destructive. `tokio::fs::write` will truncate an existing `".runtimed_write_test"` file, and the next line removes it unconditionally, so attaching a directory can silently delete real user data if that filename already exists. Please switch this to a uniquely named temp file created with `create_new` semantics instead. Ref: https://github.com/nteract/desktop/blob/c88dc92bff8c199e58fe309ee30cd728db6a53ed/crates/runtimed/src/daemon.rs#L1118-L1121

PR #1033, opened 2026-03-21
fix(session): hydrate kernel state from RuntimeStateDoc after connect

### LINKED ISSUE #1018: Session doesn't reflect auto-launched kernel state — kernel_started/kernel_type/env_source are stale
## Problem

After `create_notebook()` or `open_notebook()`, the Session's local properties (`kernel_started`, `kernel_type`, `env_source`) are all `None`/`False` — even though the daemon auto-launched a kernel. This causes:

1. Tests and agents think no kernel is running and call `start_kernel()` unnecessarily
2. `start_kernel()` gets `KernelAlreadyRunning` and silently returns the wrong kernel type
3. `execute_cell()` has a hidden `ensure_kernel_started` guard that papers over the issue, but `queue_cell()` and `stream_execute()` don't

## Root Cause

`session_core.rs` hardcodes the initial state:
```rust
SessionState {
    kernel_started: false,
    kernel_type: None,
    env_source: None,
    // ...
}
```

But the daemon auto-launches a kernel in a background `tokio::spawn` for every trusted notebook on first-peer connect. The Session never learns about it.

## Fix

Hydrate Session state from the RuntimeStateDoc after connect. The RuntimeStateDoc (added in #977) is an Automerge document that the daemon keeps updated with kernel status — it's the source of truth:

```rust
fn hydrate_kernel_state(handle: &DocHandle) -> (bool, Option<String>, Option<String>) {
    match handle.get_runtime_state() {
        Ok(rs) => {
            let running = matches!(rs.kernel.status.as_str(), "idle" | "busy" | "starting");
            let ktype = if running { Some(rs.kernel.language) } else { None };
            let env = if running { Some(rs.kernel.env_source) } else { None };
            (running, ktype, env)
        }
        Err(_) => (false, None, None),
    }
}
```

Apply this at the end of `connect_create`, `connect_open`, and `connect_with_socket` in `session_core.rs`.

## Additional fixes

- **`start_kernel()` should error on type mismatch**: If you request `kernel_type="deno"` but Python is already running, return an error instead of silently succeeding with `KernelAlreadyRunning`
- **`restart_kernel()` hardcodes "python"**: Should preserve the current kernel type or accept an override parameter
- **`queue_cell()` and `stream_execute()` should have the same `ensure_kernel_started` guard** as `execute_cell()`

## Context

Discovered during integration test stress testing. The `KernelAlreadyRunning` silent success caused Deno, Conda, and project detection tests to all fail — they thought they had the right kernel but actually had the auto-launched Python one.

rgbkrk:
## Summary

After `connect_open()`, `connect_create()`, or `connect_with_socket()`, `Session.kernel_started` / `kernel_type` / `env_source` were always `false`/`None` — even when the daemon had already auto-launched a kernel. This caused callers to redundantly call `start_kernel()`, which silently succeeded via `KernelAlreadyRunning` (potentially with the wrong kernel type).

This PR adds a `hydrate_kernel_state()` helper that reads the RuntimeStateDoc (the daemon's source of truth) at connect time and populates Session state accordingly. It also fixes three related issues:

- **`start_kernel()` now errors on type mismatch** — requesting `"typescript"` when `"python"` is already running returns an error instead of silently succeeding
- **`restart_kernel()` preserves current kernel type** — no longer hardcodes `"python"` / `"auto"`
- **`queue_cell()` gains `ensure_kernel_started` guard** — matching the existing behavior in `execute_cell()` and `stream_execute()`

Closes #1018

## Verification

- [ ] `open_notebook()` a trusted notebook → `session.kernel_started` is `True` and `session.get_runtime_state().kernel.language` matches `session.kernel_type`
- [ ] `start_kernel("typescript")` when Python is already running → raises error with mismatch message
- [ ] `restart_kernel()` on a non-Python kernel → relaunches same kernel type
- [ ] `queue_cell()` on a fresh session with no kernel → auto-starts kernel (same as `execute_cell()`)

### DIFF
--- a/crates/runtimed-py/src/async_session.rs
+++ b/crates/runtimed-py/src/async_session.rs
@@ -1097,9 +1097,10 @@ impl AsyncSession {
     /// Queue a cell for execution without waiting for the result.
     fn queue_cell<'py>(&self, py: Python<'py>, cell_id: &str) -> PyResult<Bound<'py, PyAny>> {
         let state = Arc::clone(&self.state);
+        let notebook_id = self.notebook_id.clone();
         let cell_id = cell_id.to_string();
         future_into_py(py, async move {
-            session_core::queue_cell(&state, &cell_id).await
+            session_core::queue_cell(&state, &notebook_id, &cell_id).await
         })
     }
 
--- a/crates/runtimed-py/src/session.rs
+++ b/crates/runtimed-py/src/session.rs
@@ -884,8 +884,11 @@ impl Session {
 
     /// Queue a cell for execution without waiting for the result.
     fn queue_cell(&self, cell_id: &str) -> PyResult<()> {
-        self.runtime
-            .block_on(session_core::queue_cell(&self.state, cell_id))
+        self.runtime.block_on(session_core::queue_cell(
+            &self.state,
+            &self.notebook_id,
+            cell_id,
+        ))
     }
 
     /// Stream execution events for a cell as an iterator.
--- a/crates/runtimed-py/src/session_core.rs
+++ b/crates/runtimed-py/src/session_core.rs
@@ -170,7 +170,29 @@ pub(crate) async fn connect_with_socket(
     st.blob_base_url = blob_base_url;
     st.blob_store_path = blob_store_path;
 
+    hydrate_kernel_state(&mut st);
+
     Ok(())
+}
+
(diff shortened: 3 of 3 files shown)


rgbkrk:
Folding this into #1030

PR #1049, opened 2026-03-21
fix(e2e): poll for kernel status after trust approval and wait for button enabled

rgbkrk:
Fixes the Trust Dialog E2E runner. Two bugs found by running the test locally and watching the app:

### 1. Approve button disabled during loading race

The `daemon:ready` event triggers `checkTrust()` which briefly sets `loading=true`, disabling the dialog buttons. If WebDriver tries to click during that window, `waitForClickable` times out after 5s. Fix: added `waitForEnabled({ timeout: 10000 })` before `waitForClickable`.

### 2. Kernel status read too early after approval

`handleTrustApprove` calls `launchKernel("auto", "auto")` as fire-and-forget (no `await`). The kernel status propagates via RuntimeStateDoc sync, so it's still `not_started` when read immediately. Fix: replaced the single-read `expect` with a 10s `waitUntil` poll for `starting`/`idle`/`busy`.

Local test result after fix:
```
Trust dialog appeared
Kernel status before trust approval: not_started
Clicked approve button  
Dialog dismissed in 307ms
Kernel status after dialog closed: idle
✓ should close trust dialog on single click without waiting for kernel (1.5s)
```

### DIFF
--- a/e2e/specs/trust-dialog-dismiss.spec.js
+++ b/e2e/specs/trust-dialog-dismiss.spec.js
@@ -68,7 +68,10 @@ describe("Trust Dialog Dismiss", () => {
     const declineButton = await $('[data-testid="trust-decline-button"]');
     expect(await declineButton.isExisting()).toBe(true);
 
-    // Click the approve button
+    // Wait for the button to be enabled — a checkTrust() call from the
+    // daemon:ready listener can briefly set loading=true, which disables
+    // the buttons. Poll until the disabled attribute clears.
+    await approveButton.waitForEnabled({ timeout: 10000 });
     await approveButton.waitForClickable({ timeout: 5000 });
 
     const clickTime = Date.now();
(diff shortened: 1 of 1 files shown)


rgbkrk:
## CI Results

**5/8 E2E runners pass:**

| Runner | Status |
|--------|--------|
| Smoke | ✅ |
| UI (Cell Visibility) | ✅ |
| UV Prewarmed | ✅ |
| Deno | ✅ |
| UV Pyproject | ✅ |
| Trust Dialog | ❌ trust dialog never appears |
| UV Inline Deps | ❌ trust dialog never appears |
| Conda Inline Deps | ❌ trust dialog never appears |

**Root cause of 3 failures:** The trust dialog doesn't appear on CI even though the fixture notebook (`2-uv-inline.ipynb`) has inline deps and no `trust_signature`. The daemon's trust verification may not be reading `runt.uv.dependencies` from the notebook metadata correctly, or the trust check on Linux resolves differently.

This is a daemon-side trust detection issue, not a test issue. The trust dialog fix in this PR (polling for kernel status, waiting for button enabled) is correct — it just can't be exercised because the dialog never shows.

Recommend: merge this PR for the trust dialog fix, investigate the trust detection separately.

Issue #1048, opened 2026-03-21
feat(runtimed): Execution handle abstraction — decouple "submit work" from "get results"


rgbkrk:
## Problem

The current Python API has three execution methods on `CellHandle`:

- `cell.run()` — execute and wait for collected results
- `cell.queue()` — fire and forget
- `cell.stream()` — execute and yield events as they come

All three conflate "submit work to the runtime" with "how you consume the results." The frontend model is cleaner: queue a cell for execution, then watch the CRDT for output changes as they stream in. The execution queue is a daemon concept, but there's no first-class representation of it in the Python API.

## Proposed API

```python
# Submit work — returns immediately
execution = await cell.execute()

# Check status (sync read from local state)
execution.status      # "pending" | "running" | "done" | "error"
execution.cell_id     # the cell being executed

# Wait for completion
result = await execution.result()          # blocks until done
result = await execution.result(timeout=30) # with timeout

# Stream events (alternative to waiting)
async for event in execution:
    print(event.event_type, event.output)

# Cancel
await execution.cancel()
```

The cell's outputs update independently through the CRDT regardless of whether anyone holds the `Execution` handle. This matches how the frontend works — it queues execution and observes document changes.

### `cell.run()` becomes sugar

```python
# Equivalent to:
execution = await cell.execute()
result = await execution.result(timeout=timeout_secs)
```

### `cell.queue()` becomes sugar

```python
# Equivalent to:
execution = await cell.execute()
# (discard the handle)
```

## Architecture: Execution IDs + CRDT State (no intents)

The spike in #1052 proved that threading a UUID `execution_id` through the stack works end-to-end. For 2.1, we keep the existing `ExecuteCell` request/response path (no CRDT-based intents) and focus on making execution a first-class concept in the CRDT documents.

### Why no intents

An earlier design proposed writing execution intents into the notebook doc. After review, this was dropped because:
- Intents are mutable CRDT data but semantically immutable once acknowledged — Automerge doesn't enforce this
- A client modifying an already-processed intent creates inconsistency the daemon must silently ignore
- The existing `ExecuteCell` request/response is clean and ephemeral — no lingering mutable state
- Attribution and CRDT-based cancel can be revisited later if needed

### Document ownership

| Concern | Document | Writer(s) |
|---|---|---|
| Cell source, metadata | Notebook doc | Any client |
| Cell `execution_id` pointer | Notebook doc | Daemon (on acknowledgment) |
[...]

PR #1052, opened 2026-03-22
feat(runtimed): execution handle abstraction spike

rgbkrk:
Spike for #1048. Threads a UUID-based `execution_id` through the entire execution pipeline and introduces the `Execution` handle class.

## Phase 1: Execution IDs

**Protocol** - `CellQueued` response, `ExecutionStarted`/`Done`/`Output` broadcasts, and `QueueChanged` all carry `execution_id`. New `QueueEntry { cell_id, execution_id }` struct.

**Daemon** - `QueuedCell` generates `execution_id` at queue time. `cell_id_map` stores `(cell_id, execution_id)` tuples. All broadcast sites propagate it.

**RuntimeStateDoc** - Parallel fields for execution IDs in queue schema. `QueueState`/`QueueEntry` types updated.

**Frontend** - TypeScript `QueueEntry` type, `App.tsx` and `useDaemonKernel` updated for new shape.

**Python** - `QueueState` exposes `executing_execution_id` and `queued_execution_ids`. `get_queue_state` reads from local doc replica instead of daemon request.

## Phase 2: Execution Handle

New `cell.execute()` returns an `Execution` handle:

```python
execution = await cell.execute()

# Block for result
result = await execution.result(timeout=30)

# Or stream events
async for event in execution.stream():
    print(event)

# Check status (sync)
execution.status  # "pending" | "running" | "done" | "error"

# Cancel
await execution.cancel()
```

`cell.run()`, `cell.queue()`, and `cell.stream()` are rewritten as sugar on top of `execute()`.

## Phase 3: Backwards compat

Existing `run()`, `queue()`, `stream()` work as sugar. Frontend handles the new `QueueChanged` shape.

## Phase 4: Cancel, re-execute, completion log

**Cancel** - `Execution.cancel()` sends kernel interrupt. Queue clearing needs a `ClearQueue` protocol variant (TODO - interrupt-only for now). Frontend keeps its existing interrupt button behavior.

**Re-execute** - Calling `execute()` on an already-queued cell returns a handle for the existing in-flight execution. The daemon queue is idempotent and returns the same `execution_id` in `CellQueued`.

**Completion log** - RuntimeStateDoc tracks the last 32 completed executions with `{ execution_id, cell_id, success }`. Daemon appends on `execution_done` and `kernel_died`. Late consumers calling `result()` check the completion log first, so they don't poll forever if the execution already finished. The cell's `execution_id` in metadata tells you whether your outputs are still current or were superseded.


rgbkrk (commented):
## Review: Execution Handle Abstraction

Overall this looks solid for a spike. The execution_id threading is complete and correct across all layers, CRDT ownership rules are respected (daemon owns execution_id), and the `Execution` handle API is clean and intuitive.

### What looks good

- `cell.execute()` → `Execution` handle is a great API surface
- `run()`, `queue()`, `stream()` properly rewritten as sugar on top of `execute()`
- Hybrid polling+broadcast approach in `result()` is resilient — doesn't rely on a single signal source
- Kernel death properly handled with synthetic error output
- Broadcast resubscription in `stream()` is idiomatic tokio
- PyO3 GIL handling and async patterns are correct throughout

### Findings (none blocking)

#### Medium

1. **`status()` try_lock error** (`execution.rs`) — `try_lock()` failure raises a Python error, but the API surface suggests it should be a simple sync property read. Concurrent operations (checking status while another cell executes) could surface unexpected `RuntimedError: Session state locked`. Consider either documenting the error or falling back gracefully.

2. **Empty execution_id on map lookup failure** (`kernel_manager.rs`) — If `cell_id_map.lock()` fails (unlikely), the broadcast gets `execution_id=""` while RuntimeStateDoc has the correct UUID. Creates an inconsistency between broadcast events and doc state.

3. **Silent length mismatch** (`runtime_state.rs` ~line 577) — `iter::repeat(String::new())` silently pads if `queued_execution_ids` is shorter than `queued`. This masks potential bugs in the parallel list maintenance. A debug log when lengths differ would help catch issues early.

#### Low

4. **Stale `QueueState` response** — `refreshQueueState()` in the frontend hardcodes `execution_id: ""`. Not a bug since RuntimeStateDoc is the source of truth now, but it's dead/inconsistent code that could confuse future readers.

5. **Fixed 50ms polling** — Progressive backoff in `collect_outputs_for_execution` could reduce unnecessary work for long-running kernels, but fine for a spike.

6. **Test coverage** — No tests for multiple `result()` calls, kernel death, concurrent `stream()`, or timeout edge cases. Expected for a spike.

Looks good to merge as a spike. The medium items are worth addressing before this graduates from spike status.

rgbkrk:
Addressed all three medium findings in 7f9c82fc:

1. **`status()` try_lock** — now returns `"running"` on contention instead of raising `RuntimedError`. If the async mutex is held, an operation is in flight, so `"running"` is the safe default.

2. **Empty execution_id on map lock failure** — added `warn!` on lock poisoning (was silently returning `None`). Also restructured the IOPub/shell handlers to extract `(cell_id, execution_id)` as a pair at the top, eliminating 6 scattered `unwrap_or_default()` calls. The empty-string fallback was technically unreachable (cell_id guards prevent broadcast emission when the lookup fails), but the code is now clearer about the invariant.

3. **Silent list length mismatch** — `read_state()` now logs a `warn!` when `queued` and `queued_execution_ids` have different lengths before padding.

rgbkrk:
## Review Notes (voice session with @rgbkrk)

Went through the full PR — all five commits across the Rust crates, Python bindings, and frontend. The spike successfully proves the concept end-to-end. Here's what we discussed:

### What the spike validates

- **Execution IDs work.** Threading a UUID from `queue_cell()` through IOPub routing, broadcasts, CRDT state, and the Python handle is clean and correct.
- **The `Execution` handle API is good.** `execute()` → handle with `result()`, `stream()`, `status`, `cancel()` is intuitive. Sugar methods (`run`, `queue`, `stream`) rewritten on top of it is the right layering.
- **CRDT as source of truth for queue state** — the frontend reading from `RuntimeStateDoc` instead of broadcast messages is the right call.
- **The completion log** (last 32 executions) solves the late-consumer problem elegantly.

### Design direction for production

Through the review, a cleaner architecture emerged that we want to pursue instead of graduating this spike directly:

**Separate intent from acknowledgment.** Clients write execution *intents* into the notebook doc (any client can write). The daemon writes execution *state* and *outputs* into the runtime state doc (daemon-only). The `execution_id` links the two.

Key changes from the spike:
- **Parallel lists → map of maps** in the CRDT. The spike uses parallel `queued` and `queued_execution_ids` lists, which is fragile (silent mismatch if one list is modified without the other). Map keyed by `execution_id` enforces the pairing structurally.
- **Cell gets an `execution_id` pointer.** Each cell in the notebook doc gains an `execution_id` field, set by the daemon at acknowledgment time (before kernel starts). This is how the frontend knows which outputs to display — read `outputs/{execution_id}/` from the runtime state doc.
- **Clear outputs = set `execution_id` to null.** Pure client operation, no daemon round trip. Old outputs stay in the runtime state doc until trimmed.
- **Save = snapshot both docs.** For each cell, look up its `execution_id`, pull outputs from the runtime state doc snapshot, serialize to ipynb.
- **Cancel = CRDT write.** Client sets `cancelled: true` on the intent in the notebook doc. Daemon picks it up. No new protocol message needed.

### Tokio simplification potential

The new architecture could reduce concurrency complexity in the daemon:
- The command processor task could shrink (execution lifecycle becomes CRDT writes, not channel messages)
- The IOPub task's output routing gets simpler (append to flat list by execution_id, not upsert into cell structures)
- The cell_id_map becomes a cache rather than source of truth

### Recommendation

**Keep this PR as a reference spike.** Ship Monday's release as-is (this is draft and doesn't need to land). Open a fresh PR for the execution log architecture tracked in #1048.

Design doc: [`.context/pr-1052/execution-log-design.md`](https://github.com/nteract/desktop/blob/spike/execution-handle-abstraction/.context/pr-1052/execution-log-design.md) (on this branch)
Tokio analysis: [`.context/pr-1052/tokio-analysis.md`](https://github.com/nteract/desktop/blob/spike/execution-handle-abstraction/.context/pr-1052/tokio-analysis.md) (on this branch)

rgbkrk:
Closing. Solid spike that led to new designs. Onward to new PRs after this next release.

PR #1061, opened 2026-03-22
fix(trust): add trust state to RuntimeStateDoc for race-free sync

### LINKED ISSUE #1057: bug(runtimed): trust dialog doesn't appear for notebooks with inline deps on Linux CI
The E2E trust dialog tests fail on Linux CI because the trust dialog never appears, even though the fixture notebook (`2-uv-inline.ipynb`) has inline UV dependencies and no `trust_signature` in metadata.

**Expected:** Opening a notebook with `runt.uv.dependencies` in metadata and no `trust_signature` should trigger the trust dialog.

**Actual on CI:** The notebook opens directly without trust approval. The daemon auto-launches the kernel as if the notebook is trusted.

**Fixture metadata:**
```json
{
  "runt": {
    "env_id": "fixture-uv-inline",
    "uv": {
      "dependencies": ["requests"],
      "requires-python": ">=3.10"
    }
  }
}
```
No `trust_signature` field present.

**Possible causes:**
1. `verify_trust_from_file` in `notebook_sync_server.rs` may not be reading `runt.uv.dependencies` correctly from the `.ipynb` metadata
2. The trust check returns `NoDependencies` despite inline deps being present
3. The trust behavior differs between macOS (local dev) and Linux (CI)

**Affected E2E tests:** Trust Dialog, UV Inline Deps, Conda Inline Deps — all 3 runners fail because the trust dialog never appears.

rgbkrk:
Add trust state to the RuntimeStateDoc so the frontend reads it via the same reactive Automerge sync pipeline as kernel status — no event listeners, no IPC round-trips, no race conditions.

**Problem:** The frontend's `checkTrust()` reads metadata from the Automerge doc (via IPC), which is empty until the streaming load completes. On slow I/O (Linux CI), the trust check runs before metadata arrives, returns `NoDependencies`, and the trust dialog never appears. (#1057)

**Solution:** The daemon already knows the trust status from `verify_trust_from_file()` during room creation. Write it into the RuntimeStateDoc immediately:

```
RuntimeStateDoc schema:
  trust/
    status: Str        ("trusted" | "untrusted" | "signature_invalid" | "no_dependencies")
    needs_approval: Bool
```

The frontend reads `runtimeState.trust.needs_approval` via `useRuntimeState()` — the same pipeline that delivers kernel status. When the RuntimeStateDoc syncs (which happens before the notebook doc streaming load), the trust state arrives reactively. No race.

**Changes:**
- `crates/notebook-doc/src/runtime_state.rs` — Added `TrustRuntimeState`, schema scaffolding, `set_trust()` method, snapshot reading
- `crates/runtimed/src/notebook_sync_server.rs` — Write trust state on room creation and on kernel launch (after trust approval)
- `apps/notebook/src/lib/runtime-state.ts` — Added `TrustState` to TypeScript interface
- `apps/notebook/src/hooks/useTrust.ts` — `needsApproval` now also reads from RuntimeStateDoc

Fixes #1057


rgbkrk:
## CI Results after WASM rebuild

**5/8 E2E green** — back to baseline, no regression from the RuntimeStateDoc change:

| Runner | Status |
|--------|--------|
| Smoke | ✅ |
| UI (Cell Visibility) | ✅ |
| UV Prewarmed | ✅ |
| Deno | ✅ |
| UV Pyproject | ✅ |
| Trust Dialog | ❌ |
| UV Inline Deps | ❌ |
| Conda Inline Deps | ❌ |

The initial WASM binary mismatch caused a full regression (0/8 E2E), but rebuilding the WASM with the `trust` field in `RuntimeState` fixed it.

The 3 trust-dialog failures still need investigation — the `needs_approval` flag is being written to the RuntimeStateDoc, but the E2E specs may need to wait for the RuntimeState sync before checking for the dialog. The trust dialog rendering path needs debugging on Linux.

PR #1084, opened 2026-03-23
feat(notebook-doc): standard bootstrap doc skeleton for all clients

rgbkrk:
Standard bootstrap document skeleton for all clients, with proper actor attribution.

## The Problem

Automerge's `load_incremental` has a fast-path for empty docs (`is_empty() == true`) that replaces `*self` with a freshly-loaded doc using default settings — discarding any encoding or actor we set at construction. This caused:

1. **Emoji corruption** — WASM's UTF-16 encoding was discarded on first sync, so CodeMirror's surrogate pair positions corrupted text near emoji (the 🐸 bug from [nteract/archived-desktop-app#1706](https://github.com/nteract/archived-desktop-app/issues/1706), 2017)
2. **Wrong attribution** — Actor labels were set *after* `connect()` returned, meaning bootstrap ops were attributed to a random UUID

## The Fix

`NotebookDoc::bootstrap(encoding, actor_label)` — a single constructor that every client calls before syncing:

```
ROOT/
  schema_version: 2
  cells: {}           (empty Map)
  metadata: {}        (empty Map)
```

Both parameters are required:
- **`encoding`** — `Utf16CodeUnit` for WASM/CodeMirror, `UnicodeCodePoint` for Python
- **`actor_label`** — identity for edit attribution, set before any ops

This mirrors the daemon's doc structure, so the CRDT merge converges with no conflicts. And because `is_empty()` is now false, `load_incremental` takes the normal incremental path — preserving encoding and actor.

### Simplified connect API

All three `notebook-sync` connect functions take `actor_label: &str` (required). No more post-hoc `set_actor()` calls. Python bindings default to `"runtimed-py"` when no peer label is provided.

### Multi-round sync in tests

The bootstrap skeleton means both peers have ops before sync, so frame-based tests may need multiple rounds (first round can be a heads exchange). Updated affected tests.

## Changes

- `notebook-doc`: `NotebookDoc::bootstrap(encoding, actor_label)` replaces all `empty*` / `bootstrap_with_*` variants
- `notebook-sync`: `connect()`, `connect_open()`, `connect_create()` take required `actor_label: &str`
- `runtimed-wasm`: `create_bootstrap(actor_label)` constructor
- `runtimed-py`: `DEFAULT_ACTOR_LABEL = "runtimed-py"`, wired through all connect paths
- Tests: descriptive actor labels, multi-round sync where needed

## Test results
- 227 notebook-doc ✅
- 31 notebook-sync ✅
- 123 Deno (50 smoke + 71 splice + 2 cross-impl) ✅
- Clippy, fmt, lint all clean ✅
- Workspace compiles (`cargo check --workspace --all-targets`) ✅


crates/runtimed-wasm/tests/deno_smoke_test.ts
```diff
@@ -1404,13 +1404,17 @@ Deno.test("Sync: bidirectional mutations converge", () => {
...
+  // default metadata — same structure as NotebookDoc::new().  This
+  // prevents automerge's load_incremental empty-doc fast-path from
+  // discarding encoding/actor settings on the first sync.
   assertEquals(handle.cell_count(), 0);
   assertEquals(handle.get_cells().length, 0);
   assertEquals(handle.get_cells_json(), "[]");
-  // Empty doc has no metadata
-  assertEquals(handle.get_metadata("runtime"), undefined);
+  // Bootstrap sets default runtime metadata
+  assertEquals(handle.get_metadata("runtime"), "python");
```
rgbkrk: We should not set default runtime to python.

crates/notebook-doc/src/lib.rs
```diff
@@ -650,52 +650,98 @@ impl NotebookDoc {
...
+        if let Some(label) = actor_label {
+            doc.set_actor(ActorId::from(label.as_bytes()));
+        }
+
+        // Mirror the structure from new_inner() — same keys, same values.
+        // The daemon writes these too, so the CRDT merge is a no-op.
         let _ = doc.put(automerge::ROOT, "schema_version", SCHEMA_VERSION);
+        let _ = doc.put_object(automerge::ROOT, "cells", ObjType::Map);
+        if let Ok(meta_id) = doc.put_object(automerge::ROOT, "metadata", ObjType::Map) {
+            let _ = doc.put(&meta_id, "runtime", "python");
```
rgbkrk: Let's not set the default runtime to python here.  I'd just leave this segment out, unless we're going to pass it into bootstrap.

PR #1086, opened 2026-03-23
fix(presence): clear stale selections on cursor/focus transitions

rgbkrk:
## Summary

The cursor-registry (hot path for dispatching remote presence to CodeMirror) kept stale selection highlights when receiving cursor-only updates from a peer. The `presenceSenderExtension` sends either a cursor (anchor === head) or a selection (anchor !== head), never both — so receiving a cursor-only update means the peer no longer has a selection. The registry now clears `peer.selection` on cursor-only updates, matching the existing pattern that already clears focus.

Note: the `focus` channel is intentionally left alone — it's a cell-level indicator emitted on execute/clear-outputs and doesn't imply the peer's selection is gone.

## Verification

- [ ] Open the same notebook in two windows side by side
- [ ] In window A, select text in a code cell — window B shows colored selection highlight
- [ ] In window A, click to place cursor (deselect) — window B's selection highlight disappears
- [ ] In window A, select text, then execute (Shift+Enter) — window B still shows the selection (execution doesn't change selection state)

### DIFF
--- a/apps/notebook/src/lib/cursor-registry.ts
+++ b/apps/notebook/src/lib/cursor-registry.ts
@@ -246,11 +246,6 @@ function handlePresence(payload: unknown): void {
           affectedCells.add(peer.cursor.cell_id);
           peer.cursor = undefined;
         }
-        // Focus replaces selection
-        if (peer.selection) {
-          affectedCells.add(peer.selection.cell_id);
-          peer.selection = undefined;
-        }
         if (peer.focus && peer.focus.cell_id !== data.cell_id) {
           affectedCells.add(peer.focus.cell_id);
         }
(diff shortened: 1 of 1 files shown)


apps/notebook/src/lib/cursor-registry.ts
```diff
@@ -241,6 +246,11 @@ function handlePresence(payload: unknown): void {
           affectedCells.add(peer.cursor.cell_id);
           peer.cursor = undefined;
         }
+        // Focus replaces selection
```
rgbkrk: Clearing `peer.selection` on every `focus` update regresses valid presence states. The Python session API emits focus for actions like `clear_outputs`, `execute_cell`, `queue_execute`, and `run_all_cells` without sending `clear_channel(selection)`, so a peer can still have an active selection while focus moves: https://github.com/nteract/desktop/blob/1c06c1348889027d1168013a88eedd83eece5a43/crates/runtimed-py/src/session_core.rs#L1033-L1036 and https://github.com/nteract/desktop/blob/1c06c1348889027d1168013a88eedd83eece5a43/crates/runtimed-py/src/session_core.rs#L1066-L1067. With this change, remote selection highlights disappear as soon as another client executes or clears outputs, which contradicts the PR verification case that selection should persist across execute.

PR #1098, opened 2026-03-24
fix(wasm-tests): make sync tests resilient to multi-round protocol

rgbkrk:
## Summary

Fixes the intermittent WASM test failures (`changeset: splice_source flags source as changed` and friends).

**Root cause:** Two independent issues compounding:

1. **Automerge multi-round sync:** The sync protocol can require multiple exchanges before changes are delivered (first round is often heads-only on Linux CI). The `syncViaFrame()` helper only did one round.

2. **Test parallelism:** Deno 2.4 appears to parallelize tests even within a single file. The WASM `NotebookHandle` instances share global state, so concurrent tests cause sync events to bleed across test boundaries (~8% failure rate).

**Changes:**

- `syncViaFrame()` now runs up to 5 bidirectional sync rounds, collecting events across all of them and breaking early when a `sync_applied` event with changes is found
- Collapsed the inline multi-round pattern from the deletion attribution test into the shared helper (DRY)
- Added `--parallel=false` to the CI deno test command

**Verification:** 0/80 failures with `--parallel=false` vs ~4/50 without.

## Test plan

- [x] 30/30 local runs pass with `--parallel=false`
- [x] 50/50 local runs pass with `--parallel=false`
- [ ] CI passes on this PR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,7 @@ jobs:
           deno-version: v2.x
 
       - name: Run runtimed-wasm Deno tests
-        run: deno test --allow-read --allow-env --no-check crates/runtimed-wasm/tests/
+        run: deno test --allow-read --allow-env --no-check --parallel=false crates/runtimed-wasm/tests/
 
   build-ui:
     name: Build shared UI artifacts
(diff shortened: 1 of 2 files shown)


rgbkrk (commented):
Looks good. The multi-round `syncViaFrame()` fix is the right approach — Automerge sync is a multi-step protocol and the old single-round helper was always fragile. Collapsing the inline pattern from the deletion test into the shared helper is a clean DRY improvement.

`--parallel=false` makes sense given the shared WASM global state.

Will merge once CI is green.

PR #1119, opened 2026-03-24
feat(python): improve API discoverability with _repr_markdown_ and callable guards

### LINKED ISSUE #1115: Improve property vs method discoverability on runtimed Python API
## Summary

The runtimed Python API has a clean, consistent pattern:
- **Properties** for read-only state (`ids`, `source`, `outputs`, `cell_type`, `peers`, etc.)
- **Async methods** for mutations or daemon communication (`execute()`, `save()`, `create()`, etc.)
- **Sync methods** for local-only operations (`find()`, `get_by_id()`, `snapshot()`)

However, this isn't obvious at call time. During testing, I hit `TypeError: 'list' object is not callable` by writing `cells.ids()` instead of `cells.ids`.

## Suggestions

1. **Docstrings** on properties and methods visible via `help()`
2. **`_repr_markdown_`** on key objects (Notebook, CellCollection, CellHandle) that shows a quick API reference when displayed in a notebook — markdown is readable by both agents and humans (agents don't get HTML by default)
3. **Better error messaging** — e.g., if a property returns a list and someone tries to call it, a custom descriptor could provide a hint

## Full API classification (from introspection)

```
=== Notebook ===
  Properties: cells, notebook_id, peers, presence, runtime
  Async methods: add_dependency, close, get_dependencies, interrupt, is_connected,
                 queue_state, remove_dependency, restart, run_all, save, save_as,
                 shutdown, start, sync_environment

=== CellCollection ===
  Properties: ids
  Async methods: create, insert_at
  Sync methods: find, get_by_id, get_by_index

=== CellHandle ===
  Properties: cell_type, execution_count, id, metadata, outputs, outputs_hidden, source, source_hidden, tags
  Async methods: append, clear_outputs, delete, execute, move_after, queue, run,
                 set_outputs_hidden, set_source, set_source_hidden, set_tags, set_type, splice
  Sync methods: snapshot
```

## Environment

- `runtimed==2.0.3a202603241511`

rgbkrk:
## Summary

Improves discoverability of the runtimed Python API by making the property-vs-method distinction visible at call time.

- Add `_repr_markdown_` to `Notebook`, `CellCollection`, `CellHandle`, and `Client` — renders a quick-reference table of properties (sync) vs methods (async) when displayed in a notebook cell
- Add `_HintList` wrapper for list-returning properties (`ids`, `outputs`, `tags`) that raises a clear `TypeError` when accidentally called: `'ids' is a property, not a method — drop the parentheses: .ids`

Closes #1115

## Verification

* [ ] Display a `Notebook` object in a notebook cell and verify the markdown table renders with properties and methods
* [ ] Display `notebook.cells` and a `CellHandle` to verify their markdown tables
* [ ] Call `notebook.cells.ids()` (with parentheses) and verify the error message says to drop the parentheses
* [ ] Confirm `notebook.cells.ids` (without parentheses) still returns a normal list
* [ ] Verify `cell.outputs()` and `cell.tags()` also produce the helpful error

### DIFF
--- a/python/runtimed/src/runtimed/_cell.py
+++ b/python/runtimed/src/runtimed/_cell.py
@@ -17,6 +17,21 @@ if TYPE_CHECKING:
     )
 
 
+class _HintList(list):
+    """List that gives a helpful error if accidentally called."""
+
+    __slots__ = ("_attr",)
+
+    def __init__(self, items, attr: str):
+        super().__init__(items)
+        self._attr = attr
+
+    def __call__(self, *a, **kw):
+        raise TypeError(
+            f"'{self._attr}' is a property, not a method — drop the parentheses: .{self._attr}"
+        )
+
+
 class CellHandle:
     """A live reference to a cell in the notebook document.
 
@@ -48,9 +63,9 @@ class CellHandle:
         """Resolved outputs (sync — may do disk I/O for blob resolution)."""
         try:
             cell = self._session.get_cell_sync(self._id)
-            return cell.outputs
+            return _HintList(cell.outputs, "outputs")
         except _RuntimedError:
-            return []
+            return _HintList([], "outputs")
 
     @property
     def execution_count(self) -> int | None:
@@ -78,9 +93,9 @@ class CellHandle:
     def tags(self) -> list[str]:
         """Cell tags (sync read). Uses Rust Cell helpers for key resolution."""
         try:
-            return self._session.get_cell_sync(self._id).tags
+            return _HintList(self._session.get_cell_sync(self._id).tags, "tags")
         except _RuntimedError:
-            return []
+            return _HintList([], "tags")
 
     @property
     def source_hidden(self) -> bool:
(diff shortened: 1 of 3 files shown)


python/runtimed/src/runtimed/_cell.py
```diff
@@ -187,6 +202,25 @@ async def set_outputs_hidden(self, hidden: bool) -> CellHandle:
         await self._session.set_cell_outputs_hidden(self._id, hidden)
         return self
 
+    def _repr_markdown_(self) -> str:
+        src = self.source
+        preview = (src[:60] + "...") if len(src) > 60 else src
+        preview = preview.replace("\n", "\\n")
+        return (
+            f"**Cell** `{self._id[:8]}` ({self.cell_type})\n\n"
+            f"```\n{preview}\n```\n\n"
```
rgbkrk: `_repr_markdown_()` will generate invalid markdown for cells whose source already contains triple backticks. The preview is interpolated directly inside a ``` fence here, so a markdown cell like ```python\nprint(1)\n``` closes the outer fence early and corrupts the rest of the quick-reference table. I reproduced this from the workspace with `PYTHONPATH=python/runtimed/src python` and a stub `CellHandle`.

PR #1155, opened 2026-03-24
refactor(runtimed): rename Notebook.shutdown/close to stop_runtime/disconnect

### LINKED ISSUE #1147: [runtimed] Notebook.shutdown() naming is misleading — shuts down kernel, not notebook
## Summary

`Notebook.shutdown()` calls `shutdown_kernel()` under the hood — it stops the runtime but the notebook remains active in the daemon (still shows up in `list_active_notebooks` with the daemon as a peer). `close()` is the session disconnect.

This is confusing for agents who expect `shutdown()` to fully tear down the notebook. The naming should better reflect what actually happens.

**Tested on:** `runtimed==2.0.3a202603242044`

## Where

`_notebook.py:88-90` — `shutdown()` calls `self._session.shutdown_kernel()`
`_notebook.py:112-114` — `close()` closes the session

## Suggestion

Consider renaming before a stable release to avoid a major version bump later:
- `shutdown()` → `shutdown_kernel()` or `stop_runtime()` (to match what it does)
- Or make `shutdown()` actually shut down the notebook session entirely

rgbkrk:
## Summary

Renames two methods on the `Notebook` class to better reflect what they actually do:

- `shutdown()` → `stop_runtime()` — stops the kernel, session stays connected
- `close()` → `disconnect()` — drops the session connection to the daemon

The underlying `Session.shutdown_kernel()` and `Session.close()` are unchanged — their names are already accurate at that abstraction level.

Closes #1147

## Verification

- [ ] `await notebook.stop_runtime()` stops the kernel; notebook remains in `list_active_notebooks`
- [ ] `await notebook.disconnect()` drops the session; `async with` context manager still works
- [ ] MCP server correctly disconnects when switching notebooks (`join_notebook`, `open_notebook`, `create_notebook`)

### DIFF
--- a/python/runtimed/src/runtimed/_notebook.py
+++ b/python/runtimed/src/runtimed/_notebook.py
@@ -85,8 +85,8 @@ class Notebook:
         """
         await self._session.start_kernel(runtime, env_source, notebook_path)
 
-    async def shutdown(self) -> None:
-        """Shut down the runtime."""
+    async def stop_runtime(self) -> None:
+        """Stop the runtime (kernel). The session remains connected."""
         await self._session.shutdown_kernel()
 
     async def restart(self, wait_for_ready: bool = True) -> list[str]:
@@ -109,8 +109,8 @@ class Notebook:
         """Get the execution queue state (currently executing + queued cells)."""
         return await self._session.get_queue_state()
 
-    async def close(self) -> None:
-        """Close the notebook session."""
+    async def disconnect(self) -> None:
+        """Disconnect from the notebook session."""
         await self._session.close()
 
     # ── Dependency management ────────────────────────────────────────
@@ -166,7 +166,7 @@ class Notebook:
         return self
 
     async def __aexit__(self, *args) -> None:
-        await self.close()
+        await self.disconnect()
 
     def _repr_markdown_(self) -> str:
         nid = self.notebook_id[:12]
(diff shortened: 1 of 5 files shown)


python/runtimed/src/runtimed/_notebook.py
```diff
@@ -85,8 +85,8 @@ async def start(
         """
         await self._session.start_kernel(runtime, env_source, notebook_path)
 
-    async def shutdown(self) -> None:
-        """Shut down the runtime."""
+    async def stop_runtime(self) -> None:
```
rgbkrk, line 88: This removes two documented `Notebook` methods from the public Python API without a compatibility shim. After this change, `hasattr(runtimed.Notebook, "shutdown")` and `hasattr(runtimed.Notebook, "close")` are both false, so existing callers will fail with `AttributeError` on upgrade even though the underlying behavior is unchanged. These names were part of the documented high-level API before this PR: https://github.com/nteract/desktop/blob/7d6176de54670541f8f326bb44c3647d9e234383/docs/python-bindings.md#L101-L106 and https://github.com/nteract/desktop/blob/7d6176de54670541f8f326bb44c3647d9e234383/docs/python-bindings.md#L280-L284. If the intent is a non-breaking rename, keep `shutdown()`/`close()` as aliases for at least one release; otherwise this needs to be treated as a breaking API change.

PR #1157, opened 2026-03-24
test(runtimed-py): regression test for coroutine warning in list_active_notebooks

### LINKED ISSUE #1148: [runtimed] list_active_notebooks emits spurious coroutine warning
## Summary

Calling `await client.list_active_notebooks()` works and returns results, but emits:

> RuntimeWarning: coroutine 'Client.list_active_notebooks' was never awaited

The results are correct, but the warning is confusing — especially for agents that parse stderr.

**Tested on:** `runtimed==2.0.3a202603242044`

## Where

- `async_client.rs:104-123` — Rust `future_into_py()` returns a coroutine
- `_client.py:34` — Python wrapper awaits it

Something in the chain appears to create an extra unawaited coroutine object.

**Note:** This may have been addressed in #1120 but the fix does not appear to be in this nightly build.

rgbkrk:
## Summary

Adds a regression test for #1148 — `list_active_notebooks()` used to emit a spurious `RuntimeWarning: coroutine 'Client.list_active_notebooks' was never awaited`. The root cause (a manual `IntoPyObject` impl on `RoomInfoData`) was fixed in #1120 by switching to `#[derive(IntoPyObject)]`. This test guards against the warning recurring.

The test calls `await async_client.list_active_notebooks()` inside `warnings.catch_warnings()` with `simplefilter("error", RuntimeWarning)`, so any spurious coroutine warning becomes a test failure.

Closes #1148

## Verification

- [ ] Integration test `test_list_active_notebooks_no_coroutine_warning` passes in CI

### DIFF
--- a/python/runtimed/tests/test_daemon_integration.py
+++ b/python/runtimed/tests/test_daemon_integration.py
@@ -1676,6 +1676,16 @@ class TestBasicConnectivity:
         assert "AsyncSession" in r
         assert async_session.notebook_id in r
 
+    @pytest.mark.asyncio
+    async def test_list_active_notebooks_no_coroutine_warning(self, async_client):
+        """Regression: list_active_notebooks must not emit RuntimeWarning (#1148)."""
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", RuntimeWarning)
+            rooms = await async_client.list_active_notebooks()
+        assert isinstance(rooms, list)
+
 
 class TestDocumentFirstExecution:
     """Test document-first execution pattern."""


rgbkrk:
Closing — investigation shows the underlying bug was already fixed by #1120 (`#[derive(IntoPyObject)]`). 

Reproduction attempts with proper `await` usage (simple calls, concurrent calls, GC pressure, rapid client creation/destruction) all pass without warnings. The initial test approach (`warnings.catch_warnings`) was also ineffective for GC-time warnings.

The reporter's nightly build likely had stale artifacts despite the timestamp suggesting it should have included #1120.

PR #1198, opened 2026-03-26
Add Nix flake development environment

MostlyKIGuess:
- Provides a reproducible dev shell with Rust, Node.js, and Python toolchain via `nix develop`. Includes Tauri Linux dependencies, GStreamer media codecs, and proper env variables.
- Had to add explicit color as cursor wasn't being visible.


rgbkrk (approved):
Thank you!

Probably could go without the wasm changes, but I know that somehow `wasm-pack` builds different between MacOS and Linux.

src/components/editor/extensions.ts
```diff
@@ -54,8 +54,17 @@ export const notebookEditorTheme = EditorView.theme({
...
   ".cm-cursor": {
     borderLeftWidth: "2px",
+    borderLeftColor: "var(--foreground)",
+  },
+  ".cm-focused .cm-cursor": {
+    borderLeftColor: "var(--foreground)",
+  },
+  // Ensure cursor is visible even when not focused
+  ".cm-content": {
+    caretColor: "var(--foreground)",
```
rgbkrk:
```
Run pnpm --dir apps/notebook build

> notebook-ui@0.1.0 build /home/runner/work/desktop/desktop/apps/notebook
> tsc -b && vite build

Error: ../../src/components/editor/extensions.ts(66,3): error TS1117: An object literal cannot have multiple properties with the same name.
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

You'll need to combine with the `.cm-content` from above that has 

```
  ".cm-content": {
    paddingTop: "1.375rem",
  },
```

MostlyKIGuess:
> Thank you!
> 
> Probably could go without the wasm changes, but I know that somehow `wasm-pack` builds different between MacOS and Linux.

Hmm, I can perhaps also add darwin inputs so that this is handled for MacOS as well incase someone wants to use nix there.

rgbkrk:
> Hmm, I can perhaps also add darwin inputs so that this is handled for MacOS as well incase someone wants to use nix there.

No worries on this PR. My comment was only about wasm generation not being stable per platform.

PR #1291, opened 2026-03-30
fix(sync): prevent execute race with in-flight debounced flush

rgbkrk:
## Summary

When typing quickly and pressing Shift+Enter, the daemon could execute stale cell source (e.g. `pri` instead of `print('hey')`). The root cause: `SyncEngine.flush()` (called by the 20ms debounce timer) is fire-and-forget — if the debounce claimed changes from `flush_local_changes()` but its IPC hadn't landed yet, `flushSync()` would return with nothing to send, letting `executeCell()` race past the still-in-flight sync frame.

## Changes

- **`packages/runtimed/src/sync-engine.ts`**: Track the in-flight debounced flush promise. Add `flushAndWait()` that awaits it before flushing remaining changes with an awaited send.
- **`apps/notebook/src/hooks/useAutomergeNotebook.ts`**: `flushSync()` now delegates to `engine.flushAndWait()` instead of independently calling `flush_local_changes()` (which could race with the debounce timer).
- **`apps/notebook/src/App.tsx`**: Add per-cell `executingCellsRef` dedup guard to prevent rapid Shift+Enter from flooding the daemon with duplicate ExecuteCell requests. Auto-restart kernel and retry once when `executeCell` returns `no_kernel`.

## Verification

- [ ] Create a new Untitled notebook, type `print("hello")` and immediately press Shift+Enter — verify the full source executes (not a partial string like `pri`)
- [ ] Rapidly press Shift+Enter multiple times on the same cell — verify only one execution fires, no kernel crash
- [ ] Shut down the kernel manually, then press Shift+Enter — verify the kernel auto-restarts and the cell executes
- [ ] Verify `Restart and Run All` still works correctly
- [ ] Verify save (Cmd+S) still works

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -622,7 +622,14 @@ function AppContent() {
           }
         }
       } finally {
-        executingCellsRef.current.delete(cellId);
+        // Hold the guard briefly to prevent rapid re-queueing.
+        // The daemon returns CellQueued immediately (~30ms), but
+        // the cell may still be executing. A 1s hold prevents
+        // accidental double-execution from rapid keypresses while
+        // still allowing intentional re-runs.
+        setTimeout(() => {
+          executingCellsRef.current.delete(cellId);
+        }, 1000);
       }
     },
     [flushSync, kernelStatus, tryStartKernel, executeCell],
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -184,6 +184,9 @@ function AppContent() {
   // Guard against concurrent Run All / Restart & Run All operations (#982)
   const runAllInFlightRef = useRef(false);
 
+  // Guard against duplicate per-cell execute requests (rapid Shift+Enter)
+  const executingCellsRef = useRef(new Set<string>());
+
   // Notebook runtime type — reactive read from WASM Automerge doc.
   // Re-renders automatically when metadata changes (bootstrap, sync, writes).
   const detectedRuntime = useDetectRuntime();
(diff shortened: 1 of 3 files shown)


apps/notebook/src/App.tsx
```diff
@@ -573,25 +576,53 @@ function AppContent() {
...
+              );
+            } else if (retry.result === "no_kernel") {
+              logger.error(
+                "[App] handleExecuteCell: still no kernel after restart",
+              );
+            }
+          }
+        }
+      } finally {
+        executingCellsRef.current.delete(cellId);
```
rgbkrk: This guard only covers the request round-trip, not the cells actual queued/executing lifetime. `execute_cell_via_daemon` returns as soon as the request is queued ([crates/notebook/src/lib.rs#L2184-L2200](https://github.com/nteract/desktop/blob/7e86da0d40dc935dcea0f7d6df9d261310f75fa4/crates/notebook/src/lib.rs#L2184-L2200)), and the daemon returns `CellQueued` immediately after `queue_cell` ([crates/runtimed/src/notebook_sync_server.rs#L3658-L3700](https://github.com/nteract/desktop/blob/7e86da0d40dc935dcea0f7d6df9d261310f75fa4/crates/runtimed/src/notebook_sync_server.rs#L3658-L3700)). Because the `finally` block removes the ID right away, another Shift+Enter after the first response will still enqueue the same cell again, so the duplicate-execute problem is still reproducible once the first RPC completes.

packages/runtimed/src/sync-engine.ts
```diff
@@ -610,6 +615,52 @@ export class SyncEngine {
...
+   *    changes from `flush_local_changes()`.
+   * 2. Flushes any remaining local changes and awaits delivery.
+   *
+   * Use before execute/save to guarantee the daemon has the latest source.
+   */
+  async flushAndWait(): Promise<void> {
+    // Wait for any in-flight debounced flush to land first.
+    if (this.inflightFlush) {
+      await this.inflightFlush;
+      this.inflightFlush = null;
```
rgbkrk: `flushAndWait()` only remembers one in-flight debounced send, so it can still return before the latest sync frame lands. `scheduleFlush()` can trigger another fire-and-forget `flush()` every 20ms ([packages/runtimed/src/sync-engine.ts#L544-L547](https://github.com/nteract/desktop/blob/7e86da0d40dc935dcea0f7d6df9d261310f75fa4/packages/runtimed/src/sync-engine.ts#L544-L547)), and `TauriTransport.sendFrame()` does not serialize or chain those sends ([apps/notebook/src/lib/tauri-transport.ts#L21-L25](https://github.com/nteract/desktop/blob/7e86da0d40dc935dcea0f7d6df9d261310f75fa4/apps/notebook/src/lib/tauri-transport.ts#L21-L25)). If a second debounced flush claims newer edits while this method is awaiting the first promise, line 632 clears the newer promise and line 639 sees no local changes, so execute/save can still race ahead of the newest IPC.

PR #1335, opened 2026-03-31
refactor(presence): align identity between presence and text attribution

rgbkrk:
## Summary

- **Shared editor registry**: Cell components now register their CodeMirror EditorView once via `editor-registry.ts` instead of calling both `registerEditor` and `registerAttributionEditor` separately. Both cursor-registry and attribution-registry read from the shared map.
- **Strip usePresence to send-only**: Removed the hook's dead read-side (~160 lines) — no consumer used `cursorsForCell`, `selectionsForCell`, or `peerVersion`. The cursor-registry already handles all read-side concerns with better message coverage (focus, clear_channel, reconciliation).
- **Identity alignment**: Added `actor_label: Option<String>` to the presence wire format (CBOR `Update` and `PeerSnapshot`) so attribution highlights use the exact same color as the peer's live cursor. Previously, fuzzy substring matching between actor labels and peer display names failed for generic names like "Agent". The change spans all layers: Rust wire types → WASM bindings → daemon relay → MCP server → Python bindings → frontend send/receive.

Closes #833

## Verification

- [ ] Open a notebook via MCP agent, execute a cell — agent's cursor dot and attribution underline use the same color
- [ ] Open the same notebook in two app windows, type in one — attribution underline in the other window matches the cursor color
- [ ] Verify presence still works: cursor flags, selection highlights, cell presence dots, late-joiner snapshots


crates/notebook-doc/src/presence.rs
```diff
@@ -320,6 +336,7 @@ pub fn encode_custom_update_labeled(
     encode_message(&PresenceMessage::Update {
         peer_id: peer_id.to_string(),
         peer_label: peer_label.map(|s| s.to_string()),
+        actor_label: None,
```
rgbkrk: This still drops `actor_label` for the empty-notebook announce path. Both MCP and Python fall back to `encode_custom_update_labeled(...)` when `first_cell_id()` is `None`, so the peer gets registered without an actor identity. After this refactor, attribution color matching is exact on `actor_label`, which means the first attributed edit in an empty notebook will fall back to `peerColor(actors[0])` instead of the peer color until a later focus/cursor update arrives. That is observable today because `create_cell` confirms sync before it emits cursor presence.

PR #1373, opened 2026-04-01
feat(env): surface prewarmed packages in RuntimeStateDoc and MCP tools

### LINKED ISSUE #1360: feat(env): surface prewarmed dependencies in notebook metadata
## Problem

When a notebook uses a prewarmed environment (e.g., `uv:prewarmed`), packages like `pandas`, `numpy`, `matplotlib` are available at runtime but **not declared anywhere in the notebook**. This causes two problems:

1. **No reproducibility** — sharing the notebook or opening it on a different machine (or after a pool reset) silently loses access to those packages. There's no record of what was actually imported.

2. **No visibility for agents** — MCP agents calling `get_dependencies` see an empty list, so they don't know what's available. They end up calling `add_dependency` + `sync_environment` for packages that are already installed, wasting time and potentially changing the environment.

## Observed behavior

- Create a notebook with `uv:prewarmed` environment
- `import pandas as pd` works
- `get_dependencies` returns `[]`
- Agent has no way to know pandas is available without trying to import it

## Expected behavior

Either:
- **Option A**: After execution, detect imports and inject them as inline script dependencies (`# /// script` / `[tool.uv]` metadata) so the notebook is self-contained
- **Option B**: Surface prewarmed/pool packages in the runtime state or dependency UI as "available (not pinned)" so users and agents can see what's there
- **Option C**: Both — show what's available, prompt to pin

## Context

This came up during an MCP testing session where `import plotly` failed because it wasn't in the prewarmed pool, but `import pandas` succeeded silently. The agent had no way to distinguish "not installed" from "installed via pool but not declared."

rgbkrk:
## Summary

When a notebook uses a prewarmed environment (`uv:prewarmed`), packages like `pandas` and `numpy` are available at runtime but not declared in notebook metadata. This means `get_dependencies` returns `[]` — MCP agents waste time re-adding packages that are already installed, and there's no visibility into what's available.

This PR tracks which packages were pre-installed in pooled environments and surfaces them as ephemeral runtime state through RuntimeStateDoc and MCP tools.

Closes #1360

## What changed

- **`PooledEnv`** gets a `prewarmed_packages` field, populated at pool creation time from the `install_packages` list (ipykernel, ipywidgets, anywidget, uv + user-configured defaults)
- **`LaunchedEnvConfig`** propagates prewarmed packages through kernel launch for `uv:prewarmed` and `conda:prewarmed` environments
- **`RuntimeStateDoc`** gains `env.prewarmed_packages` — written at kernel launch, cleared on shutdown
- **`get_dependencies` MCP tool** now returns `available_packages` alongside `dependencies` when prewarmed packages are present
- **`read_runtime_info`** surfaces prewarmed packages in `join_notebook`, `open_notebook`, and `show_notebook` responses
- TypeScript (`EnvState`), Python bindings (`PyEnvState`), and type stubs all updated

Prewarmed packages are **ephemeral runtime state only** — they are not persisted to `.ipynb` metadata. Users must explicitly `add_dependency` to pin packages for reproducibility.

## Verification

- [ ] Open a notebook with `uv:prewarmed` environment
- [ ] Call `get_dependencies` — should return `available_packages` list with prewarmed packages
- [ ] Call `show_notebook` — should include `prewarmed_packages` in runtime info
- [ ] Open a notebook with `uv:inline` declared deps — `available_packages` should be absent
- [ ] Shut down kernel and verify `prewarmed_packages` clears from RuntimeStateDoc


crates/runtimed/src/daemon.rs
```diff
@@ -901,6 +901,7 @@ impl Daemon {
                                 env_type: EnvType::Uv,
                                 venv_path: env_path.clone(),
                                 python_path,
+                                prewarmed_packages: vec![],
```
rgbkrk: Because `find_existing_environments()` rebuilds pooled envs with `prewarmed_packages: []`, any daemon restart drops the metadata this feature depends on. After restart, reused `uv:prewarmed` and `conda:prewarmed` envs will report no `available_packages` even though the packages are still installed. The only non-empty population in this patch is the fresh UV warmup path at https://github.com/nteract/desktop/blob/d16f513d77c8fd19546aad3143de233b83b1a989/crates/runtimed/src/daemon.rs#L2831-L2837, so the feature disappears as soon as the daemon reuses an existing pool.

crates/runtimed/src/daemon.rs
```diff
@@ -2437,6 +2439,7 @@ impl Daemon {
                     env_type: EnvType::Conda,
                     venv_path: env_path.clone(),
                     python_path,
+                    prewarmed_packages: vec![],
```
rgbkrk: This still initializes every fresh conda pool entry with an empty package list, so `conda:prewarmed` kernels can never surface `available_packages`. `build_launched_config()` now propagates `prewarmed_packages` for `conda:prewarmed` too (https://github.com/nteract/desktop/blob/d16f513d77c8fd19546aad3143de233b83b1a989/crates/runtimed/src/notebook_sync_server.rs#L200-L206), but nothing ever populates the source data on the conda side.

crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -3943,7 +3965,10 @@ async fn handle_notebook_request(
                         // Dual-write shutdown to state_doc
                         {
                             let mut sd = room.state_doc.write().await;
-                            if sd.set_kernel_status("shutdown") {
+                            let mut changed = false;
+                            changed |= sd.set_kernel_status("shutdown");
+                            changed |= sd.set_prewarmed_packages(&[]);
```
rgbkrk, line 3980: Clearing the field only on the explicit `ShutdownKernel` request is not enough. The other terminal paths still leave `env.prewarmed_packages` behind: `RoomKernel::shutdown()` only writes status/queue (https://github.com/nteract/desktop/blob/d16f513d77c8fd19546aad3143de233b83b1a989/crates/runtimed/src/kernel_manager.rs#L2553-L2556), `apply_kernel_died_to_state_doc()` only writes status/queue/comms (https://github.com/nteract/desktop/blob/d16f513d77c8fd19546aad3143de233b83b1a989/crates/runtimed/src/notebook_sync_server.rs#L605-L615), and the launch-failure branch above only sets `error` (https://github.com/nteract/desktop/blob/d16f513d77c8fd19546aad3143de233b83b1a989/crates/runtimed/src/notebook_sync_server.rs#L3731-L3737). A failed relaunch from a prewarmed env into anything else will keep reporting stale `available_packages` even though no prewarmed kernel is running.

Issue #1409, opened 2026-04-03
refactor(frontend): migrate off broadcasts toward RuntimeStateDoc sync


rgbkrk:
## Summary

The frontend currently reads some state from `NotebookBroadcast` messages (frame 0x03) and some from RuntimeStateDoc sync (frame 0x05). The goal is to migrate entirely to RuntimeStateDoc as the source of truth, eliminating broadcast-driven state.

## Why

- Runtime agents (#1333) are sync-only — they don't forward broadcasts
- Broadcasts are legacy from before RuntimeStateDoc existed
- Single source of truth reduces bugs and simplifies the protocol
- Enables removing broadcast variants from the protocol over time

## What broadcasts still drive frontend state?

Audit needed. Key candidates:
- `KernelStatus` → already in RuntimeStateDoc (`kernel.status`)
- `ExecutionStarted` → already in RuntimeStateDoc (`executions/{eid}/status = "running"`)
- `ExecutionDone` → already in RuntimeStateDoc (`executions/{eid}/status = "done"`)
- `Output` → already in RuntimeStateDoc (`executions/{eid}/outputs`)
- `QueueChanged` → already in RuntimeStateDoc (`queue.*`)
- `EnvSyncState` → already in RuntimeStateDoc (`env.*`)

The frontend may be using broadcasts for immediate UI reactivity. The question is whether RuntimeStateDoc sync is fast enough to replace them (it should be — same process, no network).

## Approach

1. Audit which broadcasts the frontend subscribes to (`subscribeBroadcast` in `notebook-frame-bus.ts`)
2. For each, check if the equivalent data is in RuntimeStateDoc
3. Migrate consumers to read from `useRuntimeState()` instead
4. Remove broadcast subscriptions one by one
5. Eventually deprecate the broadcast variants

## Files

- `apps/notebook/src/lib/notebook-frame-bus.ts` — broadcast subscriptions
- `apps/notebook/src/hooks/useDaemonKernel.ts` — likely broadcast consumer
- `apps/notebook/src/lib/runtime-state.ts` — RuntimeStateDoc store

## Test plan

- [ ] All existing E2E tests pass
- [ ] Manual: kernel status, execution, outputs all display correctly
- [ ] No regressions in widget interactions

Part of #832
Supports #1333

rgbkrk:
## Broadcast → RuntimeStateDoc Migration Audit

### Summary

[...]

Issue #1405, opened 2026-04-03
refactor(runtimed): remove execution_count from NotebookDoc — resolve at save time from RuntimeStateDoc


rgbkrk:
## Summary

`execution_count` is currently stored in **two** places:
1. **NotebookDoc** `cells/{cell_id}/execution_count` — persisted, written by coordinator (via `QueueCommand::ExecutionCountSet`)
2. **RuntimeStateDoc** `executions/{execution_id}/execution_count` — ephemeral, written by kernel IOPub

The NotebookDoc copy is a denormalized cache. The frontend reads it from the WASM-materialized cell for the gutter display (`[5]`), and `.ipynb` export reads it for Jupyter format compliance. Both can resolve from RuntimeStateDoc instead:

- **Frontend**: reads `cell.execution_id` → looks up `executions/{eid}/execution_count` from RuntimeStateDoc (via `useRuntimeState()`)
- **Save**: joins `cell.execution_id` → `executions/{eid}/execution_count` at serialization time

### 1. Save path: resolve execution_count from RuntimeStateDoc

In the `.ipynb` save code, instead of reading `cell.execution_count` from NotebookDoc, look up the cell's `execution_id` in RuntimeStateDoc's `executions` map and read `execution_count` from there.

### 2. Frontend: read from RuntimeStateDoc

The `CompactExecutionButton` currently receives `cell.execution_count` from the materialized cell. Instead, derive it from the cell's `execution_id` + the runtime state's execution entries.

### 3. Remove from NotebookDoc

- Remove `execution_count` field from `CellSnapshot` in `notebook-doc`
- Remove `set_execution_count()` method from `NotebookDoc`
- Remove `QueueCommand::ExecutionCountSet` from `kernel_manager.rs`
- Remove the handler in `start_command_loop()` that writes execution_count to NotebookDoc
- Stop writing `"null"` to execution_count at queue time in `notebook_sync_server.rs`

### 4. Remove from WASM

- Remove `execution_count` from `WasmCellSnapshot` in `runtimed-wasm`
- Remove `get_cell_execution_count()` and `set_execution_count()` from WASM API

## Why

- Eliminates the last kernel→NotebookDoc write path (via the coordinator's `QueueCommand::ExecutionCountSet`)
- Single source of truth — RuntimeStateDoc is authoritative for all execution state
- Simplifies the agent protocol — agents don't need `ExecutionCountSet` notifications
[...]

Issue #1333, opened 2026-03-31
feat(runtimed): process-isolated runtime agent with blob relay


rgbkrk:
## Summary

~~Spawn the kernel agent as a separate process, communicating with the coordinator via the existing framed protocol.~~

**Status: SHIPPED** — Agent mode is unconditional as of #1433. Every kernel runs in a separate `runtimed agent` subprocess.

### Architecture (evolved from original plan)

The original plan used stdin/stdout framing. The shipped implementation uses **Unix socket peer connections** (#1431) with **CRDT-driven execution** — cleaner, more resilient, and directly enables SSH remote runtimes.

```
Coordinator (runtimed daemon)
  ├── Resolves environments, manages trust, handles peer sync
  ├── Writes execution entries (source + seq) to RuntimeStateDoc
  └── Connects to agent via standard Unix socket protocol

Agent subprocess (runtimed agent)
  ├── Connects back to daemon socket as Automerge peer
  ├── Watches RuntimeStateDoc for queued execution entries
  ├── Owns kernel process (ZMQ), IOPub, outputs
  ├── Writes results back via RuntimeStateDoc sync
  └── Supports RestartKernel RPC for in-place restart
```

### Key differences from original plan

| Original plan | What shipped |
|--------------|-------------|
| stdin/stdout framed protocol | Unix socket peer (standard handshake + frames) |
| `AgentRequest::ExecuteCell` RPC | CRDT-driven: coordinator writes queue entries, agent watches |
| Blob relay (new frame type) | Shared filesystem blob store (agent reads/writes directly) |
| Separate `runt agent` binary | Hidden `runtimed agent` subcommand on the daemon binary |
| Agent as special peer | Agent IS a regular peer (same protocol as frontends/MCP) |

### PRs that implemented this

- **#1411** — Runtime agent module (sync-only subprocess main loop)
- **#1431** — Agent as Unix socket peer with CRDT-driven execution
- **#1433** — Make agent mode unconditional
- **#1439** — Remove `is_agent_mode_enabled` wrapper
- **#1441** — Remove agent_mode toggle infrastructure
- **#1442** — Remove dead protocol variants
- **#1443** — Delete dead local kernel dispatch code (Phase A cleanup)
- **#1449** — Agent-internal kernel restart + provenance

### What's still tracked elsewhere

- **#1436** — Remaining cleanup (remove `room.kernel` Mutex, RuntimeStateDoc-driven launch serialization)
[...]

PR #1416, opened 2026-04-03
refactor(runtimed): extract transport-agnostic logic from useDaemonKernel

rgbkrk:
## Summary

Extracts transport-agnostic kernel logic from the React `useDaemonKernel` hook into the `runtimed` JS package, making it reusable from any JS runtime (WebSocket, Node, tests) without React or Tauri dependencies.

- **Derived state** (`derived-state.ts`): `KernelStatus` types/constants, `deriveKernelInfo()`, `deriveQueueState()`, `deriveEnvSyncState()`, and a `throttleBusyStatus()` RxJS operator for the 60ms busy-idle debounce
- **Typed broadcasts** (`broadcast-types.ts`): Typed interfaces and type guards for daemon broadcasts (`OutputBroadcast`, `CommBroadcast`, `KernelErrorBroadcast`, etc.), replacing the untyped `broadcasts$` observable
- **Comm diffing** (`comm-diff.ts`): Pure `diffComms()` function for detecting widget lifecycle changes (open/update/close), `detectOutputManifestHashes()` for Output widget hash detection, and `isManifestHash()`
- **Kernel commands** (`notebook-client.ts` + `request-types.ts`): `NotebookClient` class with typed methods for all kernel operations, backed by a new `sendRequest()` method on the `NotebookTransport` interface
- **SyncEngine** gains `kernelStatus$`, typed broadcast sub-observables, and `runtime_state_snapshot` wiring
- **`useDaemonKernel`** shrinks from ~889 to ~625 lines, now a thin React wrapper delegating to runtimed APIs
- **`TauriTransport`** implements `sendRequest()` by mapping typed requests to existing `*_via_daemon` Tauri commands

## Verification

- [ ] Open a notebook, launch a kernel, execute cells — kernel status, queue indicators, and execution counts work as before
- [ ] Test Output widget rendering: run a cell with `ipywidgets.Output()`, capture outputs, verify they render
- [ ] Test widget interactions (sliders, buttons) — comm messages route correctly
- [ ] Verify env sync state shows correctly when dependencies drift
- [ ] Kill the daemon — verify disconnection/reconnection handling still works

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -10,6 +10,7 @@ import {
   useRef,
   useState,
 } from "react";
+import { NotebookClient } from "runtimed";
 import { IsolationTest } from "@/components/isolated";
 import { MediaProvider } from "@/components/outputs/media-provider";
 import { setCrdtCommWriter } from "@/components/widgets/crdt-comm-writer";
@@ -61,6 +62,7 @@ import { KERNEL_STATUS } from "./lib/kernel-status";
 import { logger } from "./lib/logger";
 import { getNotebookCellsSnapshot } from "./lib/notebook-cells";
 import { useDetectRuntime } from "./lib/notebook-metadata";
+import { TauriTransport } from "./lib/tauri-transport";
 import { startWindowFocusHandler } from "./lib/window-focus";
 import type { JupyterMessage } from "./types";
 
@@ -269,7 +271,12 @@ function AppContent() {
     [handleWidgetMessage],
   );
 
-  // Clear page payload for a cell (e.g., when dismissed or re-executed)
+  // NotebookClient for sending kernel commands via transport
+  const notebookClient = useMemo(
+    () => new NotebookClient({ transport: new TauriTransport() }),
+    [],
+  );
+
   // Daemon-owned kernel execution
   const {
     kernelStatus,
@@ -286,6 +293,7 @@ function AppContent() {
     runAllCells: daemonRunAllCells,
     sendCommMessage,
   } = useDaemonKernel({
+    client: notebookClient,
     onExecutionCount: handleExecutionCount,
     onExecutionDone: handleExecutionDone,
     onUpdateDisplayData: updateOutputByDisplayId,
--- a/apps/notebook/src/hooks/useDaemonKernel.ts
+++ b/apps/notebook/src/hooks/useDaemonKernel.ts
@@ -1,50 +1,47 @@
 /**
  * Hook for daemon-owned kernel execution.
  *
- * State (kernel status, queue, env sync) is derived from the daemon's
- * RuntimeStateDoc via `useRuntimeState()`. Broadcasts are only used for
- * event callbacks (execution lifecycle, outputs, comms).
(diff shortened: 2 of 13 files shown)


rgbkrk:
## Feedback from #1413 / #1405 work (execution_count removal + runtime agents)

### Good alignment

The transport extraction is well-aligned with the runtime agent architecture (#1333). Key observations:

1. **`execution_count` reads from RuntimeStateDoc** — The `diffExecutions` path reads `t.execution_count` from the execution entry, which is the correct source of truth. In #1413, we removed the daemon's write of `execution_count` to NotebookDoc entirely. The `CellSnapshot.execution_count` will show `"null"` for newly-executed cells since nobody writes it there anymore. If any code in this PR reads `cell.execution_count` from the materialized WASM snapshot (as opposed to `runtimeState.executions[eid].execution_count`), it will get stale data after #1413 merges.

2. **Redundant broadcasts are silently ignored** — The comment block for `execution_started`, `execution_done`, `kernel_status`, `queue_changed`, `env_sync_state` matches our broadcast audit (#1409). The daemon can stop sending these — the frontend already ignores them. After #1413 merges, runtime agents won't send them at all (agents are sync-only).

3. **`NotebookClient` + `NotebookTransport`** — This abstraction is exactly what the `runtimed` JS package needs for WebSocket/non-Tauri transports. Runtime agents communicate via RuntimeStateDoc sync, so the `sendRequest()` path won't be needed for agent-backed kernels (requests go coordinator→agent, not client→agent). But the typed request API is still correct for the coordinator path.

### Potential conflicts with #1413

- **`execution_count` in `onExecutionCount` callback** — If this reads from `CellSnapshot` (NotebookDoc), it'll get `"null"` after #1413. If it reads from `runtimeState.executions[eid].execution_count`, it's fine. The `diffExecutions` path looks correct.

- **WASM `set_execution_count` removal** — #1413 removes this from the WASM API and rebuilds artifacts. If #1416 merges first, #1413 will need to rebase but shouldn't conflict (different files).

### Suggestion

Consider whether the `onExecutionCount` callback is still needed. In the new model, `execution_count` is just a field on the execution entry in RuntimeStateDoc — the `diffExecutions` effect already detects it changing. The callback may be legacy from when it was a broadcast-driven event.

rgbkrk:
Thanks for the thorough cross-PR analysis!

**Re: `execution_count` path** — Confirmed, the `onExecutionCount` callback reads from `runtimeState.executions[eid].execution_count` via `diffExecutions()` (line 282-287). It never touches `CellSnapshot` or NotebookDoc, so #1413's removal of `set_execution_count` from WASM won't affect this path.

**Re: `onExecutionCount` callback removal** — Agreed this is a good candidate for simplification in a follow-up. The callback exists because `App.tsx` calls `applyExecutionCountFromDaemon()` to update the cell store — but if `execution_count` moves entirely to RuntimeStateDoc reads, the store projection and callback could both go away. Leaving that for a follow-up after #1413 lands.

**Re: Codex's coverage gap note** — Expected. The `sendRequest()` abstraction is designed for future WebSocketTransport; today only TauriTransport exercises it. We'll get coverage when the web client work starts.

Issue #832, opened 2026-03-15
Runtime agents: decompose kernel ownership into per-runtime Automerge peers


rgbkrk:
## Summary

Today, `runtimed` is monolithic — it manages kernels, receives Jupyter messages, and proxies kernel output (cell outputs, execution counts, kernel status) into the Automerge document. All of these writes are attributed to a single `"runtimed"` actor ID, making it impossible to distinguish "runtimed set up the document structure" from "the Python runtime produced this output."

The proposal is to introduce **runtime agents** — lightweight peers that each own a single kernel and participate as first-class Automerge peers on the notebook document.

### Current

```
┌─────────────────────────────────┐
│           runtimed              │
│  actor: "runtimed"              │
│                                 │
│  ┌───────────┐ ┌───────────┐   │
│  │ kernel A  │ │ kernel B  │   │
│  └───────────┘ └───────────┘   │
│        │              │         │
│  Jupyter msgs → CRDT writes    │
│  (all attributed to "runtimed")│
└─────────────────────────────────┘
```

### Proposed

```
┌──────────────────────┐
│      runtimed        │
│  actor: "runtimed"   │
│  (coordinator/sync)  │
└──────┬───────┬───────┘
       │       │
┌──────┴──┐ ┌──┴──────┐
│ runtime │ │ runtime │
│ agent A │ │ agent B │
│ actor:  │ │ actor:  │
│ "rt:    │ │ "rt:    │
│ python: │ │ deno:   │
│ " │ │ " │
│         │ │         │
│ owns    │ │ owns    │
│ kernel  │ │ kernel  │
│ + comms │ │ + comms │
└─────────┘ └─────────┘
```

Each runtime agent:
- Is an Automerge peer with its own actor ID (e.g., `rt:python:`)
- Owns a kernel lifecycle (launch, restart, shutdown)
- Handles Jupyter protocol messages from its kernel
- Writes outputs, execution counts, and kernel status directly to the CRDT
- **Writes widget state to `doc.comms`** (with comms-in-doc, #761)
- Owns the `OutputCaptureRouter` for its kernel's Output widget routing
- Syncs with `runtimed` via the existing Automerge sync protocol

`runtimed` becomes a coordinator:
- Manages rooms and peer sync
- Spawns/supervises runtime agents
- Handles notebook-level operations (persistence, file watching, trust, autosave)
- Routes execution requests to the correct agent
- No longer directly handles kernel Jupyter messages

## State ownership with comms-in-doc

[...]

PR #1418, opened 2026-04-03
feat(runtimed): wire RUNT_AGENT_MODE into LaunchKernel

rgbkrk:
## Summary

Activates the agent subprocess path behind `RUNT_AGENT_MODE=1` env var.

When set, LaunchKernel spawns an `AgentHandle` subprocess instead of creating a local `RoomKernel`. The coordinator still handles:
- Env resolution (pool, inline deps, PEP 723, project files)
- Trust verification
- Starting phase transitions in RuntimeStateDoc

The resolved config is sent to the agent via `AgentRequest::LaunchKernel`.

Also adds `BlobStore::root()` accessor.

## Test plan

- [ ] Default (no env var): existing behavior unchanged
- [ ] `RUNT_AGENT_MODE=1`: kernel launches via agent subprocess
- [ ] Execute cells, verify outputs appear
- [ ] `cargo xtask lint` / `cargo test`

Part of #1333, #832

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -7,3 +7,4 @@ PATH_add bin
 
 export RUNTIMED_WORKSPACE_PATH="$(expand_path .)"
 export RUNTIMED_DEV=1
+export RUNT_AGENT_MODE=1
--- a/crates/notebook-doc/src/runtime_state.rs
+++ b/crates/notebook-doc/src/runtime_state.rs
@@ -354,7 +354,7 @@ impl RuntimeStateDoc {
         let trust = doc
             .put_object(&ROOT, "trust", ObjType::Map)
             .expect("scaffold trust");
-        doc.put(&trust, "status", "untrusted")
+        doc.put(&trust, "status", "no_dependencies")
             .expect("scaffold trust.status");
         doc.put(&trust, "needs_approval", false)
             .expect("scaffold trust.needs_approval");
--- a/crates/notebook-doc/src/runtime_state.rs
+++ b/crates/notebook-doc/src/runtime_state.rs
@@ -296,6 +296,77 @@ impl RuntimeStateDoc {
         Self { doc }
     }
 
+    /// Create a new `RuntimeStateDoc` with scaffolding and a custom actor.
+    ///
+    /// Used by the runtime agent to create its own doc with a unique actor
+    /// that won't conflict with the coordinator's `"runtimed:state"` actor.
+    /// The schema is identical — all scaffolding ops use the custom actor.
+    #[allow(clippy::expect_used)]
+    pub fn new_with_actor(actor_label: &str) -> Self {
+        let mut doc = AutoCommit::new();
+        doc.set_actor(ActorId::from(actor_label.as_bytes()));
+
+        // Identical scaffolding as new(), but with a different actor
+        let kernel = doc
+            .put_object(&ROOT, "kernel", ObjType::Map)
+            .expect("scaffold kernel");
+        doc.put(&kernel, "status", "not_started")
+            .expect("scaffold kernel.status");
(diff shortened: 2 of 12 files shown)


rgbkrk:
## Codex Review Findings (latest)

Three findings, all related to the dual kernel representation (`room.kernel` vs `room.agent_handle`):

**P1: GetKernelInfo/GetQueueState/ListRooms report not_started in agent mode** — these APIs read from `room.kernel` which is None when using an agent. RuntimeStateDoc has the correct state but these legacy APIs don't read it yet.

**P2: GC can evict active agent envs** — `collect_active_env_paths()` scans `room.kernel.env_path` only. Agent-backed envs aren't tracked.

**P2: has_kernel() stays true after agent kernel crash** — `agent.is_alive()` tracks the subprocess, not the kernel inside it. KernelDied notification doesn't kill the agent process.

All three are behind `RUNT_AGENT_MODE=1` — default behavior is unchanged. These should be addressed before enabling agent mode by default, likely by having the agent report kernel status via RuntimeStateDoc (which it already does) and migrating the legacy APIs to read from RuntimeStateDoc instead of `room.kernel`.

rgbkrk:
## Codex Review (final, CI green)

Four findings, all behind `RUNT_AGENT_MODE=1` — default behavior unaffected.

**P1: RuntimeStateDoc emptied before agent startup confirmed** — if agent spawn fails, the fallback path panics because `reset_starting_state()` calls `set_kernel_status()` on an empty doc. Fix: only replace state_doc after agent launch succeeds.

**P1: Agent subprocess leaks on shutdown** — `shutdown_kernel()` stops the kernel but the `runtimed agent` process stays alive. The `Child` is owned by the reader task, which holds `child.wait()` forever. Fix: signal the reader task to kill the child on shutdown.

**P2: Widget comm state not cleared on agent shutdown** — the early return skips `state_doc.clear_comms()`. Fix: add the cleanup to the agent shutdown path.

**P2: GetKernelInfo/GetQueueState report wrong data** — still read from `room.kernel` which is None in agent mode. Known gap, documented in previous review.

All four are within the feature-flagged agent path. Default behavior (no `RUNT_AGENT_MODE`) is correct and CI-green (23/23).

rgbkrk:
## Morning Status — Agent Mode Progress

### What works
- `runt agent enable` / `runt agent disable` — runtime toggle via pool IPC, no daemon restart needed
- Agent subprocess spawns, launches kernel, executes cells with variable persistence
- Initial RuntimeStateDoc sync from agent to coordinator
- Fallback to local kernel when agent spawn fails
- Default behavior (agent mode OFF) is fully working and CI-green

### What's left
- **RuntimeStateDoc replacement breaks peer connections** — replacing `state_doc` with `new_empty()` after launch crashes existing MCP/frontend sync loops. Need to either:
  - Create the empty doc at room creation time (before peers connect)
  - Or give the agent the coordinator's existing doc to merge into
- **DuplicateSeqNumber on reverse sync** — coordinator and agent both have scaffolded docs with the same actor. Fixed by new_empty() + distinct actor, but that causes the peer disconnect above.
- The core issue is the same: both coordinator and agent create `RuntimeStateDoc::new()` with the same schema, same default actor → seq number collision.

### Files changed in PR
- `agent.rs`, `agent_handle.rs` — subprocess + coordinator handle
- `notebook_sync_server.rs` — dispatch, agent_handle field, is_agent_mode_enabled  
- `main.rs` (runtimed) — agent subcommand, startup logging
- `main.rs` (runt) — agent enable/disable/status commands
- `protocol.rs`, `client.rs` — SetAgentMode request/response
- `.envrc` — RUNT_AGENT_MODE=1 for dev
- `integration.rs` — agent mode dispatch test

PR #1433, opened 2026-04-04
feat(runtimed): make agent mode unconditional

rgbkrk:
## Summary

- `is_agent_mode_enabled()` → always `true`
- Remove local kernel launch fallback from both `auto_launch_kernel` and `LaunchKernel` handler
- Agent spawn failure is now an error, not a silent fallback
- **-150 lines** of dead local kernel code removed

Every kernel now runs in a separate `runtimed agent` subprocess connected via Unix socket (from #1431).

## Test plan

- [x] `cargo test -p runtimed` — 184 lib + 22 integration tests pass
- [x] `cargo xtask lint` — all checks pass
- [x] Manual: tested on nightly with feature flag, confirmed working
- [ ] CI: E2E tests will exercise this as the only execution path

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -3412,6 +3412,18 @@ async fn auto_launch_kernel(
                         publish_kernel_state_presence(room, presence::KernelStatus::Idle, &es)
                             .await;
 
+                        // Write kernel status + info to RuntimeStateDoc so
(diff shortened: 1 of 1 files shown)


rgbkrk:
Re-triggering CI — E2E UV Prewarmed and Pyproject failures may be flaky or related to agent launch timing on Linux CI. All local tests pass.

rgbkrk:
Re-opening to re-trigger CI

Issue #1436, opened 2026-04-04
chore: remove dead local kernel code after unconditional agent mode


rgbkrk:
## Summary

After #1433 made agent mode unconditional, dead code cleanup is tracked here.

## Status

Most items are complete. Remaining work is minimal.

### Done
- [x] is_agent_mode_enabled() function — #1439
- [x] _kernel = RoomKernel::new() — #1439, #1443
- [x] Local kernel else branches in all 11 request handlers — #1443
- [x] check_and_broadcast_sync_state local kernel body — #1443
- [x] Comm forwarding spawned task — #1443
- [x] handle_sync_environment body — #1443
- [x] has_kernel() / kernel_info() local branches — #1443
- [x] Eviction kernel shutdown (replaced with agent RPC) — #1443
- [x] daemon.agent_mode AtomicBool + initialization — #1441
- [x] SetAgentMode request handler — #1441
- [x] SetAgentMode protocol variant — #1442
- [x] AgentRequest::ExecuteCell — #1442
- [x] AgentResponse::CellQueued — #1442
- [x] agent_mode in SyncedSettings + persistence + tests — #1441
- [x] agentMode in settings UI — #1435
- [x] agentMode in useSyncedSettings.ts hook — #1441
- [x] runt agent enable/disable/status CLI — #1441

### Remaining
- [ ] room.kernel field + lock in launch paths (4 lines) — used for launch serialization. Less urgent now that #1449 handles restart via agent-internal RPC. Can be replaced with RuntimeStateDoc-driven serialization in a future PR.
- [ ] "in-process prewarming" stale comments in crates/notebook/src/lib.rs (2 lines)

## References

- #1431 — Agent as Unix socket peer
- #1433 — Unconditional agent mode
- #1435, #1439, #1441, #1442, #1443 — Cleanup PRs
- #1449 — Agent-internal restart + provenance

PR #1438, opened 2026-04-04
refactor(runtimed): remove dead local kernel launch paths

rgbkrk:
## Summary

First pass of #1436 cleanup — removes dead code from the two kernel launch paths now that agent mode is unconditional.

- Delete `is_agent_mode_enabled()` function (always returned `true`)
- Remove unused `_kernel = RoomKernel::new()` in both launch handlers
- Replace `room.kernel.lock()` with `agent_request_tx` checks for "already running" guards
- Remove `drop(kernel_guard)` statements and `unreachable!()` macro
- **-70 lines**

The remaining local kernel dispatch in individual handlers (ExecuteCell, Interrupt, Shutdown, etc.) will be a follow-up.

## Test plan

- [x] `cargo test -p runtimed` — 184 lib + 22 integration tests pass
- [x] `cargo xtask lint` — all checks pass
- [ ] CI (including E2E)

Closes part of #1436.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -87,12 +87,6 @@ type AgentRequestSender = tokio::sync::mpsc::Sender<(
     tokio::sync::oneshot::Sender<notebook_protocol::protocol::AgentResponse>,
 )>;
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Scoped down to minimal changes — see replacement PR

PR #1493, opened 2026-04-05
feat(frontend): virtualized cell list with pretext height estimation

### LINKED ISSUE #1212: fix(notebook): pin viewport to cursor when output growth pushes content off-screen
## Problem

When executing a cell with Shift-Enter, the focus moves to the next cell and `scrollIntoView` fires to bring it on screen. However, if the executed cell produces a large output (e.g. a big matplotlib plot), the output renders *after* the scroll and pushes the next cell's code below the fold. The user ends up staring at the previous cell's output instead of the code they navigated to.

## Expected behavior

After Shift-Enter, the viewport should stay pinned to the editor cursor in the newly focused cell, even as outputs render above it. The code the user is about to edit should remain visible.

## Current state (as of #1448)

Scrolling is now consolidated into `focusCell()` in `useEditorRegistry.tsx` — it's called only from keyboard navigation (`onFocusPrevious`/`onFocusNext`). The old effect-based scroll with `focusSourceRef` has been removed. This means:

- `focusCell()` calls `cellElement.scrollIntoView({ block: "nearest", behavior: "smooth" })` once
- CM6's `scrollIntoView: true` dispatch also fires for cursor positioning
- Neither re-fires when output renders above the focused cell, so the cell drifts off-screen

### ~~1. CSS `overflow-anchor` (simplest)~~ — Not viable

~~Pin the scroll position to the focused cell's element using `overflow-anchor: auto`.~~

**Ruled out:** WebKit/Safari does not support `overflow-anchor`. Since we run in WKWebView (Tauri on macOS), this is a non-starter.

- MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor
- Can I Use: https://caniuse.com/css-overflow-anchor (Safari: "No")
- WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=171099 (open since 2017)

### 2. Re-scroll after output settles (time-bounded) — Recommended

After `focusCell()`, watch for DOM changes (ResizeObserver on the notebook container or the focused cell's preceding sibling) and re-fire `scrollIntoView` on the focused cell within a time window (e.g. 2-3 seconds). This handles the common case where a plot renders quickly after execution. After the window closes, stop re-scrolling so the user's manual scroll position isn't hijacked.

### 3. Scroll anchor with ResizeObserver

Pin scroll position relative to the focused cell's top offset. Use a ResizeObserver on the notebook container to detect when content above the focused cell changes height, and adjust `scrollTop` by the delta. More precise than approach 2 but more complex.

### 4. Output-aware scroll

Hook into the output materialization pipeline (`scheduleMaterialize` / manifest resolution) and re-scroll when outputs for the cell *above* the focused cell finish rendering. Most targeted but couples scroll logic to the sync pipeline.

## Recommendation

Approach 2 (time-bounded re-scroll) is the pragmatic choice: simple, handles the 90% case, and degrades gracefully. Could be implemented as a small effect in `useEditorRegistry` or `NotebookView` that watches for resize events on the focused cell's container for a few seconds after `focusCell()` fires.

## Related

- #1448 — consolidated scroll into `focusCell`, removed fragile `focusSourceRef`
- `apps/notebook/src/hooks/useEditorRegistry.tsx` — `focusCell()` is the single scroll entry point
- `apps/notebook/src/components/NotebookView.tsx` — notebook scroll container

rgbkrk:
## Summary

- **Virtualizes the notebook cell list** — only cells in/near the viewport are mounted in the DOM
- Uses [`@chenglou/pretext`](https://github.com/chenglou/pretext) for pure-arithmetic text height measurement (~0.0002ms/cell) to compute scroll positions without DOM
- Switches from `flex + CSS order` to `position: absolute` — eliminates iframe `insertBefore` risk entirely
- Binary search for visible range (adapted from pretext's markdown-chat demo)
- Self-correcting height estimates: pretext estimates on first pass, ResizeObserver measurements replace them once rendered
- Overscan of 16 cells above/below viewport to prevent flicker on scroll entry

### New files
| File | Purpose |
|------|---------|
| `cell-height-cache.ts` | Pretext-based height estimation store — `prepare()` with `pre-wrap` for code cells, MIME-type estimation for outputs, ResizeObserver self-correction |
| `useVirtualCells.ts` | Virtual window hook — binary search visible range, force-include for focused/search/executing cells, RAF-coalesced scroll |

### Key decisions
- DnD disables virtualization during drag (short-lived, safe fallback)
- Widget remount is safe (CRDT-backed state)
- Focused, executing, and search-match cells are always force-rendered
- Absolute positioning makes iframe lifecycle safer than the old flex+order approach

Closes #1212

## Test plan

- [ ] Open a notebook with 20+ cells, scroll through — no blank gaps or jumps
- [ ] Run a cell with large output while cursor is in the editor — viewport stays pinned
- [ ] Drag-and-drop a cell — still works
- [ ] Cmd+Tab away and back — no visual glitches
- [ ] Resize window — cells reflow correctly
- [ ] Keyboard nav to offscreen cell — scrolls to it
- [ ] Global find highlights in offscreen cells — scrolls to match
- [ ] Markdown cells don't flicker when scrolling into view

### DIFF
--- a/apps/notebook/package.json
+++ b/apps/notebook/package.json
@@ -9,6 +9,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@chenglou/pretext": "^0.0.4",
     "@codemirror/autocomplete": "^6.20.0",
     "@codemirror/commands": "^6.10.2",
     "@codemirror/lang-html": "^6.4.11",
--- a/apps/notebook/src/components/NotebookView.tsx
+++ b/apps/notebook/src/components/NotebookView.tsx
@@ -445,7 +445,7 @@ function NotebookViewContent({
     useVirtualCells({
       cellIds,
       scrollContainerRef: containerRef,
-      overscan: 4,
+      overscan: 16,
       isDragging: activeId !== null,
       forceInclude,
     });
@@ -840,7 +840,7 @@ function NotebookViewContent({
   return (
     <div
       ref={containerRef}
-      className="flex-1 overflow-y-auto overflow-x-clip overscroll-x-contain pl-8 pr-2"
+      className="flex-1 overflow-y-auto overflow-x-clip overscroll-x-contain pb-16 pl-8 pr-2"
       style={{ contain: "paint", overflowAnchor: "none" }}
       onScroll={onScroll}
       data-notebook-synced={!isLoading && cellIds.length > 0}
--- a/apps/notebook/src/components/NotebookView.tsx
+++ b/apps/notebook/src/components/NotebookView.tsx
@@ -27,9 +27,15 @@ import {
   EditorRegistryProvider,
   useEditorRegistry,
 } from "../hooks/useEditorRegistry";
+import { useVirtualCells } from "../hooks/useVirtualCells";
+import {
+  setMeasuredCellHeight,
+  updateHeightCache,
+} from "../lib/cell-height-cache";
 import { useFocusedCellId, useSearchCurrentMatch } from "../lib/cell-ui-state";
 import { logger } from "../lib/logger";
 import {
+  getCellById,
   getNotebookCellsSnapshot,
   useCell,
   useMaterializeVersion,
@@ -271,6 +277,7 @@ function SortableCell({
   cellId,
   nextCellId,
   index,
+  top,
   renderCell,
   onAddCell,
   onDeleteCell,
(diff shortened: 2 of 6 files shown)


rgbkrk:
Closing until later.

PR #1505, opened 2026-04-05
fix(daemon): replace pixi init with direct manifest to fix pool warming crash

rgbkrk:
## Summary

- Pixi pool warming was failing repeatedly with `exit status: 101` (Rust panic from `pixi init`) causing exponential backoff up to 300s
- Root cause: the rattler-bootstrapped pixi binary (0.65.0) panics under the daemon's subprocess environment conditions
- Fix: write `pixi.toml` manifest directly and run `pixi install` instead of `pixi init` + `pixi add`, eliminating the crash-prone `pixi init` call entirely
- Also switches subprocess from `.status()` to `.output()` so stderr is captured and logged on failure for better diagnostics

## Test plan

- [x] `cargo test -p runtimed` passes (22 tests)
- [x] `cargo test -p kernel-env` passes (16 tests)
- [x] Manual verification: writing pixi.toml + `pixi install` creates a valid env with Python at `.pixi/envs/default/bin/python`
- [ ] Deploy to nightly and verify pixi pool reaches 2/2 ready

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -2792,11 +2792,14 @@ print("warmup complete")
         self.create_conda_env().await;
     }
 
-    /// Create a pixi environment using `pixi init` + `pixi add`.
+    /// Create a pixi environment by writing a manifest and running `pixi install`.
     ///
-    /// Creates a managed pixi project directory with ipykernel and default packages,
-    /// then adds the resulting env to the pixi pool.
+    /// Instead of `pixi init` + `pixi add` (which can panic in pixi 0.65.x),
+    /// we write the `pixi.toml` manifest directly and run `pixi install`.
+    /// This is more reliable and reduces subprocess calls from two to one.
     async fn create_pixi_env(&self) {
+        use rattler_conda_types::Platform;
+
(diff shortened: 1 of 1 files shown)


rgbkrk:
## Codex Review Findings

**Blocking: Dependency spec semantics changed.**

The new `deps_toml` generation writes every package as `"{pkg}" = "*"`, which is not equivalent to `pixi add`. This breaks:
- Versioned specs: `numpy>=2` → invalid TOML key
- Channel-qualified specs: `pytorch::pytorch-cpu` → invalid TOML key
- Duplicates with built-ins like `ipykernel`

The approach (skip `pixi init`, write manifest directly) is correct, but the manifest generation needs to properly parse dependency specs into valid pixi TOML entries.

**Non-blocking suggestions:**
- Sharpen the rationale — "direct manifest + `pixi install` is a valid Pixi flow" rather than asserting a specific upstream root cause
- Richer error propagation — include exit status, fall back to stdout when stderr is empty
- Add test coverage for: plain name, versioned spec, matchspec, duplicate, install failure with empty stderr

Converting back to draft until the spec handling is fixed.

rgbkrk:
Superseded by #1509 (library-based pixi pool warming with rattler)

PR #1507, opened 2026-04-05
feat(kernel-env): replace subprocess pixi pool warming with rattler

rgbkrk:
## Summary

- Replace the fragile `pixi init` + `pixi add` subprocess approach for pixi pool warming with direct rattler API calls
- New `kernel_env::pixi` module creates pixi-compatible project directories (pixi.toml + `.pixi/envs/default/` layout) using the same rattler solve+install pipeline as our conda module
- `RattlerReporter` now accepts a configurable `env_type` label so pixi progress events are correctly labeled
- Daemon `create_pixi_env()` reduced from ~190 lines of subprocess error handling to ~50 lines

## Motivation

The subprocess approach was fragile:
- Exit code 101 panics from pixi binary
- Lost stderr on pixi failures
- Runtime dependency on pixi binary availability
- 60s + 120s subprocess timeouts with poor error reporting

## Design decision

Using `pixi_api` as a Rust library was investigated and rejected:
- `pixi_api` -> `pixi_core` pulls in ~30 pixi subcrates + the entire `uv` resolver stack
- Requires rattler 0.40.3 (we use 0.39) — version alignment would be a major undertaking
- Would roughly double compile times and dependency count

Instead, we use rattler directly (already a dependency) and generate the pixi manifest file ourselves. This gives us the same pixi-compatible directory layout with zero new dependencies.

## Test plan

- [x] `cargo check --workspace` passes
- [x] `cargo test -p kernel-env` — all 19 tests pass (including 3 new pixi manifest tests)
- [ ] Manual: start dev daemon, verify pixi pool warms successfully
- [ ] Manual: open notebook with pixi env type, verify kernel starts

### DIFF
--- a/crates/kernel-env/src/lib.rs
+++ b/crates/kernel-env/src/lib.rs
@@ -28,6 +28,8 @@
 pub mod conda;
 #[cfg(feature = "runtime")]
 pub mod gc;
+#[cfg(feature = "runtime")]
+pub mod pixi;
 pub mod progress;
 #[cfg(feature = "runtime")]
 pub mod uv;
@@ -35,6 +37,8 @@ pub mod uv;
 // Re-export key types
 #[cfg(feature = "runtime")]
 pub use conda::{CondaDependencies, CondaEnvironment};
+#[cfg(feature = "runtime")]
+pub use pixi::PixiEnvironment;
 pub use progress::{EnvProgressPhase, LogHandler, ProgressHandler};
 #[cfg(feature = "runtime")]
 pub use uv::{UvDependencies, UvEnvironment};
--- a/crates/kernel-env/src/pixi.rs
+++ b/crates/kernel-env/src/pixi.rs
@@ -1,0 +1,542 @@
+//! Pixi-compatible environment management via rattler.
+//!
+//! Creates pixi-style project directories with conda environments managed
+//! by rattler. This replaces the subprocess-based `pixi init` + `pixi add`
+//! approach with direct rattler API calls while maintaining pixi-compatible
+//! directory layout (`.pixi/envs/default/`).
+//!
+//! ## Why not use `pixi_api` directly?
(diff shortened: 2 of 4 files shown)


rgbkrk:
## Review & Next Steps

### What's here
- Replaces ~190 lines of subprocess `pixi init` + `pixi add` with ~50 lines using rattler directly
- `kernel-env/src/pixi.rs` — solve+install via rattler, generates pixi.toml manifests
- `RattlerReporter` now supports configurable env_type ("pixi" vs "conda")
- 19 tests pass, builds clean on rattler 0.39

### What's missing
This is rattler-direct, not `pixi_api`. It only does conda-forge resolution — no PyPI unification, no pixi lockfiles, no source builds. For `pixi:prewarmed` pool warming it works, but `pixi:inline` with PyPI deps would need the full pixi stack.

### Version alignment needed
- We're on rattler 0.39, pixi uses 0.40.3 (latest on crates.io)
- `rattler_conda_types`: us 0.43, pixi 0.44.3
- `rattler_solve`: us 4.2, latest 5.1.0
- Bumping to match pixi's versions is a prerequisite for using `pixi_api` as a library

### Proposed roadmap
1. **This PR**: merge as-is to fix the immediate pixi pool crash (rattler 0.39, no new deps)
2. **Follow-up**: bump rattler to 0.40.3, align all rattler sub-crate versions
3. **Then**: attempt `pixi_api` integration for full conda+PyPI unified resolution
4. **Then**: surface env type in MCP tools so agents know which dep backend to use

### Codex review findings (from earlier)
The dep spec conversion needs fixing — currently writes `"{pkg}" = "*"` which breaks versioned specs. The manifest generation should preserve spec strings properly.

rgbkrk:
Superseded by #1509 (library-based pixi pool warming with rattler)

PR #1519, opened 2026-04-05
feat(isolated-renderer): extract markdown into on-demand renderer plugin

rgbkrk:
## Summary

Introduces a **renderer plugin API** for the isolated iframe and extracts the markdown rendering stack as the first plugin. The core isolated renderer bundle drops from **~9MB to ~4.7MB**.

### Renderer Plugin API

On-demand renderer plugins are CJS modules that export an `install(ctx)` function. The iframe loads them via a custom `require` shim that provides the shared React instance through dependency injection — no window globals. This replaces the ad-hoc `window.Plotly` / `window.L` pattern with a structured contract:

- New `nteract/installRenderer` message in the frame bridge protocol
- Module-private `RendererRegistry` in the iframe (not on `window`)
- `OutputRenderer` checks the registry before the built-in MIME type switch
- `iframe-libraries.ts` distinguishes renderer plugins (`frame.installRenderer()`) from legacy eval-based libraries (`frame.eval()`)

### Markdown Extraction

The markdown stack (react-markdown, remark-gfm, remark-math, rehype-katex, rehype-raw, KaTeX, PrismLight + 24 languages) is built as a separate CJS bundle by the Vite plugin with `react` and `react/jsx-runtime` externalized. It's loaded on-demand when `text/markdown` outputs appear — both in code cell outputs (via `OutputArea`'s existing MIME scanning) and markdown cells (via updated `MarkdownCell.tsx`).

### Future

Plotly, Vega, and Leaflet can migrate from ad-hoc window globals to this same plugin API in follow-up PRs — their React wrappers would move into plugin bundles alongside the underlying libraries.

## Verification

- [ ] Open a notebook with markdown outputs — renders correctly (GFM tables, lists, links)
- [ ] Open a markdown cell — view mode renders, double-click to edit works
- [ ] Test KaTeX math rendering in markdown (e.g., `$x^2 + y^2 = z^2$`)
- [ ] Test syntax-highlighted code blocks in markdown
- [ ] Test theme switching (light/dark) with markdown content visible
- [ ] Plotly, Vega, Leaflet, and widget outputs still work (unchanged paths)

### DIFF
--- a/apps/notebook/src/components/MarkdownCell.tsx
+++ b/apps/notebook/src/components/MarkdownCell.tsx
@@ -18,6 +18,7 @@ import { remoteCursorsExtension } from "@/components/editor/remote-cursors";
 import { searchHighlight } from "@/components/editor/search-highlight";
 import { textAttributionExtension } from "@/components/editor/text-attribution";
 import { IsolatedFrame, type IsolatedFrameHandle } from "@/components/isolated";
+import { injectLibraries } from "@/components/isolated/iframe-libraries";
 import { useDarkMode } from "@/lib/dark-mode";
 import { cn } from "@/lib/utils";
 import { usePresenceContext } from "../contexts/PresenceContext";
@@ -161,6 +162,7 @@ export const MarkdownCell = memo(function MarkdownCell({
   const presence = usePresenceContext();
   const { extension: crdtBridgeExt } = useCrdtBridge(cell.id);
   const frameRef = useRef<IsolatedFrameHandle>(null);
+  const injectedLibsRef = useRef(new Set<string>());
   const viewRef = useRef<HTMLDivElement>(null);
 
   // Register EditorView with the cursor registry when in edit mode.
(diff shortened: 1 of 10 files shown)


apps/notebook/src/components/MarkdownCell.tsx
```diff
@@ -231,10 +233,16 @@ export const MarkdownCell = memo(function MarkdownCell({
   }, [cell.source]);
 
   // Render markdown content when iframe is ready
-  const handleFrameReady = useCallback(() => {
+  const handleFrameReady = useCallback(async () => {
     if (!frameRef.current || !cell.source) return;
     // Ensure theme is in sync before re-rendering (fixes theme drift after cell moves)
     frameRef.current.setTheme(darkModeRef.current);
+    // Inject markdown renderer plugin before rendering (idempotent, cached after first load)
+    await injectLibraries(
```
rgbkrk, line 243: Because `injectedLibsRef` is never cleared here, any iframe reload will skip reinstalling the markdown plugin even though the new iframe starts with an empty renderer registry. Since `OutputRenderer` no longer has a built-in `text/markdown` branch, the next render after a reload will produce a blank markdown cell. `OutputArea` already clears its injected set on each ready for exactly this case: https://github.com/nteract/desktop/blob/main/src/components/cell/OutputArea.tsx#L352-L366

PR #1521, opened 2026-04-05
feat(isolated-renderer): migrate vega to renderer plugin API

rgbkrk:
## Summary

Migrates Vega/Vega-Lite rendering from ad-hoc `window.vegaEmbed` / `window.vega` / `window.vl` globals to the renderer plugin API introduced in #1519. The vega plugin bundles `vega-embed` (which imports `vega` and `vega-lite`) directly via esbuild as a self-contained CJS module — no more eval'd UMD strings.

### Changes

- **New `vega-renderer.tsx`** — Self-contained Vega renderer plugin that imports `vegaEmbed` directly instead of reading from `window`. Uses `registerPattern` with the `isVegaMimeType` regex to handle any Vega/Vega-Lite version. Includes theme-aware re-rendering on dark/light toggle.
- **`registerPattern` API** — Extends the renderer plugin context with pattern-based MIME matching. Exact `Map.get()` lookups run first, then pattern matchers. Needed because vega MIME types are versioned (`application/vnd.vega.v5+json`, `application/vnd.vegalite.v6+json`, etc.).
- **Separate virtual modules for code splitting** — Each renderer plugin now has its own virtual module (`virtual:renderer-plugin/markdown`, `virtual:renderer-plugin/vega`) instead of sharing `virtual:isolated-renderer`. This lets Vite code-split them into independent chunks that load only when their MIME types appear.
- **`buildRendererPlugin()` helper** — Refactored the Vite plugin to deduplicate the markdown and vega build configs. Both plugins build in parallel.
- **`iframe-libraries.ts`** — Vega moved from legacy eval path to `RENDERER_PLUGINS` set.
- **`isolated-renderer/index.tsx`** — Removed `VegaOutput` and `isVegaMimeType` imports; vega MIME types now handled by the plugin registry.

### Bundle size impact

| Chunk | Before | After |
|-------|--------|-------|
| Core isolated renderer | 5,731 KB | **2,657 KB** |
| Markdown plugin | (bundled in core) | 2,209 KB (on demand) |
| Vega plugin | (3× raw UMD evals) | 865 KB (on demand) |

### What's left

Plotly and Leaflet remain on the legacy eval path. They can follow the same migration pattern in future PRs.

## Verification

- [ ] Open a notebook that produces Vega-Lite output (e.g., Altair chart) — renders correctly
- [ ] Open a notebook that produces Vega output — renders correctly
- [ ] Toggle dark/light theme with a vega chart visible — re-renders with correct theme
- [ ] Test with uncommon vega MIME versions (v3, v4, v6) — pattern matcher handles them
- [ ] Plotly, Leaflet, markdown, and widget outputs still work (unchanged)
- [ ] Verify network tab shows vega chunk loading only when a vega output appears

### DIFF
--- a/src/isolated-renderer/vega-renderer.tsx
+++ b/src/isolated-renderer/vega-renderer.tsx
@@ -16,18 +16,8 @@ function isVegaMimeType(mime: string): boolean {
   return /^application\/vnd\.vega(lite)?\.v\d/.test(mime);
 }
 
-// --- Vega MIME types to register ---
-// Cover common versions. The regex in isVegaMimeType handles all versions,
-// but we register specific ones for the plugin registry. Additional versions
-// are caught by the fallback check in install().
-const VEGA_MIME_TYPES = [
-  "application/vnd.vega.v5+json",
-  "application/vnd.vega.v5.json",
-  "application/vnd.vegalite.v4+json",
-  "application/vnd.vegalite.v4.json",
-  "application/vnd.vegalite.v5+json",
-  "application/vnd.vegalite.v5.json",
-];
+// No hardcoded MIME list — we use registerPattern with the regex matcher
+// to handle any Vega/Vega-Lite version (v1, v2, ..., v6, future versions).
 
 // --- VegaOutput component (self-contained, no window globals) ---
 
@@ -123,8 +113,13 @@ export function install(ctx: {
     mimeTypes: string[],
     component: React.ComponentType<RendererProps>,
   ) => void;
+  registerPattern: (
+    test: (mime: string) => boolean,
+    component: React.ComponentType<RendererProps>,
+  ) => void;
 }) {
-  ctx.register(VEGA_MIME_TYPES, VegaRenderer);
+  // Use pattern matcher to handle any Vega/Vega-Lite version
+  ctx.registerPattern(isVegaMimeType, VegaRenderer);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/isolated-renderer/vega-renderer.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+function isVegaMimeType(mime: string): boolean {
+  return /^application\/vnd\.vega(lite)?\.v\d/.test(mime);
+}
+
+// --- Vega MIME types to register ---
+// Cover common versions. The regex in isVegaMimeType handles all versions,
+// but we register specific ones for the plugin registry. Additional versions
+// are caught by the fallback check in install().
+const VEGA_MIME_TYPES = [
```
rgbkrk: This narrows Vega support from "any version" to six exact MIME strings. The isolated renderer only looks up plugins with `rendererRegistry.get(mimeType)` and has no regex fallback, so anything not listed here now falls through to the generic JSON/text path. That regresses variants we already prioritize elsewhere, including `application/vnd.vegalite.v6+json`, `application/vnd.vegalite.v6.json`, `application/vnd.vegalite.v3+json`, `application/vnd.vega.v6+json`, and `application/vnd.vega.v4+json`.

References:
- https://github.com/nteract/desktop/blob/660606f1c548edf51db88c7bd3e93e5c34eb69d7/src/isolated-renderer/index.tsx#L398-L403
- https://github.com/nteract/desktop/blob/660606f1c548edf51db88c7bd3e93e5c34eb69d7/src/components/outputs/vega-mime.ts#L1-L4
- https://github.com/nteract/desktop/blob/660606f1c548edf51db88c7bd3e93e5c34eb69d7/src/components/outputs/media-router.tsx#L61-L76

PR #1545, opened 2026-04-06
fix(mcp): avoid unnecessary reconnect loop on daemon PID change

### LINKED ISSUE #1522: fix(mcp): MCP server stuck in retry loop after daemon restart
## Summary

When the daemon restarts (upgrade, crash, manual restart), the long-lived MCP server process gets stuck in an unrecoverable "Daemon is restarting" retry loop with exponential backoff that grows too large to recover (500+ seconds observed). A fresh MCP server process connects immediately.

Also leaves orphaned notebook sessions (0 peers) from failed reconnection attempts.

## Reproduction

1. Connect an MCP client (Claude Code, etc.) to `runt mcp`
2. Restart the daemon (`runt daemon stop` + wait for auto-restart, or daemon upgrade)
3. Try using any MCP tool — stuck in "Daemon is restarting" loop indefinitely

## Expected Behavior

MCP server should reconnect to the new daemon within a few seconds, with bounded retry backoff.

## Environment

- Version: `2.1.0-nightly.202604052101+545d151`
- Found during nightly QA testing

## Notes

Daemon restarts are common (nightly upgrades, dev rebuilds, crashes). This is high-impact for long-running MCP sessions.

rgbkrk:
## Summary

When the daemon restarts with the same version but a different PID, the health monitor's ping succeeds (the new daemon is reachable), yet it transitions to `Reconnecting` state. This blocks all MCP tools with "Daemon is restarting" and risks cascading exponential backoff if subsequent pings fail during the brief daemon initialization window — observed growing to 500+ seconds elapsed.

Two changes:

1. **Stay `Connected` on PID change** — when a PID change is detected via a successful ping, update the daemon info and rejoin the session inline instead of entering `Reconnecting`. Tools remain unblocked throughout.

2. **Retry `auto_rejoin_session`** — the daemon may answer pings before notebook sync connections are fully ready. `auto_rejoin_session` now retries up to 3 times with 1-second delays, covering this warm-up window. This benefits both the PID-change path and the existing `Reconnecting → Connected` path.

Closes #1522

## Verification

- [ ] Start an MCP session (`runt mcp` or via Claude Code)
- [ ] Use any tool to confirm connectivity
- [ ] Restart the daemon (`runt daemon stop`, wait for auto-restart)
- [ ] Use a tool again — should succeed within ~5 seconds without "Daemon is restarting" error
- [ ] Kill the daemon process to verify genuine-outage backoff still works (tools should show "Daemon is restarting" and recover once daemon returns)

### DIFF
--- a/crates/runt-mcp/src/health.rs
+++ b/crates/runt-mcp/src/health.rs
@@ -102,18 +102,18 @@ pub async fn daemon_health_monitor(
                                 return EXIT_DAEMON_UPGRADED;
                             }
                             if current.pid != info.pid {
-                                // Same version, different PID — daemon restarted.
-                                // Transition to Reconnecting so we rejoin the notebook
-                                // session once the new daemon is fully ready.
+                                // Same version, different PID — daemon restarted but is
+                                // already reachable (this ping succeeded). Stay Connected
+                                // so tools aren't blocked, and rejoin the session inline.
                                 info!(
-                                    "Daemon restarted (same version {}, PID {} → {}), will rejoin session",
+                                    "Daemon restarted (same version {}, PID {} → {}), rejoining session",
                                     info.version, info.pid, current.pid
                                 );
-                                *state = DaemonState::Reconnecting {
-                                    since: Instant::now(),
-                                    attempt: 0,
-                                    last_info: Some(info.clone()),
+                                *state = DaemonState::Connected {
+                                    info: current.clone(),
                                 };
+                                drop(state);
+                                auto_rejoin_session(&socket_path, &session, &peer_label).await;
                             }
                         }
                     }
--- a/crates/runt-mcp/src/health.rs
+++ b/crates/runt-mcp/src/health.rs
@@ -210,7 +210,16 @@ pub async fn daemon_health_monitor(
     }
 }
 
+/// Maximum retries for auto-rejoin. The daemon may answer pings before notebook
+/// connections are fully ready, so we retry a few times with short delays.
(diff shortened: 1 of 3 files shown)


crates/runt-mcp/src/health.rs
```diff
@@ -102,18 +102,18 @@ pub async fn daemon_health_monitor(
...
                                 info!(
-                                    "Daemon restarted (same version {}, PID {} → {}), will rejoin session",
+                                    "Daemon restarted (same version {}, PID {} → {}), rejoining session",
                                     info.version, info.pid, current.pid
                                 );
-                                *state = DaemonState::Reconnecting {
-                                    since: Instant::now(),
-                                    attempt: 0,
-                                    last_info: Some(info.clone()),
+                                *state = DaemonState::Connected {
```
rgbkrk, line 112: Setting `daemon_state` back to `Connected` before the rejoin succeeds changes this path from retryable to one-shot. [`auto_rejoin_session()` clears `session` and leaves it `None` on any `connect()` error](https://github.com/nteract/desktop/blob/5e79d2105078c8df09873ac996641531b03b2870/crates/runt-mcp/src/health.rs#L231-L252), and after this hunk there is no later `Reconnecting` state to trigger another attempt. That means a daemon that answers `ping()` before notebook reconnects are ready can strand the MCP session until the user manually calls `join_notebook`, which is the same warm-up window described in the PR. I think this still needs a retry path or a delayed transition to `Connected`.

PR #1547, opened 2026-04-06
feat(wasm): compute requiredPlugins as derived data in RuntimeStateDoc

### LINKED ISSUE #1542: feat: compute requiredPlugins as derived data in runtimed/WASM
## Context

#1539 added `requiredPlugins` to `CodeCell` by computing it in the app-layer materialization functions. This works for the initial render case but has a known gap: late-arriving widget outputs (e.g., `OutputModel.outputs` gaining a plotly chart after the iframe is mounted) can miss plugin installation.

## What changed: #1558

#1558 inlined output manifests into the RuntimeStateDoc CRDT. MIME types are now readable directly from the CRDT without any blob fetch:

```
executions/{eid}/outputs/[i]:
  output_type: "display_data"
  data:
    "text/plain":      { inline: "hello" }
    "image/png":       { blob: "def456...", size: 45000 }
    "application/vnd.plotly.v1+json": { blob: "789abc...", size: 12400 }
```

Widget outputs use the same format in `comms/{comm_id}/outputs`.

## Remaining work

`requiredPlugins` is derived data from output MIME types. Now that MIME types are visible in the CRDT, the computation simplifies to scanning `output.data` keys — no blob fetches or output resolution needed.

Concretely:

1. **Wire `computeRequiredPlugins()` to read from manifest data keys** instead of scanning resolved output bundles. The WASM `get_cell_outputs()` already returns structured manifests with MIME keys visible.
2. **Move the computation out of app-layer materialization** (`materialize-cells.ts`, `frame-pipeline.ts`) into the data layer so any frontend gets it for free.
3. **Cover widget outputs** — scan `comms/{comm_id}/outputs` the same way.
4. **Remove the imperative scanning** in `OutputArea.handleFrameReady`.

### What to keep from #1539

- The shared `renderer-registry.ts` and `renderer-plugins.ts` modules
- `MediaRouter` checking the registry (so output widgets use installed plugins)
- The `preWarmPlugins()` cache warming pattern
- The `computeRequiredPlugins()` function (used by whatever layer computes it)

### What to replace

- App-layer computation in `materialize-cells.ts` and `frame-pipeline.ts`
- Widget state scanning in `OutputArea.handleFrameReady`
- The `requiredPlugins` field on `CodeCell` (it would come from the data layer instead)

## Related

- #1519 — Renderer plugin API (markdown)
- #1521 — Vega migration
- #1534 — Plotly migration
- #1537 — Leaflet migration + docs
- #1539 — Pre-compute requiredPlugins at materialization
- #1552 / #1558 — Inline manifests into RuntimeStateDoc (landed)

rgbkrk:
## Summary

Moves renderer plugin computation (`requiredPlugins`) from the app layer to the data layer by storing output MIME types in RuntimeStateDoc. The daemon extracts MIME types when writing outputs, and any consumer — WASM frontend, runt-mcp, future web viewers — can derive plugin requirements without blob store access.

- **`notebook-doc`**: New `required_plugins` module with Rust MIME→plugin mapping (`plugin_for_mime`, `compute_required_plugins`) and 8 unit tests
- **`RuntimeStateDoc`**: New `mime_types: Vec<String>` field on `ExecutionState` with `set_mime_types`, `append_mime_types`, `get_mime_types` methods
- **Daemon** (`kernel_manager`, `notebook_sync_server`): Writes MIME types alongside outputs at all `append_output` and `set_outputs` call sites
- **WASM**: New `get_cell_output_mime_types()` and `get_cell_required_plugins()` accessors; `RuntimeStateSyncApplied` includes `required_plugins_by_cell` for reactive updates
- **runt-mcp**: `cell_structured_content` and `get_all_cells` (json) include `mime_types` and `required_plugins` — enables MCP Apps (e.g. Claude Cowork) to know output types without parsing
- **Frontend**: `frame-pipeline.ts` and `materializeCellFromWasm` use WASM-provided plugins; `OutputArea` removes imperative widget output scanning (WASM now covers widget outputs via RuntimeStateDoc)

## Verification

- [ ] Open a notebook with plotly/vega/leaflet outputs — plugins load correctly, no flickers
- [ ] Execute a cell producing a widget output with plotly — plugin installs without the old widget scanning path
- [ ] `get_cell` via MCP on a cell with plotly output — response includes `mime_types` and `required_plugins`
- [ ] Verify no extra iframe re-renders when `requiredPlugins` changes (it always changes alongside `outputs`)

Closes #1542


crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -6714,6 +6717,8 @@ where
                     let synthetic_eid = uuid::Uuid::new_v4().to_string();
                     sd.create_execution(&synthetic_eid, &cell.id);
                     let _ = sd.set_outputs(&synthetic_eid, output_refs);
+                    let mime_types = output_store::extract_all_mime_types(output_refs);
```
rgbkrk: `output_refs` are manifest hashes here, but `extract_all_mime_types()` only extracts keys from raw nbformat JSON and explicitly skips hashes. That means the streaming load path will always persist `mime_types=[]`, so pre-existing Plotly/Vega/Leaflet/markdown outputs opened from disk lose their derived `requiredPlugins` and won’t preload/install the right renderer plugin.

src/components/cell/OutputArea.tsx
```diff
@@ -358,38 +357,10 @@ export function OutputArea({
     // Clear the tracking set on each call — a reloaded iframe has a fresh registry.
     injectedLibsRef.current.clear();
 
-    // Collect plugins from the cell's direct outputs
+    // Plugins are pre-computed by the WASM layer from RuntimeStateDoc MIME types,
```
rgbkrk: This drops the only code that accounted for plugin requirements inside OutputModel widget captures. get_cell_required_plugins() only reads the cell execution mime_types, but widget-captured outputs live under comm entries instead, so Plotly/Vega/etc. inside an Output widget will no longer trigger plugin installation in the iframe.

crates/runtimed/src/kernel_manager.rs
```diff
@@ -1706,6 +1706,14 @@ impl RoomKernel {
                                                 e
                                             );
                                         }
+                                        let mime_types =
+                                            output_store::extract_mime_types(&nbformat_value);
+                                        if let Err(e) = fork.append_mime_types(&eid, &mime_types) {
```
rgbkrk, line 1717: The append paths update execution.mime_types, but the sibling update_display_data path below still only replaces the stored output hash. Since requiredPlugins now comes entirely from mime_types, a display update that changes its MIME bundle will leave the derived plugin set stale and can miss plugin installation for the updated output.

PR #1559, opened 2026-04-06
feat(notebook-doc): compute requiredPlugins from CRDT output MIME keys

### LINKED ISSUE #1542: feat: compute requiredPlugins as derived data in runtimed/WASM

rgbkrk:
Compute `requiredPlugins` from CRDT output MIME keys in Rust. Every consumer gets it for free — no JS-side output scanning needed.

## What

`CellSnapshot` now includes `required_plugins: Vec<String>`, derived from inline manifest MIME type keys via `plugin_for_mime()` in `notebook-doc`. Computed once in Rust, available to WASM frontend, MCP server, and Python client.

## How it works

With #1558, output manifests are inline in the CRDT. MIME type keys are visible without any blob fetch:

```
outputs/[i]/data: { "text/plain": ..., "application/vnd.plotly.v1+json": ... }
```

`compute_required_plugins()` scans these keys through a small MIME→plugin mapping (plotly, vega, leaflet, markdown) and returns the deduplicated set.

## Frontend simplification

- `materialize-cells.ts` reads `snap.required_plugins` from the snapshot — no `computeRequiredPlugins(outputs)` calls
- `frame-pipeline.ts` uses `handle.get_cell_required_plugins(cellId)` WASM accessor
- `useAutomergeNotebook.ts` pre-warms from snapshot data before blob resolution
- Plugin pre-warming starts **before** async blob fetches, not after

## What's kept

- `pluginForMime()` in `renderer-plugins.ts` — still used by `iframe-libraries.ts`
- `computeRequiredPlugins()` in `renderer-plugins.ts` — still used by `OutputArea` for widget output scanning
- Widget scanning in `OutputArea.handleFrameReady` — separate data path (widget store, not CRDT)

Closes #1542


rgbkrk:
Closing in favor of a simpler approach: the MIME type *is* the plugin identifier — no need for an intermediate "plugin name" concept. With inline manifests (#1558), MIME keys are already right there in the CRDT. Instead of pushing the plugin-name abstraction deeper into Rust, we'll collapse it entirely on the TS side.

Plan: `.context/plans/media-plugins-simplification.md`

PR #1669, opened 2026-04-09
feat(runtimed): ephemeral notebooks for MCP agents

rgbkrk:
## Summary

Closes #1668

MCP agents create many short-lived notebooks that leave orphaned `.automerge` files on disk, causing daemon startup delays. This adds an `ephemeral` mode to `create_notebook` — ephemeral notebooks exist only in memory with no persistence overhead, promotable to file-backed via `save_notebook(path=...)`.

- **Protocol**: `ephemeral: Option<bool>` on `CreateNotebook` handshake, `ephemeral: bool` on `NotebookConnectionInfo` response
- **NotebookRoom**: `persist_tx` becomes `Option` — ephemeral rooms skip the persist debouncer and `doc.save()` serialization entirely
- **MCP default**: `create_notebook` defaults to `ephemeral=true` for agents; GUI always creates persistent notebooks
- **Save transition**: `save_notebook(path=...)` promotes ephemeral → persistent, clears metadata flag, spawns autosave + file watcher. Emergency persist on save failure for data recovery
- **Redirect map**: After rekey, peers reconnecting with the old UUID get redirected to the new path-keyed room (1h TTL)
- **RoomInfo**: `ephemeral` flag in listing so `launch_app` can warn about ephemeral rooms
- **GC**: Orphaned `.automerge` files older than 24h cleaned up in the existing GC loop
- **Tool cleanup**: Removed dead `show_notebook` backward-compat alias

## Test plan

- [x] 223 lib tests pass (`cargo test -p runtimed --lib`)
- [x] Protocol tests pass (`cargo test -p notebook-protocol`)
- [x] New tests: `test_ephemeral_room_skips_persistence`, `test_session_room_persists`, `test_ephemeral_room_has_metadata_flag`
- [x] Lint clean (`cargo xtask lint`)
- [ ] Manual: `create_notebook()` via MCP → verify no `.automerge` in `notebook-docs/`
- [ ] Manual: `save_notebook(path=...)` → verify `.ipynb` written, room re-keyed
- [ ] Manual: GUI `open_notebook` → still persistent, `.automerge` written
- [ ] Manual: GUI joins ephemeral room → ephemeral banner visible

## Follow-ups

**From code review:**
- [ ] Cap redirect map to ~1000 entries (spec mentions it, only TTL pruning implemented)
- [ ] Add comment in `rekey_ephemeral_room` explaining promoted rooms intentionally lack `.automerge` persistence
[...]

PR #1671, opened 2026-04-09
fix(runtimed): ephemeral notebook review cleanups

quillaid:
## Summary

Follow-up from #1669 code review — three small cleanups:

1. **Cap redirect map at 1000 entries** — prevents unbounded growth under extreme agent churn (thousands of save-and-disconnect cycles within an hour). Evicts oldest entries when the cap is hit.

2. **Add `delete_metadata()` to NotebookDoc** — replaces `set_metadata("ephemeral", "")` with a clean key deletion. No more empty string keys left in the CRDT.

3. **Document promoted room persistence** — clarifies that `persist_tx` remains `None` after ephemeral→persistent promotion. The `.ipynb` autosave (via file watcher) is the source of truth for file-backed notebooks, not `.automerge` persistence.

## Test plan

- [x] 224 lib tests pass (`cargo test -p runtimed --lib`)
- [x] Tokio mutex lint passes
- [x] Ephemeral room tests pass
- [x] Lint clean (`cargo xtask lint`)

### DIFF
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -1723,6 +1723,21 @@ impl NotebookDoc {
         Ok(())
     }
 
+    /// Delete a metadata key. Returns `true` if the key existed and was removed.
+    pub fn delete_metadata(&mut self, key: &str) -> Result<bool, AutomergeError> {
+        let meta_id = match self.metadata_map_id() {
+            Some(id) => id,
+            None => return Ok(false),
+        };
+        match self.doc.get(&meta_id, key)? {
+            Some(_) => {
+                self.doc.delete(&meta_id, key)?;
+                Ok(true)
+            }
+            None => Ok(false),
+        }
+    }
+
     // ── Sync protocol ───────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


rgbkrk (approved):

crates/notebook-doc/src/lib.rs
```diff
@@ -1723,6 +1723,21 @@ impl NotebookDoc {
...
+            Some(id) => id,
+            None => return Ok(false),
+        };
+        match self.doc.get(&meta_id, key)? {
+            Some(_) => {
+                self.doc.delete(&meta_id, key)?;
+                Ok(true)
+            }
+            None => Ok(false),
+        }
```
rgbkrk, line 1738: I have to wonder if there's a pattern to "delete in" with automerge

quillaid:
Re: the `delete_metadata` pattern — Automerge `doc.delete(obj, key)` errors if the key does not exist (no "delete if present" variant). The get-then-delete is necessary to return `Ok(false)` instead of propagating an error for missing keys. The error variant for "key not found" is not well-documented in automerge 0.7, so I went with the explicit check rather than matching on error types.

PR #1672, opened 2026-04-10
feat(runtimed): blob store GC for orphaned blobs

quillaid:
## Summary

Follow-up from #1669 (ephemeral notebooks). Adds mark-and-sweep garbage collection for orphaned blobs in the content-addressed blob store (`~/.cache/runt/blobs/`).

Ephemeral notebooks produce blob references (images, plots, rich outputs) that become orphaned when the room is evicted. Without GC, the blob store grows unbounded.

## How it works

Runs in the existing 6-hour GC loop (alongside env cache and notebook-docs cleanup):

1. **Mark phase:** Collect all blob hashes referenced by output manifests across ALL active rooms. Walks both execution outputs and widget comm outputs in each room's RuntimeStateDoc.

2. **Sweep phase:** List all blobs on disk, delete unreferenced ones older than 1 hour.

The 1-hour grace period prevents racing with in-flight outputs (a blob might be written to disk but not yet referenced in the CRDT manifest).

## Key design decisions

- **Scans all rooms, not just ephemeral:** Blobs are content-addressed, so the same hash could be referenced by multiple rooms. Must check ALL rooms before deleting.
- **Rooms lock dropped before state_doc reads:** Avoids holding Tokio mutex across `.await` (enforced by `tokio_mutex_lint` test).
- **Helper function `collect_blob_hashes`:** Walks manifest JSON structure — `data.{mime}.blob`, `text.blob`, `traceback.blob`.
- **Grace period uses `BlobMeta.created_at`:** The metadata sidecar stores creation time, so we don't rely on filesystem mtime.

## Test plan

- [x] 224 lib tests pass (`cargo test -p runtimed --lib`)
- [x] Tokio mutex lint passes (initially failed — fixed lock ordering)
- [x] Clippy clean (`cargo clippy -p runtimed -- -D warnings`)
- [x] Lint clean (`cargo xtask lint`)
- [ ] Manual: create ephemeral notebook, generate outputs, evict room, verify blobs cleaned after GC cycle

- Expose GC stats in `runt daemon status` (blob count, total size)
- Configurable grace period and GC interval
- On-eviction cleanup for ephemeral rooms (immediate, not waiting for 6h cycle)

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -4262,4 +4262,184 @@ mod tests {
         collect_blob_hashes_recursive(&value, &mut hashes);
         assert!(hashes.is_empty());
     }
+
+    // ── Blob GC edge case tests ────────────────────────────────────
+
+    #[test]
+    fn test_collect_blob_hashes_empty_manifest() {
+        let manifest = serde_json::json!({});
+        let mut hashes = std::collections::HashSet::new();
+        collect_blob_hashes(&manifest, &mut hashes);
+        assert!(hashes.is_empty());
+    }
+
+    #[test]
+    fn test_collect_blob_hashes_inline_only_no_blobs() {
+        // Manifests with only inline content should yield no blob hashes
+        let manifest = serde_json::json!({
+            "output_type": "display_data",
+            "data": {
+                "text/plain": {"inline": "hello world"},
+                "text/html": {"inline": "<b>hello</b>"}
+            }
+        });
+        let mut hashes = std::collections::HashSet::new();
+        collect_blob_hashes(&manifest, &mut hashes);
+        assert!(hashes.is_empty());
+    }
+
+    #[test]
+    fn test_collect_blob_hashes_mixed_inline_and_blob() {
+        // Same output with both inline and blob MIME types
+        let manifest = serde_json::json!({
+            "output_type": "execute_result",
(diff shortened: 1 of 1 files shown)


crates/runtimed/src/daemon.rs
```diff
@@ -2415,11 +2415,117 @@ impl Daemon {
...
+            // multiple rooms. We must scan ALL rooms before deleting anything.
+            {
+                let blob_max_age = std::time::Duration::from_secs(3600); // 1 hour grace period
+
+                // Collect room Arcs under the rooms lock, then drop it before
+                // any async state_doc reads (deadlock prevention).
+                let room_arcs: Vec<_> = {
+                    let rooms = self.notebook_rooms.lock().await;
+                    rooms.values().cloned().collect()
+                };
```
rgbkrk, line 2435: I feel like we need to do this in batches so we don't overload with everything all at once.

quillaid:
Good call — pushed a fix. Both phases are now batched:

- **Mark phase**: reads room state_docs in batches of 10, `yield_now()` between batches
- **Sweep phase**: processes blob deletions in batches of 50, `yield_now()` between batches

This prevents the GC from monopolizing the tokio runtime when there are many rooms or blobs.

rgbkrk:
Merged. This will be worth testing out well as a nightly.

Issue #1675, opened 2026-04-10
bug(runt-mcp-proxy): boolean and array params dropped when forwarded via mcpb proxy


quillaid:
## Problem

When using the nteract MCP server via the `mcpb-runt` proxy (e.g., `nteract-nightly`), boolean and array parameters on `create_notebook` are silently dropped. String parameters work fine.

## Reproduction

```
# Via nteract-nightly (proxy) — FAILS
create_notebook(runtime="python", ephemeral=false, dependencies=["requests"])
# Returns: ephemeral=true, dependencies=[]

# Via nteract-dev (supervisor, same code) — WORKS  
create_notebook(runtime="python", ephemeral=false, dependencies=["requests"])
# Returns: ephemeral=false, dependencies=["requests"]

# String params via nightly — WORKS
create_notebook(kernel="python")  
# Returns: info="Used 'kernel' parameter (alias for 'runtime')"
```

## Analysis

- `kernel` (string) → forwarded correctly
- `runtime` (string) → forwarded correctly
- `ephemeral` (boolean) → arrives as None, defaults to true
- `dependencies` (array) → arrives as empty/None

The MCP server reads ephemeral via `v.as_bool()` (session.rs:392). If the param arrives as the string `"false"` instead of JSON boolean `false`, `as_bool()` returns `None` and `unwrap_or(true)` kicks in.

## Hypothesis

Possible type coercion issue in how Claude Code MCP client serializes non-string parameters through the proxy. The proxy does `client.call_tool(params.clone())` — a straight pass-through. Both nteract-dev and nteract-nightly use the same proxy code (`runt-mcp-proxy`). The difference is:
- nteract-dev: supervisor wraps proxy, runs `cargo run -p mcp-supervisor`
- nteract-nightly: standalone `mcpb-runt-nightly` binary, env vars stripped via `env -u`

## Environment

- Daemon: 2.1.3+8391471 (verified correct version)
- Proxy tool cache: has correct schema including `ephemeral: boolean|null`
- Platform: Linux
- MCP client: Claude Code CLI

## Workaround

Use `add_dependency` + `sync_environment` after creating the notebook. Use `save_notebook` to persist (instead of `ephemeral=false`).

quillaid:
## Root Cause Found

[...]

PR #1676, opened 2026-04-10
fix(runt-mcp): add type coercion for boolean and array MCP params

quillaid:
## Summary

Ref #1675

Add `arg_bool()` and `arg_string_array()` helpers to `runt-mcp` that handle type coercion for MCP tool parameters. MCP clients may send boolean values as strings (`"true"`/`"false"`) and arrays as JSON-encoded strings rather than native JSON types.

## Root Cause Discovery

The original issue (#1675) where `ephemeral` and `dependencies` appeared dropped via the nightly proxy was actually caused by **the proxy caching the resolved binary path at startup**. Replacing `runt-nightly` on disk doesn't affect the running proxy — it keeps spawning the old binary. A `/mcp` reconnect (which restarts the proxy process) is required to pick up new binaries.

The type coercion fix is still valuable as **defense-in-depth** for MCP client compatibility, since different MCP transports may serialize types differently.

## Changes

- Added `arg_bool()` — reads JSON boolean or coerces `"true"`/`"false"` strings
- Added `arg_string_array()` — reads JSON array or parses JSON-encoded string arrays
- Updated `create_notebook` to use these for `ephemeral` and `dependencies`
- 7 unit tests covering native types, string coercion, empty, missing, and invalid values

## Test plan

- [x] 7 new tests pass (`cargo test -p runt-mcp --lib -- tools::tests`)
- [x] Lint clean
- [x] All existing MCP tests pass

### DIFF
--- a/crates/runt-mcp/src/tools/cell_crud.rs
+++ b/crates/runt-mcp/src/tools/cell_crud.rs
@@ -12,7 +12,7 @@ use notebook_protocol::protocol::NotebookRequest;
 use crate::execution;
 use crate::NteractMcp;
 
-use super::{arg_str, tool_error, tool_success};
+use super::{arg_bool, arg_f64, arg_str, tool_error, tool_success};
 
 #[allow(dead_code)]
 #[derive(Debug, Deserialize, JsonSchema)]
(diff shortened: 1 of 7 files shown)


quillaid:
Closing — the root cause is the proxy running stale binaries, not type coercion. The `/proc/PID/exe` investigation confirmed my proxy children are running a `(deleted)` binary with a different MD5 than what's on disk. The `and_run` issue on `replace_match` likely works fine with the current binary. Will focus on the proxy binary resolution issue instead.

PR #1677, opened 2026-04-10
fix(runt-mcp-proxy): re-resolve child binary on every restart

### LINKED ISSUE #1673: fix(runt-mcp-proxy): circuit breaker trips on daemon restart
## Problem

When the nightly daemon is restarted (binary upgrade), the `mcpb-runt-nightly` proxy's circuit breaker trips and the proxy becomes permanently unresponsive until the MCP client is manually reconnected (`/mcp`).

## Root Cause

1. Daemon stops → socket disappears
2. `runt-nightly mcp` child hangs on the broken socket (doesn't exit cleanly)
3. Proxy detects tool call failure → calls `restart_child()`
4. `state.child_client.is_none()` returns `false` (child still "alive" but unresponsive)
5. Circuit breaker counts it as a crash (`!was_dead` path)
6. 5 rapid restarts in <30s trips the breaker permanently

The breaker has no auto-reset after cooldown — once tripped, it stays tripped.

## Expected Behavior

After daemon restart, the proxy should:
1. Detect the child is unresponsive (not just dead)
2. Either skip the circuit breaker for daemon-restart scenarios, or
3. Auto-reset the breaker after a cooldown period and retry

## Possible Fixes

1. Have `runt mcp` exit cleanly when its daemon socket disappears (child exits → `was_dead = true` → breaker skipped)
2. Add a cooldown-then-retry after the breaker trips (e.g., 60s cooldown, then reset and try again)
3. Treat "child alive but tool calls failing" as equivalent to "child dead" for breaker purposes

## Reproduction

```bash
# With mcpb-runt-nightly running as MCP server:
systemctl --user stop runtimed-nightly.service
cp new-binary ~/.local/share/runt-nightly/bin/runtimed-nightly
systemctl --user start runtimed-nightly.service
# Proxy now returns "Transport closed" on all tool calls
```

## Context

Found during nightly shipping workflow. The proxy (`runt-mcp-proxy`) is designed to handle child restarts transparently, and the circuit breaker correctly skips for clean exits (`was_dead` path at proxy.rs:182). The issue is specifically with unclean child states during daemon restarts.

quillaid:
## Summary

Closes #1673

The proxy's entire purpose is upgrade transparency: the MCPB bundle ships a stable `mcpb-runt` binary while the user upgrades the nteract app (new `runt` binary). But the proxy resolved the binary path ONCE at startup and cached it in `ProxyConfig.child_command`, so upgrades were never picked up.

Confirmed via `/proc/PID/exe` — proxy children were running `(deleted)` binaries with a different MD5 than the current binary on disk.

## Fix

Replace `ProxyConfig.child_command: PathBuf` with `resolve_child_command: Box<dyn Fn() -> Result<PathBuf, String> + Send + Sync>` — a resolver closure called fresh on every `init_child()` and `restart_child()`.

Each consumer passes its own resolution strategy:
- **mcpb-runt**: `find_runt_binary()` — searches PATH + platform-specific locations
- **mcp-supervisor**: `cargo_binary()` — respects the debug/release atomic flag

This also fixes a secondary issue: after `supervisor_set_mode` switches between debug/release, the next child restart now correctly picks up the new binary path.

## Changes

- `crates/runt-mcp-proxy/src/proxy.rs` — `ProxyConfig` struct, `init_child()`, `restart_child()`, test helpers
- `crates/mcpb-runt/src/main.rs` — Pass closure; still validates at startup for clear error messages
- `crates/mcp-supervisor/src/main.rs` — Pass closure wrapping `cargo_binary()`

## Test plan

- [x] 93 proxy tests pass
- [x] All three crates compile cleanly
- [x] Lint clean
- [ ] Manual: ship nightly, verify `/proc/PID/exe` shows current binary after proxy child restart

### DIFF
--- a/crates/runt-mcp-proxy/src/proxy.rs
+++ b/crates/runt-mcp-proxy/src/proxy.rs
@@ -1011,7 +1011,7 @@ mod tests {
         );
 
         let config = ProxyConfig {
-            resolve_child_command: Box::new(|| Ok(PathBuf::from("/usr/local/bin/runt"))),
+            resolve_child_command: Box::new(|| Ok(PathBuf::from("/nonexistent/runt"))),
             child_args: vec!["mcp".to_string()],
             child_env: env,
             server_name: "nteract".to_string(),
--- a/crates/runt-mcp-proxy/src/proxy.rs
+++ b/crates/runt-mcp-proxy/src/proxy.rs
@@ -27,8 +27,11 @@ use crate::version::{self, ReconnectionEvent};
 
 /// Configuration for the MCP proxy.
 pub struct ProxyConfig {
-    /// Command to spawn the child (path to `runt` or `runt-nightly`).
-    pub child_command: PathBuf,
+    /// Resolver that returns the path to the child binary.
+    /// Called on every spawn (init and restart) so it picks up binary upgrades on disk.
+    /// This is the core reason the proxy exists: the MCPB bundle is stable while
+    /// the user upgrades the nteract app, and the proxy transparently picks up the new binary.
+    pub resolve_child_command: Box<dyn Fn() -> Result<PathBuf, String> + Send + Sync>,
     /// Arguments to the child (e.g., `["mcp"]`).
     pub child_args: Vec<String>,
     /// Environment variables for the child process.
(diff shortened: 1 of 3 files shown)


crates/runt-mcp-proxy/src/proxy.rs
```diff
@@ -990,7 +1011,7 @@ mod tests {
         );
 
         let config = ProxyConfig {
-            child_command: PathBuf::from("/usr/local/bin/runt"),
+            resolve_child_command: Box::new(|| Ok(PathBuf::from("/usr/local/bin/runt"))),
```
rgbkrk: Now that it does full resolution, this test will likely not pass on other platforms.

quillaid:
Fixed — switched to `/nonexistent/runt` to match the other test helpers. The resolver is never called in this test (it only checks env vars), but the hardcoded platform path was unnecessary.

rgbkrk (approved):

PR #1686, opened 2026-04-10
experiment(gladius): 26 tools, minimal descriptions — bytes-only test

quillaid:
Same 26 tools as baseline, descriptions trimmed to 8-15 words. Tests whether schema SIZE (not count) drives Claude Code deferral.


quillaid:
Closing — superseded by #1688 which includes the same description trimming plus restored behavioral context for replace tools and create_notebook. This was the raw Gladius experiment; #1688 is the refined version.

PR #1687, opened 2026-04-10
experiment(centurion): alwaysLoad on top 5 core tools — zero-regression test

quillaid:
Same 26 tools and verbose descriptions as baseline. Only change: add alwaysLoad meta to create_notebook, save_notebook, get_all_cells, create_cell, execute_cell. Tests whether the meta hint alone eliminates deferral.


quillaid:
Closing — A/B experiment concluded. Centurion/Hawk (alwaysLoad on top 5) helps but doesn't match description trimming: 4 replace_match + 1 set_cell on refiner (vs 9 + 0 for Cheetah). alwaysLoad alone is insufficient. Shipping #1688 instead.

PR #1699, opened 2026-04-11
feat(kernel-env): offline-first environment resolution

quillaid:
## Summary

Makes environment resolution work offline when packages are already cached. Tries local cache first (`uv --offline` and rattler `ForceCacheOnly`), falls back to network on miss.

**Impact:** Users with pre-cached packages (e.g., working in the mountains with a local model 🏔️) can create environments and sync dependencies without network access.

### Phase 1: UV --offline

Added offline-first logic to three functions in `crates/kernel-env/src/uv.rs`:
- `prepare_environment_in()` - inline deps
- `create_prewarmed_environment_in()` - pool warming
- `sync_dependencies()` - hot-sync

Pattern:
1. Clone install args, insert `--offline` after "install"
2. Run offline command
3. If success (exit 0) → emit `OfflineHit`, return early
4. If fails → log at debug, fall through to existing network install

### Phase 2: Conda/Pixi ForceCacheOnly

Created `crates/kernel-env/src/repodata.rs` with shared offline-first helper:
```rust
pub async fn query_repodata_offline_first(
    channels, platforms, specs,
    rattler_cache_dir, download_client,
    handler, env_type
) -> Result<Vec<RepoData>>
```

Logic:
1. Build Gateway with `CacheAction::ForceCacheOnly`
2. Try query
3. If success → emit `OfflineHit`, return
4. If fails → build normal Gateway with retry loop (existing behavior)

Wired into:
- `conda.rs`: `install_conda_env()`, `sync_dependencies()`
- `pixi.rs`: `install_pixi_env()`

### Phase 3: Progress Events

Added `OfflineHit` variant to `EnvProgressPhase` in `progress.rs`:
```rust
/// Environment resolved from local package cache without network access.
OfflineHit,
```

Emitted from both UV and Conda/Pixi offline paths. Updated `LogHandler` and `runtimed-py` session_core match arm.

## Verification

- ✅ `cargo xtask lint` passes
- ✅ `cargo test -p kernel-env` - all 25 tests pass
- ✅ Compiles cleanly across workspace

## Testing Plan

Manual verification (requires real network + cached packages):

1. **UV offline path:**
   ```bash
   # Pre-cache pandas
   uv pip install pandas --cache-dir ~/.cache/uv
   
   # Open notebook with pandas dep, disconnect network
   # Should resolve from cache with OfflineHit event
   ```

2. **Conda offline path:**
   ```bash
   # Pre-cache numpy via rattler
   # Open notebook with conda deps, disconnect network
   # Should resolve from cache with OfflineHit event
   ```

3. **Cache miss (expected failure):**
   ```bash
   # Open notebook with never-seen deps while offline
   # Should fail with clear "not found in cache" error
   ```

## Non-Goals (Not Doing)

- **"Clone closest env + sync delta"**: UV's `--offline --link-mode hardlink` already achieves zero-copy reuse
- **pixi_api integration**: ~30 transitive deps, not worth it for this
- **Daemon warming with network-failure detection**: Covered in Phase 4 of the plan, but deferring for now

## Backward Compatibility

Fully backward compatible:
- Online users: no behavior change (offline attempt is fast, network path unchanged)
- Offline users with cache: now works
- Offline users without cache: clear error message (same as before, just faster)

## Related

Implements offline-first approach discussed in #offline-env-resolution planning session.

Enables the "nteract in the mountains with local model" use case 🏔️

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -6,12 +6,11 @@
 
 use anyhow::{anyhow, Result};
 use log::{info, warn};
-use rattler::{default_cache_dir, install::Installer, package_cache::PackageCache};
+use rattler::{default_cache_dir, install::Installer};
 use rattler_conda_types::{
     Channel, ChannelConfig, GenericVirtualPackage, MatchSpec, ParseMatchSpecOptions, Platform,
     PrefixRecord,
 };
-use rattler_repodata_gateway::Gateway;
 use rattler_solve::{resolvo, SolverImpl, SolverTask};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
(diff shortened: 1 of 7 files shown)


quillaid:
## Code Review Complete ✅

Ran local code review via `superpowers:code-reviewer` agent.

### Critical Bug Fixed (1accda50)

**Issue:** `record_count` was counting RepoData objects instead of total package records
- **Impact:** Would show "Loaded 2 packages" instead of "Loaded 50,000 records"
- **Fix:** Sum records across all RepoData objects: `repo_data.iter().map(|r| r.len()).sum()`

### Minor Issues Addressed (1accda50)

- Added comments explaining the magic index 2 for `--offline` flag insertion in UV code

### Remaining (Non-Blocking)

**Important: Missing test coverage**
- Offline paths are untested (will rely on manual verification + monitoring in production)
- Should add integration tests in follow-up work

**Minor: Error message clarity**
- Offline repodata failures log "expected if not cached" but don't distinguish cache miss vs corruption
- Low priority, debug-level logging

### Review Assessment

**Status:** Ready to merge after critical fix ✅

The implementation is architecturally sound, follows the plan, and maintains backward compatibility. The critical bug is fixed. Missing tests are a concern but acceptable given the low-risk nature (online path unchanged, offline is opportunistic).

PR #1704, opened 2026-04-11
fix(runt): socket-aware daemon stop with signal escalation

quillaid:
## Summary

Fixes `runt daemon stop` and `runt daemon restart` to work with orphaned daemons (those not managed by systemd/launchd).

**Root cause:** Prior implementation only tried the service manager, which fails silently for manually-started or supervisor-managed daemons, leaving orphaned processes blocking the socket path.

**Solution:** Three-step hybrid stop that prioritizes graceful shutdown:

1. **Socket shutdown** (graceful, 3s timeout) — daemon stops cleanly, closes kernels/rooms
2. **Service manager** (compatibility) — preserves existing behavior for system-managed daemons  
3. **Signal escalation** (last resort) — SIGTERM → wait 5s → SIGKILL → wait 2s

### New Helper Functions

- `wait_for_pid_exit(pid, timeout_secs)` — polls `/proc/{pid}` or `kill -0` until exit or timeout
- `stop_process_by_pid(pid, timeout_secs)` — SIGTERM → wait → SIGKILL with timeout
- `cleanup_stale_daemon_info(channel)` — removes `daemon.json` and `daemon.lock` after forced stop
- `stop_daemon_smart(channel)` — orchestrates three-step hybrid stop

### Updated Commands

- `runt daemon stop` — uses `stop_daemon_smart()` instead of service manager only
- `runt daemon restart` — calls `stop_daemon_smart()` before starting

### Tests

Added 3 integration tests:

- `test_stop_orphaned_daemon` — socket-only stop (no service manager)
- `test_stop_graceful` — happy path (daemon responds to socket shutdown)
- `test_stop_signal_escalation` — SIGTERM → SIGKILL fallback

All 12 runt-cli tests pass. Manual testing confirmed socket shutdown is chosen first and daemon stops gracefully.

## Test Plan

- [x] All existing tests pass
- [x] New integration tests cover all three stop paths
- [x] Manual testing: start daemon via `cargo xtask dev-daemon`, stop via `./target/debug/runt daemon stop` → graceful socket shutdown
- [x] Manual testing: restart via `./target/debug/runt daemon restart` → stop + start with new PID

## Notes

- Socket shutdown is the **happy path** — daemon closes kernels/rooms gracefully, no data loss
- Service manager stop is **compatibility layer** for system-installed daemons
- Signal escalation is **last resort** for unresponsive daemons only
- Cleanup removes stale files to prevent "daemon running but socket path gone" confusion

### DIFF
--- a/crates/runt/src/main.rs
+++ b/crates/runt/src/main.rs
@@ -1702,6 +1702,238 @@ async fn pool_command(command: PoolCommands) -> Result<()> {
 // Daemon management commands
 // =============================================================================
 
+/// Wait for a process to exit by checking if the PID still exists.
+/// Returns true if the process exited within the timeout, false otherwise.
+#[cfg(unix)]
+async fn wait_for_pid_exit(pid: u32, timeout: Duration) -> bool {
+    use std::time::Instant;
+
+    let start = Instant::now();
+    let pid_i32 = pid as i32;
+
+    while start.elapsed() < timeout {
+        // Use kill with signal 0 to check if process exists
+        let exists = unsafe { libc::kill(pid_i32, 0) } == 0;
+        if !exists {
+            // Process doesn't exist
+            return true;
+        }
+        tokio::time::sleep(Duration::from_millis(100)).await;
+    }
+
+    false // Timeout
+}
+
+#[cfg(not(unix))]
+async fn wait_for_pid_exit(_pid: u32, _timeout: Duration) -> bool {
+    // On Windows, we'd need different process checking logic
+    // For now, just wait the full timeout
+    tokio::time::sleep(_timeout).await;
+    false
+}
+
+/// Stop a process by PID using signal escalation (SIGTERM → SIGKILL).
+/// Returns Ok(()) if the process was stopped, Err if it couldn't be stopped.
+#[cfg(unix)]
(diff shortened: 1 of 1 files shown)


quillaid:
## Code Review Updates

Addressed feedback from local code review:

### Fixed: PID Reuse Vulnerability (Critical)
**Issue:** After graceful socket shutdown, `daemon.json` could persist with stale PID if the Drop handler didn't run before CLI exit. If the PID gets reused, future `runt daemon stop` commands could signal wrong processes.

**Fix:** After socket shutdown succeeds:
1. Sleep 100ms to give Drop handler time
2. Verify `daemon.json` is gone
3. Explicitly clean up if it still exists

### Fixed: Cleanup Consistency (Important)
**Issue:** `cleanup_stale_daemon_info()` was only called on signal escalation error path, not success path. After SIGKILL, the daemon didn't run Drop, so `daemon.json`/`daemon.lock` remained stale.

**Fix:** Call `cleanup_stale_daemon_info()` after successful signal stop (not just error path).

Both fixes ensure that stale daemon info files are always cleaned up, preventing future confusion and accidental signal delivery to wrong PIDs.

Ready for @rgbkrk review.

PR #1723, opened 2026-04-12
fix(runtimed): env cleanup on room eviction and GC improvements

quillaid:
## Summary

- Eagerly clean up environment directories when notebook rooms are evicted
- Reduce GC interval from 6 hours to 30 minutes and max cache age from 7 days to 1 day
- Add GC safety net that scans for orphaned pool env directories (`runtimed-uv-*`, `runtimed-conda-*`, `runtimed-pixi-*`)
- Fix pool env leaks in `try_uv_pool_for_inline_deps` and `try_conda_pool_for_inline_deps` error paths
- Add `pool_env_root()` helper to normalise pixi's nested venv_path for correct GC and eviction

## Root cause

During sustained gremlin runs, 353+ stale env directories (11 GB) accumulated because:

1. **No cleanup on room eviction.** Pool envs are removed from the pool's VecDeque on `take()` and mutated with notebook deps. When the room is evicted, the env directory was orphaned — never returned to the pool, never cleaned up.
2. **GC too slow and too narrow.** 6-hour interval with 7-day max_age, and only scanned content-addressed (16-char hex) dirs — never touched pool-format dirs.
3. **Delta install leak.** `try_uv_pool_for_inline_deps` and `try_conda_pool_for_inline_deps` leaked taken pool envs on error/independent code paths.

### Eager cleanup on room eviction (`notebook_sync_server.rs`)
Delete the environment directory (any type) when a room is evicted after idle timeout. Uses `pool_env_root()` to normalise pixi paths so the top-level `runtimed-pixi-*` dir is deleted, not just the nested `.pixi/envs/default`.

### GC tuning (`daemon.rs`)
- Interval: 6h → 30min (faster response to env accumulation)
- Max age: 7d → 1d (content-addressed caches don't need week-long retention)

### Orphaned pool dir scanning (`daemon.rs`)
New GC pass that scans `cache_dir` for `runtimed-{uv,conda,pixi}-*` dirs not tracked by any pool and not in use by running kernels. Safety net for any envs that slip through eager cleanup.

### `pool_env_root()` normalisation (`daemon.rs`)
Pixi pool envs have nested venv_paths (`runtimed-pixi-{uuid}/.pixi/envs/default`) but GC and eviction operate on top-level dirs. The helper walks up from any venv_path to find the `runtimed-*` ancestor. Applied in GC tracked set, `collect_active_env_paths`, and room eviction.

### Delta install leak fix (`notebook_sync_server.rs`)
`remove_dir_all` on taken pool envs when delta install fails or deps are independent, in both UV and Conda paths.

## Test plan

- [x] Deployed as nightly `2.1.3+b1dfcfb`, verified under sustained gremlin workload
- [x] Pool stability verified: UV 10/10, Conda 10/10, Pixi 10/10 — disk matches in-memory
- [x] GC cycle confirmed: zero false pool deletions post-fix
- [x] Reviewed by systems-engineer-3 (pixi regression caught and fixed before PR)
- [ ] `cargo xtask lint` passes
- [ ] CI passes

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -4,7 +4,7 @@
 //! notebook windows via IPC (Unix domain sockets on Unix, named pipes on Windows).
 
 use std::collections::{HashMap, VecDeque};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::Arc;
 use std::time::Instant;
@@ -30,7 +30,10 @@ use crate::notebook_sync_server::NotebookRooms;
 use crate::protocol::{BlobRequest, BlobResponse, Request, Response};
 use crate::settings_doc::SettingsDoc;
 use crate::singleton::DaemonLock;
-use crate::{default_blob_store_dir, default_cache_dir, default_socket_path, EnvType, PooledEnv};
+use crate::{
+    default_blob_store_dir, default_cache_dir, default_socket_path, is_pool_env_dir,
+    is_within_cache_dir, pool_env_root, EnvType, PooledEnv,
+};
 use runtimed_client::singleton::DaemonInfo;
 
 /// Configuration for the pool daemon.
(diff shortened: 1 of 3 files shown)


crates/runtimed/src/daemon.rs
```diff
@@ -272,6 +272,33 @@ fn pixi_prewarmed_packages(extra: &[String]) -> Vec<String> {
...
+///
+/// Returns the path unchanged if no `runtimed-*` ancestor is found (e.g. for
+/// content-addressed envs that are already top-level).
+pub(crate) fn pool_env_root(path: &Path) -> PathBuf {
+    let mut cur = path;
+    loop {
+        if let Some(name) = cur.file_name().and_then(|n| n.to_str()) {
+            if name.starts_with("runtimed-uv-")
+                || name.starts_with("runtimed-conda-")
+                || name.starts_with("runtimed-pixi-")
```
rgbkrk: These should be known and validated constants within the runt workspace crate. Revisit that crate to see shared patterns.

crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -3352,12 +3384,14 @@ async fn try_conda_pool_for_inline_deps(
...
                         e
                     );
+                    tokio::fs::remove_dir_all(&env.venv_path).await.ok();
                     Err(())
                 }
             }
         }
         crate::inline_env::PoolDepRelation::Independent => {
             debug!("[notebook-sync] Conda pool env doesn't match inline deps, falling back");
+            tokio::fs::remove_dir_all(&env.venv_path).await.ok();
```
rgbkrk: We better gate this on the venv path definitely being defined and a known cache dir. I'd hate to see this clean up anything other than the cache dirs.

rgbkrk (changes requested):
Thanks for the contribution. There's some safety and alignment to fix up then I'd say this is good to go.

rgbkrk (approved):

PR #1732, opened 2026-04-13
fix(runtimed): guard reset_starting_state against stale agent clobbering

quillaid:
## Summary

- Guard `reset_starting_state` with expected agent ID to prevent stale timeouts from clobbering newer agents
- Clear `current_runtime_agent_id` on reset to reject zombie agent connections via provenance check
- Set `current_runtime_agent_id` in `handle_notebook_request` spawn path (was missing — pre-existing bug)

## Root cause

During rapid kernel restart cycles (observed in gremlin stress testing):

1. Agent A spawns, sets `current_runtime_agent_id = A`, starts 30s connect timer
2. Agent A is killed (gremlin SIGKILL or rapid close/reopen)
3. Retry spawns Agent B, sets `current_runtime_agent_id = B`
4. Agent A's stale 30s timer fires → `reset_starting_state` unconditionally clears `runtime_agent_handle` (dropping Agent B's handle → SIGKILL on its process group) and sets `kernel_status = "not_started"`

This caused 6 consecutive restart failures in iteration 29 — each retry spawned a new agent that got killed by the previous agent's stale timeout.

## Fix

`reset_starting_state` now accepts `Option<&str>` for the expected agent ID:
- **Post-spawn failures** pass `Some(&runtime_agent_id)` — reset is skipped if a newer agent has taken over
- **Pre-spawn failures** pass `None` — unconditional reset (safe, no agent spawned)
- On successful reset, `current_runtime_agent_id` is cleared to prevent zombie agent acceptance

Also fixes a pre-existing bug: `handle_notebook_request`'s spawn path never set `current_runtime_agent_id`, so the provenance check at connection time couldn't protect against stale agents.

## Test plan

- [x] `cargo check -p runtimed` clean
- [x] 21 integration tests pass
- [x] Tokio mutex lint passes
- [x] Reviewed by systems-engineer-3
- [x] Deployed as nightly `2.1.3+a50dd85`
- [ ] Gremlin before/after comparison (in progress)
- [ ] CI passes

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -3217,7 +3217,25 @@ fn is_untitled_notebook(notebook_id: &str) -> bool {
 
 /// Reset runtime state to "not_started" (clears any stale starting phase).
 /// Used when an early exit prevents kernel launch after status was set to "starting".
-async fn reset_starting_state(room: &NotebookRoom) {
+/// Reset kernel state after a failed launch or connect timeout.
+///
+/// When `expected_agent_id` is provided, the reset is skipped if a newer agent
+/// has already taken over `current_runtime_agent_id`. This prevents a stale
+/// timeout (from agent A) from clobbering a newer agent B that connected in the
+/// meantime.
+async fn reset_starting_state(room: &NotebookRoom, expected_agent_id: Option<&str>) {
+    // Guard: if a newer agent has replaced ours, don't clobber its state.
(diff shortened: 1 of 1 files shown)


quillaid:
Closing in favor of a CancellationToken-based approach — see #1734 for the issue describing the better fix. The agent ID guard here has a TOCTOU race (separate lock acquisitions between the check and the state mutation). The `handle_notebook_request` provenance fix (setting `current_runtime_agent_id` after spawn) will be extracted into its own PR.

PR #1736, opened 2026-04-13
fix(runt-mcp): reconnect in-place on daemon upgrade instead of exiting

quillaid:
## Summary

- `runt mcp` previously exited (code 75) when it detected a daemon version change, expecting the mcpb proxy to respawn it with the new binary
- This created a brief window where MCP clients got "Transport closed" errors, requiring manual `/mcp` reconnection
- Now reconnects to the upgraded daemon in-place and rejoins the notebook session — MCP transport stays alive, no interruption

## Why

The experiments-manager agent can't manually run `/mcp` to reconnect. When we deploy daemon upgrades during gremlin testing, the analyst's MCP connection breaks and they can't recover without human intervention.

## What changed

- `health.rs`: Both version-change detection paths (Connected and Reconnecting states) now reconnect + rejoin instead of returning `EXIT_DAEMON_UPGRADED`
- `main.rs`: Updated the health monitor `select!` branch since the monitor now runs indefinitely
- Removed `EXIT_DAEMON_UPGRADED` constant (dead code)

## Test plan

- [x] `cargo check -p runt-cli -p runt-mcp` clean
- [x] Deployed to nightly — new children spawned and serving notebooks
- [ ] CI passes
- [ ] Verify: deploy a daemon upgrade while MCP client is connected → client stays connected

### DIFF
--- a/crates/runt-mcp/src/health.rs
+++ b/crates/runt-mcp/src/health.rs
@@ -17,10 +17,6 @@ use tracing::{info, warn};
 
 use crate::session::NotebookSession;
 
-/// Exit code when the daemon has been upgraded and the MCP server should restart.
-/// EX_TEMPFAIL (sysexits.h) — "temporary failure; try again."
-pub const EXIT_DAEMON_UPGRADED: i32 = 75;
-
 /// Current connection state to the daemon.
 pub enum DaemonState {
     /// Connected and healthy.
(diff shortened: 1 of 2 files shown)


quillaid:
Wrong approach — the child should exit on version mismatch so the proxy can respawn with the new binary. The real fix is making the proxy's restart transparent to MCP clients.

PR #1746, opened 2026-04-13
fix(runtimed): gremlin-found bugs — interrupt race, conda dep loss, project-root detection

quillaid:
## Summary

Three bugs found and verified by autonomous gremlin agents during overnight stress testing:

- **Interrupt + execute race condition** — `InterruptExecution` cleared the entire execution queue, racing with concurrent `ExecuteCell` requests. Cells queued in the same window as an interrupt were silently dropped. Fix: remove queue clearing from interrupt path; SIGINT only targets the running cell, queued cells proceed normally via `execution_done()` → `process_next()`.

- **Conda dependency loss on restart** — `sync_dependencies()` built solver specs without existing packages, so the rattler Installer removed pre-installed packages (pandas, numpy, matplotlib) after `add_dependency(after="restart")`. Fix: add installed package names as preservation specs before solving.

- **Systemic project-root detection failure** — MCP `create_notebook()` didn't set `working_dir` when the caller omitted it, so untitled notebooks had no path for project file detection. `pyproject.toml`, `pixi.toml`, and `environment.yml` were never discovered. Fix: default `working_dir` to `std::env::current_dir()`.

## Found by

Autonomous gremlin overnight rotation (2026-04-13):
- Breaker gremlin, cycle 1 → interrupt race + conda dep loss
- Pixi gremlin, cycle 3 → project-root detection
- UV-data-scientist gremlin, cycle 2 → project-root detection

## Test plan

- [x] All three fixes verified by gremlin replay against nightly `2.1.3+ab12f14`
- [x] `cargo check` passes on all affected crates (`runtimed`, `kernel-env`, `runt-mcp`)
- [x] Rebased on latest main (`6586cd5f`) with no conflicts
- [ ] CI passes

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -48,7 +48,7 @@ use crate::protocol::{
 };
 use notebook_doc::diff::diff_metadata_touched;
 use notebook_doc::presence::{self, PresenceState};
-use notebook_doc::runtime_state::{QueueEntry as DocQueueEntry, RuntimeStateDoc};
+use notebook_doc::runtime_state::RuntimeStateDoc;
 
 /// Capacity for the per-room kernel broadcast channel. Sized to absorb bursts
 /// of output messages (e.g. fast-printing cells) so slower peers trigger a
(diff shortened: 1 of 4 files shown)


crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -758,31 +758,30 @@ pub(crate) async fn apply_interrupt_to_state_doc(
...
-    // Read the currently-executing entry from the CRDT (it stays — the
-    // kernel will send an execute_reply for it via the normal IOPub path).
-    let state = fork.read_state();
-    let exec = state.queue.executing.as_ref().map(|e| DocQueueEntry {
-        cell_id: e.cell_id.clone(),
-        execution_id: e.execution_id.clone(),
-    });
-
-    // Clear the queued entries, keeping only the executing one.
-    fork.set_queue(exec.as_ref(), &[]);
```
rgbkrk, line 778: Interrupt clearing the queue is by design. Let's say you have these two cells in the document:

```python
import time
time.sleep(10)
x = 10
```

```python
x + 3
```

If you run both then interrupt the first one, you shouldn't expect `x + 3` to run. If it does, it will be a `NameError`.

quillaid:
Splitting into separate PRs per review feedback. Interrupt race fix (commit 942a9f22) withdrawn — needs a more targeted approach that preserves queue clearing but avoids sweeping concurrently-queued cells.

PR #1749, opened 2026-04-13
fix(kernel-env): preserve existing conda packages during sync_dependencies

quillaid:
## Summary

`sync_dependencies()` built solver specs without including packages already installed in the pool environment. The rattler Installer treated the solved lock as authoritative and removed any package not in the spec list, causing conda environments to lose pre-installed packages (pandas, numpy, matplotlib) after `add_dependency(after="restart")`.

Fix: add all installed package names as preservation specs before solving, so the solver keeps existing packages alongside newly requested ones.

## Found by

Autonomous gremlin overnight rotation — breaker gremlin, cycle 1 (2026-04-13).

## Test plan

- [x] Verified by gremlin replay against nightly `2.1.3+ab12f14`
- [x] `cargo check -p kernel-env` passes
- [ ] CI passes

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -663,6 +663,10 @@ pub async fn sync_dependencies(env: &CondaEnvironment, deps: &CondaDependencies)
 
     let match_spec_options = ParseMatchSpecOptions::strict();
 
+    // Read currently installed packages first — we need them both for the
+    // solver's locked_packages and to build preservation specs below.
+    let installed_packages = PrefixRecord::collect_from_prefix::<PrefixRecord>(&env.env_path)?;
+
     // Always include base runtime packages — the solver only returns packages
     // needed to satisfy specs, and locked_packages are "preferred" not "required".
     // Without these, the Installer will remove ipykernel etc from the env.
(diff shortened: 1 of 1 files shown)


crates/kernel-env/src/conda.rs
```diff
@@ -673,6 +677,26 @@ pub async fn sync_dependencies(env: &CondaEnvironment, deps: &CondaDependencies)
...
+    // prewarmed pool packages (e.g. pandas, matplotlib) that the user expects
+    // to survive a `add_dependency(after="restart")` cycle.
+    let base_names: std::collections::HashSet<&str> =
+        ["ipykernel", "ipywidgets", "anywidget", "nbformat"]
+            .iter()
+            .copied()
+            .collect();
+    for record in &installed_packages {
+        let name = record.repodata_record.package_record.name.as_normalized();
+        if !base_names.contains(name) {
```
rgbkrk: This seems very crude compared to letting rattler resolve adding in the other package(s).

quillaid:
You're right — that was crude. Updated to use `pinned_packages` instead of `locked_packages` in the solver task. `pinned_packages` tells the solver these packages are required and version-locked, so they stay in the solution without needing explicit specs for each one. The diff is now just a one-field rename (locked → pinned) + removing the spec loop.

quillaid:
Closing in favor of a unified conda dependency management redesign. The pinned_packages fix here is correct (prevents solver from removing existing packages during sync_dependencies), but it needs to land alongside the env.yaml merge fix (#1751) as part of a cohesive dep provenance model.

Key learning preserved on branch `fix/conda-dep-preservation`:
- sync_dependencies must use pinned_packages (not locked_packages) to make the solver purely additive — locked_packages are "preferred" but the solver can drop them

Will consolidate with #1751 into a unified PR covering the full conda dep lifecycle.

PR #1751, opened 2026-04-13
fix(runtimed): merge env.yaml deps with CRDT deps on conda:env_yml restart

quillaid:
## Summary

`add_dependency(after="restart")` for `conda:env_yml` notebooks lost the environment.yaml baseline packages (pandas, numpy, matplotlib) on restart. The handler had either/or logic: "if CRDT has deps, use CRDT; else read env.yaml." After `add_dependency("scipy")`, CRDT contained only `["scipy"]`, so the restart built an env missing the env.yaml packages.

### Fixes

**Daemon — merge env.yaml + CRDT deps on restart** (the core fix):
- Both the `LaunchKernel` and `auto_launch` `conda:env_yml` handlers now always read `environment.yaml` as the baseline, then merge CRDT-added deps on top.
- `merge_env_yml_and_crdt_deps` deduplicates by package name using `extract_package_name`; CRDT version specs override env.yaml entries when both have the same package.
- Bootstrap `environment.yaml` deps into CRDT conda metadata on auto_launch detection, so `SyncEnvironment` drift computation works.

**Daemon — conda solver preservation**:
- `try_conda_pool_for_inline_deps` now passes the full dep list (not just the delta) to `sync_dependencies`, preventing the solver from removing unlisted packages.

**MCP — restart correctness**:
- Bootstrap prewarmed pool packages into CRDT on first `add_dependency` call.
- `confirm_sync` before restart to prevent CRDT race.
- Map prewarmed→inline `env_source` when CRDT has deps.
- Sync environment after restart as a safety net.

**MCP — dependency provenance** (`get_dependencies` enhancement):
- Returns `env_source` and classified `mode` (`"project"`, `"prewarmed"`, `"inline"`) so callers know where packages originate.
- For project-file-backed sources, includes `source_file` (e.g., `"environment.yaml"`).
- `dependencies` = what `add/remove_dependency` manages (CRDT). The project file provides the baseline; runtime deps layer on top.

### Relationship with #1749

This PR and #1749 fix different layers of the same problem:
- **#1749** (kernel-env): Solver uses `pinned_packages` to preserve existing packages during `sync_dependencies`
- **#1751** (this PR): Daemon merges env.yaml + CRDT deps when building restart envs; MCP surfaces provenance

They don't conflict and can land independently. Together they define the full conda dep lifecycle: where deps come from (#1751 provenance), how they survive restart (#1751 merge), and how the solver preserves them during sync (#1749 pinning).

## Test plan

- [x] Local manual test: `create_notebook(conda)` from CWD with `environment.yaml` → baseline OK → `add_dependency(scipy, after=restart)` → all packages present after restart
- [x] Gremlin `conda-sequential`: 5/5 phases pass
- [x] Gremlin `conda-lifecycle`: 6/6 phases pass
- [x] Gremlin `conda-conflict`: 5/5 phases pass

### DIFF
--- a/crates/runt-mcp/src/tools/deps.rs
+++ b/crates/runt-mcp/src/tools/deps.rs
@@ -444,27 +444,47 @@ pub async fn get_dependencies(
     let deps = get_deps_for_manager(&handle, &manager);
 
     // Include prewarmed packages from RuntimeStateDoc when available
-    let prewarmed = handle
-        .get_runtime_state()
-        .ok()
-        .map(|s| s.env.prewarmed_packages)
+    let state = handle.get_runtime_state().ok();
+    let prewarmed = state
+        .as_ref()
+        .map(|s| s.env.prewarmed_packages.clone())
         .unwrap_or_default();
 
-    // Indicate whether deps come from a project file or notebook metadata
-    let env_source = handle
-        .get_runtime_state()
-        .ok()
-        .map(|s| s.kernel.env_source.clone());
-    let mode = match env_source.as_deref() {
-        Some("pixi:toml") | Some("uv:pyproject") => "project",
-        _ => "inline",
+    let env_source = state
+        .as_ref()
+        .map(|s| s.kernel.env_source.clone())
+        .unwrap_or_default();
+
+    // Classify the dependency mode by env_source so the caller knows where
+    // packages originate and whether they can be add/removed via MCP.
+    //
+    // - "project": deps come from a project file (environment.yaml, pyproject.toml,
+    //   pixi.toml). The project file is the source of truth for baseline deps.
+    //   add_dependency adds ON TOP of the baseline; remove_dependency only
+    //   removes runtime-added deps.
+    // - "prewarmed": deps come from the kernel pool's default packages.
+    //   add_dependency adds on top; baseline changes with pool configuration.
+    // - "inline": deps are self-contained in notebook metadata (CRDT).
+    //   add_dependency/remove_dependency have full control.
+    let (mode, source_file) = match env_source.as_str() {
+        "conda:env_yml" => ("project", Some("environment.yaml")),
+        "uv:pyproject" => ("project", Some("pyproject.toml")),
+        "pixi:toml" => ("project", Some("pixi.toml")),
+        "conda:prewarmed" | "uv:prewarmed" | "pixi:prewarmed" => ("prewarmed", None),
+        _ => ("inline", None),
     };
 
     let mut result = serde_json::json!({
         "dependencies": deps,
         "package_manager": manager,
         "mode": mode,
+        "env_source": env_source,
     });
+
+    if let Some(file) = source_file {
+        result["source_file"] = serde_json::json!(file);
+    }
+
     if !prewarmed.is_empty() {
         result["available_packages"] = serde_json::json!(prewarmed);
     }
--- a/crates/runt-mcp/src/tools/deps.rs
+++ b/crates/runt-mcp/src/tools/deps.rs
@@ -218,6 +218,12 @@ pub async fn add_dependency(
 
     let manager = detect_package_manager(&handle);
 
+    // When adding the first dep to a prewarmed kernel, bootstrap the pool's
+    // user-default packages into the CRDT so they survive restarts/rebuilds.
+    // Without this, a restart builds a fresh env from CRDT deps alone (just the
+    // newly added package), losing pandas/numpy/etc. that were in the pool env.
+    bootstrap_prewarmed_to_crdt(&handle, &manager);
+
     add_dep_for_manager(&handle, package, &manager)
         .map_err(|e| McpError::internal_error(e, None))?;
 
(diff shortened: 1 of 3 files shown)


crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -288,6 +288,126 @@ fn extract_pyproject_deps(content: &str) -> Vec<String> {
...
+/// Merge environment.yml base deps with CRDT-added deps.
+///
+/// The env.yml deps form the baseline (e.g. pandas, numpy, matplotlib).
+/// CRDT deps are packages added at runtime via `add_dependency` (e.g. scipy).
+/// Returns the union, preferring the CRDT version spec if a package appears in both.
+fn merge_env_yml_and_crdt_deps(yml_deps: Vec<String>, crdt_deps: Vec<String>) -> Vec<String> {
+    let mut merged = yml_deps;
+    for dep in crdt_deps {
+        let pkg = conda_spec_pkg_name(&dep);
+        if !merged.iter().any(|d| conda_spec_pkg_name(d) == pkg) {
```
rgbkrk: `merge_env_yml_and_crdt_deps` says it prefers the CRDT spec when a package appears in both places, but this implementation does the opposite: if `environment.yml` already has `numpy`, a later `add_dependency("numpy>=2")` is ignored because the existing YAML entry prevents the CRDT spec from being pushed. That means version overrides or pin changes still won’t survive restart/launch for packages already present in the project file. I think this needs to replace the existing entry (or merge through a map keyed by normalized package name) so runtime edits actually win.

rgbkrk (commented):
Left two inline comments: one on `merge_env_yml_and_crdt_deps` not actually honoring CRDT overrides for existing package names, and one on the remaining portability gap from only materializing prewarmed deps into notebook metadata on first mutation.

crates/runt-mcp/src/tools/deps.rs
```diff
@@ -471,6 +528,62 @@ pub async fn sync_environment(
...
+
+/// When the kernel is prewarmed, copy the pool's user-default packages
+/// (e.g. pandas, numpy, matplotlib) into the CRDT metadata so they survive
+/// a restart that builds a fresh environment from CRDT deps alone.
+///
+/// This is a no-op when:
+/// - the kernel isn't prewarmed (deps already in CRDT from inline/project)
+/// - prewarmed_packages is empty
+/// - all prewarmed packages are base packages (ipykernel etc.)
+fn bootstrap_prewarmed_to_crdt(handle: &notebook_sync::handle::DocHandle, manager: &str) {
```
rgbkrk, line 564: This fixes restart semantics once the user mutates dependencies, but it still leaves the underlying portability problem unsolved: a notebook opened on a prewarmed env and then saved/shared without calling `add_dependency` still serializes none of the baseline packages. On another machine we’ll just re-resolve whatever that machine’s pool defaults are. I think the product model wants these prewarmed deps written into notebook metadata at env selection/launch time, not lazily on first mutation.

crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -288,6 +288,126 @@ fn extract_pyproject_deps(content: &str) -> Vec<String> {
     deps
 }
 
+/// Extract the bare package name from a conda dependency spec.
+/// e.g., "pandas>=2.0" → "pandas", "numpy" → "numpy", "python=3.12" → "python"
+fn conda_spec_pkg_name(spec: &str) -> &str {
```
rgbkrk: This looks like duplicate parsing logic. We already have `notebook_doc::metadata::extract_package_name()` for metadata dedupe, and it handles more cases than this helper (for example channel-qualified specs like `conda-forge::numpy>=1.24`). Since `runtimed` also already depends on `rattler`, another option is to use `MatchSpec`-backed normalization if we want conda to be the source of truth. Either way, I’d rather reuse one of those than add a second ad hoc package-name parser here.

rgbkrk (commented):
Left one inline follow-up on the local conda-spec parsing helper.

crates/runtimed/src/notebook_sync_server.rs
quillaid: Fixed in 8924c67 — replaced `conda_spec_pkg_name` with `notebook_doc::metadata::extract_package_name`, which handles channel qualifiers, extras, and case normalization.

crates/runtimed/src/notebook_sync_server.rs
quillaid: Fixed in 8924c67 — `merge_env_yml_and_crdt_deps` now replaces the env.yml entry when a CRDT dep has the same package name, so `add_dependency("numpy>=2")` properly overrides the env.yml's bare `numpy`. Also switched to `extract_package_name` for the name comparison.

crates/runt-mcp/src/tools/deps.rs
quillaid: Agreed — this bootstrap is a bandaid for the restart path. The right product behavior is materializing prewarmed packages into notebook metadata at env selection/launch time, so the notebook is self-contained from the start. That's a separate scope though (touches the auto_launch and create_notebook paths, plus the notebook save/export flow). I'll file a follow-up issue for it.

quillaid:
Closing in favor of a unified conda dependency management redesign. The fixes here (env.yaml + CRDT merge on restart, solver preservation, provenance in get_dependencies) are validated by all three conda gremlins, but they touch different layers of the same flow and need a cohesive design pass before landing as separate pieces.

Key learnings preserved on branch `fix/sync-after-restart`:
- conda:env_yml handler must merge env.yaml baseline with CRDT runtime deps (not either/or)
- CRDT sync race requires confirm_sync before restart
- Solver needs full dep list (not delta) to avoid removing unlisted packages
- get_dependencies needs provenance: mode, env_source, source_file

Will consolidate with #1749 (solver pinned_packages) into a unified PR that addresses the full dep lifecycle: provenance tracking, baseline vs. runtime distinction, restart behavior, and solver interaction.

PR #1759, opened 2026-04-14
feat(mcp): parquet summaries for LLM + nteract-predicate/sift-wasm crate split

rgbkrk:
## Summary

Two related changes bundled so the compute crate can be used server-side:

### 1. Split \`nteract-predicate\` into two crates
- **\`nteract-predicate\`** (rlib only) — pure Rust compute kernels (summary, filter, histogram, parquet summarization). Returns plain Rust types so it can be used server-side by repr-llm / runt-mcp without dragging in wasm-bindgen.
- **\`sift-wasm\`** (cdylib + rlib) — WASM bindings that wrap the core crate's functions with \`#[wasm_bindgen]\`. Owns the stateful \`DataStore\` (JS-lifecycle).

The WASM output filename stays \`nteract_predicate.wasm\` via \`wasm-pack --out-name\` so sift's imports and the daemon's \`/plugins/nteract-predicate.wasm\` URL don't need to change.

### 2. Parquet summaries in \`text/llm+plain\`

When a cell output includes \`application/vnd.apache.parquet\`, the daemon now resolves the bytes, reads the schema + column stats, and synthesizes a compact \`text/llm+plain\` summary so agents understand dataframe shape without rendering.

Before:
\`\`\`
Content output (application/vnd.apache.parquet, 7 KB)
http://localhost:56766/blob/...
\`\`\`

After:
\`\`\`
Parquet dataset (200 rows × 6 columns, 7.1 kB)

Columns:
  id (int64) · range 0 – 199
  name (string) · 5 distinct, top: \"Dan\" (49), \"Carol\" (47), \"Alice\" (38)
  age (int64) · range 18 – 79
  salary (float64) · range 7785.4 – 207531.89
  active (bool) · true 71% / false 29%
  department (string) · 4 distinct, top: \"Support\" (55), \"Sales\" (53), \"Marketing\" (48)
\`\`\`

Parquet bytes are resolved transiently then dropped so they don't ship through MCP.

## Test plan
- [x] All Rust tests pass (\`cargo test -p nteract-predicate -p repr-llm -p runtimed-client\`)
- [x] All JS tests pass (\`npx vp test run\`)
- [x] Sift still renders correctly in the notebook (crate split is transparent)
- [x] Agent gets rich parquet summary via MCP for a pandas.to_parquet output
- [ ] Handles edge cases: empty parquet, non-string dict columns, binary columns, date32


rgbkrk:
Addressed code review findings in the latest commit:

**Done in this PR:**
- ✅ Decimal128/Decimal256 handled
- ✅ Bounded string accumulator (10k distinct × 256 bytes/value) — prevents OOM on high-cardinality columns. The summary now signals "≥N distinct (sampled)" when the cap triggers so the LLM knows the sample is partial.
- ✅ Better type labels for Binary/List/Struct/Map so nested columns at least get named
- ✅ `Send + Sync` on the error type
- ✅ Trimmed duplicate arrow deps from sift-wasm
- ✅ Dropped dead `summarize_record_batches`
- ✅ Fixed mangled doc comment

**Follow-up worth doing (not in this PR):**

1. **Parquet footer-only fast path.** Today we do a full scan through every row group. Parquet stores per-column-chunk `Statistics` (min, max, null_count, distinct_count) in the footer. For just a schema + cheap stats, we could read only the footer via `ParquetMetaDataReader` and skip the scan entirely. Worth implementing when the daemon starts seeing large (~100MB+) parquet outputs. Inline TODO left in `summarize_parquet`'s doc.

2. **Python-side metadata stamping.** When we emit dataframes from our own Python display handler (not user-generated `to_parquet`), we could write extra metadata into the parquet KV metadata so the daemon summary can include pandas-specific context (index name, categorical levels, timezone info, etc.) without inferring.

3. **`wasm-pack` direct invocation trap.** If someone runs `wasm-pack build crates/sift-wasm` without `--out-name nteract_predicate`, they get `sift_wasm.{js,wasm}` and sift's imports break silently. Cheapest mitigation is to rename the crate to `nteract-predicate-wasm` so the default output already matches. Leaving as-is for now since xtask and CI both pass the flag.

PR #1768, opened 2026-04-14
docs: blob GC correctness + ipynb save without base64 (two specs)

rgbkrk:
Two design specs for storage correctness in the dx era. Implementation will land via separate PRs (one per spec, dispatched to worktree agents).

- `docs/superpowers/specs/2026-04-14-blob-gc-correctness-design.md` — fail-safe when zero rooms, walk persisted notebook-docs/*.automerge, extend grace from 1h to 30d.
- `docs/superpowers/specs/2026-04-14-ipynb-save-blob-refs-design.md` — large binary outputs save as `application/vnd.nteract.blob-ref+json` instead of base64; keeps small images inline for vanilla-Jupyter compat.

Spec 1 is a prerequisite for Spec 2 — once .ipynb holds refs instead of bytes, GC becomes a correctness dependency.

### DIFF
--- a/docs/superpowers/specs/2026-04-14-blob-gc-correctness-design.md
+++ b/docs/superpowers/specs/2026-04-14-blob-gc-correctness-design.md
@@ -1,0 +1,66 @@
+# Blob GC correctness
+
+**Status:** Design
+**Date:** 2026-04-14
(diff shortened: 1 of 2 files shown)


rgbkrk:
Superseded — the two spec files are carried inside the implementation PRs that reference them (#1770 for spec 1, #1769 for spec 2). Closing so we land docs + impl in one merge per spec.

PR #1771, opened 2026-04-14
feat(mcp): inline image outputs in get_cell responses

quillaid:
## Summary

- `get_cell()` now returns image outputs (png, jpeg, gif, webp) as inline `Content::image()` base64 items instead of `[N output(s) with non-text content]` text summaries
- The LLM output resolver (`resolve_display_for_llm`) now populates `blob_urls` and `blob_paths` on the Output struct so downstream formatters can resolve binary content
- New `outputs_to_content_items_with_images()` reads image blobs from disk (up to 10 MB) and base64-encodes them for MCP responses
- `get_all_cells()` unchanged — stays text-only for overview use

**Problem**: Agents calling `get_cell()` saw `[1 output(s) with non-text content (image/png)]` and had to manually curl blob URLs to see images — wasting 2+ turns per occurrence. 70% of evaluator reports cited this as friction.

**Fix**: Three-layer change:
1. `output_resolver.rs` — track blob metadata on LLM-path outputs
2. `formatting.rs` — read image blobs from disk, return as `Content::image()`
3. `cell_read.rs` — use the new function in `get_cell()`

## Test plan

- [x] All 176 existing tests pass (45 runt-mcp + 131 runtimed-client)
- [x] Manual verification: created cell with `IPython.display.Image(data=png_bytes)`, called `get_cell` — image returned inline as `Content::image()`
- [ ] Gremlin regression suite (experiments-manager)


quillaid:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Reviewed: structured.rs blob file path approach, all call sites for `cell_structured_content_from_manifests`, git blame history, previous PR comments, and code comment accuracy. Minor stale-comment observations (doc comments still say "emits blob URLs" after the image file path change) did not meet the confidence threshold.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1772, opened 2026-04-14
fix(runtimed): add wall-clock launch deadline and retry for kernel port collisions

quillaid:
## Summary

- **Adds a 45-second hard wall-clock deadline** around the shell-connect + `kernel_info_reply` exchange in `JupyterKernel::launch()`, preventing indefinite hangs when a kernel is alive but unresponsive due to partial ZMQ bind failure.
- **Adds `launch_kernel_with_retry()`** in the runtime agent with a 60-second per-attempt timeout and up to 3 retry attempts, each with fresh port allocation. Both `LaunchKernel` and `RestartKernel` handlers use this wrapper.
- **Derives `Clone`** on `KernelLaunchConfig` and `KernelSharedRefs` to support retry cloning.

## Root cause

Discovered by the gremlin suite during v2.2.0 testing: 19 concurrent pixi notebook launches caused one kernel to hang for 20+ minutes, blocking 10/19 gremlins.

Two interacting bugs:

1. **TOCTOU port race** -- `peek_ports_with_listeners()` reserves 5 ports via TCP listeners, but `drop(listeners)` releases them before the kernel process binds its ZMQ sockets. Under high concurrency, the OS can reassign a port to another process during this window.

2. **Missing timeout recovery** -- When a kernel partially binds (e.g., 4/5 ZMQ ports succeed), it stays alive but broken. The existing `select!` racing `shell.read()` (30s timeout) vs `died_rx` (process death) never fires because neither condition is met: the kernel is alive but unresponsive on the shell channel due to the stolen port.

The specific incident: port 33425 (`stdin_port`) was stolen during concurrent launch. The kernel started (4/5 ports bound) but the runtime agent hung in the `kernel_info_reply` check for 20+ minutes until the gremlin suite timed out.

## What changed

| File | Change |
|---|---|
| `crates/runtimed/src/jupyter_kernel.rs` | Wrap shell-connect + kernel_info_reply in `tokio::time::timeout(45s)`. Inner `select!` still handles fast path (kernel dies or responds within 30s); outer timeout catches alive-but-broken kernels. |
| `crates/runtimed/src/runtime_agent.rs` | New `launch_kernel_with_retry()`: 60s per-attempt timeout, 3 attempts max, 500ms backoff between retries. Fresh `KernelLaunchConfig` clone per attempt so port allocation restarts. |
| `crates/runtimed/src/kernel_connection.rs` | `#[derive(Clone)]` on `KernelLaunchConfig` and `KernelSharedRefs`. |

## Test plan

- [x] `cargo clippy -p runtimed` passes with zero warnings
- [x] Stress test: 10 concurrent pixi notebooks during pool warmup (4/10 pixi envs ready) -- 10/10 passed, 2.3s each
- [x] Stress test: 15 concurrent pixi notebooks (exceeds pool of 10) -- 15/15 passed
- [x] Stress test: 10 concurrent pixi notebooks after daemon restart during warmup -- 10/10 passed
- [x] Zero ZMQ port collisions or retry messages in daemon log across all runs (normal path unaffected)
- [ ] Full gremlin suite pass (pending)

### DIFF
--- a/crates/runtimed/src/jupyter_kernel.rs
+++ b/crates/runtimed/src/jupyter_kernel.rs
@@ -43,6 +43,84 @@ use notebook_protocol::protocol::LaunchedEnvConfig;
 type PendingCompletions =
     Arc<StdMutex<HashMap<String, oneshot::Sender<(Vec<CompletionItem>, usize, usize)>>>>;
 
+/// RAII guard for resources spawned during `JupyterKernel::launch()`.
+///
+/// Holds the kernel PID and abort handles for all spawned background tasks.
+/// If `launch()` fails — either via an explicit error return **or** because the
+/// outer timeout in `launch_kernel_with_retry` drops the future — the guard's
+/// `Drop` kills the child process and aborts every background task, preventing
+/// orphan kernels.
+///
+/// On success, call [`LaunchGuard::disarm`] so the guard's `Drop` becomes a
+/// no-op (the `JupyterKernel` struct takes ownership of the resources via its
+/// own `Drop` impl).
+struct LaunchGuard {
+    /// Kernel PID for SIGKILL on cleanup (Unix only).
+    #[cfg(unix)]
+    kernel_pid: Option<i32>,
+    /// Abort handles for spawned tasks — invoked on drop.
+    abort_handles: Vec<tokio::task::AbortHandle>,
+    /// Whether the guard has been disarmed (resources transferred to JupyterKernel).
+    disarmed: bool,
+}
+
+impl LaunchGuard {
+    #[cfg(unix)]
+    fn new(kernel_pid: Option<i32>) -> Self {
+        Self {
+            kernel_pid,
+            abort_handles: Vec::new(),
+            disarmed: false,
+        }
+    }
+
+    #[cfg(not(unix))]
+    fn new() -> Self {
+        Self {
+            abort_handles: Vec::new(),
+            disarmed: false,
+        }
+    }
+
+    /// Register a spawned task's abort handle so it will be aborted on cleanup.
+    fn track(&mut self, handle: &JoinHandle<()>) {
+        self.abort_handles.push(handle.abort_handle());
+    }
+
+    /// Disarm the guard — caller takes ownership of the resources.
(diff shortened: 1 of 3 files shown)


rgbkrk:
Looked at this from a different angle. The 45s deadline + 60s retry + zombie SIGKILL machinery all exist because we're doing port reservation the wrong way: bind 5 TCP listeners, read the port numbers, drop the listeners, then hope the kernel binds before anything else grabs those ports. That's an inherent TOCTOU race — we can make it rarer with retries and tighter timeouts, but we can't close it while we're pre-reserving ports from a different process than the one that ultimately binds.

The deeper fix is to stop using TCP for daemon↔kernel transport. Jupyter's connection file format natively supports `"transport": "ipc"` with filesystem socket paths. We'd:

- `mkdir -p ~/.cache/runt/kernels/{kernel_id}/` with 0700
- Fill `ConnectionInfo` with `transport: IPC`, `ip: <that dir>`, and `stdin_port=1, control_port=2, …` as socket-path suffixes
- Spawn the kernel — it binds to filesystem sockets we own, no OS port allocation happens at all
- `rm -rf` the dir on teardown

No race window, no retries, no deadlines, no zombie sweep. `peek_ports_with_listeners` goes away. The `launch_kernel_with_retry` wrapper, `LaunchGuard`, and outer 45s timeout from this PR become unnecessary. **We'd delete more code than we'd add.**

Side wins: permission-isolated via fs perms, no firewall surface (not listening on 127.0.0.1), slightly faster than TCP over loopback, and the daemon↔kernel link is always colocated so we lose nothing by dropping cross-machine support.

Support caveat: PyZMQ has supported IPC everywhere for years on Unix; Windows Named Pipes support landed in PyZMQ 22.x (2021). If our bundled Windows build has any issue, the fallback is option 2: write the connection file with `port=0` and let the kernel rewrite it after binding (standard Jupyter idiom, adds a "wait for file rewrite" step but still TOCTOU-free).

Closing this PR in favor of pursuing the IPC switch. The kernel-hang bug it addresses is real; the fix framing just isn't deep enough.

PR #1794, opened 2026-04-15
ci(xtask): run `ty check python/` as part of `cargo xtask lint`

rgbkrk:
## Summary

Closes task #13. The python-package CI workflow already runs `uv run ty check python/`, but `cargo xtask lint` only ran ruff — so local lint could pass while CI failed on type errors.

Adds a `=== Python (ty) ===` step right after ruff, scoped to `python/` (matches the CI `Type check` step exactly). `scripts/` is intentionally excluded because it carries 206 historical ty diagnostics; expand the scope when those are cleaned up.

CLAUDE.md's description of what `cargo xtask lint` does is updated to match.

## Test plan

- [x] `cargo xtask lint` runs ty and passes on a clean tree
- [x] `cargo xtask lint --fix` runs ty and passes
- [x] New `=== Python (ty) ===` section appears in the output
- [x] `codex review --base main` — no issues found

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -210,7 +210,7 @@ Run this command before every commit. CI will reject PRs that fail formatting ch
 cargo xtask lint --fix
 ```
 
-This formats Rust, lints/formats TypeScript/JavaScript with Biome, and lints/formats Python with ruff.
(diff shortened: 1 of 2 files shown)


rgbkrk:
Superseded by #1783 which shipped the same `ty check` addition to `cargo xtask lint`. The only remaining delta was a one-line AGENTS.md doc update — not worth a dedicated PR.

PR #1809, opened 2026-04-15
feat(dx): richer per-column stats + Dataset handler in text/llm+plain

quillaid:
## Summary

Implements the enrichment plan from #1807:

- **Per-column stats** in `summarize_dataframe`: numeric range (min–max), string distinct/top-3, null counts with percentages, temporal range, all-null detection — matching the shape of `repr-llm::summarize_parquet` so dx-provided summaries carry equivalent information density
- **Head truncation**: manual per-cell truncation (40 chars in head preview, 80 chars in stats) keeps summaries under 1 KB for inline CRDT storage, preventing the blob-ref resolution failures that caused empty `{"data": {}}` MCP outputs for text-heavy DataFrames
- **Wide DataFrame capping**: columns beyond 40 are elided with `…[+N more columns]`
- **`datasets.Dataset` handler**: metadata-only summary (features, num_rows, single-row peek via `ds[0]`) — no `.to_pandas()`, preserves lazy Arrow-backed design
- **Polars parity**: dedicated `_polars_column_stats` extractor covering numeric, temporal (Date/Datetime/Duration), string/categorical, and all-null columns

Works in tandem with #1808 (merged) which gives author-provided `text/llm+plain` priority over Rust-side synthesis in the output resolver.

## Files changed

- `python/dx/src/dx/_summary.py` — core rewrite with `_truncate_cell`, per-column stat extractors, `summarize_dataframe`, `summarize_dataset`
- `python/dx/src/dx/_format_install.py` — `datasets.Dataset` handler registration in `install_formatters()`
- `python/dx/tests/test_summary_unit.py` — 20 new unit tests (truncation, numeric/string/temporal/null stats, wide DataFrames, polars parity, Dataset summary)
- `python/dx/tests/test_dx_integration.py` — 6 new integration tests (1 KB size guard, truncation markers, Dataset mimebundle/display)

## Test plan

- [x] `pytest python/dx/tests/test_summary_unit.py` — 20 tests pass (4 skip without `datasets`)
- [x] `pytest python/dx/tests/test_dx_integration.py` — 6 tests pass (4 skip without `datasets`)
- [ ] Gremlin validation against nightly with dx wheel installed (in progress)

### DIFF
--- a/python/dx/src/dx/_format_install.py
+++ b/python/dx/src/dx/_format_install.py
@@ -383,8 +383,9 @@ def _emit_dataframe(df: Any, *, total_rows: int) -> dict | None:
     as a fallback bundle for hosts that don't understand the ref MIME;
     nteract frontends pick the parquet renderer via the ref MIME.
 
-    Returns ``None`` when serialization fails — lets IPython's default
-    chain render unchanged.
+    Returns ``None`` only when both parquet serialization and summary
+    generation fail. When parquet fails but the summary succeeds
+    (e.g. pyarrow missing), returns a summary-only bundle.
     """
     try:
         data, content_type = serialize_dataframe(df, max_bytes=_MAX_PAYLOAD_BYTES)
--- a/python/dx/src/dx/_format_install.py
+++ b/python/dx/src/dx/_format_install.py
@@ -55,7 +55,7 @@ from typing import Any
 from dx._env import Environment, detect_environment
 from dx._format import serialize_dataframe
 from dx._refs import BLOB_REF_MIME, BlobRef, build_ref_bundle
-from dx._summary import summarize_dataframe
+from dx._summary import summarize_dataframe, summarize_dataset
 
 log = logging.getLogger("dx")
 
@@ -153,6 +153,14 @@ def install_formatters() -> None:
         ipython_display.for_type(nw.DataFrame, _narwhals_ipython_display)
     except ImportError:
         pass
+
+    try:
+        import datasets  # noqa: PLC0415  # ty: ignore[unresolved-import]
+
+        mimebundle.for_type(datasets.Dataset, _dataset_mimebundle)
+        ipython_display.for_type(datasets.Dataset, _dataset_ipython_display)
+    except ImportError:
+        log.debug("dx: datasets not installed, skipping handler")
 
     _install_display_pub_hook()
     _enable_third_party_nteract_renderers()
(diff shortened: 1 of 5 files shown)


python/dx/src/dx/_summary.py
```diff
@@ -9,34 +9,272 @@
...
     if mod in ("pandas", "polars"):
         return mod
     return "unknown"
 
 
-def _pandas_dtypes(df: Any) -> list[tuple[str, str]]:
-    return [(str(col), str(df[col].dtype)) for col in df.columns]
+# ── Per-column stat extractors ──────────────────────────────────────
+
+
```
rgbkrk, line 42: All this makes me wonder if we should make narwhals a dep.

quillaid: Good instinct — I just tested it and narwhals has everything we need: `min()`, `max()`, `n_unique()`, `value_counts(sort=True)`, `null_count()`, `dtype.is_numeric()`, `dtype.is_temporal()`, `dtype == nw.String`. The API surface maps 1:1 to what both extractors do today.

The win: collapse `_pandas_column_stats` + `_polars_column_stats` (~120 lines) into a single `_column_stats` (~60 lines) that works on any narwhals-compatible backend (pandas, polars, modin, cudf, pyarrow tables). dx already depends on narwhals for the `_format_install.py` display handler, so it's not a new dep — just promoting it from "optional formatter target" to "stat extraction backbone."

I'd do this as a follow-up rather than in this PR:
- This PR is validated by gremlin and ready to ship as-is
- The narwhals refactor touches the same code paths and would need its own test cycle
- We'd also want to think about whether `summarize_dataframe` should accept `nw.DataFrame` directly (right now it sniffs flavor via `type(df).__module__`)

Happy to open the follow-up immediately after this lands.

python/dx/src/dx/_format_install.py
```diff
@@ -154,6 +154,14 @@ def install_formatters() -> None:
     except ImportError:
         pass
 
+    try:
+        import datasets  # noqa: PLC0415  # ty: ignore[unresolved-import]
```
rgbkrk, line 158: `ty: ignore[unresolved-import]` is unused — `datasets` is in the workspace dev deps so ty resolves it cleanly. `cargo xtask lint` (which runs `ty check python/` per #1794) flags it as a warning. `cargo xtask lint --fix` strips it. Same pattern at 8 other sites across this PR.

python/dx/src/dx/_summary.py
```diff
@@ -9,34 +9,272 @@
...
+    if len(s) <= max_chars:
+        return s
+    keep = max_chars - 1  # room for the ellipsis char
+    suffix = f"…[+{len(s) - keep} chars]"
+    # Adjust: we need the prefix + suffix to fit reasonably
+    prefix_len = max_chars - len(suffix)
+    if prefix_len < 1:
+        prefix_len = 1
+        suffix = f"…[+{len(s) - prefix_len} chars]"
+    return s[:prefix_len] + suffix
```
rgbkrk, line 30: Small off-by-N in the `+N chars` count. `keep = max_chars - 1` builds the suffix, so `+N` claims `len(s) - (max_chars - 1)` chars were dropped. But the actual prefix length below is `max_chars - len(suffix)`, which is smaller — so more chars get dropped than the suffix claims. Example: `s = "x"*500`, `max_chars=80` → suffix says `+421 chars` but 433 were actually dropped.

Cosmetic — truncation marker still renders correctly. One-line fix:

```python
# Upper-bound suffix length to size the prefix, then compute the true suffix.
suffix_upper = f"…[+{len(s)} chars]"
prefix_len = max(max_chars - len(suffix_upper), 1)
suffix = f"…[+{len(s) - prefix_len} chars]"
return s[:prefix_len] + suffix
```

rgbkrk (commented):
Nice work — matches the plan, tests are thorough, and the pyarrow-missing fallback that still emits `text/llm+plain` is a quietly great addition. Ran locally: 31/31 passing.

Three small cleanups before merge, none blocking:

1. **Stale `ty: ignore[unresolved-import]` directives** — see inline comment. 9 total across the PR; `cargo xtask lint --fix` handles them.
2. **Stale docstring in `_emit_dataframe`** (around line 387) — still reads "Returns `None` when serialization fails — lets IPython's default chain render unchanged," but the new fallback (good change!) now returns `{text/llm+plain: summary}`. Worth updating to "Returns `None` only when both parquet serialization and summary generation fail."
3. **Off-by-N in `_truncate_cell` truncation count** — see inline comment.

python/dx/src/dx/_summary.py
quillaid: Fixed in 6e14f06e — using the upper-bound suffix approach you suggested. The count is now accurate.

python/dx/src/dx/_format_install.py
quillaid: Checked — `datasets` is actually not in the workspace dev deps, so ty can't resolve it. The `ty: ignore[unresolved-import]` directives are needed. `cargo xtask lint` fails with 9 `unresolved-import` diagnostics without them. Left them in place.

Updated the `_emit_dataframe` docstring in the same commit (6e14f06e).

PR #1810, opened 2026-04-15
fix(runtimed): check env sync drift after kernel restart

quillaid:
## Summary

- After `restart_kernel`, the daemon was unconditionally marking the environment as in-sync (`set_env_sync(true, ...)`) without checking whether the CRDT metadata had deps not yet installed in the venv
- Packages added via `add_dependency` silently disappeared after restart — the venv persisted but the hot-install drift watcher never fired because it was told everything was already synced
- Now both restart paths (in-place restart via runtime agent and fallback new-agent spawn) run `compute_env_sync_diff` against the current CRDT metadata, correctly marking out-of-sync when drift exists so `SyncEnvironment` hot-installs the missing packages

Found by gremlin validation during dx enrichment work (PR #1809) — the gremlin lost ~40 turns re-adding deps after each kernel restart.

## Test plan

- [x] `cargo test -p runtimed -- compute_env_sync_diff` — all 7 tests pass
- [x] `cargo test -p runtimed --test tokio_mutex_lint` — no locks held across await
- [ ] Gremlin replay to confirm deps persist across restart

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -5681,13 +5681,11 @@ async fn handle_notebook_request(
                                 }
                             }
 
-                            // Fresh kernel is in sync with its launched config
-                            {
-                                let mut sd = room.state_doc.write().await;
-                                if sd.set_env_sync(true, &[], &[], false, false) {
-                                    let _ = room.state_changed_tx.send(());
-                                }
-                            }
+                            // Compute env sync state against the freshly
+                            // stored launched_config (updated above).
+                            // Covers both inline-dep drift and the
+                            // prewarmed-with-added-inline-deps case.
+                            check_and_broadcast_sync_state(room).await;
 
                             return NotebookResponse::KernelLaunched {
                                 kernel_type: resolved_kernel_type,
@@ -5845,13 +5843,9 @@ async fn handle_notebook_request(
                                     }
                                 }
 
-                                // Fresh kernel is in sync with its launched config
-                                {
-                                    let mut sd = room.state_doc.write().await;
-                                    if sd.set_env_sync(true, &[], &[], false, false) {
-                                        let _ = room.state_changed_tx.send(());
-                                    }
-                                }
+                                // Compute env sync state against the freshly
+                                // stored launched_config (updated above).
+                                check_and_broadcast_sync_state(room).await;
 
                                 NotebookResponse::KernelLaunched {
                                     kernel_type: resolved_kernel_type,


crates/runtimed/src/notebook_sync_server.rs
```diff
@@ -5681,12 +5681,40 @@ async fn handle_notebook_request(
...
+                                        ),
+                                        None => sd.set_env_sync(true, &[], &[], false, false),
+                                    };
+                                    if changed {
+                                        let _ = room.state_changed_tx.send(());
+                                    }
                                 }
+                                let _ = room
+                                    .kernel_broadcast_tx
+                                    .send(NotebookBroadcast::EnvSyncState { in_sync, diff });
```
rgbkrk: This block and the sibling at line 5906 are a subset of the existing `check_and_broadcast_sync_state(room)` helper at line 691. The helper also handles the prewarmed-kernel-with-inline-deps-added case, which `compute_env_sync_diff` alone doesn't detect (see the top-level review comment).

`runtime_agent_launched_config` is already written at line 5661 just above, so calling the helper here picks up the right state.

rgbkrk (changes requested):
Good catch on the bug — the `set_env_sync(true, ...)` on restart was definitely wrong. But there's an existing helper, `check_and_broadcast_sync_state(room)` at `notebook_sync_server.rs:691`, that already does this exact computation, **including the prewarmed-with-inline-deps-added case** that this PR doesn't cover.

## The gap

`compute_env_sync_diff` only reports drift for deps the **launched** config was already tracking (`launched.uv_deps.is_some()`, etc.). So this scenario silently regresses:

1. Launch prewarmed kernel — `launched.uv_deps = None`
2. User `add_dependency("pandas")` — metadata gains inline deps; hot-install fires; everything fine
3. User restarts kernel — venv persists; metadata still has the inline dep
4. **This PR's new code**: `compute_env_sync_diff` returns `None` (because `launched.uv_deps` is still `None`), so the kernel is marked in_sync even though the CRDT metadata now requires inline tracking

The existing helper's `else` branch at line 755 specifically handles this by calling `check_inline_deps` + reading the deps from metadata.

## Suggested shape

Replace the two inline blocks with:

```rust
// runtime_agent_launched_config is already updated a few lines above
// (lines 5661 / 5839), so this picks up the latest launched state.
check_and_broadcast_sync_state(room).await;
```

That's ~2 lines instead of ~30, both restart paths get identical behavior, and the prewarmed case is covered.

One caveat to verify when switching: `check_and_broadcast_sync_state` bails early if kernel status isn't `idle`/`busy` (line 717). The restart paths set status to `idle` at lines 5670 / 5851 before the env-sync block, so the status check should pass — but worth a local trace.

Marking as request-changes for the prewarmed gap; the DRY suggestion is nice-to-have but the missed case is the real issue.

crates/runtimed/src/notebook_sync_server.rs
quillaid: Good catch on the prewarmed gap — replaced both blocks with `check_and_broadcast_sync_state(room).await`. Net result: +8/-14 lines instead of +61/-8.

Verified the status precondition: both restart paths set `set_kernel_status("idle")` before the env-sync block (lines 5670 and 5859), and `runtime_agent_launched_config` is written at lines 5661/5839 — so the helper sees the right state. All 23 runtimed tests pass.

rgbkrk (approved):
Great — this is exactly the refactor I had in mind. Net result is a *smaller* diff than the original PR that also covers a case the original missed (prewarmed + later inline deps).

Verified locally:
- `cargo check -p runtimed` clean
- `cargo test -p runtimed --lib` — 291 passing
- `cargo test -p runtimed --test tokio_mutex_lint` — 1 passing (confirms no locks held across the `.await` call to `check_and_broadcast_sync_state`)
- `cargo xtask lint` clean

The two ordering invariants the fix depends on both hold in the current file:
1. `runtime_agent_launched_config` is written just above each call site (lines 5661 / 5839), so the helper reads the latest config.
2. Kernel status is set to `idle` just above too (line 5670 / ~5851), so the helper's status-guard at line 717 won't early-bail.

PR #1824, opened 2026-04-15
chore(dx): strip stale ty: ignore[unresolved-import] for datasets

rgbkrk:
Main is currently red on `cargo xtask lint` because of 9 unused-ignore-comment warnings left over from #1809 — the `# ty: ignore[unresolved-import]` suppressions on `datasets` imports are unused since `datasets` IS in the workspace dev deps (pyproject.toml:58), so ty resolves the import cleanly.

Strips all 9 of those. Leaves the suppression on `altair` intact since altair is genuinely not in workspace deps.

## Verification

- `uv run ty check python/dx/` — "All checks passed!"
- `cargo xtask lint` — "All checks passed!" (exit 0)

## Scope

Trivial — 9 occurrences, single-character-class sed. No behavior change.

### DIFF
--- a/python/dx/src/dx/_format_install.py
+++ b/python/dx/src/dx/_format_install.py
@@ -155,7 +155,7 @@ def install_formatters() -> None:
         pass
 
     try:
-        import datasets  # noqa: PLC0415  # ty: ignore[unresolved-import]
+        import datasets  # noqa: PLC0415
 
         mimebundle.for_type(datasets.Dataset, _dataset_mimebundle)
         ipython_display.for_type(datasets.Dataset, _dataset_ipython_display)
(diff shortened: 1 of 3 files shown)


rgbkrk:
Closing as no-op. After investigation (see 76264665), the `ty: ignore[unresolved-import]` directives are legitimately needed: `datasets` is not declared in any `pyproject.toml`, only installed in my local `.venv` as a transitive leftover. CI's fresh `uv sync` can't resolve it, so the suppressions are required for CI to pass. Locally ty reports them as 'unused' because my venv happens to have datasets — that's a local-vs-CI drift, not a cleanup opportunity.

If we want to get rid of the warnings entirely, the right path is either (a) add `datasets` to the dx dev deps so both local and CI see it, or (b) wait for ty to treat unresolved-import suppressions consistently across environments.

PR #1828, opened 2026-04-15
fix(runt-mcp): use inline env_source when deps are pre-declared

quillaid:
## Summary

- When `create_notebook(package_manager="uv", dependencies=["pandas"])` runs from a directory with an unrelated `pyproject.toml`, the daemon's auto-launch resolves to `uv:pyproject` via `room.working_dir` project-file detection. This causes the kernel to fail (e.g. hash mismatch on local wheel deps), bootstraps 25+ wrong deps into the CRDT, and the MCP relaunch with `auto:uv` re-detects the same project file.
- Fix: when inline deps exist, all three MCP tool paths (`create_notebook`, `restart_kernel`, `add_dependency(after="restart")`) now use `*:inline` instead of `auto:*`, skipping project-file detection entirely.
- Also replaces (rather than appends) the dep list in `create_notebook` to override any auto-bootstrapped pyproject deps.

## Reproduction

The nightly gremlin test suite (run from `/home/ubuntu/projects/nightly-tester/`, which has its own `pyproject.toml` with 25+ deps) consistently hit:
1. `create_notebook(package_manager="uv", dependencies=["pandas", "numpy"])` → kernel never launches (`has_kernel: false`)
2. `execute_cell` hangs (kernel status stuck at `not_started`)
3. `restart_kernel` fails with "Kernel process exited immediately: exit status: 1"

Root cause: the MCP server inherits the gremlin harness's cwd. The daemon detects the harness's `pyproject.toml` → `uv:pyproject` → `uv run --with ipykernel` fails on a `dx` wheel with stale hash → exit status 1.

## Changes

| File | Change |
|------|--------|
| `session.rs` | `create_notebook` relaunch uses `*:inline` when deps provided; `set_deps_for_manager()` replaces bootstrapped deps |
| `kernel.rs` | `restart_kernel` maps prewarmed→`*:inline` when inline deps exist; fallback also checks inline deps |
| `deps.rs` | `add_dependency(after="restart")` maps prewarmed→`*:inline`; new `has_any_inline_deps()` and `set_deps_for_manager()` helpers |

## Test plan

- [x] E2E test via MCP subprocess from problematic cwd (`/home/ubuntu/projects/nightly-tester/`):
  - `create_notebook` returns `deps=["pandas","numpy"]` (not 25+ pyproject deps)
  - `execute_cell` succeeds with `uv:inline` kernel
  - `restart_kernel` uses `uv:inline` (not `auto:uv`)
  - `execute_cell` after restart succeeds
- [x] Daemon logs confirm: auto-launch → `uv:pyproject` (fails), MCP relaunch → `uv:inline` (succeeds)
- [ ] Nightly gremlin suite validation (experiments-manager)

### DIFF
--- a/crates/kernel-env/src/uv.rs
+++ b/crates/kernel-env/src/uv.rs
@@ -166,6 +166,9 @@ pub async fn prepare_environment_in(
 
     let mut venv_cmd = tokio::process::Command::new(&uv_path);
     venv_cmd.arg("venv").arg(&venv_path);
+    // Set explicit cwd so `uv` doesn't fail when the daemon's inherited cwd
+    // has been deleted (e.g. a cleaned-up gremlin temp directory).
+    venv_cmd.current_dir(cache_dir);
 
     if let Some(ref py_version) = deps.requires_python {
         let version = py_version
@@ -234,6 +237,7 @@ pub async fn prepare_environment_in(
 
     let offline_output = tokio::process::Command::new(&uv_path)
         .args(&offline_args)
+        .current_dir(cache_dir)
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .output()
@@ -268,6 +272,7 @@ pub async fn prepare_environment_in(
 
     let install_output = tokio::process::Command::new(&uv_path)
         .args(&install_args)
+        .current_dir(cache_dir)
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .output()
@@ -287,6 +292,7 @@ pub async fn prepare_environment_in(
         retry_args.insert(2, "--refresh".to_string());
         tokio::process::Command::new(&uv_path)
             .args(&retry_args)
+            .current_dir(cache_dir)
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
             .output()
@@ -352,8 +358,13 @@ pub async fn sync_dependencies(env: &UvEnvironment, deps: &[String]) -> Result<(
     // Insert --offline after "pip install" (index 2)
     offline_args.insert(2, "--offline".to_string());
 
+    // Use the venv's parent directory as cwd so `uv` doesn't fail when the
+    // daemon's inherited cwd has been deleted.
+    let cwd = env.venv_path.parent().unwrap_or_else(|| Path::new("/tmp"));
+
     let offline_output = tokio::process::Command::new(&uv_path)
         .args(&offline_args)
+        .current_dir(cwd)
         .output()
         .await?;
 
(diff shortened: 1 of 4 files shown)


crates/kernel-env/src/uv.rs
```diff
@@ -166,6 +166,9 @@ pub async fn prepare_environment_in(
 
     let mut venv_cmd = tokio::process::Command::new(&uv_path);
     venv_cmd.arg("venv").arg(&venv_path);
+    // Set explicit cwd so `uv` doesn't fail when the daemon's inherited cwd
+    // has been deleted (e.g. a cleaned-up gremlin temp directory).
+    venv_cmd.current_dir(cache_dir);
```
rgbkrk, line 171: this part is smart

rgbkrk:
I'm not sure if this is the right product direction as we should use the local environment if the user is launching a notebook in a project. What should we think about for inline deps?

quillaid:
Good call — forcing `*:inline` when deps are explicit was too aggressive. It solved the immediate crash (stale `pyproject.toml` with a bad wheel hash) but at the cost of ignoring the user's project environment entirely.

The right product direction is:

1. **Respect project environments** — if there's a `pyproject.toml` or `environment.yml`, that should be the base environment. An agent calling `create_notebook(dependencies=["pandas"])` in a project dir should get "project env + pandas", not an isolated inline env with only pandas.

2. **Add agent deps on top** — explicit deps from `create_notebook` or `add_dependency` layer onto whatever environment was resolved, not replace it.

3. **Inline only as fallback** — when there's no project context (no pyproject, no conda env), then inline is the right default.

4. **Better AX** — `create_notebook` and `get_dependencies` responses should give agents visibility into what was resolved: "uv:pyproject, project has N deps, added your M deps on top." This lets agents make informed decisions rather than operating blind.

The original crash (kernel exiting immediately due to a broken pyproject.toml) is still a real bug — but the fix should be better error handling/fallback in that path, not bypassing project detection wholesale.

**Plan:** I'll rework the `session.rs` / `kernel.rs` / `deps.rs` env_source changes to respect this direction. The `.current_dir()` fix in `kernel-env/src/uv.rs` (commit `69eed8e1`) is orthogonal and stays as-is.

quillaid:
Splitting into two focused PRs: one for the .current_dir() bug fix (UV subprocesses failing when daemon cwd is deleted), one for the env resolution rework (respect project env, layer agent deps). See the new PRs linked below.

quillaid:
Split into two focused PRs:
- **#1830** — `.current_dir()` bug fix for UV subprocesses (daemon-side)
- **#1831** — Env resolution rework: respect project env, enrich create_notebook response (MCP-side)

PR #1854, opened 2026-04-17
fix(sift): collapse filtered histogram to single bin when slice is constant (closes #1847)

### LINKED ISSUE #1847: fix(sift): column header range shows phantom max (min + bin_count) when filtered slice is constant
## Summary

When a filter narrows a numeric column to rows where all values are identical (`min == max`), the Sift header displays a nonsense range like `0.46 – 25.46` for danceability when every filtered row is `0.459`, or `0 – 25` for popularity when every filtered row is `0`.

## Repro

1. `pnpm --filter @nteract/sift dev`
2. Open http://localhost:5174/ (Spotify Tracks dataset)
3. Click any repeated `track_id` cell, e.g. the first row (`5SuOikwiRyPMVoIQDJUgSV`) — this auto-filters to rows matching that value.
4. Observe the numeric column headers: `popularity` shows `0 – 25`, `duration_ms` shows `423,680 – 423,705`, `danceability` shows `0.46 – 25.46`.

All filtered rows have `popularity = 0`, `duration_ms = 423680`, `danceability = 0.459` — so the headers should read `0 – 0`, `423,680 – 423,680`, `0.459 – 0.459` (or collapse to a single value).

## Root cause

Two-layer bug:

**Rust** (`crates/nteract-predicate/src/summary.rs`, the `histogram` fn around line 143–154):

```rust
let min = values.iter().cloned().fold(f64::INFINITY, f64::min);
let max = values.iter().cloned().fold(f64::NEG_INFINITY, f64::max);
let bin_width = if (max - min).abs() < f64::EPSILON {
    1.0                                   // ← fallback when all values equal
} else {
    (max - min) / num_bins as f64
};

let mut bins: Vec<HistogramBin> = (0..num_bins)
    .map(|i| HistogramBin {
        x0: min + i as f64 * bin_width,
        x1: min + (i + 1) as f64 * bin_width,
        count: 0,
    })
    .collect();
```

When `min == max`, `bin_width = 1.0`, so the last bin's `x1 = min + num_bins * 1.0`. With `num_bins = 25`, that produces `min + 25`.

**TypeScript** (`packages/sift/src/wasm-table-data.ts:231–232`):

```ts
tableData.columnSummaries[c] = {
  kind: colType as "numeric" | "timestamp",
  min: bins[0].x0,
  max: bins[bins.length - 1].x1,   // ← consumes the 25-wide phantom range
  bins,
};
```

So `summary.min = 0.459` (correct), `summary.max = 25.459` (wrong), and the sparkline footer renders `formatNum(min) – formatNum(max)` → `0.46 – 25.46`.

The same thing happens for the unfiltered `histogram()` path when the full column is constant, but in practice nobody notices because real columns aren't constant until a filter narrows them.

## Proposed fix

Two options, rank-ordered:

1. **(Preferred)** Return `min`/`max` explicitly from Rust alongside `bins`, and have TS use those instead of inferring from bin edges. This also helps the single-value and single-bin cases cleanly. Change the wasm-bindgen shape of `histogram` / `store_filtered_histogram` to `{ min, max, bins }`.

2. **(Cheap)** In Rust, when `(max - min).abs() < f64::EPSILON`, emit a single bin `[min, min]` (or `[min, min + small_epsilon]`) with the full count instead of 25 synthetic bins. TS already has `bins[0].x0` and `bins[last].x1` logic; no TS change needed.

Option 1 is more robust long-term (unblocks other "I need true min/max" features); option 2 is a one-liner.

## Observed UX impact

- Users see fabricated numbers in the column header that don't match any row in the table.
- The `values hidden` label is shown below the range but is too generic — doesn't tell the user that the numeric range has actually collapsed to a single value.
- The histogram area visually appears mostly empty because 25 bins over a tiny range put all counts in the first bin.

## Related

- PR #1843 `fix(sift): handle nested list values in table cells` (recent sift fix — unrelated but adjacent)
- Memory note "sift renderer follow-ups" tracks other sift polish items after PR #1712

rgbkrk:
## Summary

Closes #1847.

Both histogram helpers (`nteract-predicate::summary::histogram` and `sift-wasm::store_filtered_histogram`) had the same degenerate path when every value was identical:

```rust
let bin_width = if (max - min).abs() < f64::EPSILON {
    1.0
} else {
    (max - min) / num_bins as f64
};
```

With `bin_width = 1.0` and `num_bins = 25`, the emitted bins spanned `[min, min + 25]`, and the TS consumer reads `bins[last].x1` as the column max. A column where every filtered row was `0.459` would show **"0.46 – 25.46"** in the header - a phantom value that doesn't exist anywhere.

## Fix

Detect the constant case and return a single degenerate bin `[min, min, values.len()]`. `wasm-table-data.ts` already reads `bins[0].x0` and `bins[last].x1` as min/max - both resolve to `min` now, so the downstream summary is correct.

Also added a `formatNumRange(min, max)` helper in `packages/sift/src/sparkline.tsx` that collapses `"0.46 – 0.46"` → `"0.46"`. Mirrors the same logic already in use for filter-pill labels, and the timestamp histogram's label pair gets the same treatment.

## Verified

Demo dataset "Spotify Tracks" → click a rare `track_id` in the header category list to filter to ~8 identical rows:

- **Before**: `danceability` header showed `"0.46 – 25.46"`, `popularity` showed `"0 – 25"`, `duration_ms` showed `"423,680 – 423,705"` (wrong - 25 added to min).
- **After**: `danceability` shows `"0.46"` (collapsed). `popularity` and `duration_ms` drop the range entirely because they hit the existing binary-like heuristic - their "values hidden" footer still makes the filter state obvious.

verified

## Test plan

- [x] `cargo check -p sift-wasm --target wasm32-unknown-unknown` clean
- [x] `cargo check -p nteract-predicate` clean
- [x] Manual: Spotify demo, filter to constant slice, header range collapses correctly
- [x] Renderer plugin + MCP assets rebuilt and committed
- [ ] Reviewer: spot-check other datasets (Airbnb, Credit Card) to confirm no regression on non-constant filtered slices


rgbkrk:
Superseded by #1859 which consolidates with #1855.

PR #1855, opened 2026-04-17
fix(sift): cast Utf8View to Utf8 before emitting IPC to JS (closes #1853)

### LINKED ISSUE #1853: fix(sift): apache-arrow-js can't decode Utf8View (type 24), breaking polars-produced .arrow files
## Summary

Loading an Arrow IPC stream that contains a `Utf8View` column (IPC type code 24) through the sift demo throws at decode time:

```
Error: Unrecognized type: "undefined" (24)
  at decodeFieldType (apache-arrow.js:7973)
  at Field.decodeField (apache-arrow.js:7901)
  ...
  at tableFromIPC (apache-arrow.js:12742)
  at Object.prefetchViewport (src/wasm-table-data.ts:54)
```

The WASM/Rust side (`sift-wasm`) loads the same bytes fine - `num_rows`, column names, schema classification, and the header-row sparklines all render correctly. The failure is on the TS side when `wasm-table-data.ts` calls `tableFromIPC(ipcBytes)` against the viewport bytes returned by `get_viewport_by_indices` to render table cells.

User-visible symptom: the footer reports `N rows` but the table body is empty and `sift-empty-state` shows "No matching rows".

## Repro

1. Write an Arrow IPC stream from recent polars (any string column triggers it - polars defaults to `Utf8View` for strings in the Rust backend).
2. Register it as a `"local"` dataset in `packages/sift/src/datasets.ts` and point at it.
3. Open the demo → header row renders, body is empty.

## Why this matters

- Polars is the most obvious Python path into the sift demo for regression fixtures and ad-hoc exploration. Its default string encoding lands users in this hole with no actionable error shown in the UI.
- `apache-arrow@21.1.0` is installed in `pnpm-lock.yaml`; Utf8View was added to the Arrow format spec in 1.4 and has been in the apache-arrow npm package for a while. Expected it to work.

## Related context

Discovered while building `packages/sift/public/timestamp-fixture.arrow` as the regression fixture for PR #1852 (filtered timestamp histograms). Worked around it by building the fixture directly with pyarrow and forcing `pa.string()` (type code 13). That covers the fixture but not the general case.

## Fix directions

1. Update `apache-arrow` to a version that decodes Utf8View, if one exists, or pin to a version known to work.
2. Optionally transform Utf8View columns to Utf8 in `sift-wasm`'s `get_viewport_by_indices` before emitting IPC to the JS side, so the JS consumer only sees types it understands.
3. Surface a clearer user-facing error when `tableFromIPC` throws - today the body just goes empty with no console-linked message in the UI.

rgbkrk:
## Summary

Closes #1853.

`apache-arrow@21.x` (the npm package the sift frontend uses) doesn't decode `Utf8View` (IPC wire type 24). When `get_viewport` or `get_viewport_by_indices` in `sift-wasm` emits IPC bytes containing a Utf8View column, `tableFromIPC` throws `Unrecognized type: "undefined" (24)` and silently empties the table body. The footer still reports the correct row count because the WASM side knows the total, but no cells render.

This hits any polars-written `.arrow` file with string columns (polars' default string encoding is Utf8View).

## Fix

Add a `downgrade_view_types(RecordBatch)` helper in `crates/sift-wasm/src/store.rs` that:

1. Walks schema fields once - if none are `Utf8View`/`BinaryView`, clones and returns (fast path).
2. Otherwise, casts `Utf8View` → `Utf8` and `BinaryView` → `Binary` via `arrow_cast::cast`, rebuilds the batch with the transformed schema, and returns it.

Called in both `get_viewport` and `get_viewport_by_indices` before handing IPC bytes to JS. Zero overhead for the common case (no view types in the schema).

### Scope

Top-level columns only. Nested occurrences (`List<Utf8View>`, `Struct{s: Utf8View}`) aren't handled here - I haven't seen a real data source that produces them, and traversing arbitrary nested types deserves its own design pass. If it comes up, the same decode error will surface and we can follow up.

## Regression fixture

`packages/sift/public/polars-utf8view.arrow` (100 rows, polars-written with two string columns + emoji) registered as "Polars Utf8View (100)" in the demo picker.

Before this change: footer reports "100 rows", body is empty, console shows `Error: Unrecognized type: "undefined" (24)` from `tableFromIPC`.
After: all 100 rows render, strings + emoji display correctly.

## Test plan

- [x] `cargo check -p sift-wasm --target wasm32-unknown-unknown` clean
- [x] Manual: "Polars Utf8View (100)" dataset in the demo renders all rows with string columns
- [x] Manual: existing datasets (Spotify, Generated, Airbnb) still render correctly (no regression on non-view schemas)
- [x] WASM + renderer plugin artifacts rebuilt and committed


rgbkrk:
Superseded by #1859 which consolidates with #1854.

PR #1863, opened 2026-04-17
fix(widgets): inline text-MIME comm state blobs

ktaletsk:
[pywidget](https://github.com/ktaletsk/pywidget) widgets (CounterWidget, StatsWidget) use `_py_render` and `_esm` traits that carry Python/JS source code via Automerge blob refs. The WASM resolver previously returned these as URL strings, same as binary buffers — but widget code reads synced string traits directly and can't fetch URLs.

walk_and_resolve_comm_state now classifies blob refs by media_type and returns text_paths alongside buffer_paths. The sync-engine fetches each text URL and substitutes the decoded body in place before emitting to commChanges$, with exponential-backoff retry on transient failures.


rgbkrk:
Interesting gap you've found here. `_esm` I assumed never needs to be a string unless less than 1 KB (I don't want big strings in the CRDT/Automerge document). I think _that_ one I'll want to keep a special case for but you definitely found a big bug in text handling on these fields though.

rgbkrk (approved):

PR #1870, opened 2026-04-17
feat(widgets): native ipyleaflet support (Phase 1)

### LINKED ISSUE #1497: feat(widgets): ipyleaflet support
ipyleaflet renders as "Unsupported widget: LeafletMapModel" because it's a classic `DOMWidget` extension — no `_esm` field, no anywidget compatibility.

## Current state

- 57+ built-in ipywidgets are supported via hand-written React components
- anywidget widgets work automatically (ESM + AFM spec)
- No plugin/extension system — widget support requires source changes

## Options

1. **ipyleaflet adopts anywidget upstream** — would work automatically, no nteract changes needed. Best path but depends on upstream.
2. **Dedicated built-in implementation** — significant effort given ipyleaflet's complexity (Map, TileLayer, Marker, GeoJSON, draw controls, many layer types)
3. **Generic DOMWidget compatibility layer** — would support ipyleaflet and other classic widget extensions, but substantial architecture addition

## Workaround

Use folium instead for interactive maps — it renders as HTML output (now working with the CSP fix in #1496). No widget protocol needed.

rgbkrk:
## Summary

Implements native rendering for ipyleaflet widgets by reading their state fields directly and rendering with Leaflet.js — no ipyleaflet JavaScript is loaded or executed. This is Phase 1, covering the core widgets that handle ~90% of ipyleaflet usage.

**Functional widgets (full two-way binding):**
- `LeafletMapModel` — creates `L.map`, syncs center/zoom/bounds back to kernel, forwards mouse events
- `LeafletTileLayerModel` — reactive tile URL, opacity, visibility
- `LeafletMarkerModel` — draggable with echo suppression, custom icon resolution (Icon/AwesomeIcon/DivIcon)
- `LeafletGeoJSONModel` — feature click/hover events, hover_style support

**Architecture:** The Map widget shares its `L.map` instance with child layer widgets via React context, so each layer (a separate widget comm) can call `layer.addTo(map)`. All remaining ipyleaflet model names (~30) are stub-registered as null renders to prevent "Unsupported widget" fallback.

Closes #1497

## Verification

- [ ] Create a notebook with `ipyleaflet` installed, run `from ipyleaflet import Map, Marker, GeoJSON; m = Map(center=[45.5, -122.5], zoom=10); m` — map renders with tiles
- [ ] Pan/zoom the map, then check `m.center` and `m.zoom` in Python — values update
- [ ] Run `m.center = [40.7, -74.0]` — map pans to New York
- [ ] Add a `Marker(location=[45.5, -122.5], draggable=True)` to the map, drag it, check `marker.location` updates
- [ ] Add a `GeoJSON` layer with a FeatureCollection — features render, clicks fire
- [ ] Verify dark/light theme doesn't break map rendering
- [ ] Verify existing `application/geo+json` output renderer still works (different code path)
- [ ] Run a notebook using unsupported ipyleaflet features (e.g. `DrawControl`) — should render blank, not "Unsupported widget"

### DIFF
--- a/src/components/widgets/ipyleaflet/index.ts
+++ b/src/components/widgets/ipyleaflet/index.ts
@@ -1,4 +1,11 @@
 import { registerWidget } from "../widget-registry";
+import {
+  LeafletAttributionControlWidget,
+  LeafletFullScreenControlWidget,
+  LeafletLayersControlWidget,
+  LeafletScaleControlWidget,
+  LeafletZoomControlWidget,
+} from "./leaflet-controls";
 import { LeafletGeoJSONWidget } from "./leaflet-geojson";
 import { LeafletMapWidget } from "./leaflet-map-widget";
 import { LeafletMarkerWidget } from "./leaflet-marker";
@@ -37,11 +44,11 @@ registerWidget("LeafletVideoOverlayModel", NullWidget);
 registerWidget("LeafletWMSLayerModel", NullWidget);
 registerWidget("LeafletPopupModel", NullWidget);
 registerWidget("LeafletControlModel", NullWidget);
-registerWidget("LeafletZoomControlModel", NullWidget);
-registerWidget("LeafletScaleControlModel", NullWidget);
-registerWidget("LeafletAttributionControlModel", NullWidget);
-registerWidget("LeafletFullScreenControlModel", NullWidget);
-registerWidget("LeafletLayersControlModel", NullWidget);
+registerWidget("LeafletZoomControlModel", LeafletZoomControlWidget);
+registerWidget("LeafletScaleControlModel", LeafletScaleControlWidget);
+registerWidget("LeafletAttributionControlModel", LeafletAttributionControlWidget);
+registerWidget("LeafletFullScreenControlModel", LeafletFullScreenControlWidget);
+registerWidget("LeafletLayersControlModel", LeafletLayersControlWidget);
 registerWidget("LeafletDrawControlModel", NullWidget);
 registerWidget("LeafletGeomanDrawControlModel", NullWidget);
 registerWidget("LeafletMeasureControlModel", NullWidget);
--- a/src/components/widgets/ipyleaflet/index.ts
+++ b/src/components/widgets/ipyleaflet/index.ts
@@ -1,0 +1,57 @@
+import { registerWidget } from "../widget-registry";
+import { LeafletGeoJSONWidget } from "./leaflet-geojson";
+import { LeafletMapWidget } from "./leaflet-map-widget";
+import { LeafletMarkerWidget } from "./leaflet-marker";
+import { LeafletTileLayerWidget } from "./leaflet-tile-layer";
+
+// Phase 1: Map + TileLayer + Marker + GeoJSON
+registerWidget("LeafletMapModel", LeafletMapWidget);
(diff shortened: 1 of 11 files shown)


src/components/widgets/ipyleaflet/leaflet-map-widget.tsx
```diff
@@ -0,0 +1,161 @@
...
+      });
+      setTimeout(() => {
+        localUpdateRef.current = false;
+      }, 200);
+    };
+
+    map.on("moveend", syncBounds);
+    map.on("zoomend", syncBounds);
+
+    for (const eventType of [
```
rgbkrk, line 79: Upstream `ipyleaflet` forwards more map interaction types here: `mousemove`, `contextmenu`, and `preclick` in addition to the six currently listed ([source](https://github.com/jupyter-widgets/ipyleaflet/blob/master/python/jupyter_leaflet/src/Map.ts#L336-L345)). `Map.on_interaction(...)` callbacks that rely on those event types will silently stop firing with this renderer.

src/components/widgets/ipyleaflet/index.ts
```diff
@@ -0,0 +1,57 @@
...
+registerWidget("LeafletMarkerClusterModel", NullWidget);
+registerWidget("LeafletLayerGroupModel", NullWidget);
+registerWidget("LeafletFeatureGroupModel", NullWidget);
+registerWidget("LeafletHeatmapModel", NullWidget);
+registerWidget("LeafletImageOverlayModel", NullWidget);
+registerWidget("LeafletVideoOverlayModel", NullWidget);
+registerWidget("LeafletWMSLayerModel", NullWidget);
+registerWidget("LeafletPopupModel", NullWidget);
+registerWidget("LeafletControlModel", NullWidget);
+registerWidget("LeafletZoomControlModel", NullWidget);
```
rgbkrk: `Map` still auto-adds `ZoomControl` and `AttributionControl` when `zoom_control` / `attribution_control` are true ([source](https://github.com/jupyter-widgets/ipyleaflet/blob/master/python/ipyleaflet/ipyleaflet/leaflet.py#L2986-L2994)). In this patch those control models are stubbed to `NullWidget`, while `leaflet-map-widget.tsx` also disables Leaflet's native controls. The result is that default ipyleaflet maps lose both the zoom buttons and the attribution UI entirely.

src/components/widgets/ipyleaflet/leaflet-map-widget.tsx
rgbkrk: Fixed in 1c76312 — added `mousemove`, `contextmenu`, and `preclick` to match upstream's full `'click dblclick mousedown mouseup mouseover mouseout mousemove contextmenu preclick'` set.

src/components/widgets/ipyleaflet/index.ts
rgbkrk: Fixed in 1c76312 — implemented `LeafletZoomControlModel`, `LeafletAttributionControlModel`, `LeafletScaleControlModel`, `LeafletLayersControlModel`, and `LeafletFullScreenControlModel` as real Leaflet controls in `leaflet-controls.tsx`. They consume position/options from widget state and call `L.control.zoom()` etc.

PR #1871, opened 2026-04-17
feat(outputs): detect and inject plotly.js for text/html outputs

### LINKED ISSUE #1533: feat(outputs): detect and inject plotly.js for text/html outputs
## Context

When Plotly uses non-default renderers (`notebook`, `iframe_connected`), it emits `text/html` output containing `<script>` tags that reference `window.Plotly` or `require('plotly')`. Currently, plotly.js is only injected into iframes when `application/vnd.plotly.v1+json` is detected in the output MIME types.

This means `text/html` Plotly outputs from these backends silently fail — the scripts execute but `window.Plotly` is undefined.

## Possible Approach

Scan `text/html` output content for known Plotly signatures (e.g., `Plotly.newPlot`, `require(['plotly']`, `require("plotly")`) before rendering. If detected, inject plotly.js into the iframe before the HTML is rendered.

This would be similar to how `getRequiredLibraries()` scans MIME types today, but extended to peek at HTML content for known library references.

## Notes

- The default `plotly_mimetype` renderer outputs `application/vnd.plotly.v1+json` which works correctly via the renderer plugin API
- Self-contained HTML outputs (e.g., Plotly's `cdn` renderer which includes a CDN `<script>` tag) already work
- The gap only affects renderers that expect plotly.js to be pre-loaded
- Related: #1255 (original plotly MIME type support)

rgbkrk:
## Summary

When Plotly uses non-default renderers like `notebook`, it emits `text/html` containing inline scripts that reference `window.Plotly` or AMD `require("plotly")`. These silently fail because plotly.js is only injected when `application/vnd.plotly.v1+json` appears in the output MIME bundle.

This PR scans `text/html` content for Plotly signatures and triggers the plotly plugin injection, which now also exposes `window.Plotly` and an AMD `require` shim inside the sandboxed iframe.

### Changes

- **`iframe-libraries.ts`** — new `htmlNeedsPlotly()` function that regex-scans HTML for Plotly method calls and AMD require patterns
- **`OutputArea.tsx`** — after the existing MIME-based plugin scan, adds a content scan of `text/html` outputs; if Plotly signatures detected, triggers the plotly plugin load
- **`plotly-renderer.tsx`** — `install()` now exposes `window.Plotly` globally in the iframe, registers named AMD modules if RequireJS is present, and installs a minimal AMD/CJS `require` shim

### Timing

The plugin `install()` runs before `renderBatch()`, so `window.Plotly` is guaranteed to be set before any HTML scripts execute.

Closes #1533

## Verification

- [ ] Create a notebook, `import plotly.express as px`, set `plotly.io.renderers.default = "notebook"`, render a chart — it should display
- [ ] Default renderer (`plotly_mimetype`) still works as before
- [ ] Non-plotly `text/html` outputs (e.g. pandas DataFrames) do NOT trigger unnecessary plotly plugin loading (check console for `[iframe-libraries] installing renderer plugin` messages)

### DIFF
--- a/src/components/cell/OutputArea.tsx
+++ b/src/components/cell/OutputArea.tsx
@@ -346,8 +346,8 @@ export function OutputArea({
       }
     }
 
-    // Detect Plotly in text/html content from non-default renderers (e.g. "notebook",
-    // "iframe_connected") that emit only text/html without the structured MIME type.
+    // Detect Plotly in text/html content from non-default renderers (e.g. "notebook")
+    // that emit inline scripts without the structured MIME type.
     if (!pluginMimes.has("application/vnd.plotly.v1+json")) {
       for (const output of outputs) {
         if (
--- a/src/components/cell/OutputArea.tsx
+++ b/src/components/cell/OutputArea.tsx
@@ -6,7 +6,11 @@ import {
   IsolatedFrame,
   type IsolatedFrameHandle,
 } from "@/components/isolated";
-import { injectPluginsForMimes, needsPlugin } from "@/components/isolated/iframe-libraries";
+import {
+  htmlNeedsPlotly,
+  injectPluginsForMimes,
+  needsPlugin,
+} from "@/components/isolated/iframe-libraries";
 import { AnsiErrorOutput, AnsiStreamOutput } from "@/components/outputs/ansi-output";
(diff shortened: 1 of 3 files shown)


rgbkrk:
I think I want to just push people towards using the proper mimetype (or `dx` does it for them).

PR #1877, opened 2026-04-17
fix(runtimed): reset execution_count on Clear Outputs

### LINKED ISSUE #1876: Clear All Outputs leaves stale execution counters
## Summary

**Cell → Clear All Outputs** removes cell outputs but keeps the execution counter `[N]:` on each cell. Cleared cells look like they've been run when they haven't. Jupyter's equivalent action clears both.

## Steps to reproduce

1. Create a few code cells and run them — they show `[1]:`, `[2]:`, `[3]:`.
2. **Cell menu → Clear All Outputs.**
3. Counters remain `[1]:`, `[2]:`, `[3]:` instead of resetting.

## Expected behavior

Jupyter's [`clearExecution()`](https://github.com/jupyterlab/jupyterlab/blob/main/packages/cells/src/model.ts) nulls `executionCount` alongside clearing outputs:

```typescript
clearExecution(): void {
  this.outputs.clear();
  this.executionCount = null;      // ← we skip this
  ...
}
```

Cleared cells should look like never-run cells.

## Root cause

`crates/runtimed/src/notebook_sync_server.rs` handles `ClearOutputs` by calling `clear_execution_outputs(eid)`, which only empties the `outputs` list. The `execution_count` field on the execution entry is never nulled, so the frontend's per-cell count resolver still finds the old number and keeps displaying it.

## Proposed fix

Extend `clear_execution_outputs` in `crates/notebook-doc/src/runtime_state.rs` to also null `execution_count` (and `success`) alongside the outputs list. Single call site, unit-tested.

ktaletsk:
## Summary

- `ClearOutputs` now nulls `execution_count` and `success` alongside emptying the outputs list, matching Jupyter's `clearExecution()` semantics.
- Single-site change in `notebook-doc::runtime_state::clear_execution_outputs`; no call sites need to change.
- Added a regression test covering the new contract; the original test still passes.

Closes #1876

## Why

Before this, Clear Outputs only emptied an execution's `outputs` list. The frontend's per-cell count resolver walks every execution for a cell and returns the latest non-null count — so cleared cells kept showing their stale `[N]:` counter. See #1876 for repro + screenshots.

As a side effect, Run All with a mid-sequence error now leaves the trailing cells at `[▶]:` naturally (they never started, their `execution_id` pointer is null, no fresh execution overwrote the counter). No "cancelled" UI state needed.

## Test plan

- [x] `cargo test -p notebook-doc --lib` — 328 passed (includes new `test_clear_execution_outputs_resets_count_and_success`).
- [x] `cargo clippy -p notebook-doc --lib -- -D warnings` — clean.
- [x] Manual: run 3 cells → Clear All Outputs → counters snap to `[▶]:`.
- [x] Manual: Run All with `raise ValueError("boom")` in the middle → erroring cell shows `[N]:` with traceback, trailing cells show `[▶]:`.

### DIFF
--- a/crates/notebook-doc/src/runtime_state.rs
+++ b/crates/notebook-doc/src/runtime_state.rs
@@ -1192,7 +1192,16 @@ impl RuntimeStateDoc {
         Ok(true)
     }
 
-    /// Clear all outputs for an execution.
+    /// Reset an execution entry to a cleared state.
+    ///
+    /// Empties the outputs list and nulls `execution_count` and `success`, so
+    /// the cell backed by this execution reads as "never run" from the
+    /// frontend's point of view. Mirrors JupyterLab's `clearExecution()` which
+    /// clears outputs *and* resets `executionCount` to `null`.
+    ///
+    /// Without nulling `execution_count`, the frontend's per-cell resolver
+    /// (which walks all executions for a cell looking for a non-null count)
+    /// would keep displaying the stale `[N]:` counter after a Clear Outputs.
     #[allow(clippy::expect_used)]
     pub fn clear_execution_outputs(&mut self, execution_id: &str) -> Result<bool, AutomergeError> {
         let Some(executions) = self.get_map("executions") else {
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
Thanks!

PR #1880, opened 2026-04-18
fix+feat(sync): widget sync stall — detection, recovery, headless regression harness

rgbkrk:
Fixes the ipywidget sync stall reproduced with matplotlib `@interact`: slider thumb moves, kernel runs, CRDT has the right state, but the UI freezes on a stale output until reload. This PR lands **both tracks** from the design doc — safety net for detecting/recovering from silent stalls, and the CRDT-first architectural refactor that eliminates the stall's root cause.

Full context and rationale: [`docs/superpowers/specs/2026-04-17-widget-sync-stall-design.md`](./docs/superpowers/specs/2026-04-17-widget-sync-stall-design.md)

### Track B — safety net

1. **`fix(widgets): abort hung blob fetches, log swallowed subscriber errors`** — 5s `AbortController` on per-attempt blob fetches (hung fetches were silently poisoning the serial `commEmitQueue`). `frame-bus.ts` / `runtime-state.ts` no longer swallow subscriber errors.
2. **`test(sync): real-WASM harness for widget comm-state sync`** — extends `WasmHarness` with runtime-state primitives + `put_comm_for_test` on WASM. 5 regression tests in `widget-sync-stall.test.ts` lock in `commChanges$` emission shape for the `@interact` scenario.
3. **`feat(sync): surface WASM sync recovery via top-level banner`** — new `SyncEngine.syncErrors$` observable + `SyncRecoveryBanner`. Recurring recoveries bump a counter ("Recovered N times recently — connection may be unhealthy"). No behavior change for healthy sessions.
4. **`feat(sync): watchdog resets sync state on silent runtime-state stall`** — 3s watchdog armed on outbound `RUNTIME_STATE_SYNC` flush. On timeout, `reset_sync_state` + re-flush. Detects pipe drops and silent daemon failures. Fixes applied inline from codex review (watchdog resets on each flush to avoid false positives under sustained traffic; clears on any inbound frame).

### Track A — architectural fix (CRDT-first)

5. **`feat(sync): project local comm writes through commChanges$ (A1)`** — `SyncEngine.projectLocalState()` fires the projection pipeline after a frontend-initiated `set_comm_state_batch`. The widget store now sees local writes from the same path as remote changes.
6. **`refactor(widgets): CRDT-first — drop optimistic path + echo suppression (A2)`** — `WidgetUpdateManager` collapses from ~180 lines to ~60. Deletes `shouldSuppressEcho`, `optimisticKeys`, per-comm debounce timers. Bootstrap fallback queues patches until the CRDT writer is registered, drains in insertion order. Binary buffers still mirror to the local store since the CRDT doesn't carry ArrayBuffers.
7. **`refactor(widgets): route jslink propagation via injected writer (A3)`** — abstracts link-subscriptions behind a `LinkWriter` callback. Default wiring is plain `store.updateModel` (jslink is documented as frontend-only in ipywidgets; it mirrors values client-side without touching the kernel). Codex-flagged: initial A3 routed through CRDT, which would have made jslink kernel-persisted; reverted.

### Supporting

8. **`perf(notebook): hoist MediaProvider renderers out of App render`** — module-scope `WidgetViewRenderer` + `mediaRenderers` object. Prevents remount cascades if App ever starts re-rendering.
9. **`docs: widget sync stall — design notes for tracks A + B`** — full design doc including PR #1580 archaeology explaining why the optimistic dual-write existed and what Track A does instead.

## What this PR is NOT

- Not persisting sync state across reloads (deferred; today's rebuild-from-scratch works fine).
- Not a full reconnect on stall — `reset_sync_state()` is cheaper and sufficient.
- Not a Deno-specific harness — the vitest-based real-WASM harness achieves the same "no React/Tauri" isolation goal without needing a second test runner.

## Test plan

- `pnpm vp test run packages/runtimed apps/notebook src/components/widgets` — 764 tests pass.
- `packages/runtimed/tests/widget-sync-stall.test.ts` scripts the matplotlib-@interact scenario against real WASM. Locks in the `commChanges$` invariants that Track A preserves.
- `packages/runtimed/tests/sync-engine.test.ts` — stall watchdog tests (fire on timeout, clear on inbound, re-arm on next flush) + `syncErrors$` emission tests for all three doc recoveries.
- `src/components/widgets/__tests__/widget-update-manager.test.ts` — rewritten for post-A2 semantics: happy path, bootstrap queue drain, buffer mirroring, no-op lifecycle.
- `cargo xtask lint` — clean.
- Manual: matplotlib `@interact` + FloatSlider, hammer arrow keys. Before: silent stall requiring reload. After: no drift possible between optimistic store and CRDT; if the pipe does drop a frame, banner surfaces the recovery and watchdog resets sync state within 3s.

## Codex review

Two codex passes caught four issues, all fixed inline:
- P1: watchdog clearing on unrelated daemon traffic — fixed via the `generate_runtime_state_sync_reply == null` signal (subsequently reverted for a simpler "any inbound" clear after a follow-up P1).
- P2: bootstrap patches dropped when no writer — fixed via pending queue with 50ms drain.
- P2: stall watchdog false-positive under sustained local traffic — fixed via reset-on-each-flush.
- P1: watchdog fires after every healthy interaction (strict null-reply signal) — fixed via "any inbound clears" + P2: jslink routed through CRDT — reverted to store-only write.

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -114,7 +114,7 @@ const _outputResolveGen = new Map<string, number>();
 function diffResolvedState(
   existing: Record<string, unknown>,
   next: Record<string, unknown>,
-  isPendingKey?: (key: string) => boolean,
+  isEchoOfPending?: (key: string, candidate: unknown) => boolean,
 ): Record<string, unknown> {
   const out: Record<string, unknown> = {};
   for (const [key, value] of Object.entries(next)) {
@@ -129,11 +129,12 @@ function diffResolvedState(
     } catch {
       // Non-serializable values (rare in comm state) → treat as changed.
     }
-    // Drop keys with a pending local write. The projected echo
-    // carries the daemon's pre-flush view of the CRDT — the store's
-    // newer in-flight value wins until the throttle trailing-edge
-    // flush reaches the daemon and the daemon's sync frame catches up.
-    if (isPendingKey?.(key)) continue;
+    // Drop stale echoes of our own local writes. An authoritative
+    // update (kernel validator, peer edit) carries a *different*
+    // value than what we wrote — those still flow through. Only
+    // suppresses keys whose projected value matches a local write
+    // currently in flight through the CRDT.
+    if (isEchoOfPending?.(key, value)) continue;
     out[key] = value;
   }
   return out;
@@ -429,8 +430,8 @@ function AppContent() {
         // microtask.
         const existing = widgetStore.getModel(comm.commId);
         const diff = existing
-          ? diffResolvedState(existing.state, comm.state, (key) =>
-              updateManager.hasPendingKey(comm.commId, key),
+          ? diffResolvedState(existing.state, comm.state, (key, candidate) =>
+              updateManager.isEchoOfPendingWrite(comm.commId, key, candidate),
             )
           : comm.state;
         if (Object.keys(diff).length > 0) {
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -1411,18 +1411,31 @@ const updateManager = new WidgetUpdateManager({
   getCrdtWriter: getCrdtCommWriter,
 });
 
+/**
+ * Renderer for `application/vnd.jupyter.widget-view+json` outputs.
+ * Hoisted to module scope so its identity is stable across App renders —
+ * an inline component type would remount the entire widget subtree
+ * every time React re-ran the surrounding JSX.
+ */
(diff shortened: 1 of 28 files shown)


rgbkrk:
Closing — superseded by the current investigation stack:

- Observability primitives (abort hung blob fetches, log swallowed subscriber errors) shipped as #1883 (merged).
- Design + plan captured in #1882.
- Instrumentation for the actual diagnosis in #1884 / #1886.

The branch stays accessible if anyone needs to reference the full iteration history.

PR #1881, opened 2026-04-18
feat(sync): stall detection + recovery banner (Track B)

rgbkrk:
Narrow safety-net for the widget-sync stall class. Makes two silent failure modes visible and self-healing, without changing the widget-store write path.

See the full design note at `docs/superpowers/specs/2026-04-18-widget-sync-recovery-design.md`.

## Summary

1. **`SyncEngine.syncErrors$`** — public observable forwarding each WASM auto-recovery (notebook / runtime_state / pool_state) to the UI. The engine already handles the recovery reply; this just stops the event from being silent.
2. **`SyncRecoveryBanner`** — top-level banner mounted in App.tsx. Auto-dismisses after 5s, counts bursts while visible, resets counter on dismissal.
3. **Runtime-state stall watchdog** — after a `flush_runtime_state_sync`, arm a 3s timer. Any inbound runtime-state frame clears it. On timeout, call `reset_sync_state()`, re-flush, and emit on `syncErrors$`. Clears on `resetForBootstrap`/`stop` so stale timers can't fire against replaced handles.
4. **Observability primitives** — 5s `AbortController` timeout on blob fetches (one hung fetch was poisoning the serial comm emit queue); log subscriber errors in `frame-bus` + `runtime-state` instead of silently swallowing them.

## Not in this PR

- The CRDT-first write-path refactor (Track A) that the earlier branch attempted. The investigation showed this sync-layer safety net is separable and maintainable on its own; Track A stays deferred until we see whether this is enough.
- Sync-state persistence across reloads — still the user's last-resort recovery, intentionally.

## Tests

- 4 new SyncEngine unit tests: `syncErrors$` emission per doc kind, watchdog clear-on-inbound, watchdog fire-on-timeout, watchdog clear-on-bootstrap.
- 1018 existing tests unaffected.

## Test plan

- [ ] `cargo xtask lint` clean
- [ ] `pnpm test:run` passes
- [ ] Manual: matplotlib `@interact` slider — verify the banner appears on sync-layer recovery events and the UI no longer needs a reload after a silent flush drop.

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -3,7 +3,7 @@ import { listen } from "@tauri-apps/api/event";
 import { getCurrentWebview } from "@tauri-apps/api/webview";
 import { getCurrentWindow } from "@tauri-apps/api/window";
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
-import { NotebookClient } from "runtimed";
+import { NotebookClient, type SyncErrorEvent } from "runtimed";
 import { IsolationTest } from "@/components/isolated";
 import { MediaProvider } from "@/components/outputs/media-provider";
 import { getCrdtCommWriter, setCrdtCommWriter } from "@/components/widgets/crdt-comm-writer";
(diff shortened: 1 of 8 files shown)


rgbkrk:
Closing — the approach here doesn't match what we now know.

#1881's watchdog was designed around "the daemon went silent, reset our sync state and retry." The real-world repro showed it firing every 3 seconds in a loop: resetting sync state didn't help, because the underlying wedge isn't in sync state. Instrumentation via #1884 / #1886 is aimed at localizing where the actual wedge lives.

The `syncErrors$` observable and the SyncRecoveryBanner are still potentially useful — they surface WASM auto-recovery, which does happen and is currently silent. If we want them later, they're recoverable from this branch as a smaller PR without the watchdog.

See #1882 for the updated direction.

PR #1882, opened 2026-04-18
proposal: widget sync — diagnose before architecting

rgbkrk:
The proposal now leads with "we don't actually know what the stall is, measure it first." A previous revision claimed the fix was a five-line `try/catch` — that turned out to be wrong (Tauri 2.10.3's `runCallback` doesn't unregister listeners on throw), so the listener-death theory is unproven.

The `try/catch` itself did ship as #1883 as defensive hygiene, and it'll surface an exception if one is firing — but it isn't the cure.

## Structure

- **(0) Instrumentation** — counters on both sides of the pipe plus a daemon-side correlating log, so the next repro localizes the break to outbound / inbound / WASM-state-corruption. Nothing below this is worth building until we know.
- **(0.5) Tauri unlisten verification** — conditional on (0) pointing at inbound. 15 minutes.
- **(3) Reconnect path** — conditional on (0) + (0.5). Honest about the API surface required: transport is currently fixed at `SyncEngine` construction time, so this isn't just wiring — it needs `reattach` or equivalent. Options laid out.
- **(1) Actor-ID echo detection** — separable code-smell refactor. Prototype the `get_changes_added` + per-key walk cost before committing; the existing text-attribution precedent computes "any local actor in delta," not "last writer for key K," so this is a design spike, not an extension.
- **(2) Heads-gossip** — deferred. Same dead-pipe critique as before.

## Corrections from prior drafts

Folds in two rounds of reviewer corrections:
- Tauri `runCallback` does not unregister on throw (verified in `scripts/core.js:39`).
- `automerge` 0.8.0 does have `get_changes_added` (`autocommit.rs:591`).
- Text-attribution precedent is "set of actors in delta," not "last writer per key."
- Reconnect is not "mostly wiring" — engine/transport are coupled at construction; real API surface is required.

## Test plan

N/A — documentation only.

### DIFF
--- a/docs/superpowers/specs/2026-04-18-widget-sync-next-architecture.md
+++ b/docs/superpowers/specs/2026-04-18-widget-sync-next-architecture.md
@@ -1,4 +1,4 @@
-# Widget sync: immediate fix + cleanup plan
+# Widget sync: diagnose, then redesign
 
 **Status:** Proposal. Two existing PRs are open as reference
 (#1880, #1881); neither is merged. Don't treat them as current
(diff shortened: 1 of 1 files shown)


rgbkrk:

PR #1884, opened 2026-04-18
diag: trace every frame crossing the Tauri/webview boundary

rgbkrk:
Trace instrumentation for the Tauri ↔ daemon frame pipe. No behavior change; opt-in tracing only. Goal is to localize where the sync stall reported in recent reproductions is actually wedging.

## What

Seven matched trace lines across four layers, all at `trace!` / `logger.debug` level and off by default.

**Webview (`apps/notebook/src/lib/tauri-transport.ts`):**
- `[frame-trace] js-out` — at the `invoke("send_frame")` call site.
- `[frame-trace] js-in` — at the `listen("notebook:frame")` callback, before the try/catch.

**Tauri command layer (`crates/notebook/src/lib.rs`):**
- `[frame-trace] out` — enters `send_frame_bytes` before forwarding.
- `[frame-trace] in` — in the relay's emit loop, before `emit_to_label`.

**Relay handle (`crates/notebook-sync/src/relay.rs`):**
- `relay forward_frame enter` — before sending to the bounded command channel. Includes `queue_cap` (available slots in the 32-slot channel; falling toward 0 means backpressure).
- `relay forward_frame queued` — after the command channel `send` resolves; `send_wait_ms`.
- `relay forward_frame done` — after the relay task's reply resolves; `reply_wait_ms` + ok flag.

**Relay task (`crates/notebook-sync/src/relay_task.rs`):**
- `relay outbound` — before `send_typed_frame` to the daemon socket.
- `relay outbound socket write` — after the socket write; `write_ms` + ok.
- `relay inbound` — on every frame pulled off the daemon socket; `frame_type` + `len`.

## Enable

- **Rust:** `RUST_LOG=trace` (or `RUST_LOG=notebook=trace,notebook_sync=trace` to scope).
- **Webview:** `localStorage.setItem("nteract.frameTrace", "1")` in devtools + reload. Gated locally because `logger.debug` fires an IPC per call regardless of Rust-side level.

No daemon rebuild required — all traces are on the Tauri-app side of the socket.

## What the traces can distinguish

| Observation | Likely failure location |
|---|---|
| `js-out` fires, no `out` | Tauri invoke path |
| `out` fires, no `forward_frame enter` | Never reached the relay handle (impossible given the code, but worth confirming) |
| `forward_frame enter` fires, no `queued` | Bounded mpsc channel full; relay task stalled. `queue_cap` will show 0. |
| `queued` fires, no `done` | Relay task picked it up but the socket write or daemon response didn't complete |
| `done` fires quickly, but no `relay inbound` | Daemon is accepting but has stopped sending back |
| `relay inbound` fires, no `in` | Impossible — `pipe_frame` synchronously emits |
| `in` fires, no `js-in` | Tauri event emit is losing frames to the webview |

## Test plan

- [x] `pnpm test:run` passes
- [ ] Manual: reproduce the slider stall with tracing on; grep `[frame-trace]` and localize the break.

### DIFF
--- a/apps/notebook/src/lib/tauri-transport.ts
+++ b/apps/notebook/src/lib/tauri-transport.ts
@@ -15,6 +15,25 @@ import type {
 } from "runtimed";
 import { logger } from "./logger";
 
+/**
+ * Frame-boundary tracing. OFF by default: `logger.debug` would
+ * otherwise fire an IPC to the Tauri log plugin for every frame,
+ * roughly doubling the hot-path cost on a hammered slider and
+ * potentially introducing the sync stalls this trace exists to
+ * diagnose. Enable per-session by setting
+ * `localStorage.setItem("nteract.frameTrace", "1")` in devtools
+ * and reloading. Runs alongside Rust-side `[frame-trace]` logs
+ * (controlled via `RUST_LOG=debug`) so the two can be correlated
+ * across the boundary.
+ */
+const FRAME_TRACE_ENABLED: boolean = (() => {
+  try {
+    return typeof localStorage !== "undefined" && localStorage.getItem("nteract.frameTrace") === "1";
+  } catch {
+    return false;
+  }
+})();
+
 export class TauriTransport implements NotebookTransport {
   private _connected = true;
   private unlisteners: Array<() => void> = [];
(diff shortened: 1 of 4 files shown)


rgbkrk:

PR #1885, opened 2026-04-18
test(runtimed): stall reproducer harness

rgbkrk:
Adds an `#[ignore]`-gated integration test that hammers the daemon's sync path with rapid writes to give us a daemon-side reproducer for the widget-sync stall. Lets us iterate on the stall without Tauri / webview / kernel in the loop.

## First run

3-second hammer at ~115 Hz, 347 writes from client A:

```
[stall-repro] hammered 347 writes in 3008 ms (115.4 Hz)
[stall-repro] drain: converged=true after 15 ms, last_seen="x = 347"
```

Client B converges in 15 ms. The base `NotebookDoc` sync path handles this load without wedging.

That's a useful narrowing. The widget-sync stall we've been chasing is NOT a generic "daemon sync can't keep up with rapid writes" problem. Narrows the suspect list to:

- The `RuntimeStateDoc` path specifically (different frame type, different writer flow, kernel-driven).
- Tauri's `invoke` / `listen` machinery specifically.
- Something in the kernel IOPub path that's in the loop for real widget interactions.

## What it does

- Spawns a real `runtimed` daemon in a subprocess (temp cache dir).
- Opens two clients on the same notebook.
- Hammers `update_source` at ~120 Hz for 3 seconds.
- Measures drain time for the second client.

## Why `#[ignore]`

- It's a reproducer *seeker*, not a regression guarantee. Asserting `converged == true` would turn a future intermittent stall into a flaky test rather than a loud signal.
- It hammers I/O for several seconds — not appropriate for the default test run.

## How to run

```
cargo test -p runtimed --test integration -- \
  reproduce_stall_via_rapid_edits --ignored --nocapture
```

Pair with `RUST_LOG=trace` and the `[frame-trace]` instrumentation from #1884 for a layer-by-layer timeline.

## Follow-up iterations

Noted in the commit message:
- Extend to `RuntimeStateDoc` sync (requires a client-side write API, doesn't exist in `notebook-sync` today).
- Launch a real kernel and exercise the comm-state write path.
- Run concurrent with other per-notebook traffic.

## Test plan

- [x] Test compiles and passes when run `--ignored`
- [x] Standalone on `main` (rebased off `#1884`)

### DIFF
--- a/crates/runtimed/tests/integration.rs
+++ b/crates/runtimed/tests/integration.rs
@@ -1590,3 +1590,164 @@ async fn error_blob_spill_is_renderable_by_llm_resolver() {
     assert!(tb[1].contains("500"));
     assert!(tb[1].contains("traceback frames"));
 }
+
+/// Stall reproducer — hammers the daemon's sync path with rapid
+/// source edits to see whether the observed widget-sync stall class
+/// also shows up without any Tauri / webview / kernel in the loop.
+///
+/// Signature of the user's repro (`docs/superpowers/specs/2026-04-18-*.md`):
+/// outbound `RuntimeStateSync` frames grow one byte per message
+/// because `sent_hashes` accumulates unacknowledged changes; inbound
+/// frames dry up entirely. Reload fixes it.
+///
+/// If this test trips the same class of stall (client A keeps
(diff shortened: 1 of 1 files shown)


rgbkrk:

PR #1886, opened 2026-04-18
diag: add daemon-side frame-trace

rgbkrk:
Third layer of the frame-trace instrumentation. #1884 covered the Tauri app side (webview and relay); this adds the daemon side so the next widget-sync stall repro can be walked end-to-end.

## What

Three `tracing::trace!` lines inside `handle_notebook_sync_connection` and its sync loop:

- `[frame-trace] daemon recv` — any typed frame received from a client. Includes `peer`, `notebook`, frame type, payload length.
- `[frame-trace] daemon state-sync apply` — inside the `RuntimeStateSync` handler, reports `lock_ms` for the `state_doc.write()` acquisition and `recv_ms` for `receive_sync_message_with_changes`. High `lock_ms` localizes the wedge to lock contention against the state doc.
- `[frame-trace] daemon state-sync reply sent` / `no reply` — after the reply path completes. The `no reply` case is the specific shape we've been chasing: client keeps sending RuntimeStateSync frames, daemon thinks the protocol converged, so nothing goes back — matching the user's observation of outbound frames piling up with no inbound arriving.

## How to use with the other layers

This PR plus #1884 gives a complete timeline for each frame:

```
[frame-trace] js-out type=0xXX len=N          (webview)
[frame-trace] out window=... type=0xXX len=N  (Tauri command)
[frame-trace] relay forward_frame enter       (relay handle)
[frame-trace] relay forward_frame queued      (mpsc channel)
[frame-trace] relay outbound                  (relay task picks up)
[frame-trace] relay outbound socket write     (socket write complete)
[frame-trace] daemon recv                     (daemon read from socket)  ← this PR
[frame-trace] daemon state-sync apply         (daemon processed)         ← this PR
[frame-trace] daemon state-sync reply sent    (daemon wrote reply)       ← this PR
[frame-trace] relay inbound                   (Tauri relay pulled it)
[frame-trace] in window=... type=0xXX len=N   (Tauri emit to webview)
[frame-trace] js-in type=0xXX len=N           (webview listen callback)
```

Gaps between any two consecutive lines localize the wedge precisely.

## Enable

- **Daemon:** rebuild + `RUST_LOG=notebook_sync_server=trace` (or `RUST_LOG=trace`). Required because this is in the daemon binary.
- **Tauri app:** `RUST_LOG=trace` (already on for #1884's traces).
- **Webview:** `localStorage.setItem("nteract.frameTrace", "1")` + reload.

## Test plan

- [x] `cargo check -p runtimed` clean
- [ ] Manual: with daemon rebuilt, reproduce the stall, grep `[frame-trace]` in `runtimed.log` and `notebook.log`, compare sequences.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -2791,8 +2791,13 @@ where
                                 // arrow-key slider drags. If `lock_ms` is ever
                                 // seconds long, some other task is holding
                                 // `state_doc.write()` — that's the wedge site.
+                                //
+                                // The trace itself is emitted AFTER the write guard
+                                // drops so a slow tracing subscriber can't extend
+                                // the critical section and cause the contention it
+                                // was meant to diagnose.
                                 let stage_start = std::time::Instant::now();
-                                let reply_encoded = {
+                                let (reply_encoded, lock_ms, recv_ms) = {
                                     let mut state_doc = room.state_doc.write().await;
                                     let lock_ms = stage_start.elapsed().as_millis();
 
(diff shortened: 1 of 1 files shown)


rgbkrk:

PR #1887, opened 2026-04-18
diag: slider-stall reproducer harness (headless)

rgbkrk:
Headless harness I can iterate with to reproduce the widget-sync stall — so diagnosis stops depending on someone at a keyboard.

## What

`e2e/slider-stall-repro.mjs` — standalone WebDriverIO script in the same style as `e2e/stress-sync.mjs`. Drives the Tauri app via the `e2e-webdriver` feature's built-in WebDriver server on port 4445.

Steps:

1. Waits for the notebook + kernel to be ready.
2. Rewrites the first cell to a matplotlib `@interact FloatSlider` scenario.
3. Executes the cell → slider renders inside the output iframe.
4. Focuses the slider, hammers `ArrowRight` at ~60 Hz for a configurable duration.
5. Reads both the slider value and the plot title (`sin(X.YYx)`) from the iframe DOM. They should converge within a step after hammering ends. A persistent gap is the stall signature.

Exit code is `1` on non-convergence, so a CI job or a watch loop can detect it.

## How to run

Prereqs in `e2e/README.md` apply. Quick form:

```bash
cargo build --features e2e-webdriver -p notebook
cargo xtask dev-daemon                                     # terminal 1
RUST_LOG=trace ./target/debug/notebook <scratch.ipynb>     # terminal 2
node e2e/slider-stall-repro.mjs                            # terminal 3
```

Pairs with the frame-trace instrumentation in #1884 (app/webview) and #1886 (daemon). With `RUST_LOG=trace` on both sides, running this harness produces an end-to-end `[frame-trace]` timeline in the logs that localizes the stall.

## Why standalone (not a WebdriverIO spec)

Same reasoning as `stress-sync.mjs`: it's a reproducer, not a correctness regression. Running it via the full `pnpm test:e2e:native` would miss the point — we want to poke at it, tweak parameters, watch logs.

## Test plan

- [ ] Run once end-to-end against a live app to confirm it reaches all the waits and produces either converged or stall output.

## Follow-ups if useful

- A flag to dump the `[frame-trace]` counts from the logs alongside the convergence result, so the script output is self-contained rather than needing log grepping.
- A "no kernel" variant that creates a dummy widget via direct RuntimeStateDoc writes — narrower test of just the sync path, faster feedback loop.

### DIFF
--- a/e2e/slider-stall-repro.mjs
+++ b/e2e/slider-stall-repro.mjs
@@ -2,27 +2,44 @@
 /**
  * Widget-sync stall reproducer.
  *
- * Drives the same input pattern that reproduces the slider stall in
- * manual testing — a `matplotlib @interact FloatSlider` with rapid
- * arrow-key input — but does it headlessly via WebDriver so it can be
- * iterated on without someone sitting at a keyboard.
+ * Drives the slider-arrow-key input pattern that reproduces the
+ * widget-sync stall in manual testing, headlessly via WebDriver so
+ * it can be iterated on without someone at a keyboard.
+ *
+ * Scope:
+ *
+ *   This harness drives the *input* that causes the stall and
+ *   exits when the hammer phase completes. It does NOT try to
+ *   detect convergence itself. The diagnosis lives in the
+ *   `[frame-trace]` logs from the sibling instrumentation PRs
+ *   (#1884 app/webview/relay, #1886 daemon). After a run, grep
+ *   `notebook.log` + `runtimed.log` for `[frame-trace]` and
+ *   compare outbound vs inbound counts.
+ *
+ *   Why not detect convergence in-harness? The output widget is
+ *   rendered inside a sandboxed iframe (`allow-same-origin` is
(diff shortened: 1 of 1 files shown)


rgbkrk:
Closing — actually running this revealed a hard blocker: WebKit's WebDriver implementation refuses to script into sandboxed cross-origin iframes, so `switchToFrame` throws 500s and the harness can drive the app up through "execute cell" but cannot verify the slider rendered or send keys into it.

See the conversation thread for the full walk-through (canonical build recipe is `cargo xtask e2e build`, not plain `cargo build`; binary loads the correct UI; cell selection and source-set work; iframe access is where it stops).

Takeaway for future contributors: full-stack E2E reproduction of widget-sync stalls is probably not the right tool — either the reproduction happens below the iframe layer (see #1885 for the Rust-daemon-socket shape) or it requires a dev-only app hook that dispatches keys into iframe-focused widgets, which is its own piece of work.

Commits stay on the branch as reference.

PR #1888, opened 2026-04-18
feat(dev-harness): Electron harness + Playwright for headless stall repro

rgbkrk:
## Summary

- Add `apps/dev-harness-electron/` — a dev-only Electron shell that opens the runtimed daemon's Unix socket from Node and loads the existing notebook frontend from Vite. Playwright's `_electron` API drives Chromium, which handles sandboxed output iframes cleanly (WebKit's WebDriver did not).
- Add `ElectronTransport` + a transport factory in the notebook-ui so the same React app mounts unchanged under Tauri (production) or Electron (dev harness), picking the transport based on `window.electronAPI`.
- Add a Tauri API shim in the preload so `@tauri-apps/api` imports don't crash: `invoke` is routed to Electron IPC → daemon request queue, Tauri-chrome commands return safe defaults.
- First Playwright smoke test passes end-to-end on my machine: Unix-socket handshake, notebook-ui loads, `#root` renders, zero renderer exceptions.

## Why

PR #1887 hit a hard wall: `tauri-plugin-webdriver` can't `switchToFrame` into the sandboxed output iframe where ipywidgets renders. Without iframe drilling we can't drive the slider that reproduces the widget-sync stall we've been chasing in #1880/#1881/#1884/#1886. Playwright + Chromium does iframe drilling natively.

We considered a dev-only WebSocket bridge first but rejected it on security grounds — a localhost WS listener is reachable from any browser tab (see https://nteract.io/blog/security). Electron's main process is Node and opens the existing Unix socket directly, so no new network listener is introduced.

## Architecture

```
┌─────────────────────────────────┐                                   ┌──────────┐
│  Electron dev harness           │                                   │          │
│                                 │                                   │          │
│  ┌──────────┐  IPC   ┌───────┐  │   Unix socket (existing)          │ runtimed │
│  │ Renderer │◄──────►│ Main  │──┼──────────────────────────────────►│ (daemon) │
│  │ (React,  │        │ (Node)│  │                                   │          │
│  │  Chromium│        └───────┘  │                                   └──────────┘
│  │  iframe) │                   │
│  └──────────┘                   │
└─────────────────────────────────┘
          ▲
          │ CDP (Playwright _electron)
          ▼
    ┌──────────────┐
    │  Playwright  │
    │  test script │
    └──────────────┘
```

- **Main** (`src/main/index.js`) ports the wire protocol in `crates/notebook-protocol/src/connection.rs` to Node: 5-byte preamble, length-prefixed typed frames, one-at-a-time Request/Response queue mirroring `crates/notebook-sync/src/relay_task.rs`.
- **Preload** (`src/preload/index.js`) uses `contextBridge` to expose `window.electronAPI` (sendFrame / onFrame / sendRequest) and a `window.__TAURI_INTERNALS__` shim (invoke routed to main, metadata stub).
- **Frontend** gets a new `ElectronTransport` next to `TauriTransport` with identical surface. A factory in `apps/notebook/src/lib/transport.ts` picks based on `window.electronAPI`. Swap happens at two construction sites only — `App.tsx` and `useAutomergeNotebook.ts`.

## Security posture

| Check | Status |
|---|---|
| New network listener | **No** — Unix socket only, reuses existing daemon socket (mode 0600) |
| contextIsolation | `true` |
| nodeIntegration | `false` |
| sandbox | `true` |
| Release artifact includes harness? | **No** — separate workspace package; `pnpm --filter notebook-ui build`, `cargo xtask build-app`, `install-nightly`, `.mcpb` all ignore it |
| Electron in production bundle? | **No** — devDependency of the harness package only |

The notebook-ui side of the change is ~50 lines of `ElectronTransport` (pure shim over `window.electronAPI`), a 10-line transport factory, and two one-line swaps. No Electron runtime is bundled into Tauri.

## Usage

```bash
# Term 1 — dev daemon
cargo xtask dev-daemon

# Term 2 — notebook frontend (Vite dev server)
RUNTIMED_VITE_PORT=5176 pnpm --filter notebook-ui dev

# Term 3 — Electron harness
RUNTIMED_VITE_PORT=5176 pnpm --filter @nteract/dev-harness-electron dev
```

Playwright:
```bash
RUNTIMED_VITE_PORT=5176 pnpm --filter @nteract/dev-harness-electron test:e2e
```

## What's deliberately not here

- **Full Tauri-API coverage.** The shim handles the boot path and daemon requests. Things like auto-updater, window-state persistence, and file dialogs are no-ops with a one-time warning log per unknown command. I'll extend on demand as real tests surface gaps.
- **The stall repro itself.** This PR lands the harness infrastructure. The actual slider-drives-stall test comes in a follow-up now that we have a path for it.
- **`cargo xtask dev-harness`.** The plan called for an orchestrator subcommand. Skipped for now — the three-terminal manual workflow is fine for iteration; I'll add xtask once we know we're living with this harness.

## Test plan

- [x] `pnpm --filter @nteract/dev-harness-electron install` clean
- [x] Playwright smoke test passes: `#root` appears, no renderer errors, no pageerrors
- [x] Typecheck clean (`tsc -b` on notebook-ui)
- [ ] **Reviewer**: verify `pnpm --filter notebook-ui build` output contains no `electron` references
- [ ] **Reviewer**: verify `cargo xtask build-app` bundle contains no Electron runtime
- [ ] **Reviewer**: sanity-check the wire protocol port in `src/main/index.js` against `crates/notebook-protocol/src/connection.rs`

1. Slider-stall Playwright test: open `notebooks/ipywidgets-demo.ipynb` via `HARNESS_NOTEBOOK_PATH`, launch kernel, drive arrow keys into the IntSlider, assert inbound/outbound frame counters from #1886 keep advancing.
2. `cargo xtask dev-harness` that starts daemon + Vite + Electron in one command.
3. Extend the Tauri shim as the harness hits more unknown commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,11 @@ dist/
 **/dist/
 *.tsbuildinfo
 
+# Dev-only Electron harness artifacts
+apps/dev-harness-electron/test-results/
+apps/dev-harness-electron/playwright-report/
+apps/dev-harness-electron/.cache/
+
 # Generated isolated-renderer bundle (built from src/isolated-renderer)
 apps/notebook/public/isolated/
 
--- a/apps/dev-harness-electron/README.md
+++ b/apps/dev-harness-electron/README.md
@@ -1,0 +1,36 @@
+# @nteract/dev-harness-electron
+
+**Dev-only.** Never shipped. Never referenced by the production Tauri build.
+
+Electron shell that mounts the same notebook frontend Tauri mounts, but opens the daemon's Unix socket directly from the main process so Playwright (on Chromium) can drive the UI headlessly. Exists because Tauri's WebKit runtime blocks WebDriver from stepping into sandboxed output iframes, which stopped us from reproducing the widget-sync stall end-to-end.
+
+## Run it
+
+Prereqs: repo-level `pnpm install`, direnv active, dev daemon running.
+
+```bash
+# Terminal 1 — daemon (or use `up` from nteract-dev MCP)
+cargo xtask dev-daemon
+
+# Terminal 2 — notebook frontend (Vite dev server on :5174)
+pnpm --filter notebook-ui dev
+
+# Terminal 3 — Electron harness
+pnpm --filter @nteract/dev-harness-electron dev
+```
+
+The harness launches Electron, opens the dev daemon's Unix socket from the main process, and loads `http://localhost:5174` in a `BrowserWindow`. The renderer detects `window.electronAPI` and uses `ElectronTransport` instead of `TauriTransport`; everything else is the same frontend code.
+
+## Security
+
+- No network listener is added. The main process reads the existing `runtimed` Unix socket (mode 0600, user-scoped) via Node's `net.createConnection`.
+- `BrowserWindow` runs with `contextIsolation: true`, `nodeIntegration: false`, `sandbox: true`. The preload's `contextBridge` surface is the only path between renderer JS and Node.
+- This package is not compiled into any release artifact. `pnpm --filter notebook-ui build`, `cargo xtask build-app`, `cargo xtask install-nightly`, and the `.mcpb` packager all ignore it.
+
+## Playwright
+
+```bash
+pnpm --filter @nteract/dev-harness-electron test:e2e
+```
+
+Tests live in `tests/`. They use Playwright's `_electron` API to launch the harness, then drive the renderer with full Chromium sandbox-iframe support.
--- a/apps/dev-harness-electron/fixtures/int-slider.ipynb
+++ b/apps/dev-harness-electron/fixtures/int-slider.ipynb
@@ -65,7 +65,7 @@
             "application/vnd.jupyter.widget-view+json": {
               "version_major": 2,
               "version_minor": 0,
-              "model_id": "7fb7fd25cd9a47e58b60bcbec21ea683"
+              "model_id": "f8811edcea094a0eb114736f1e172738"
             },
             "text/plain": "IntSlider(value=0, description='value', max=1000000)"
           },
@@ -88,10 +88,10 @@
         {
           "output_type": "stream",
           "name": "stdout",
-          "text": "kernel_slider_value=55650\n"
+          "text": "kernel_slider_value=53283\n"
         }
       ],
-      "execution_count": 13
+      "execution_count": 11
     }
   ]
 }
--- a/apps/dev-harness-electron/fixtures/int-slider.ipynb
+++ b/apps/dev-harness-electron/fixtures/int-slider.ipynb
@@ -48,7 +48,7 @@
       ],
       "metadata": {},
       "outputs": [],
-      "execution_count": null
+      "execution_count": 1
     },
     {
       "id": "slider-cell",
@@ -60,18 +60,19 @@
       "metadata": {},
       "outputs": [
         {
-          "output_type": "error",
-          "ename": "NameError",
-          "evalue": "name 'widgets' is not defined",
-          "traceback": [
-            "\u001b[31m---------------------------------------------------------------------------\u001b[39m",
-            "\u001b[31mNameError\u001b[39m                                 Traceback (most recent call last)",
-            "\u001b[36mCell\u001b[39m\u001b[36m \u001b[39m\u001b[32mIn[1]\u001b[39m\u001b[32m, line 1\u001b[39m\n\u001b[32m----> \u001b[39m\u001b[32m1\u001b[39m slider = widgets.IntSlider(value=\u001b[32m0\u001b[39m, min=\u001b[32m0\u001b[39m, max=\u001b[32m1000000\u001b[39m, step=\u001b[32m1\u001b[39m, description=\u001b[33m\"value\"\u001b[39m)\n\u001b[32m      2\u001b[39m display(slider)\n",
-            "\u001b[31mNameError\u001b[39m: name 'widgets' is not defined"
-          ]
+          "output_type": "display_data",
+          "data": {
+            "application/vnd.jupyter.widget-view+json": {
+              "version_major": 2,
+              "version_minor": 0,
+              "model_id": "bec52a5c78cf458d9cd93df09ae30c75"
+            },
+            "text/plain": "IntSlider(value=0, description='value', max=1000000)"
+          },
+          "metadata": {}
         }
       ],
-      "execution_count": 1
+      "execution_count": 2
     }
   ]
 }
--- a/apps/dev-harness-electron/fixtures/int-slider.ipynb
+++ b/apps/dev-harness-electron/fixtures/int-slider.ipynb
@@ -65,7 +65,7 @@
             "application/vnd.jupyter.widget-view+json": {
               "version_major": 2,
               "version_minor": 0,
-              "model_id": "f8811edcea094a0eb114736f1e172738"
+              "model_id": "b77aa99cd9ae42dcbc8e9d14b7f669c3"
             },
             "text/plain": "IntSlider(value=0, description='value', max=1000000)"
           },
@@ -88,10 +88,10 @@
         {
           "output_type": "stream",
           "name": "stdout",
-          "text": "kernel_slider_value=53283\n"
+          "text": "kernel_slider_value=569\n"
         }
       ],
-      "execution_count": 11
+      "execution_count": 15
     }
   ]
 }
--- a/apps/dev-harness-electron/fixtures/int-slider.ipynb
+++ b/apps/dev-harness-electron/fixtures/int-slider.ipynb
@@ -1,0 +1,77 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 5,
+  "metadata": {
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3",
+      "language": "python"
+    },
+    "language_info": {
+      "name": "python",
+      "version": "3.12.0",
+      "codemirror_mode": null
+    },
+    "uv": {
+      "dependencies": [
+        "ipywidgets>=8.0"
+      ],
+      "requires-python": ">=3.10"
+    },
+    "runt": {
+      "schema_version": "1",
+      "uv": {
+        "dependencies": [
+          "ipywidgets>=8.0"
+        ],
+        "requires-python": ">=3.10"
+      }
+    }
+  },
+  "cells": [
+    {
+      "id": "heading",
+      "cell_type": "markdown",
+      "source": [
+        "# IntSlider fixture\n",
+        "\n",
+        "Minimal notebook for reproducing the widget-sync stall under the Electron harness."
+      ],
+      "metadata": {}
+    },
+    {
+      "id": "imports",
+      "cell_type": "code",
+      "source": [
+        "import ipywidgets as widgets\n",
+        "from IPython.display import display"
+      ],
+      "metadata": {},
+      "outputs": [],
+      "execution_count": null
+    },
+    {
+      "id": "slider-cell",
+      "cell_type": "code",
+      "source": [
+        "slider = widgets.IntSlider(value=0, min=0, max=1000000, step=1, description=\"value\")\n",
+        "display(slider)"
+      ],
+      "metadata": {},
+      "outputs": [
+        {
+          "output_type": "error",
+          "ename": "NameError",
+          "evalue": "name 'widgets' is not defined",
+          "traceback": [
+            "\u001b[31m---------------------------------------------------------------------------\u001b[39m",
+            "\u001b[31mNameError\u001b[39m                                 Traceback (most recent call last)",
+            "\u001b[36mCell\u001b[39m\u001b[36m \u001b[39m\u001b[32mIn[1]\u001b[39m\u001b[32m, line 1\u001b[39m\n\u001b[32m----> \u001b[39m\u001b[32m1\u001b[39m slider = widgets.IntSlider(value=\u001b[32m0\u001b[39m, min=\u001b[32m0\u001b[39m, max=\u001b[32m1000000\u001b[39m, step=\u001b[32m1\u001b[39m, description=\u001b[33m\"value\"\u001b[39m)\n\u001b[32m      2\u001b[39m display(slider)\n",
+            "\u001b[31mNameError\u001b[39m: name 'widgets' is not defined"
+          ]
+        }
+      ],
+      "execution_count": 1
+    }
+  ]
+}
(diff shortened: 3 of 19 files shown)


rgbkrk:
## Codex review round 1 — addressed

Pushed [`07a5ba43`](../commit/07a5ba43). Fixes for each finding:

**MUST FIX**
- ✅ **Frame-drop race.** Main now buffers inbound frames in `pendingFrames` until the renderer calls `dev-harness:ready` from `ElectronTransport.onFrame`. Mirrors Tauri's `notify_sync_ready` gate.
- ✅ **`--renderer-url` security.** Restricted to localhost origins; added `will-navigate` and `setWindowOpenHandler` guards so a renderer can't be redirected to a non-local origin and inherit preload IPC.
- ⏳ **Tauri APIs in renderer (`startWindowFocusHandler`, `fromTauriEvent`, `webview.listen`).** Deferred — these all silently no-op via the shim today (`__TAURI_EVENT_PLUGIN_INTERNALS__.unregisterListener` is a no-op, `plugin:event|listen` through invoke returns undefined). No renderer errors. The correct long-term fix is a host abstraction (`useDaemonLifecycle`, `useWindowFocus`) — tracked as follow-up in this PR's description. None of these are on the widget-sync-stall hot path.
- ✅ **Blob port.** Verified in this run — `get_blob_port` returns the port from daemon status JSON, no "no-op for unknown command" warning anymore.

**NICE-TO-HAVE**
- ✅ **Unsolicited Response frames dropped**, not forwarded to renderer.
- ✅ **Frame size caps** (100 MiB outer / 1 MiB Presence) in `FrameBuffer.tryTakeOne` matching `crates/notebook-protocol/src/connection.rs`.
- ✅ **`protocol: "v2"`** now sent on `notebook_sync` handshake.

Smoke test still passes:
```
[main] handshake ok: { notebook_id: '387892f8-…', cell_count: 0 }
[test] harness info={"notebookId":"387892f8-…","cellCount":0}
  ✓ tests/smoke.spec.ts:16:5 (846ms)
```

Remaining open item from the review: the `open_notebook.spec.ts` I added returns `cell_count: 0` when pointing at `notebooks/ipywidgets-demo.ipynb`. Daemon reformats the file on open (seen in git diff) but the handshake response doesn't reflect its cells. Separate daemon-side investigation, not a blocker for this PR.

rgbkrk:
## Codex review round 2 — addressed

Pushed [`e769a17f`](../commit/e769a17f).

**P1 — NotebookClient request shape** (MUST FIX)
- ✅ `NotebookClient.sendRequest({ type: "launch_kernel", ... })` was going out on the wire with `type` instead of the serde tag `action`, so every kernel-lifecycle call (launch, execute, interrupt, run-all, send-comm) silently failed at the daemon.
- Main now runs `toWireRequest()` on every `send-request`: `{ type, ... }` → `{ action, ... }`, and `type: "interrupt"` → `action: "interrupt_execution"`. Tauri-shim path (which already builds `{ action }`) passes through unchanged.

**P2 — Tauri event subscriptions** (nice-to-have, deferred)
- The renderer's `webview.listen("daemon:ready")`, `"daemon:disconnected"`, `"cells:outputs_cleared"` go through `plugin:event|listen` which hits the generic no-op. Net effect: lifecycle events never fire in the harness.
- Not on the widget-stall hot path. For disconnect specifically, the harness already surfaces it via `window.electronAPI.onFrame` → `__daemonDisconnected` sentinel in `ElectronTransport`. Fleshing out a full Tauri event bridge is a follow-up.

**Test state**
```
✓ tests/smoke.spec.ts                           (5.2s)
✓ tests/open-notebook.spec.ts (90 cells)        (10.0s)
- tests/slider-stall.spec.ts (skipped: uv:inline requires a prepared environment)
  — reaches launch_kernel → execute_cell → looks for iframe; iframe never materializes
    because the env prep step isn't declared in the fixture. Separate follow-up.
```

Summary of state: harness smoke-tests green, notebook files open and render 90 cells under Electron, full `NotebookClient` request path works end-to-end now that `action`/`type` are reconciled. Still draft until the slider-stall env prep is sorted and the remaining Tauri event bridge lands — but the wire-protocol and security findings from both codex passes are fully addressed.

rgbkrk:
## Progress update

Landed [`08cf1bce`](../commit/08cf1bce): fixture + kernel-launch path for stall repro.

- Added `apps/dev-harness-electron/fixtures/int-slider.ipynb` with `metadata.uv.dependencies: ["ipywidgets>=8.0"]` (top-level legacy form; nesting under `metadata.runt.uv` trips the trust check into `NoDependencies`).
- Slider test now walks: launch_kernel ✓ → kernel_launched (uv:inline, cached env hit) → execute_cell ✓ → cell_queued. Widget iframe appears in the cell DOM.
- Inside-iframe drilling (slider control + arrow-key drive) still pending — the iframe mounts but the widget's DOM hasn't populated by the time we query. Likely the comm state isn't reaching the iframe's WidgetStore via the postMessage bridge yet; follow-up investigation in the next iteration.

End-to-end path now validated headlessly:

```
Playwright → Electron main (Node)
           → Unix socket → runtimed
           → agent → jupyter kernel (uv:inline)
           → execute → comm_open → iopub
           → back to renderer via RuntimeStateDoc sync (0x05)
           → WidgetStore → IsolatedRenderer iframe
```

Everything through `WidgetStore` works; the last mile (iframe's widget mount) is the remaining gap. That's where the actual stall repro lives anyway, so getting there is the whole point of this harness.

rgbkrk:
## 🎉 Slider stall repro scaffold now passes end-to-end

Landed [`989e38e7`](../commit/989e38e7).

Added an inline-widgets bypass in the harness. When `window.__NTERACT_DEV_HARNESS_INLINE_WIDGETS__` is set by the Electron preload, `OutputArea` skips iframe isolation for widget-only outputs. The existing `MediaProvider` renderer in `App.tsx` then mounts `<WidgetView />` directly in parent DOM, where Playwright can drive controls natively.

```
✓ drive an ipywidgets IntSlider and observe frame flow (2.4s)
  [test] slider host: iframe=true direct=true
  [test] 200 ArrowRight presses in 282ms
  [test] slider value after drive: 200
```

- Flag is set only by the dev-harness preload. Tauri's preload doesn't expose it — production widget isolation is unchanged.
- Scope: for harness repro of built-in ipywidgets controls. Anywidget custom ESM and untrusted HTML still keep iframe isolation in production.

200 presses didn't hit the stall on a single run — not surprising since the real bug is intermittent. The infrastructure is now in place to run the drive at higher rates, larger ranges, and in a loop. That's follow-up iteration; the harness itself is done.

## PR state

Six commits, all green tests. Ready for review.

```
092edaa2 test: wait for DOM cells instead of handshake cell_count
e769a17f fix: translate {type} → {action} on NotebookClient requests
08cf1bce test: fixture + kernel-launch path for slider stall repro
c059e8a8 test: diagnose iframe host state in slider-stall
989e38e7 feat: inline widget rendering bypass — slider test passes
```

rgbkrk:
## Stall hunt: harness drive-stress results

Answered your question: **no**, execution never got stuck across any drive pattern I tried.

### Keyboard stress — 20 rounds × 800 presses, chaos mode on

```
round 1: drove 800 in 1474ms → displayed=342 kernel=342 diverged=false
…
round 20: drove 800 in 1332ms → displayed=5746 kernel=5746 diverged=false
16,000 keystrokes total, 0/20 rounds diverged
```

### Mouse drag — 10 rounds × 3-5s each, back-and-forth

```
round 1: drove in 3331ms → displayed=1412 kernel=1412 diverged=false
…
round 10: drove in 3567ms → displayed=55650 kernel=55650 diverged=false
~50s of continuous drag, values up to 55k, 0/10 rounds diverged
```

Every single round, `slider.value` in the kernel matched `aria-valuenow` in the DOM.

### Candidate explanation

The inline-widgets bypass we added in [`989e38e7`](../commit/989e38e7) routes built-in widgets *directly to parent DOM via MediaProvider*, skipping the iframe + postMessage CommBridgeManager entirely. If the original back-and-forth stall lives in that bridge specifically, this harness has accidentally engineered around the exact code path that breaks.

The original reason we added the bypass was that the iframe's bootstrap postMessage never fully worked under Electron — the iframe rendered at 0x0 with no widget inside. So next iteration to actually repro the Tauri stall:

1. Figure out why the iframe's bootstrap postMessage doesn't complete under Electron (parent ↔ iframe handshake that loads the React renderer bundle)
2. Remove the inline-widgets bypass (or make it opt-in via a second flag)
3. Rerun this stress test against the iframe path

That's follow-up. For this PR, the harness is end-to-end green and ready for review — it lands the infrastructure Kyle asked for and demonstrates via stress testing that the `ElectronTransport` + daemon round-trip is solid.

Full commit log:

```
989e38e7 feat: inline widget rendering bypass — slider test passes
a054aa05 test: kernel round-trip verification for stall repro
e5dd3f1c test: stress modes — mouse drag + chaos keyboard
```

rgbkrk:
## Big finding: the iframe path works under Electron

I retracted my earlier claim that the iframe bootstrap was broken under Electron. Added [tests/iframe-bootstrap.spec.ts](../blob/dev/electron-harness/apps/dev-harness-electron/tests/iframe-bootstrap.spec.ts) which disables the bypass (`HARNESS_INLINE_WIDGETS=0`) and traces the parent ↔ iframe postMessage handshake. Result:

```
parent received (24 entries):
  nteract/evalResult x7        (CSS + JS bundle injected)
  nteract/widgetReady x5
  nteract/rendererReady x4
  resize x3
  render_complete x4
iframe __REACT_RENDERER_ACTIVE__ = true
iframe dims: 1240 x 24
```

Full bootstrap completes. My earlier "iframe is 0x0" observations were timing — checking too early, before the bundle injection round-trip finished.

### Stress test through the iframe path

With `HARNESS_INLINE_WIDGETS=0 HARNESS_STALL_MODE=mouse HARNESS_STALL_ROUNDS=8`:

```
round 1: drove in 4916ms → displayed=12709 kernel=12709 diverged=false
round 2: drove in 3847ms → displayed=13940 kernel=13940 diverged=false
…
round 8: drove in 3562ms → displayed=53283 kernel=53283 diverged=false
~35s of mouse drag, slider to 53k, 0/8 rounds diverged
```

The iframe-based rendering path under Electron is **equally solid** as the inline path. No stall.

### Conclusion for this PR

- The harness + kernel round-trip infrastructure works in **both** paths (inline and iframe).
- Sync doesn't break under heavy keyboard or mouse drive in either path.
- The manual back-and-forth stall Kyle saw is not reproducible via Playwright-driven input in this environment.

Three remaining possibilities for why we can't hit it headlessly:
1. **Platform-specific**: stall is in Tauri/WebKit-specific code (e.g., a WebKitWebView postMessage throttle) that Chromium handles cleanly.
2. **Input-driver-specific**: Playwright's keyboard/mouse events don't reproduce the exact event cadence human input produces (key repeat timing, trackpad-specific event patterns).
3. **Fixed in flight**: between when Kyle saw the stall and now, some commit on main closed the underlying race.

All three are separate investigations from the harness scaffolding this PR lands. **The harness itself is done** and proves two things we previously didn't know:

- The Chromium+Electron+Unix-socket path for the full notebook frontend works end-to-end.
- Widget sync under sustained load (thousands of events) does NOT spontaneously break in this path.

If the stall is platform-specific to Tauri, we'd need a separate investigation in the production app. If not, the harness gives us a headless surface to stress it with different patterns in follow-ups.

Commits since the last comment: `8e42247d` (iframe bootstrap proof + sync-IPC for config).

rgbkrk:
Closing this PR — spiking the Electron harness design taught us a lot but the final shape shouldn't live as a harness-only patch on top of Tauri-coupled code. Reopening the work as a three-PR series aimed at making the notebook frontend genuinely host-agnostic (Tauri today, Electron next, other embedders in the future).

## What we validated on this branch

- The full notebook frontend boots inside an Electron renderer pointed at the same Vite dev server the Tauri app uses.
- Electron main can open the daemon's Unix socket from Node and speak the existing wire protocol. Preamble, length prefix, typed frames, handshake, request/response all round-trip identically to the Tauri relay.
- Playwright's `_electron.launch()` drives the renderer reliably; mouse drag + keyboard spam work in both the parent DOM and inside sandboxed output iframes.
- The iframe bootstrap handshake (`eval` CSS + JS bundle → `renderer_ready` → `widgetReady` → `widgetSnapshot`) completes cleanly under Electron. My earlier "iframe is 0x0" was a timing misread.
- An `@interact` + matplotlib fixture reliably reproduces the widget backpressure signature Kyle saw manually: slider visible position gets ahead of what the kernel has rendered, because `plt.figure(...)` per frame is slower than the comm_msg firehose.

## Why we're not merging this as-is

- `src/components/cell/OutputArea.tsx` grew a harness-only runtime check to skip iframe isolation for widgets. That's shared production code; it shouldn't branch on a dev flag.
- `apps/dev-harness-electron/src/main/index.js` grew a `tauri-shim:invoke` router that reimplements 15 Tauri commands as a compatibility layer. Every time we touch the real Tauri command surface, the shim drifts. The right fix is to move those calls behind an abstraction the frontend owns.
- The frontend still imports `@tauri-apps/api/core` / `window` / `webview` / plugin-log / plugin-updater / plugin-dialog / plugin-shell from ~18 different places. Any headless consumer (Electron today, browser-hosted viewer tomorrow) will have to re-solve this every time.

## Salvage list (all referenced in this PR's history)

- Wire protocol port to Node: `apps/dev-harness-electron/src/main/index.js` — preamble, `FrameBuffer.tryTakeOne`, `DaemonConnection` (serialized request/response queue mirroring `crates/notebook-sync/src/relay_task.rs`), size caps from `crates/notebook-protocol/src/connection.rs`.
- Preload `electronAPI` bridge shape (contextBridge, sandbox:true, onFrame / sendFrame / sendRequest): `apps/dev-harness-electron/src/preload/index.js`.
- Playwright `_electron.launch()` scaffolding + verbose main/renderer log capture: `apps/dev-harness-electron/tests/smoke.spec.ts`, `open-notebook.spec.ts`, `iframe-bootstrap.spec.ts`.
- Backpressure fixture + probe cell for kernel round-trip: `apps/dev-harness-electron/fixtures/int-slider.ipynb`.
- Catalog of every `invoke(...)` call site that needs an abstraction — see the `tauriShimNoop` + `mapTauriCommandToNotebookRequest` functions in main/index.js.
- Two codex review passes of findings already addressed (race, size caps, protocol:v2, {type}→{action}, localhost guard).

## Next steps (each its own PR)

1. **`packages/notebook-host`** — interface + `createTauriHost()` (no behavior change, just wrap what exists).
2. **Migrate direct `invoke(...)` and `webview.listen(...)` call sites** onto `useNotebookHost()` — one hook at a time.
3. **`createElectronHost()`** — the real Electron implementation, with the harness rebuilt as a thin consumer. At that point the stall investigation becomes a byproduct of having clean Electron support, not the driver.

PR #1896, opened 2026-04-19
fix(widgets): break echo amplification loop in slider comm state sync

quillaid:
## Summary

- **Root cause:** Stale kernel echoes round-tripping through the CRDT created exponential message amplification on the ZMQ shell channel during rapid slider arrow-key input. The runtime agent re-forwarded echoes as if they were new frontend changes, burying the kernel under hundreds of stale `comm_msg(update)` messages.
- **Daemon fix:** Track values last forwarded to the kernel per comm. `filter_echo_keys()` strips CRDT diff keys matching previously sent values, breaking the amplification loop.
- **Frontend fix:** Keep `optimisticKeys` alive for 500ms after CRDT flush (was: cleared immediately). Covers the full round trip so stale echoes are suppressed before reaching the WidgetStore.
- Adds E2E spec (`widget-slider-stall.spec.js`) and test fixture (`16-widget-slider.ipynb`) to reproduce the stall via rapid alternating arrow keys on a FloatSlider widget.
- 4 new unit tests for `filter_echo_keys`.

## Test plan

- [x] `cargo test -p runtimed --lib runtime_agent` — 8/8 tests pass (4 new)
- [x] `cargo xtask lint --fix` — all checks pass
- [x] Nightly daemon installed with fix (`2.2.0+28e7722`)
- [ ] E2E: `cargo xtask e2e test-fixture 16-widget-slider` — verify slider stall repro
- [ ] Manual: open `@interact` FloatSlider, rapidly alternate left/right arrow keys for 5s, verify kernel remains responsive to new cell execution

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -363,6 +363,22 @@ function AppContent() {
     };
   }, [getHandle, triggerSync]);
 
+  // Expose widget update pipeline on window for E2E tests.
+  // Allows tests to drive slider values through the real code path
+  // (store update + debounced CRDT write + echo suppression) without
+  // needing to reach into the security-isolated iframe.
+  useEffect(() => {
+    const w = window as unknown as Record<string, unknown>;
+    w.__nteractWidgetUpdate = (commId: string, patch: Record<string, unknown>) => {
+      updateManager.updateAndPersist(commId, patch);
+    };
+    w.__nteractWidgetStore = widgetStore;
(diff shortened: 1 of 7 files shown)


rgbkrk:
Superseded by #1900 — same bug, cleaner architectural fix: kernel actor on the coalescing writer + per-field foreign view so per-comm-field echoes are filtered without relying on frontend-side dedup tracking.

Issue #1667, opened 2026-04-09
refactor(notebook-doc): unified DocChangeset from single patch walk


rgbkrk:
## Problem

Three functions independently call `doc.diff(before, after)` and walk the same patches:

| Function | Crate | Calls `doc.diff()`? | Produces |
|----------|-------|---------------------|----------|
| `diff_cells()` | notebook-doc | Yes | `CellChangeset` |
| `diff_metadata_touched()` | notebook-doc | Yes | `bool` |
| `compute_text_attributions()` | runtimed-wasm (local fn) | Yes | `Vec<TextAttribution>` |

The WASM `receive_frame()` path calls `diff_cells` + `compute_text_attributions` = **2 `doc.diff()` calls with identical patches**. The daemon calls `diff_metadata_touched` separately.

Additionally, the WASM path invalidates `metadata_fingerprint_cache` on every doc change (including cell source edits and output streaming), forcing re-serialization ~30/sec during execution. See the TODO at `runtimed-wasm/src/lib.rs:1440`.

## Proposal

One function in `notebook-doc` that diffs once, walks patches once, classifies everything:

```rust
pub struct DocChangeset {
    pub cells: CellChangeset,
    pub metadata_changed: bool,
    pub text_patches: Vec<TextPatch>,  // raw splice/delete per cell source
}

pub fn diff_doc(
    doc: &mut AutoCommit,
    before: &[ChangeHash],
    after: &[ChangeHash],
) -> DocChangeset
```

### `text_patches`

Captures the raw splice/delete operations on cell source text — the same data `compute_text_attributions` currently extracts, but without actor labels (those come from `extract_change_actors` via `get_changes`, a separate query). The WASM consumer combines text patches + actors into `TextAttribution`.

### Consumer changes

**WASM (`runtimed-wasm`):**
```rust
let changeset = diff_doc(doc, &before, &after);
let actors = extract_change_actors(doc, &before);
let attributions = build_attributions(&changeset.text_patches, &actors);

if changeset.metadata_changed {
    self.metadata_fingerprint_cache = None;  // fixes the TODO
}
```

**Daemon (`runtimed`):**
```rust
let changeset = diff_doc(doc, &before, &after);
if changeset.metadata_changed {
    check_and_broadcast_sync_state(room).await;
}
```

### What moves where

- Patch-walking guts of `compute_text_attributions` move into `diff_doc()` in notebook-doc
[...]

PR #1910, opened 2026-04-19
docs: spec for map-keyed execution outputs

rgbkrk:
## Summary

Proposal doc to replace `executions[eid].outputs` from a pure Automerge `List<manifest>` to a `Map<output_id, { seq, manifest, display_id? }>`.

## Why now

- The `DuplicateSeqNumber` bug that caused missing outputs in #1905 is specifically a list-insert-under-concurrent-forks failure mode. The fix there (unique actor per fork) papers over it. Map-keyed outputs structurally avoid the class of bug on this code path.
- `update_output_by_display_id` is currently an O(N) scan over every execution's output list. With a first-class `display_id` index it becomes a direct lookup.
- Outputs are kernel-emission-ordered — nothing reorders them — so the List CRDT's RGA ordering guarantees aren't buying anything.
- Cells already learned this lesson. They're a `Map<cell_id, Cell>` with fractional indexing for exactly this reason.

## Scope

Internal CRDT storage change only. No on-the-wire changes, no `.ipynb` changes, no change to how the frontend consumes outputs. Migration plan lives behind an env-var feature flag with a rollback that's literally "unset the flag, restart the daemon."

## What's in the doc

- Schema proposal + rationale for monotonic seq vs fractional index.
- Concurrent merge semantics walk-through (appends, display_id updates, stream upserts, execution deletes).
- Migration plan with feature flag.
- What becomes easy (O(1) display_id lookup, per-output change diffs that extend the unified `DocChangeset` from #1667).
- What this *doesn't* solve (doesn't eliminate the concurrent-fork-actor-reuse invariant for other mutations).
- Tradeoffs and a test matrix.

Reviewing this is reading a design doc, not a diff. No code ships in this PR.

### DIFF
--- a/docs/superpowers/specs/2026-04-19-map-keyed-outputs.md
+++ b/docs/superpowers/specs/2026-04-19-map-keyed-outputs.md
@@ -1,0 +1,120 @@
+# Map-keyed execution outputs
+
+**Status:** Proposed
+**Date:** 2026-04-19
+**Related:** #1905 (DuplicateSeqNumber root cause), #1558 (outputs moved to `RuntimeStateDoc`), #1667 (unified `DocChangeset`)
+
+## Motivation
+
+`RuntimeStateDoc::executions[eid].outputs` is a pure Automerge `List` of manifests. Lists are great for user-editable sequences (cells in a notebook, characters in source text) where stable ordering is a user-visible property. Output streams are not that. Three recent patterns make the list choice look costly:
+
+1. **List-insert under concurrent forks is fragile.** The IOPub handler forks the state doc, inserts an output, releases the lock, does async work (blob-store, manifest creation), re-acquires the lock, and merges. If two forks share an Automerge actor ID, the second merge returns `DuplicateSeqNumber` and the insert is silently dropped. That's what the missing-output bug in #1905 was. The fix (one actor per fork) papers over the class of bug, but the underlying sharpness — "reordering a list under a CRDT is more subtle than it looks" — remains.
+
+2. **`replace_output` and `update_display_data` are O(N) scans.** Finding the output to replace means walking the list looking for a matching `display_id`. With a map keyed on something addressable, this is a direct lookup.
+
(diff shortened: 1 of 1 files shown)


rgbkrk:
Codex stress-test — 3 high, 2 medium. Spec needs rework before implementing.

**High — ordering undefined under concurrent writes.** `sort_by_key(seq)` only works if seq is unique per execution. In the fork+merge model, two forks can derive the same "next seq" from the same base heads and append distinct outputs with identical seq. Map semantics preserve both entries, but read-time ordering becomes map iteration order, not emission order. Need an explicit tie-breaker or a non-CRDT monotonic allocator.

**High — `display_id` semantics regression.** Current behavior is intentionally one-to-many: `update_output_by_display_id_with_manifests()` updates all matching outputs across executions (kernel_manager.rs:300, :528). A `display_id` field on each map entry doesn't give O(1) lookup — still needs a secondary index `display_id -> set<output_id>` or a scan.

**High — stream coalescence likely broken.** Coalescence today only updates in place if the stream output is still the tail of the output list (runtime_state.rs:1232, jupyter_kernel.rs:865). That's what preserves `stdout, stderr, stdout` as three visible chunks. A cached `output_id` per `(execution_id, stream_name)` would incorrectly rewrite the earlier stdout unless the spec adds tail-tracking/invalidation.

**Medium — migration cost understated.** `read_state()` hardcodes outputs as a list (runtime_state.rs:1068), output diffing snapshots assume `Vec` equality (:2103), WASM/frontend consume an ordered array. Every reader needs to be dual-shape aware before a daemon-only flag is safe.

**Medium — motivation is weaker than stated.** The real problems are actor reuse (fixed in #1905/#1911/#1913) and addressable updates. Append-only ordering itself isn't the issue. Simpler design: keep the list as source of truth, add stable `output_id`s to entries, optionally maintain a side index for `display_id`. Avoids inventing a second ordering system.

I think (medium 2) is the punch: the motivation that drove this spec has largely been fixed elsewhere. Worth rethinking whether map-keyed is still the right move.

PR #1911, opened 2026-04-19
fix(runtimed): eliminate shared-actor fork merges in notebook_sync_server

rgbkrk:
## Summary

Audit sibling of #1905. That PR fixed the IOPub handler's shared-actor bug (two forks setting the same Automerge actor ID silently dropping one merge with `DuplicateSeqNumber`). This PR fixes the three untreated siblings in `crates/runtimed/src/notebook_sync_server.rs`.

### Vulnerable — fixed here

1. **`process_markdown_assets`** uses `"runtimed:assets"`. Called on every peer sync; two peers racing will each fork, async-resolve markdown assets, then merge with the same actor. Silent drop.

2. **`apply_interrupt_to_state_doc`** uses `"runtimed:state:interrupt"`. Rare in practice, but two peers hitting interrupt within the same async gap would collide.

3. **Two formatter sites** use `formatter_actor(runtime)` which is `"runtimed:ruff"` or `"runtimed:deno"`:
   - Save-time bulk format — one user triggering two saves rapidly.
   - Per-cell background format spawned on execute — this one fires on **every code-cell execution** and `tokio::spawn`s. N concurrent executions = N concurrent forks on the same actor. Actively losing format ops in production.

### Safe — left alone, verified

- **`runtimed:filesystem`** at both call sites: fork and merge both happen inside a single held `doc.write()` lock. No async gap. Actor reuse is fine because forks are serialized on the lock.

## Fix

Small helper:

```rust
fn unique_fork_actor(base: &str) -> String {
    format!("{}:{}", base, uuid::Uuid::new_v4())
}
```

Four call sites updated. Every fork gets its own `(actor, seq)` series; merges are disjoint; `DuplicateSeqNumber` can no longer be raised at these sites.

## Related

- #1905 (IOPub variant of the same bug, with the invariant documented in a test in `notebook-doc`).
- Per-task stable actor refactor tracked for a follow-up. UUID-per-fork is correct but bloats the doc's actor list; the better long-term shape is per-logical-writer actors. Not in scope for this PR.

## Test plan

- [x] `cargo test -p runtimed --lib`: 325 pass
- [x] `cargo xtask lint`: clean
- The invariant this fix relies on is already asserted in `crates/notebook-doc/src/runtime_state.rs::merging_two_forks_with_shared_actor_returns_duplicate_seq_error` (landed with #1905).

### DIFF
--- a/crates/runtimed/src/notebook_sync_server.rs
+++ b/crates/runtimed/src/notebook_sync_server.rs
@@ -640,7 +640,7 @@ async fn process_markdown_assets(room: &NotebookRoom) {
             .map(|cell| (cell.id, cell.source, cell.resolved_assets))
             .collect();
         let mut fork = doc.fork();
-        fork.set_actor("runtimed:assets");
+        fork.set_actor(&unique_fork_actor("runtimed:assets"));
         (cells, fork)
     };
 
@@ -815,7 +815,7 @@ pub(crate) async fn apply_interrupt_to_state_doc(
     let mut fork = {
         let mut sd = room_state_doc.write().await;
         let mut f = sd.fork();
-        f.set_actor("runtimed:state:interrupt");
+        f.set_actor(&unique_fork_actor("runtimed:state:interrupt"));
         f
     };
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Codex review — no correctness findings.

- notebook_sync_server.rs:642, :817, :5883, :7157 are the right sites to fix; each fork+async+merge now uses a UUID-suffixed actor before any fork mutations.
- :8985, :9185 (runtimed:filesystem) are safe to leave — both fork+mutate+merge happen while holding the same doc.write() guard with no .await inside, so same-actor forks can't overlap concurrently.
- No other raw fork() callsites in runtimed/src that match the bug pattern.

**Design note (codex + Kyle agree)**: `unique_fork_actor` here + `unique_kernel_actor` in jupyter_kernel.rs is duplication of a still-manual invariant. Better API: `NotebookDoc::fork_with_actor(base)` / `RuntimeStateDoc::fork_with_actor(base)` that bake uniqueness in. Tracked as task #67, follow-up after this lands.

PR #1906, opened 2026-04-19
chore(lints): allow unwrap/expect in test code workspace-wide

rgbkrk:
## Summary

- Add `#![cfg_attr(test, allow(clippy::unwrap_used, clippy::expect_used))]` at each crate root so tests no longer need boilerplate `#[allow(clippy::unwrap_used, clippy::expect_used)]` on every `#[cfg(test)] mod tests`.
- Remove 59 now-redundant per-module allow attributes in `#[cfg(test)]` modules.
- Preserve any allow attributes that carry extra lint names (`module_inception`, `approx_constant`, `new_without_default`) — either stripping just the unwrap/expect tokens or leaving the line untouched.

Workspace-level `clippy::unwrap_used = "warn"` / `expect_used = "warn"` still applies to non-test code, so production code is unchanged.

## Why not Cargo.toml alone?

`[workspace.lints.clippy]` in Cargo.toml has no per-profile / per-`cfg` mechanism — lints apply uniformly across all builds of a crate. `#![cfg_attr(test, allow(...))]` at each crate root is the standard Rust way to gate lint allowance on the test cfg.

Follow-up in a separate PR: audit the non-test `#[allow(clippy::unwrap_used)]` / `#[allow(clippy::expect_used)]` sites still scattered through production code.

## Test plan
- [x] `cargo xtask lint` passes (`cargo clippy --workspace --all-targets -- -D warnings`, fmt, biome, ruff, ty)

PR #1916, opened 2026-04-19
chore(runt-trust): opt into workspace clippy lints

rgbkrk:
## Summary

#1906 added workspace-wide `clippy::unwrap_used` / `clippy::expect_used` warnings, but eight crates never opted in via `[lints] workspace = true`. `runt-trust` is one with two pre-existing violations in non-test code.

Both sites call `HmacSha256::new_from_slice(key).expect(...)`. That API is infallible for HMAC - the crate accepts keys of any length. The fix is scoped `#[allow(clippy::expect_used)]` with a safety argument, so the intent is documented where it lives rather than blanket-silenced at the crate root.

- Append `[lints] workspace = true` to `crates/runt-trust/Cargo.toml`
- Add scoped allows with safety comments at the two HMAC init sites

## Test plan

- [x] `cargo clippy -p runt-trust --all-targets -- -D warnings`
- [x] `cargo test -p runt-trust`
- [x] `cargo xtask lint`

### DIFF
--- a/crates/runt-trust/Cargo.toml
+++ b/crates/runt-trust/Cargo.toml
@@ -18,3 +18,6 @@ sha2 = "0.11"
 [dev-dependencies]
 serial_test = "3"
 tempfile = "3"
+
+[lints]
+workspace = true
--- a/crates/runt-trust/src/lib.rs
+++ b/crates/runt-trust/src/lib.rs
@@ -177,7 +177,10 @@ pub fn get_pixi_metadata(
 pub fn compute_signature(key: &[u8; 32], metadata: &HashMap<String, serde_json::Value>) -> String {
     let content = extract_signable_content(metadata);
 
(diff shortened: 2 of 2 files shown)


rgbkrk:
## Codex review

**Bottom line: LGTM.**

- `HmacSha256::new_from_slice` is effectively infallible in `hmac 0.13.0` - both `HmacCore<D>::new_from_slice` and `SimpleHmac<D>::new_from_slice` always return `Ok(...)`, and the crate's own `KeyInit::new` impl uses the same `expect("HMAC accepts keys of any length")` pattern. Scoped allows are defensible.
- No better idiom - `unwrap_or_else` would still just panic with more ceremony; `?` would require changing the API to propagate an error that is unreachable for `Hmac<Sha256>` in this dep version.
- Minor nit: the comment says "Safety" but there's no `unsafe` involved - "Invariant" or "Justification" would be more accurate.

Not blocking. Leaving as-is unless reviewer prefers the rename.

PR #1917, opened 2026-04-19
chore(runtimed-wasm): opt into workspace clippy lints

rgbkrk:
## Summary

Workspace lints (#1906) added `clippy::unwrap_used` and `clippy::expect_used` as warnings, but eight crates didn't inherit them. `runtimed-wasm` had one pre-existing `unwrap()` in the RUNTIME_STATE_SYNC frame handler, on a value that was Some by construction whenever `changed` was true.

Added `[lints] workspace = true` to `crates/runtimed-wasm/Cargo.toml` and rewrote the offending block:

```rust
// before
let output_changed_cells = if changed {
    let current_state = state.as_ref().unwrap();
    ...
};

// after
let output_changed_cells = if let Some(current_state) = state.as_ref() {
    ...
};
```

Same control flow, no panic path. Rebuilt the checked-in WASM bundle.

## Test plan

- [x] `cargo clippy -p runtimed-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings`
- [x] `cargo test -p runtimed-wasm` (11 tests pass)
- [x] `cargo xtask lint`
- [x] `wasm-pack build crates/runtimed-wasm --target web --out-dir ../../apps/notebook/src/wasm/runtimed-wasm`

### DIFF
--- a/apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm
+++ b/apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d4ec97572f2e115c1ab499a66bb96d65e49ca37b40328431282fe063c34ee726
+oid sha256:fcbf4fedc4c78dcb0cb5cf1d9456015853dd44171d1b20069924647789bf3935
 size 1668760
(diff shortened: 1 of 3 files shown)


rgbkrk:
## Codex review

**Verdict: ship.**

- `crates/runtimed-wasm/src/lib.rs:1615` is semantically equivalent. `state` is constructed immediately from `changed` and not mutated before the branch, so `if let Some(current_state) = state.as_ref()` fires exactly when the old `if changed { state.as_ref().unwrap() }` did. The only behavioral difference is on an impossible-by-construction invariant break: the old code would panic, the new code would quietly return `Vec::new()`. Not a practical concern given the local construction.
- `crates/runtimed-wasm/Cargo.toml` `[lints] workspace = true` matches the existing repo pattern and inherits the `[workspace.lints.clippy]` block in the root manifest.

Residual risk: limited to future refactors weakening the `changed -> state.is_some()` invariant.

PR #1919, opened 2026-04-19
chore(mcp-supervisor): opt into workspace clippy lints

rgbkrk:
## Summary

- Adds `[lints] workspace = true` to `crates/mcp-supervisor/Cargo.toml`, completing one of the eight remaining crates from #1906.
- Fixes six resulting violations in non-test code.

## Diagnosis

Four violations were `unwrap()` on `serde_json::to_value(schemars::schema_for!(...))` in the `list_tools` handler. Serializing a schemars schema can't fail outside OOM, but `clippy::unwrap_used` and `expect_used` both flag those sites. Folded the four call sites into a `schema_object::<T>()` helper that returns an empty `Map` on error rather than panicking. Behavior on the happy path is identical.

Two violations were `expect()` on `std::env::current_dir()` in `resolve_project_root`. That function is called from `main` before anything else, so the cwd really could be gone or unreadable. The walk now falls back to `PathBuf::from(".")` and continues, matching the existing fallback at the top of the loop.

## Test plan

- [x] `cargo clippy -p mcp-supervisor --all-targets -- -D warnings`
- [x] `cargo test -p mcp-supervisor` (9 passed)
- [x] `cargo xtask lint`

### DIFF
--- a/crates/mcp-supervisor/Cargo.toml
+++ b/crates/mcp-supervisor/Cargo.toml
@@ -27,3 +27,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
+
+[lints]
+workspace = true
--- a/crates/mcp-supervisor/src/main.rs
+++ b/crates/mcp-supervisor/src/main.rs
@@ -1403,27 +1403,19 @@ impl ServerHandler for Supervisor {
         let mut tools = Vec::new();
 
         // Supervisor's own tools
-        let empty_schema = serde_json::to_value(schemars::schema_for!(EmptyParams))
-            .unwrap()
-            .as_object()
-            .cloned()
-            .unwrap_or_default();
+        // `serde_json::to_value(schemars::schema_for!(...))` only fails on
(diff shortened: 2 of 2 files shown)


rgbkrk:
## Codex review

Codex flagged one medium issue: the original `PathBuf::from(".")` fallback in `resolve_project_root` silently degrades if `current_dir()` fails. `"." -> pop() -> ""` means the walk terminates without finding the workspace root, and the bad relative path gets reused for logging and subprocess spawning. That converts an explicit failure into a half-initialized supervisor.

Fixed: kept `expect()` with a targeted `#[allow(clippy::expect_used)]` and a `reason` comment. cwd-read failure is genuinely unrecoverable for this function (its whole job is path bootstrap), so loud-failure is correct. Force-pushed as a9be28dd.

Clippy cleanups in `schema_object` were found clean.

rgbkrk:
Closing — scoped `#[allow(clippy::expect_used)]` ducks the lint instead of addressing the intent. Plan at `.context/superpowers/2026-04-19-no-panic-error-propagation.md` proposes real error propagation. Fresh PR coming.

PR #1920, opened 2026-04-19
refactor(notebook-doc): fork_with_actor bakes unique-actor invariant into the API

rgbkrk:
## Why

Automerge tracks ops by `(actor, seq)`. Two concurrent forks that share an actor each produce ops at seq N, N+1, …; the first merge lands and the second returns `DuplicateSeqNumber`. If that error is caught and ignored (as under `catch_automerge_panic`), writes are silently dropped. #1905 fixed the IOPub variant; #1911 covered the rest of `notebook_sync_server`.

The invariant was still manual at the callsite. Every fork that merges across an `.await` had to follow the pattern `doc.fork()` then `f.set_actor(&unique_*_actor(...))`, with two near-identical local helpers — one in `jupyter_kernel.rs`, one in `notebook_sync_server.rs` — that just formatted `"{base}:{uuid}"`. Easy to forget on the next fork, and the failure mode is silent.

## What

Add `fork_with_actor(actor: impl AsRef<str>)` to `NotebookDoc` and `RuntimeStateDoc`. It forks and sets the actor in one step. The actor is set verbatim, so the caller chooses:

- **Stable per-task** (`"rt:kernel:abc:iopub"`) for long-running loops that fork many times per task — each unique actor takes space in Automerge's internal actor list.
- **Unique per-fork** (`format!("runtimed:assets:{}", Uuid::new_v4())`) for one-shot sites where concurrent forks from the same logical path can overlap across an async gap.

Pre-empted a `fork_with_unique_actor(base)` variant that would append a UUID internally: it would pull `uuid` into `notebook-doc` (which also compiles to wasm), and the one-shot callsites compose the UUID cleanly with the single-helper design.

### Before / After — one migrated site

Before (jupyter_kernel.rs, iopub stream handler):

```rust
let mut fork = {
    let mut sd = state_doc_for_iopub.write().await;
    let mut f = sd.fork();
    f.set_actor(&iopub_actor_id);
    f
};
```

After:

```rust
let mut fork = {
    let mut sd = state_doc_for_iopub.write().await;
    sd.fork_with_actor(&iopub_actor_id)
};
```

### Migrated sites

- `crates/runtimed/src/jupyter_kernel.rs` — 5 iopub/shell sites with the existing stable actor IDs. `coalesce` stays on `fork_and_merge` (synchronous).
- `crates/runtimed/src/notebook_sync_server.rs` — `apply_assets_to_doc`, `apply_interrupt_to_state_doc`, and the format-all-cells site use `fork_with_actor(format!("{base}:{uuid}"))`. Two file-watcher sites (synchronous fork+merge inside the write guard) use a stable `"runtimed:filesystem"` actor.
- The formatter fork for `ExecuteCell` still sets its actor inside the spawned task, because the formatter identity depends on the runtime detected after the fork is created. A comment documents why.
- `unique_fork_actor` local helper deleted. No equivalent ever existed in `jupyter_kernel.rs`; actor IDs there are computed once at kernel start.

### Safety net

The regression test `merging_two_forks_with_shared_actor_returns_duplicate_seq_error` in `runtime_state.rs` still pins the underlying Automerge invariant, independent of whether callers use `fork_with_actor` or the raw pair.

### WASM

`runtimed_wasm_bg.wasm` rebuilt because `notebook-doc` exports changed.

## Test plan

- [x] `cargo build -p notebook-doc -p runtimed`
- [x] `cargo test -p notebook-doc` (341 pass + regression test)
- [x] `cargo test -p runtimed --lib` (325 pass)
- [x] `cargo xtask lint`
- [x] `wasm-pack build crates/runtimed-wasm --target web --out-dir ../../apps/notebook/src/wasm/runtimed-wasm`

### DIFF
--- a/apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm
+++ b/apps/notebook/src/wasm/runtimed-wasm/runtimed_wasm_bg.wasm
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:9af72c2ced0b75f213b722ca268bdf3e81458c0a361f92b62ca6476e7b7e1719
+oid sha256:c7a64287c5865f7b0d985fa5a7299d64ad08d966dc804e73b29d6799fa830d9c
 size 1669510
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -201,6 +201,37 @@ impl NotebookDoc {
         }
     }
 
+    /// Fork the document and set a distinct actor ID on the fork in one step.
+    ///
+    /// Forks inherit the parent's actor ID, and Automerge tracks ops by
+    /// `(actor, seq)`. Two concurrent forks that share an actor will each
+    /// produce ops at seq `N`, `N+1`, …; the first merge lands and the
+    /// second returns `DuplicateSeqNumber` — silently dropping writes if
+    /// the error is ignored. See the regression test
+    /// `merging_two_forks_with_shared_actor_returns_duplicate_seq_error`
+    /// in `runtime_state.rs`.
+    ///
+    /// Use this for any fork whose merge crosses an `.await` point. The
+    /// `actor` argument is set verbatim — the caller controls whether
+    /// the actor is stable per task (e.g. `"rt:kernel:abc:iopub"`) or
+    /// unique per fork (e.g. `format!("runtimed:assets:{}", Uuid::new_v4())`).
+    ///
+    /// Prefer a stable per-task actor for long-running loops that fork
+    /// many times in sequence — each unique actor consumes space in
+    /// Automerge's internal actor list. Use a UUID suffix for one-shot
+    /// sites where concurrent forks from the same logical task can
+    /// overlap across the async gap.
+    ///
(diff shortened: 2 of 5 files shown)


rgbkrk:
Codex review (re-run, original hung):

**No findings.**

- Every async fork+merge site touched by the PR uses `fork_with_actor`. Only remaining raw `fork()` in the diff is notebook_sync_server.rs:5875 (formatter, documented exception).
- Stable-vs-UUID choices match concurrency model:
  - Kernel loops (jupyter_kernel.rs:820, 1022, 1066, 1236, 1671) → stable per-task actors (single long-lived task).
  - One-shot async jobs (notebook_sync_server.rs:639, 813, 7157) → UUID-suffixed (overlapping forks possible).
  - File-watcher (notebook_sync_server.rs:8988, 9189) → stable `runtimed:filesystem` (synchronous inside write-guarded critical section).
- Formatter exception justified: actor prefix depends on detected runtime, and no ops are emitted before `set_actor` at :5881-5883.
- Regression test `merging_two_forks_with_shared_actor_returns_duplicate_seq_error` still uses raw `fork()` + explicit `set_actor`, pinning the underlying Automerge invariant rather than just the helper API.

PR #1922, opened 2026-04-19
chore(notebook-doc): opt into workspace clippy lints

rgbkrk:
## Summary

Opts `notebook-doc` into `[workspace.lints]` (introduced in #1906), covering the 21 pre-existing `clippy::unwrap_used` / `clippy::expect_used` violations in non-test code. Each site sits in a path where failure is ruled out by construction, so the fixes are documentation rather than behavior change: function-scoped `#[allow(clippy::expect_used)]` with a rationale comment, plus `.unwrap()` -> `.expect("<reason>")` where a call-site message clarifies the invariant.

## Fixes by category

- **CBOR encoders (9 in `presence.rs`)** — fixed-shape structs serialized into a `Vec<u8>` writer. I/O failure is impossible, and the struct definitions rule out unserializable values at compile time. Tagged each public encoder with `#[allow(clippy::expect_used)]` and added a section-level comment explaining the rationale.
- **Automerge puts (9 in `runtime_state.rs::create_execution_with_source`)** — operates on a map and object the function either just created via `put_object` or read unconditionally via `get_map`. The sibling `create_execution` already uses this pattern with a function-level allow; this PR applies it to the `_with_source` variant.
- **Metadata path walk (3 in `lib.rs::update_cell_metadata_at`)** — each unwrap is guarded by an `is_object()` check or a `contains_key` insertion on the immediately preceding line. Converted to `.expect()` with messages describing the invariant, with a single function-level allow.

Eight crates still need to opt in after this merges.

## Test plan

- [x] `cargo clippy -p notebook-doc --all-targets -- -D warnings` clean
- [x] `cargo xtask lint` passes (Rust, JS/TS, Python)
- [x] `cargo test -p notebook-doc` passes (5/5 + doctests)
- [ ] CI green

### DIFF
--- a/crates/notebook-doc/Cargo.toml
+++ b/crates/notebook-doc/Cargo.toml
@@ -29,3 +29,6 @@ default = []
 # Enable file persistence methods (load_or_create, save_to_file, notebook_doc_filename)
 # and logging. Not available for wasm32 targets.
 persistence = ["dep:log", "dep:sha2", "dep:hex"]
+
+[lints]
+workspace = true
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -1573,6 +1573,10 @@ impl NotebookDoc {
     /// Note: This performs a read-modify-write on the JSON string. Concurrent updates
     /// to different paths may conflict (last-write-wins), but this is rare in practice
     /// since metadata updates are typically user-initiated actions.
+    // Allow reason: the `expect` calls below are safe by construction —
+    // each is guarded by an `is_object()` check or a `contains_key` insertion
+    // on the immediately preceding line, so the `Option` is provably `Some`.
+    #[allow(clippy::expect_used)]
     pub fn update_cell_metadata_at(
         &mut self,
         cell_id: &str,
@@ -1593,11 +1597,15 @@ impl NotebookDoc {
             if !current.is_object() {
                 *current = serde_json::json!({});
             }
-            let obj = current.as_object_mut().unwrap();
+            let obj = current
+                .as_object_mut()
+                .expect("current was just set to an object above");
             if !obj.contains_key(*key) {
                 obj.insert((*key).to_string(), serde_json::json!({}));
             }
-            current = obj.get_mut(*key).unwrap();
+            current = obj
+                .get_mut(*key)
+                .expect("key was just inserted above if missing");
         }
 
         // Set the final key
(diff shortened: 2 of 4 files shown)


rgbkrk:
## Codex review

Codex flagged two items. Both are real but in my reading out-of-scope for this PR, which is strictly a lint cleanup for `notebook-doc`.

### Finding 1 (medium): `create_execution_with_source` panic path on unscaffolded docs

> A concrete panic path is: `let mut doc = RuntimeStateDoc::new_empty(); doc.create_execution_with_source(...)`. If that public API is meant to be valid only on scaffolded docs, the precondition should be encoded more strongly; otherwise this signature should return `Result<bool, E>` instead of panicking.

This is a pre-existing issue and applies identically to the neighboring `create_execution` (line 832), `set_execution_running` (line 932), and every other method in this file that calls `get_map("executions").expect(...)`. The PR matches that established pattern rather than changing the contract. Converting the full surface to `Result` or documenting/enforcing the scaffolded-doc precondition is a worthwhile follow-up, but it touches callers across `runtimed` and the coordinator and doesn't belong in a clippy-cleanup PR.

### Finding 2 (medium): `encode_snapshot` hides protocol-level frame-size failure

> Snapshot size is a real protocol error: receivers drop oversized presence payloads in sync_task.rs:476, while senders currently transmit snapshot bytes without outbound validation.

Also pre-existing and distinct from the lint cleanup. The `expect()` guards CBOR serialization (infallible for `Vec<u8>` writer + fixed-shape structs); frame-size validation is a separate concern, and `validate_frame_size()` already exists. The fix is to call it at the three sender sites codex cites in `notebook_sync_server.rs` (2540, 2741, 3109). Worth filing, but again not a lint-cleanup change.

### Assessment Codex agrees with

> The new `expect()` calls in `lib.rs` look mechanically safe: each one is immediately dominated by `is_object()` normalization or an insertion that makes the subsequent `Option` non-`None`. The CBOR `expect()` calls in `presence.rs` also look sound for current message types.

Keeping this PR scoped to the lint opt-in. Filing a follow-up for the `RuntimeStateDoc` scaffolding contract and the presence snapshot send-site validation.

rgbkrk:
Closing — scoped `#[allow(clippy::expect_used)]` ducks the lint instead of addressing the intent. Plan at `.context/superpowers/2026-04-19-no-panic-error-propagation.md` proposes real error propagation. Fresh PR coming.

PR #1921, opened 2026-04-19
chore(sift-wasm): opt into workspace clippy lints

rgbkrk:
## Summary

Opts `sift-wasm` into `[workspace.lints]` (one of eight crates missed by #1906) and fixes the five resulting `unwrap_used` violations. All five are safe by construction:

- `STORES` / `NEXT_HANDLE` mutex locks — wasm32 is single-threaded, so `PoisonError` would require a prior panic holding the same lock.
- `ms_values.iter().min()` / `.max()` — guarded by `is_empty()` one line above.
- `NaiveDate::and_hms_opt(0, 0, 0)` — hardcoded valid arguments.

Each uses `expect(...)` with `#[allow(clippy::expect_used)]` plus a one-line justification, matching the pattern already used in `blob_server`, `kernel_client`, `output_store`, etc.

## Test plan

- [x] `cargo clippy -p sift-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings` passes
- [x] `cargo build -p sift-wasm --target wasm32-unknown-unknown` passes
- [x] `cargo xtask lint` passes
- [ ] No `sift-wasm` tests exist; WASM build artifact unchanged semantically

PR #1923, opened 2026-04-19
chore(sift-wasm): opt into workspace clippy lints with structural fixes

rgbkrk:
Replaces #1921. That PR opted into the workspace clippy lints but dodged five `unwrap_used` warnings with `#[allow(clippy::expect_used)]` + safety comments. The intent of the lint is "don't panic in non-test code," and scoped-allow respects the lint, not the intent. This PR fixes each site at the root.

## Fixes by shape

**`STORES` / `NEXT_HANDLE` mutex locks.** wasm32 is single-threaded, so `PoisonError` only occurs after a prior panic inside a guarded section. Real recovery is to take the inner value and continue, not to re-panic. Swapped `.unwrap()` for `.unwrap_or_else(|e| e.into_inner())`.

Why this beats a scoped allow: the allow was claiming "this can't happen in practice" while still panicking on the off chance it did. `into_inner()` removes the panic path entirely.

**`ms_values.iter().min() / .max()` in `store_temporal_histogram`.** The unwrap was guarded by an `is_empty()` check one line up — the compiler can't see that. Restructured as `match (min, max)` so the Option variants are exhaustive: `(Some, Some)` destructures both, any other pattern returns the empty-array result the `is_empty()` branch already did. The failure path is gone, not suppressed.

**`NaiveDate::and_hms_opt(0, 0, 0).unwrap()`.** The args are hardcoded valid, but the API returns Option regardless. Switched to `date.and_time(NaiveTime::default())`. `NaiveTime::default()` is midnight and returns `NaiveTime` directly; `and_time` produces `NaiveDateTime` with no Option step. Failure case disappears at the type level.

(Note: `NaiveTime::MIDNIGHT` const would be cleaner but isn't exposed in the chrono version in use here — `Default` is equivalent and non-deprecated.)

## Verification

- `cargo clippy -p sift-wasm --target wasm32-unknown-unknown --all-targets -- -D warnings` — clean
- `cargo build -p sift-wasm --target wasm32-unknown-unknown` — builds
- `cargo xtask lint` — all checks pass
- Grep for `#[allow(clippy::unwrap_used)]` / `#[allow(clippy::expect_used)]` in the diff — none

## Test plan

- [x] Clippy passes with `-D warnings` on wasm32 target
- [x] sift-wasm builds for wasm32
- [x] Workspace lint passes

### DIFF
--- a/crates/sift-wasm/Cargo.toml
+++ b/crates/sift-wasm/Cargo.toml
@@ -28,3 +28,6 @@ parquet = { version = "54", default-features = false, features = ["arrow", "snap
 
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = false
+
+[lints]
+workspace = true
--- a/crates/sift-wasm/src/store.rs
+++ b/crates/sift-wasm/src/store.rs
@@ -75,7 +75,9 @@ fn with_stores<F, R>(f: F) -> R
 where
     F: FnOnce(&mut HashMap<u32, DataStore>) -> R,
 {
-    let mut guard = STORES.lock().unwrap();
+    // wasm32 is single-threaded, so poisoning only occurs if a prior call
+    // panicked while holding the lock. Recover the inner value and carry on.
+    let mut guard = STORES.lock().unwrap_or_else(|e| e.into_inner());
     let stores = guard.get_or_insert_with(HashMap::new);
     f(stores)
 }
@@ -110,7 +112,7 @@ fn store_batches(batches: Vec<RecordBatch>, schema: &arrow::datatypes::Schema) -
     }
 
     let handle = {
-        let mut h = NEXT_HANDLE.lock().unwrap();
+        let mut h = NEXT_HANDLE.lock().unwrap_or_else(|e| e.into_inner());
         let id = *h;
         *h += 1;
         id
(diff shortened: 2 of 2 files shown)


rgbkrk:
## Codex review

Codex flagged the `PoisonError::into_inner()` fix for `STORES` as unsound — its argument: on host test builds (`panic = unwind`), a panic mid-mutation inside `with_stores` (e.g. an index-out-of-bounds in `cast_column`) would poison the mutex, and `into_inner()` would then expose a partially-mutated `DataStore` to the next caller instead of failing fast.

The min/max match and `NaiveTime::default()` fixes were approved. No scoped `#[allow]` escapes found.

## Response

The concern is valid in host-test builds. In wasm32 (the actual runtime) Rust panics abort the worker via `console_error_panic_hook` — there's no unwinding, so the mutex never becomes poisoned in practice. But the code also compiles for host tests where unwinding applies.

The reference plan (`.context/superpowers/2026-04-19-no-panic-error-propagation.md`) explicitly specifies `PoisonError::into_inner()` for this site as category-(4) "recover" — and notes that the single-threaded wasm32 target is the reason poisoning requires a prior panic. The alternative (keep `.unwrap()`, i.e. re-panic on poison) is strictly worse: it panics twice instead of once, and still exposes no guarantees about state consistency because the first panic already happened.

If reviewers want stronger guarantees, the right follow-up is:
1. Split `STORES` into per-handle `Mutex<DataStore>` so a poisoned entry is localized, or
2. On poison, clear the store and return an error from `with_stores` (would require changing its signature from `R` to `Result<R, String>` — ripples through all callers).

Both are larger refactors than this clippy-opt-in PR is scoped for. Keeping `into_inner()` matches the plan's directive and doesn't regress behavior vs. `.unwrap()` — in wasm it's unreachable; in host tests it recovers instead of cascading a second panic. Leaving as-is for this PR, filing a follow-up for per-handle locks if wanted.

PR #1938, opened 2026-04-19
feat(runt-mcp): add debug logging for MCP response payload analysis

quillaid:
## Summary

- Instruments `call_tool()` in `runt-mcp` to log payload size, content item breakdown, structured content size, timing, and problematic byte detection (null bytes, control characters) for every MCP tool response
- Normal responses log at `debug` level (zero noise in production); payloads with null bytes, control chars, or >512KB escalate to `warn` with content previews
- Nightly channel defaults to `info,runt_mcp=debug` so payload analysis is always captured; stable channel keeps `info` default

**Goal:** Determine whether runtimed sends problematic data that triggers SDK message reader crashes (null bytes, malformed UTF-8, extremely long lines), or if it's purely an SDK-side size limitation.

### What's logged per response

| Field | Description |
|-------|-------------|
| `payload` | Total serialized JSON size in bytes |
| `items` | Number of Content items |
| `structured` | Whether structured_content is present + size |
| `null_bytes` | Count of null bytes in serialized JSON |
| `control_chars` | Count of control characters (excl. `\n`, `\r`, `\t`) |
| `elapsed` | Wall-clock time from dispatch entry to return |

Debug path includes per-item breakdown (200-char preview each). Warn path adds 500-char structured content preview.

### Files changed

- `crates/runt-mcp/src/lib.rs` — `log_mcp_response()` + call site in `call_tool()`
- `crates/runt/src/main.rs` — nightly default log filter includes `runt_mcp=debug`

## Test plan

- [x] `cargo test -p runt-mcp` — 67 tests pass
- [x] Deployed to nightly, verified `[mcp-response]` strings present in binary
- [x] Gremlins harness updated to use `runt_mcp=debug` filter

### DIFF
--- a/crates/runt-mcp/src/lib.rs
+++ b/crates/runt-mcp/src/lib.rs
@@ -182,11 +182,12 @@ impl ServerHandler for NteractMcp {
                 }
             }
         }
-        let tool_name = request.name.clone();
         let start = std::time::Instant::now();
         let result = tools::dispatch(self, &request).await;
-        let elapsed = start.elapsed();
-        log_mcp_response(&tool_name, elapsed, &result);
+        if tracing::enabled!(tracing::Level::DEBUG) {
+            let elapsed = start.elapsed();
+            log_mcp_response(&request.name, elapsed, &result);
+        }
         result
     }
 
--- a/crates/runt-mcp/src/lib.rs
+++ b/crates/runt-mcp/src/lib.rs
@@ -8,6 +8,7 @@
 
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::time::Duration;
 
 use rmcp::model::{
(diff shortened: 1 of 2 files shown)


crates/runt-mcp/src/lib.rs
```diff
@@ -181,7 +182,12 @@ impl ServerHandler for NteractMcp {
                 }
             }
         }
-        tools::dispatch(self, &request).await
+        let tool_name = request.name.clone();
+        let start = std::time::Instant::now();
+        let result = tools::dispatch(self, &request).await;
+        let elapsed = start.elapsed();
+        log_mcp_response(&tool_name, elapsed, &result);
```
rgbkrk: This should only get called if RUST_LOG is high enough, otherwise a bunch of serialization and deserialization happens that shouldn't be run in a release mode.

PR #1983, opened 2026-04-21
refactor: extract notebook-persistence crate for ipynb roundtrip

rgbkrk:
## Summary

`runt recover` and the daemon save/load path each had their own
NotebookDoc <-> `.ipynb` conversion. Same source line-splitting rules,
same nbformat_minor bumping, same fallback cell IDs for pre-4.5 files -
all copied. The CLI copy was a strict subset (no merge-with-existing,
no RuntimeStateDoc outputs, no per-cell attachments) and was quietly
drifting as the daemon got richer.

Extract a new `notebook-persistence` crate. Default features stay
WASM-clean. Both consumers now call it.

## Public surface

- `doc_to_ipynb(&NotebookDoc) -> Value` - the CLI's minimal synchronous
  export.
- `build_ipynb(BuildInputs)` - the daemon's merge-aware build. Takes
  pre-resolved outputs/execution counts (the caller does the async
  blob-store resolution) plus optional existing `.ipynb` JSON, returns
  the final notebook object.
- `parse_cells_from_ipynb`, `parse_nbformat_attachments_from_ipynb`,
  `parse_metadata_from_ipynb` - pure snapshot parsers shared with the
  daemon's load and file-watcher paths.

## What did I find between the two implementations?

One real behavioral gap, documented but not fixed here:

- `runt recover` reads outputs from `CellSnapshot.outputs`, which has
  been empty since schema v3 (outputs moved to `RuntimeStateDoc`).
  Recover still ships the cells, just without outputs. The daemon's
  save path reads outputs from `RuntimeStateDoc` + blob store, so
  `SaveNotebook` remains correct. Fixing recover requires reading
  the sibling `RuntimeStateDoc` file - not in scope for a
  behavior-preserving extraction, but `doc_to_ipynb`'s doc-comment
  now calls this out so a follow-up can land on top.

Other than that, the routes match bit-for-bit. I kept the CLI's
best-effort `_ = snapshot.merge_into_metadata_value(...)` (warns were
only on serialization edge cases, and `runt recover` prints to stderr
via the calling path). `build_ipynb` surfaces the same error as
`BuildError::MetadataMerge` so the daemon can log it explicitly.

## What about `ipynb_to_doc`?

There is no symmetric `ipynb_to_doc` helper. The daemon's load path
interleaves parsing with blob-store writes and `RuntimeStateDoc`
population - a "pure" converter would silently drop outputs on every
schema-v3+ doc. I did add one briefly (`build_notebook_doc`), Codex
caught the same gap, and I removed it. The three `parse_*` helpers
are the sharable seam; daemon owns composition.

## Scope not in this PR

- No autosave / file-watcher / snapshot-rotation extraction.
- `runt recover` output fidelity (see above).
- No new behavior, just moves.

## Test plan

- [x] `cargo test -p notebook-persistence --lib` (14 tests, all moved
  from runt + daemon)
- [x] `cargo test -p runt-cli --bin runt` (9 passed)
- [x] `cargo test -p runtimed --lib` (381 passed, 1 ignored)
- [x] `cargo test -p runtimed --test integration` (25 passed)
- [x] `cargo check --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo xtask lint`
- [x] `codex review --base main` - no actionable findings after the
  `build_notebook_doc` cut


rgbkrk:
Closing. With #1986 merged, the only consumer left is the daemon itself, so a standalone crate is premature. We'll do more cleanup in place first (especially the CellSnapshot.outputs redesign in task #17) and revisit extraction once there's a stable seam worth promoting.

PR #1985, opened 2026-04-21
fix(runt): recover resolves outputs from RuntimeStateDoc + blobs

rgbkrk:
## Summary

`runt recover` was shipping `.ipynb` files with empty `outputs` arrays on every cell. Root cause: the export path read outputs from `CellSnapshot.outputs`, which has been unconditionally empty since schema v3 moved cell outputs into the daemon-authoritative `RuntimeStateDoc`. `RuntimeStateDoc` had never been persisted to disk, so offline recovery had no path back to the output manifests.

Fix ships both sides end to end:

- **Daemon now persists `RuntimeStateDoc` to a `{stem}.state.automerge` sibling** next to the notebook doc. A new debouncer mirrors the main one; a forwarder task subscribes to `state_changed_tx` and feeds the debouncer on every state mutation.
- **`runt recover` reads that sidecar** (when present) and hands it to `doc_to_ipynb_with_state`, which follows each cell's `execution_id` pointer into the state doc, pulls the output manifests, and resolves inline / blob `ContentRef` entries. Blob resolution is MIME-aware: text MIMEs come back as strings, JSON MIMEs round-trip as structured JSON, binary MIMEs base64-encode the raw bytes. The blob store root is derived from the matching `.automerge` path so cross-namespace recoveries (stable / nightly / worktree) hit the right `blobs/` directory.
- **Lifecycle is coherent**: snapshots pair the state sidecar with the notebook doc; eviction flushes synchronously; blob GC scans state-sidecar files (including snapshots) so blob-backed outputs survive the grace window; the orphaned-doc sweep reclaims stranded sidecars; untitled→saved promotion disables the forwarder and deletes the sidecar so the UUID-keyed path doesn't get rewritten after the `.ipynb` takes over.
- **Fallback** keeps pre-v3 captures alive: when no state sidecar exists and the NotebookDoc's `schema_version < 3`, recovery extracts cell-local outputs via `extract_cell_outputs` and resolves their ContentRefs the same way. A schema-v3 doc with a cleared `execution_id` does NOT fall back — that's a user-initiated clear, not recoverable data.

TDD order: failing regression test lands first, then the receiver side, then daemon-side state persistence. Codex review surfaced five rounds of lifecycle gaps (GC, eviction, promotion, lag, discriminator); each round has commits + tests.

Commits (top-down):

- `fix(runtimed): disable state-sidecar forwarder after untitled promotion`
- `fix: reclaim orphaned state sidecars, tighten ContentRef discriminator`
- `fix: stop pairing stale exec_count with empty outputs, reserialize on lag`
- `fix(runtimed): scan notebook-docs/snapshots during blob GC`
- `fix(runtimed): teach GC and eviction about state sidecars`
- `fix(runtimed): load state sidecar on reopen, gate recover fallback on v2`
- `fix(recover): fall back to legacy outputs when state sidecar is stale`
- `fix(recover): pair state sidecar lifecycle with notebook doc`
- `fix(runt): MIME-aware ContentRef resolution for display_data bundles`
- `feat(runtimed): persist RuntimeStateDoc to sibling .state.automerge`
- `fix(runt): load RuntimeStateDoc and resolve blob refs in recover`
- `test(runt): regression test for recover dropping cell outputs`

## Edge cases + follow-ups

- `looks_like_content_ref` is a strict structural check. The daemon always wraps user JSON in a ContentRef before writing, so in practice the check buys the common case without corrupting user data. If someone hand-crafts a state doc containing a raw JSON value shaped exactly like `{"inline":"<string>"}` at a ContentRef position, recovery would unwrap it. No such path exists through the current codebase.
- Eviction flushes the state sidecar synchronously by writing bytes directly, then kicks the debouncer. No dedicated ack channel yet; a disk error is logged but doesn't block eviction. Adding a `flush_request_tx` analogue is reasonable follow-up.
- Binary MIMEs (PNG, audio, video) resolve by reading raw bytes and base64-encoding. Runt-cli keeps the bytes in memory during resolution — fine for typical notebooks, potentially large for > 100MiB images. No streaming path yet.
- Recovery of active-session state: if the daemon is running when `runt recover` is invoked, the sidecar may be mid-flush, so the recovered outputs reflect the last debounced snapshot (up to ~500ms stale). Offline recovery after daemon shutdown is exact.
- Missing blob file on disk: ContentRef passes through unresolved rather than being dropped, so recovered `.ipynb` still shows some structure. Could be tightened to a placeholder stream output.

## Test plan

- [x] `cargo test -p runt-cli --bin runt` — 18 tests, including new regression tests covering inline/blob ContentRefs, MIME-aware display_data, stale-sidecar tolerance, ContentRef discriminator, and the cache-root derivation.
- [x] `cargo test -p runtimed --lib` — 383 tests, including new snapshot-pairing test and tokio-mutex lint.
- [x] `cargo test -p notebook-doc --lib` — 354 tests.
- [x] `cargo xtask lint` — rustfmt, clippy, biome, ruff, ty all green.
- [x] `codex review --base main` — ran 5 rounds; all P1/P2 findings addressed with code+tests; remaining theoretical notes documented above.
- [ ] Manual smoke: temp daemon → execute cell → shut down → `runt recover` against the `{hash}.automerge` → verify outputs present. (Not run locally due to worktree daemon isolation overhead; the regression tests exercise the same code paths end-to-end.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6884,6 +6884,7 @@ name = "runt-cli"
 version = "2.2.0"
 dependencies = [
  "anyhow",
+ "automerge",
  "chrono",
  "clap",
  "colored",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6885,6 +6885,7 @@ version = "2.2.0"
 dependencies = [
  "anyhow",
  "automerge",
+ "base64 0.22.1",
  "chrono",
  "clap",
  "colored",
--- a/crates/runt/Cargo.toml
+++ b/crates/runt/Cargo.toml
@@ -24,6 +24,7 @@ runtimelib = { workspace = true, features = ["tokio-runtime", "aws-lc-rs"] }
 runtimed-client = { path = "../runtimed-client" }
 notebook-doc = { path = "../notebook-doc", features = ["persistence"] }
 automerge = "0.8"
+base64 = "0.22"
 runt-workspace = { path = "../runt-workspace" }
 runt-mcp = { path = "../runt-mcp" }
 kernel-env = { path = "../kernel-env" }
--- a/crates/runt/Cargo.toml
+++ b/crates/runt/Cargo.toml
@@ -23,6 +23,7 @@ jupyter-protocol = { workspace = true }
 runtimelib = { workspace = true, features = ["tokio-runtime", "aws-lc-rs"] }
 runtimed-client = { path = "../runtimed-client" }
 notebook-doc = { path = "../notebook-doc", features = ["persistence"] }
+automerge = "0.8"
 runt-workspace = { path = "../runt-workspace" }
 runt-mcp = { path = "../runt-mcp" }
 kernel-env = { path = "../kernel-env" }
--- a/crates/runt/src/main.rs
+++ b/crates/runt/src/main.rs
@@ -5233,49 +5233,45 @@ fn resolve_cell_outputs_for_recovery(
 ) -> (Vec<serde_json::Value>, Option<serde_json::Value>) {
     // Preferred path: schema v3+. The NotebookDoc points at an execution_id
     // in the RuntimeStateDoc; output manifests (inline or blob ContentRef)
-    // live there. The notebook doc and state sidecar persist independently,
-    // so a stale sidecar (or one missing the referenced execution_id) is a
-    // real case — fall through to the legacy path instead of returning
-    // empty outputs when the state doc can't help.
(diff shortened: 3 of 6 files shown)


rgbkrk:
Closing per architectural discussion. We deliberately chose not to persist outputs for ephemeral notebooks; adding a lifecycle chain to recover them works against that decision. The real fix is the opposite direction: delete the stopgap recover tool and the dead CellSnapshot.outputs field. New PR incoming.

PR #2004, opened 2026-04-21
fix(mcp): drop runt-mcp health loop, drive state from DaemonConnection

### LINKED ISSUE #2000: MCP proxy stays wedged in Reconnecting state while daemon is healthy; short-circuits all tool calls
## Observed

On lab1 during iteration 1 of the nightly build-test loop:

- Team-lead's `runt-nightly mcp` process (the plugin's MCP server) reported `Daemon is restarting (attempt 28, 730s elapsed)` for 12+ minutes on every `create_notebook`, `list_active_notebooks`, etc.
- Same host, same daemon, at the same time: experiments-manager's gremlin sessions were happily connecting and driving the daemon through a full batch of concurrent fixture runs.
- `systemctl --user status runtimed-nightly`: `active (running) since 17:42 UTC`, PID 11921 stable, no restarts, uptime 27 minutes and counting.
- `~/.cache/runt-nightly/daemon.json`: fresh, PID 11921, version 2.2.0+86ca610 — no evidence of restart.

So the daemon was never restarting. One `runt mcp` process just thought it was, and refused every tool call for the duration.

## Root cause (working theory)

`crates/runt-mcp/src/tools/mod.rs` dispatches on `DaemonState`:

```rust
let state = server.daemon_state().read().await;
if let Some(msg) = state.reconnecting_message() {
    return tool_error(&msg);
}
```

The state transitions live in `crates/runt-mcp/src/health.rs::daemon_health_monitor`. The monitor:

1. Sleeps `backoff_duration(attempt)` (exponential, capped at 30s).
2. Calls `PoolClient::ping()`, which opens a fresh `UnixStream` to the daemon socket (2s connect_timeout + 3s request/response).
3. On any error — `ClientError::Timeout`, `ConnectionFailed(io::Error)`, etc. — flips/stays in `Reconnecting`, bumping `attempt`.
4. On success, flips back to `Connected`.

Under sustained concurrent gremlin load, the daemon's accept queue can spike above 2s occasionally. Once this client trips to `Reconnecting`, all subsequent tool calls are short-circuited (see dispatch above). The monitor keeps pinging on the growing backoff schedule — which means as backoff hits 30s between pings, each retry has to catch a narrow good-moment window. An unlucky phase alignment can keep the client in `Reconnecting` indefinitely while other clients that don't happen to hit bad moments talk to the daemon fine.

## Why this matters

The short-circuit is actively harmful in this failure mode:

- The daemon is up and serving. The actual tool call would almost certainly succeed.
- But the dispatch returns `Daemon is restarting` without even trying, and the user has no way to force a retry.
- The proxy is the only process that thinks anything is wrong. The state is local to this `runt mcp` process's monitor loop.

## Suggested fixes

Any of these alone probably resolves it; doing all three makes the proxy robust:

1. **Don't short-circuit on `Reconnecting`.** Try the tool call anyway. If the daemon is actually down, the call returns a meaningful error with the daemon-level context. If the daemon is up and the monitor was just stale, the call succeeds and we transition back to `Connected`. The `reconnecting_message` can be prepended to the result instead of replacing it — the proxy already has a reconnection-message prepend machinery for genuine restarts (`runt-mcp-proxy/src/proxy.rs::prepend_reconnection_message`).
2. **Corroborate `Reconnecting` against daemon.json.** Before entering or extending `Reconnecting`, check `~/.cache/runt-{channel}/daemon.json`. If the file exists, is fresh, and the PID is alive, the daemon is almost certainly fine and our ping is the problem. Log loudly and try the tool call anyway.
3. **Bound `Reconnecting` timeouts.** After N consecutive timeouts with a live-looking `daemon.json`, rebuild client state (or just log and reset `attempt` to 0 so we try again with fresh backoff). Right now the loop can run forever without self-diagnosing.

## Evidence

Session log on lab1 (team-lead): `~/.claude/projects/-home-ubuntu/99d0f36b-561d-443d-a79f-da773a04fd26.jsonl` — 28+ tool_result errors with `Daemon is restarting (attempt N, Ms elapsed)`. Lines around 102, 108, 111, 128, 140, 148, 155.

Gremlin DB finding #15 (`Daemon dies after extended uptime under sustained concurrent load`) describes a different bug — there the daemon actually died. This is the opposite: the daemon is fine, the proxy believes it's not.

rgbkrk:
Closes #2000.

## Summary

- Delete `runt-mcp/src/health.rs` (ping loop + `DaemonState` enum) and the dispatch short-circuit in `tools/mod.rs` that turned a transient ping timeout into "Daemon is restarting" errors on every tool call.
- Subscribe to `runtimed-client`'s `DaemonConnection` event stream via a new `daemon_watch.rs`. `Upgraded` with a version change returns `EXIT_DAEMON_UPGRADED` (75) for proxy respawn; same-version restarts and reconnects trigger session rejoin. `Disconnected` is a no-op — tool calls now see the real daemon error.
- Collapse rejoin to one caller. The proxy hands off the previous notebook target via `NTERACT_MCP_REJOIN_NOTEBOOK` when respawning the child. The child's `daemon_watch` consumes it on the first `Connected` event, so we no longer dial `open_notebook` over the child's MCP channel.

Net ~-260 lines. `-389 health.rs`, `-151 proxy session.rs` (auto_rejoin/build_rejoin_params/looks_like_untitled_notebook_id), `+337 daemon_watch.rs` (with tests).

## Why

From #2000: on lab1, a single `runt-nightly mcp` process reported \`Daemon is restarting (attempt 28, 730s elapsed)\` for 12+ minutes while the daemon had been up with stable PID the whole time and other clients were happily driving it. Root cause: the local ping (2s connect + 3s RPC) tripped on accept-queue pressure, flipped state to `Reconnecting`, and backoff windows (up to 30s) kept missing the healthy intervals. The dispatch short-circuit made every tool call a synthetic error.

The `DaemonConnection` supervisor (already live in the proxy) is better ground truth than a parallel ping loop. Using its events removes an entire state machine.

## Behavior

| Event | Child action |
|---|---|
| `Upgraded` (version changed) | `exit(75)` → proxy respawns |
| `Upgraded` (same version, pid changed) | rejoin session |
| `Connected` (first event after `Disconnected`, or seeded `NTERACT_MCP_REJOIN_NOTEBOOK` on fresh child) | rejoin session |
| `Disconnected` | no-op; next tool call surfaces real IPC error |

## Test plan

- [x] \`cargo xtask lint --fix\` clean
- [x] \`cargo xtask clippy\` clean
- [x] \`cargo test -p runt-mcp -p runt-mcp-proxy\` — 163 passed, includes 6 new tests for the watch-loop classifier (upgrade, same-version restart, initial-target consumption, disconnect no-op, UUID detection)
- [ ] Lab: force daemon restart under concurrent gremlin load, confirm MCP tool calls keep responding without the "Daemon is restarting" wedge
- [ ] Lab: bump daemon version, confirm child exits 75 and proxy respawns with reconnection message

## Scope

Intentionally small. Keeps: proxy circuit breaker, child monitor, tool cache, reconnect tool, tool-divergence exit. `DaemonConnection` itself unchanged — new consumer only. Future work (#TBD) evaluates whether the proxy stack (`runt-mcp-proxy` + `mcp-supervisor` + `runt-proxy`) still reads as layered patches after this ships.


rgbkrk:
## Lab smoke results

Built `fix/mcp-proxy-reconnect-wedge` on lab1, installed via `./scripts/install-nightly`, pulled through a headless MCP probe that issues `initialize` + `notifications/initialized` + `tools/call list_active_notebooks` on a FIFO-backed stdin.

### ✅ The wedge is gone (primary goal)

Pre-fix reproducer was "restart the daemon, next tool call returns *Daemon is restarting (attempt N, Ms elapsed)* indefinitely." On this branch:

- Before restart → `list_active_notebooks` returns `[]`.
- `systemctl --user restart runtimed-nightly` (same version, new PID).
- After restart → `list_active_notebooks` returns `[]`. Twice in a row, cleanly, no wedge, no synthetic "daemon restarting" error.
- Logs show `[daemon-connection] daemon upgraded: version=2.2.0+86ca610 pid=102812 → pid=103724` (same-version restart, no child exit). Tool calls served from the same child without a blip.

This matches the behavior target in the issue.

### ⚠️ Upgrade → respawn path is broken post-exit(75)

Forced a real version bump to exercise the `EXIT_DAEMON_UPGRADED` path (edited `crates/runtimed/Cargo.toml` `2.2.0 → 2.2.1`, cargo build -p runtimed, hot-swapped the binary, `systemctl restart`).

What worked:
- `DaemonConnection` correctly detected `version=2.2.0+86ca610 → 2.2.1+80ad263`.
- Child `runt-nightly mcp` emitted `Daemon upgraded, exiting for restart (exit code 75)` and the child process exited cleanly.
- Proxy saw `Child exited gracefully exit status: 75`.

What didn't:
- The next three tool calls (IDs 3, 4, 5 at 2s intervals) all returned `{"code":-32603,"message":"Child tool call failed: Transport closed"}`.
- Repeating log line: `Tool call failed but child still connected, not restarting: -32603: Child tool call failed: Transport closed`.
- No `child_spawned` log after the exit.

So the child exited 75, but neither the child-monitor task (`spawn_child_monitor` in `proxy.rs`) nor the tool-call fallback (`forward_tool_call`) respawns it. `forward_tool_call` at line 510 checks `child_client.as_ref().is_some_and(|c| !c.is_closed())` — that evaluates to "alive" even though the OS process is gone, so it short-circuits the restart path. The monitor task presumably should be the authoritative respawn, but didn't fire in this run.

I'll happily re-test whatever shape you land on. Not blocking per se — the wedge is the real user-visible bug and that's fixed — but a clean handoff after `exit(75)` is the whole reason the proxy exists.

### Environment

- Host: lab1 (ubuntu 24.04, x86_64)
- Binary: `/home/ubuntu/.local/share/runt-nightly/bin/runt-proxy-nightly` from branch HEAD `80ad263f`
- Probe: `/home/ubuntu/.claude/plugins/cache/nteract/nteract-nightly/0.1.0/bin/runt-proxy` (symlink target identical)
- Full stderr for both runs in `/tmp/mcp-{upg,upg2}-err.log` on lab1 if it helps

PR #2008, opened 2026-04-21
fix(mcp-proxy): respawn child after transport closes (follow-up to #2004)

rgbkrk:
Follow-up to #2004 based on the [lab smoke finding](https://github.com/nteract/desktop/pull/2004#issuecomment-4291230026): primary wedge is fixed, but the daemon-upgrade path still wedges after `exit(75)`.

## Root cause

rmcp 1.4's `RunningService::is_closed()` returns true only when:

1. Someone consumed the handle via `.waiting()` (the proxy never does — it stores the handle in `ProxyState`), OR
2. The `CancellationToken` was explicitly cancelled (the proxy doesn't cancel on child exit)

Neither fires when the child exits naturally. So after `runt-nightly mcp` logs `Daemon upgraded, exiting for restart (exit code 75)` and the rmcp service loop breaks out on EOF, `is_closed()` keeps returning false. Both the child-monitor poll loop and the `forward_tool_call` fallback gate their respawn on `is_closed()`, so neither fires. Result: every tool call returns `Transport closed` indefinitely.

The right signal is `Peer::is_transport_closed()` (`RunningService: Deref<Target=Peer>`, so it's already reachable on `RunningChild`). It flips true when the peer→service mpsc channel closes — which happens when the service loop exits, which happens when the child's stdout hits EOF.

## Changes

| File | Change |
|---|---|
| `crates/runt-mcp-proxy/src/proxy.rs` | Swap `is_closed()` → `is_transport_closed()` at 3 sites: monitor poll, `forward_tool_call` fallback, `forward_read_resource` fallback |
| `crates/runt-mcp-proxy/src/child.rs` | +98 lines of tests |

## Tests

Two new tokio tests in `child::tests` (both passing):

- `is_transport_closed_flips_when_child_exits` — spins up a `RunningChild` over `tokio::io::duplex`, aborts the server side, asserts `is_transport_closed()` becomes true within 2s.
- `is_closed_does_not_flip_on_child_exit` — documents rmcp 1.4's behavior. Asserts `is_transport_closed()` is true but `is_closed()` stays false after the same teardown. If upstream ever fixes `is_closed()`, this test fails loudly and we can remove the workaround.

No external processes, no daemon, no flake surface — just duplex pipes.

## Test plan

- [x] `cargo xtask clippy` clean
- [x] `cargo test -p runt-mcp-proxy` — 85 passed (+2 from baseline)
- [ ] Lab re-test of the upgrade path: bump daemon version, confirm proxy respawns child cleanly after `exit(75)` without the "Transport closed" loop

## Notes

Small, surgical. Doesn't touch #2004's design choices or the health loop deletion. The name `is_closed()` is [misleading in rmcp 1.4](https://docs.rs/rmcp/1.4.0/rmcp/service/struct.RunningService.html#method.is_closed) — worth an upstream issue, but not blocking.

### DIFF
--- a/crates/runt-mcp-proxy/src/child.rs
+++ b/crates/runt-mcp-proxy/src/child.rs
@@ -80,3 +80,101 @@ pub async fn spawn_child(
 
     Ok(client)
 }
+
+#[cfg(test)]
+mod tests {
+    //! Regression test for the post-`exit(75)` respawn gap.
+    //!
+    //! `RunningService::is_closed()` in rmcp 1.4 returns true only when the
+    //! handle has been consumed by `.waiting()` OR the cancellation token
+    //! has been cancelled — neither happens when the child process exits on
+    //! its own. `is_transport_closed()` (via `Peer::is_transport_closed`)
+    //! *does* flip when the service loop breaks because the transport
+    //! reader returned None.
+    //!
+    //! This matters for `proxy.rs`: the monitor loop and the
+    //! `forward_tool_call` fallback use this check to decide whether the
+    //! child is gone and a restart is needed. Using `is_closed()` wedges
+    //! the proxy after a daemon-upgrade exit(75) — the child is dead but
+    //! the proxy never respawns it. See the lab finding on PR #2004.
+    use super::*;
+    use rmcp::{ServerHandler, ServiceExt};
+    use std::time::{Duration, Instant};
+
+    /// Minimal `ServerHandler` — all methods use rmcp's defaults.
+    struct NoopServer;
+    impl ServerHandler for NoopServer {}
+
+    /// Bring up a `RunningChild` over an in-memory duplex pipe, then drop
+    /// the server side. The client's service loop should observe EOF on
+    /// read and break out.
+    async fn dead_transport_child() -> RunningChild {
+        let (server_side, client_side) = tokio::io::duplex(1024);
+
+        // Server handshake + hold the transport alive until we drop it.
+        let server_handle = tokio::spawn(async move {
+            let server = NoopServer.serve(server_side).await.expect("serve server");
+            // Park until the client disconnects.
+            let _ = server.waiting().await;
+        });
+
+        let handler = ChildClientHandler {
+            upstream_name: "test".to_string(),
+            upstream_title: None,
+        };
+        let client = handler
+            .serve(client_side)
+            .await
+            .expect("client handshake completes");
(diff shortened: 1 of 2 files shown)


rgbkrk:
## ✅ Lab verified — upgrade path recovers cleanly

Re-ran the upgrade scenario that wedged after #2004. Built `fix/mcp-proxy-respawn-after-exit-75` on lab2 (fresh from `main`, commit `111f2772`), bumped daemon version `2.2.0 → 2.2.1`, hot-swapped the binary, restarted the user-systemd unit. Same headless MCP probe as before.

**Result:** proxy respawns the child after `exit(75)` and the next tool call serves content:

```
{"id":3,"result":{"content":[
  {"type":"text","text":"[nteract] Daemon upgraded (2.2.0+111f277 → 2.2.1+2d4b2da)\n"},
  {"type":"text","text":"[]"}
],"isError":false}}
```

Note the reconnection-message prefix — the end-to-end machinery (detect upgrade → exit 75 → respawn → prepend message on next call) is working now, not just the fix for `is_closed()`.

Relevant stderr lines (ordered):
```
Daemon upgraded, exiting for restart (exit code 75).
[daemon-connection] daemon upgraded: version=2.2.1+2d4b2da pid=32249
Child exited gracefully exit status: 75
Tool call failed and child transport closed, attempting restart: -32603: Child tool call failed: Transport closed
Child process exited, attempting automatic restart event="child_exited" uptime_secs=10
Child monitor: restart successful, exiting (new monitor spawned)
Spawning child: /usr/local/bin/runt-nightly mcp (client="upg-probe-v3")
```

Both respawn paths (monitor poll + `forward_tool_call` fallback) fired, as designed. Subsequent calls (ids 4, 5) clean, no Transport-closed artifacts.

### Same-version restart still works

Sanity check that the #2004 fix wasn't regressed. `systemctl --user restart runtimed-nightly` with no version change: DaemonConnection reported `daemon upgraded: version=... pid=...→...` (same version, new PID) and tool calls flowed through the existing child without a blip — no child exit, no respawn, which is what should happen for a same-version restart.

LGTM.

### Environment

- Host: lab2 (ubuntu 24.04, x86_64)
- Binary: `/home/ubuntu/.local/share/runt-nightly/bin/runt-proxy-nightly` from `2d4b2dab`
- Full stderr in `/tmp/mcp-upg3-err.log` on lab2

PR #2028, opened 2026-04-22
feat(telemetry): add opt-out daily heartbeat pings

rgbkrk:
## Summary

Wire the desktop app, daemon, and MCP server to send anonymous daily heartbeat pings to the deployed `telemetry.runtimed.com` endpoint. Each source (`app`, `daemon`, `mcp`) sends one ping per day with six fields: `install_id`, `source`, `version`, `channel`, `platform`, `arch`.

- Shared gate logic and payload builder in `runtimed-client` - long-running processes (daemon, MCP) loop hourly, app fires once per launch
- Opt-out via onboarding toggle, `runt config telemetry disable`, or `NTERACT_TELEMETRY_DISABLE=1`
- Dev mode, CI, and unfinished onboarding all suppress pings - no data sent before the user sees the disclosure
- `docs/telemetry.md` documents the full schema, retention (400-day raw, indefinite anonymized rollups), and all opt-out paths

## Verification

- [ ] `runt config telemetry status` shows expected gates and last-ping times
- [ ] `runt config telemetry disable` then restart daemon - confirm no ping fires
- [ ] `runt config telemetry enable` then restart daemon - confirm ping fires
- [ ] Fresh install: onboarding shows telemetry disclosure with toggle before "Get Started"
- [ ] Toggling telemetry off during onboarding persists as `telemetry_enabled: false`
- [ ] Under `RUNTIMED_DEV=1` (direnv), `runt config telemetry status` shows dev mode gate
- [ ] Point `NTERACT_TELEMETRY_ENDPOINT` at a local echo server, confirm POST body has exactly 6 keys
- [ ] Leave daemon running >20h (or clear `telemetry_last_daemon_ping_at`), confirm second ping fires

### DIFF
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -4308,6 +4308,11 @@ pub fn run(
             )?;
             app.set_menu(menu)?;
 
+            // Fire-and-forget app telemetry heartbeat
+            tokio::spawn(async {
+                runtimed::telemetry::heartbeat_once("app", "telemetry_last_app_ping_at").await;
+            });
+
             let has_session_to_clear = restored_session.is_some();
 
             // Ensure runtimed is running (required for daemon-only mode)
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -4308,11 +4308,6 @@ pub fn run(
             )?;
             app.set_menu(menu)?;
 
-            // Fire-and-forget app telemetry heartbeat
-            tokio::spawn(async {
-                runtimed::telemetry::heartbeat_once("app", "telemetry_last_app_ping_at").await;
-            });
-
             let has_session_to_clear = restored_session.is_some();
 
             // Ensure runtimed is running (required for daemon-only mode)
@@ -4359,6 +4354,16 @@ pub fn run(
                 // Check if CLI symlinks are current and silently update if stale.
                 // This handles app reinstalls, bundle path changes, and channel switches.
                 cli_install::ensure_cli_current(&app_for_daemon);
+
+                if daemon_available {
+                    tokio::spawn(async {
+                        runtimed::telemetry::heartbeat_once(
+                            "app",
+                            "telemetry_last_app_ping_at",
+                        )
+                        .await;
+                    });
+                }
 
                 // Start settings sync subscription (reconnects automatically)
                 // Spawn as separate task since it runs forever
--- a/crates/runt/src/main.rs
+++ b/crates/runt/src/main.rs
@@ -839,7 +839,7 @@ async fn run_mcp_server(no_show: bool) -> Result<()> {
         runt_mcp::daemon_watch::watch(daemon_conn, watch_socket, session, peer_label).await
     });
 
-    let heartbeat_handle = tokio::spawn(async {
+    tokio::spawn(async {
         runtimed_client::telemetry::heartbeat_loop("mcp", "telemetry_last_mcp_ping_at").await;
     });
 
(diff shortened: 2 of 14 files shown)


crates/runt/src/main.rs
```diff
@@ -842,6 +859,7 @@ async fn run_mcp_server(no_show: bool) -> Result<()> {
             tokio::time::sleep(std::time::Duration::from_millis(200)).await;
             std::process::exit(code);
         }
+        _ = heartbeat_handle => {}
```
rgbkrk: `heartbeat_loop()` is not a daemon task here; it can return immediately when telemetry is suppressed (`RUNTIMED_DEV`, `CI`, `NTERACT_TELEMETRY_DISABLE`) or if the HTTP client cannot be built. Because this `select!` treats `_ = heartbeat_handle` as a terminal branch, `runt mcp` falls through `Ok(())` and exits right after startup in those cases. That breaks the repo-default dev path, since `nteract-dev` runs the MCP child under `RUNTIMED_DEV=1`. See the early returns in [`crates/runtimed-client/src/telemetry.rs`](https://github.com/nteract/desktop/blob/fe9636abc9d820cc36f14b2ed322c2f23aa80fb4/crates/runtimed-client/src/telemetry.rs#L151-L162).

crates/runtimed/src/daemon_telemetry.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    loop {
+        try_send_daemon_heartbeat(&daemon, &client).await;
+        tokio::time::sleep(std::time::Duration::from_secs(60 * 60)).await;
+    }
+}
+
+async fn try_send_daemon_heartbeat(daemon: &Arc<Daemon>, client: &reqwest::Client) {
+    let (settings, install_id) = {
+        let mut settings_doc = daemon.settings.write().await;
```
rgbkrk, line 37: These writes never go through the settings sync persistence path, so they stay in memory only. The existing sync flow only saves/broadcasts settings when a client mutation hits `persist_settings()` and `settings_changed.send(())` in [`crates/runtimed/src/sync_server.rs`](https://github.com/nteract/desktop/blob/fe9636abc9d820cc36f14b2ed322c2f23aa80fb4/crates/runtimed/src/sync_server.rs#L75-L80). As written, every daemon restart drops `install_id`/`telemetry_last_daemon_ping_at`, so the next daemon boot mints a fresh install id and `runt config telemetry status` still reports `Last daemon ping: never` because it reads `settings.json` from disk.

crates/notebook/src/lib.rs
```diff
@@ -4308,6 +4308,11 @@ pub fn run(
             )?;
             app.set_menu(menu)?;
 
+            // Fire-and-forget app telemetry heartbeat
+            tokio::spawn(async {
```
rgbkrk: This still races the app heartbeat ahead of daemon startup. `heartbeat_once()` generates and immediately uses a fresh `install_id`, but its persistence path only tries the daemon socket and has no disk fallback ([`crates/runtimed-client/src/telemetry.rs#L212-L215`](https://github.com/nteract/desktop/blob/d2d04a40154095930f62fdd847d813af7c677296/crates/runtimed-client/src/telemetry.rs#L212-L215), [`#L256-L271`](https://github.com/nteract/desktop/blob/d2d04a40154095930f62fdd847d813af7c677296/crates/runtimed-client/src/telemetry.rs#L256-L271)). On launches where `runtimed` is not already running, the app can send an `app` ping with a transient UUID before `ensure_daemon_via_sidecar` brings the daemon up; the daemon then starts its own heartbeat immediately and persists a different `install_id` ([`crates/runtimed/src/main.rs#L433`](https://github.com/nteract/desktop/blob/d2d04a40154095930f62fdd847d813af7c677296/crates/runtimed/src/main.rs#L433)). That splits one installation into two distinct installs in telemetry.

Issue #1816, opened 2026-04-15
Design: streaming Arrow — dx emits Arrow IPC stream for incremental rendering


rgbkrk:
Pair to the interactive-query backend #1815 but focused on emission, not query.

### Idea

Instead of always serializing a whole DataFrame to parquet up front, emit an Arrow IPC stream that renderers and the MCP can pull from incrementally. Useful for:

- Large DataFrames where the user only ever looks at the first page
- Progressive rendering in the sift grid
- Pushing incremental rows while a long `to_pandas()` or remote fetch is still streaming

### Design questions

1. Wire-level: new MIME `application/vnd.apache.arrow.stream` vs. continuing with parquet + chunked parquet?
2. Does the kernel produce it directly (via pyarrow) or does dx buffer and re-encode?
3. What does the frontend do with it? Sift today reads parquet; would it read Arrow IPC?
4. Interaction with blob store — one blob per stream chunk, or one blob per complete stream?
5. Reader library story on the frontend (WASM Arrow reader already exists via sift).

### Likely scope breakdown

- Python-side: new optional emission path in dx (alongside the current parquet path).
- Runtime agent: forward chunks through the existing buffer mechanism.
- Frontend/sift: consume the Arrow IPC stream incrementally.

### Related

- #1815 — interactive query backend (both involve efficient access to cell outputs)
- dx current parquet path: `python/dx/src/dx/_format.py`, `python/dx/src/dx/_format_install.py`

rgbkrk:
Closing as parked product exploration. This is a nice-to-have design idea, but we should shape this from concrete user feedback rather than keeping it in the active issue list.

Issue #1815, opened 2026-04-15
Design: interactive query backend for dx (SQL over parquet outputs)


rgbkrk:
dx currently emits parquet blobs + a text/llm+plain summary. A natural next capability: let the agent (or a user) run interactive queries against those outputs without re-running the cell.

### Idea

- A running cell produces parquet (already done).
- dx keeps a reference (blob hash) and exposes a query tool/interface.
- The agent can call `dx.query("SELECT ... FROM cell_<id>")` or the MCP can expose a `query_cell_output` tool.
- Execution backend options:
  - duckdb over the stored parquet bytes (no extra daemon state, probably the simplest)
  - in-kernel via dataframe library (`df.query()`, `pl.SQLContext`, etc.)

### Design questions to answer in a spec

1. Does the query run inside the kernel (re-using the user's Python env) or daemon-side (in a dedicated duckdb sandbox)?
2. Scope: per-cell, per-notebook, or notebook-wide (joinable outputs)?
3. MCP surface: one tool, one input schema?
4. Blob store implications: how long do we retain parquet blobs for query?
5. Write-side: does dx need to annotate output MIME with a table name for later lookup?

### Related

- Streaming Arrow design (#1815? — sibling issue)
- dx blob retention/GC (task #18)

rgbkrk:
Closing as parked product exploration. This is a nice-to-have design idea, but we should shape this from concrete user feedback rather than keeping it in the active issue list.

PR #2044, opened 2026-04-22
docs(specs): streaming Arrow IPC for DataFrame repr (#1816)

rgbkrk:
Design spec for #1816.

**Shape:** dx emits a small Parquet head immediately (~100 rows) plus a pull handle over a reserved `nteract.dx.stream.<id>` comm. The runtime agent runs a pull loop outside the execution hot path, putting each Arrow IPC chunk in the blob store and appending a ref to a new `arrow-stream-chunks` manifest field on the same `output_id`. Late joiners replay from the CRDT like any other output.

**Why this shape:**
- Head stays Parquet so sift's existing load path lights up immediately and metadata/schema discovery works as today.
- Continuation is Arrow IPC because record-batch-per-chunk is lighter than re-serializing Parquet per chunk and sift-wasm already has `load_ipc` + append.
- Runtime agent owns the loop — not the kernel — so the cell's busy spinner clears at the head. No kernel threads.
- CRDT routing (not a side channel) means late joiners and refreshes get progressive rendering for free.
- No ContentRef shape change, no blob-store mutability — the chunk list is JSON inside the manifest, and each chunk is its own immutable content-addressed blob.

**Shared with #1815:** The stream handle is first-class. Repr is the first producer; a query result (#1815) is the second. Same transport.

**Not in scope:**
- Query pushdown (#1815).
- Mutable blobs.
- Replacing Parquet.

**Open for implementation-time decisions** (spec's final section): threshold source (rows vs bytes), pull concurrency, back-pressure, handle lifetime across cell re-execution.

## Review plan

- [ ] Read [`docs/superpowers/specs/2026-04-22-streaming-arrow-ipc-design.md`](../blob/docs/streaming-arrow-ipc-design/docs/superpowers/specs/2026-04-22-streaming-arrow-ipc-design.md)
- [ ] Confirm the output model (one `output_id` + growing chunk list vs. one-output-per-chunk) is the right trade
- [ ] Confirm the comm namespace use (`nteract.dx.stream.*`) fits the reserved-namespace rule in CLAUDE.md
- [ ] Sanity-check the four open questions at the bottom — pin any you'd rather decide now

### DIFF
--- a/docs/superpowers/specs/2026-04-22-streaming-arrow-ipc-design.md
+++ b/docs/superpowers/specs/2026-04-22-streaming-arrow-ipc-design.md
@@ -1,0 +1,199 @@
+# Streaming Arrow IPC for DataFrame repr
+
+**Status:** Draft
+**Date:** 2026-04-22
+**Related issues:** #1816 (this spec), #1815 (query backend — same transport), #1905 / #1911 / #1913 (per-task stable actors that make fork+merge on async paths safe), [2026-04-19 addressable execution outputs](./2026-04-19-map-keyed-outputs.md)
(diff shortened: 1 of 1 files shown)


rgbkrk:
Nice to have later after other stuff stabilizes.

PR #2085, opened 2026-04-23
feat(runtime-doc): RuntimeLifecycle CRDT keys + typed writers (phase 2)

rgbkrk:
## Summary

Phase 2 of the `RuntimeLifecycle` refactor spec'd at `docs/superpowers/specs/2026-04-23-runtime-lifecycle-enum-design.md`. Adds typed CRDT keys and writers alongside the existing string-form setters. No caller migration — readers of either shape see consistent state through a reconciliation function.

Ready for review. All Phase 2 scoped work is green.

## What lands

- **CRDT scaffold** — `kernel/lifecycle`, `kernel/activity`, `kernel/error_reason` added to `new()` / `new_with_actor()` alongside the existing `status` / `starting_phase`.
- **Typed writers** — `set_lifecycle`, `set_activity`, `set_lifecycle_with_error` on `RuntimeStateDoc`. Typed setters mirror into the string keys; string setters don't touch typed keys.
- **`KernelState.error_reason: Option<String>`** — `None` when the `error_reason` key is absent (pre-Phase-2 docs), `Some("")` when scaffolded but unset, `Some("reason")` when populated.
- **`resolve_lifecycle`** in `types.rs` — the priority rule. When the typed lifecycle's `to_legacy()` projection disagrees with the string pair, a string setter ran more recently; trust the string shape. When they agree, typed wins (preserves `Running(Unknown)` through the lossy string channel).
- **`RuntimeLifecycle::to_legacy`** — inverse of `from_legacy`, used internally by the mirror writes and independently tested.
- **`read_str_if_present`** helper — returns `None` only when the key is absent, preserving the `Option<String>` contract on `error_reason`.

## Dual-shape invariants

- Typed setters always mirror into string keys.
- String setters don't touch typed keys.
- `read_state` reconciles via `resolve_lifecycle` — whichever shape was written most recently wins.
- `set_activity` throttle skips only when BOTH typed activity AND mirrored string status already match.
- `set_lifecycle` never touches `error_reason`; only `set_lifecycle_with_error` owns it.
- Leaving `Running` always clears `activity` to `""`.

## Review history (codex review --base main, 3 passes)

Each pass found dual-shape correctness bugs the prior commit introduced. All landed with targeted tests.

**Pass 1 (commit `a10588b8`)**
- [P2] Scaffolded docs reported `lifecycle = NotStarted` while `status = "busy"` when only string setters had run. Fixed by extracting reconciliation into `resolve_lifecycle` (types.rs).
- [P2] `set_activity` throttle was blind to string-key drift. Fixed by checking both typed + string state before no-op'ing.

**Pass 2 (commit `85ad9edd`)** — the fixes landed.

**Pass 3 (commit `600c7b96`)**
- [P2] `read_str` collapsed absent `error_reason` to `""`, breaking the documented `Option<String>` contract for pre-Phase-2 docs. Fixed by adding `read_str_if_present` that returns `None` only when the key is absent.

## Known Phase-3 prerequisites (NOT addressed here)

Codex's final pass (against `600c7b96`) flagged two remaining compat gaps. Both belong in Phase 3 because a clean fix requires introducing a new type, not another stringly layer:

### 1. `missing_ipykernel` pixi-install-prompt compat

The current frontend gates the pixi-install prompt on `starting_phase == "missing_ipykernel"`. Today a caller writes `set_kernel_status("error")` + `set_starting_phase("missing_ipykernel")`; with Phase 2 they'd migrate to `set_lifecycle_with_error(Error, Some("missing_ipykernel"))`. My `to_legacy()` for `Error` returns `("error", "")`, so the frontend prompt breaks until Phase 3 migrates `NotebookToolbar`.

**A quick fix would mirror the reason string into `starting_phase` when lifecycle is `Error`. Kyle flagged this as "very string like" — right call. The real fix is a typed `KernelErrorReason` enum (variants `MissingIpykernel`, `KernelDied`, …, plus `Other(String)`) rather than free-form strings. That's a Phase 3 API change: `set_lifecycle_with_error` takes `Option<&KernelErrorReason>`, the frontend matches on the enum variant instead of a magic string.**

Deferring to Phase 3 so we don't bolt another stringly layer onto Phase 2 that we'd just rip out.

### 2. `set_activity` clearing stale `starting_phase`

After the old launch path ends with `set_starting_phase("connecting")`, the first `set_activity(Idle)` IOPub event leaves `starting_phase = "connecting"` while mirroring `status = "idle"`. `set_kernel_status` itself would clear `starting_phase` in this case. Fix is straightforward (have `set_activity` clear the phase when mirroring a non-starting status) and small; folding it in is fine in Phase 2 or Phase 3.

I held off in this PR because Kyle was heading to bed and I wanted the PR at a known-good commit boundary rather than racing another codex round.

## Test plan

- [x] `cargo test -p runtime-doc --lib` — 138 passing (35 new).
- [x] `cargo check --workspace` — clean.
- [x] `cargo xtask lint` — clean.
- [x] `cargo test -p notebook-sync -p runtimed --lib` — 430 passing.

### Adversarial coverage

- `set_lifecycle(Error)` preserves `error_reason`; `set_lifecycle_with_error(_, None)` clears it.
- `set_activity(Idle)` after `set_kernel_status("busy")` re-mirrors the string key.
- Truly redundant `set_activity(Busy)` does not advance heads.
- `Running(Unknown)` survives round-trip through the CRDT.
- String setter on a scaffolded doc produces correct `lifecycle` via the reconciliation fallback.
- Pre-Phase-2 doc (kernel map without error_reason key) reads as `error_reason == None`.
- Scaffolded doc reads as `error_reason == Some("")`.
- Fork + merge with concurrent lifecycle writes pins Automerge's conflict behavior.
- Every lifecycle variant round-trips through write + read.

## Follow-ups (Phase 3 backlog)

- Introduce `KernelErrorReason` enum; migrate `set_lifecycle_with_error` signature.
- `set_activity` clears stale `starting_phase` when mirroring non-starting status.
- Move `read_str_if_present` to `automunge` — it's a general Automerge primitive, doesn't belong in `runtime-doc`.
- Migrate daemon callers (jupyter_kernel IOPub, launch_kernel, peer.rs auto-launch, runtime_agent kernel-died) from the string setters to the typed setters.

### DIFF
--- a/crates/runtime-doc/src/doc.rs
+++ b/crates/runtime-doc/src/doc.rs
@@ -661,6 +661,21 @@ impl RuntimeStateDoc {
             })
     }
 
+    /// Read a string scalar from a map object, returning `None` only when
+    /// the key itself is absent. Unlike `read_opt_str`, an explicit `""`
+    /// value returns `Some("")` so callers can distinguish "scaffolded
+    /// but unset" from "key not present at all."
+    fn read_str_if_present(&self, obj: &automerge::ObjId, key: &str) -> Option<String> {
+        let (value, _) = self.doc.get(obj, key).ok().flatten()?;
+        match value {
+            Value::Scalar(s) => match s.as_ref() {
+                ScalarValue::Str(s) => Some(s.to_string()),
+                _ => None,
+            },
+            _ => None,
+        }
+    }
+
     /// Read a bool scalar from a map object.
     fn read_bool(&self, obj: &automerge::ObjId, key: &str) -> bool {
         self.doc
@@ -2011,9 +2026,11 @@ impl RuntimeStateDoc {
                 );
                 // error_reason is Option<String> so callers can tell "no
                 // kernel map at all" (None) from "scaffolded but unset"
-                // (Some("")). The unwrap_or_default above only fires when
-                // the whole kernel map is missing.
-                let error_reason = Some(self.read_str(k, "error_reason"));
+                // (Some("")). read_str_if_present returns None only when
+                // the key itself is absent — important for docs scaffolded
+                // before Phase 2 added the key, which have the `kernel`
+                // map but no `error_reason` field yet.
+                let error_reason = self.read_str_if_present(k, "error_reason");
                 KernelState {
                     status,
                     starting_phase,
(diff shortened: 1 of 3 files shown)


rgbkrk:
## Codex review trajectory

Three codex review passes against main. Findings and resolutions:

### Pass 1 (commit a10588b8)
- **[P2]** `read_state` trusted the typed `lifecycle` key whenever non-empty, so scaffolded docs mutated only through `set_kernel_status` reported `lifecycle = NotStarted` while `status = "busy"`. Fixed by extracting the reconciliation rule into `resolve_lifecycle` (types.rs) — typed vs. string shape agreement check via `to_legacy()`.
- **[P2]** `set_activity` throttle looked only at the typed activity key, missing string drift from a prior `set_kernel_status` call. Fixed by checking both keys before no-op'ing.

### Pass 2 (commit 85ad9edd)
- **[P2]** `read_str("error_reason")` collapsed an absent key to `""`, making pre-Phase-2 docs indistinguishable from Phase-2 scaffold state. The `Option<String>` contract was documented but not honored. Fixed by adding `read_str_if_present` helper (returns `None` only when the key is absent) and using it for `error_reason` in `read_state`.

### Pass 3 (commit 600c7b96)

Pending. Running now against the latest commit.

All three findings were genuine correctness issues in the dual-shape behavior this phase is setting up. The test suite grew from 124 → 138 as each fix landed with targeted tests pinning the new contract.

PR #2088, opened 2026-04-23
fix(runt-mcp): disconnect previous session on notebook switch

quillaid:
## Summary

- Adds `disconnect_previous_session()` helper to MCP session tools that explicitly shuts down the kernel and drops the old DocHandle before switching notebooks
- Called from both `open_notebook()` and `create_notebook()` before connecting to the new session
- Sends a best-effort `ShutdownKernel` request (3s timeout) so the daemon releases path mappings promptly, then drops the session to trigger sync task shutdown

## Problem

When an MCP agent switches notebooks (e.g. gremlin calling `create_notebook` or `open_notebook` while already in a session), the old session was simply replaced without cleanup. The daemon's `path_index` retained the old path → room mapping during the 30s eviction delay, causing "Cannot save: file already open in session X" errors on subsequent `save_notebook` calls.

This affected any gremlin that switched sessions mid-run (hit at turn 63 in Refiner suite).

## Design

- **Same-notebook reconnection is a no-op**: If `open_notebook(notebook_id=X)` is called while already connected to X, the disconnect is skipped
- **Path-based opens always disconnect**: We don't know the target notebook_id until after `connect_open`, so the old session is always cleaned up first
- **Best-effort shutdown**: `ShutdownKernel` errors are logged at debug level and ignored — the kernel may not be running, or the room may have already been evicted
- **No new protocol types**: Uses existing `ShutdownKernel` request + natural DocHandle drop behavior

## Test plan

- [x] All 95 existing `runt-mcp` tests pass
- [x] Release build succeeds
- [x] Nightly install succeeds (daemon running 2.2.1+f3b42ed)
- [ ] Gremlin replay: Refiner suite should no longer hit session lock conflict at turn 63

### DIFF
--- a/crates/runt-mcp/src/tools/session.rs
+++ b/crates/runt-mcp/src/tools/session.rs
@@ -23,6 +23,51 @@ async fn previous_notebook_id(server: &NteractMcp) -> Option<String> {
         .map(|s| s.notebook_id.clone())
 }
 
+/// Cleanly disconnect the previous session before switching to a new one.
+///
+/// Clears the session — closing channels and triggering sync task shutdown.
+/// The daemon handles kernel lifecycle: when the MCP peer disconnects and
+/// the peer count drops, the eviction timer starts. Other peers (e.g. the
+/// desktop app) may still be connected, so we do NOT send `ShutdownKernel`.
+///
+/// This prevents "Cannot save: file already open in session X" errors when an
+/// agent switches notebooks and then tries to `save_notebook` to a path that
+/// was held by the previous session. Without this, the daemon's `path_index`
+/// retains the old mapping during the 30s eviction delay.
+///
+/// When `new_notebook_id` is `Some`, the cleanup is skipped if we're
+/// reconnecting to the same notebook (no-op switch). When `None` (e.g.
+/// `create_notebook`), the old session is always disconnected.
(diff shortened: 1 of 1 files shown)


crates/runt-mcp/src/tools/deps.rs
rgbkrk, line 54: All this is in your prior pr. Probably should rebase

crates/runt-mcp/src/tools/session.rs
```diff
@@ -23,6 +23,67 @@ async fn previous_notebook_id(server: &NteractMcp) -> Option<String> {
...
+    if let Some(old) = old_session {
+        tracing::info!(
+            "[mcp] Disconnecting previous session {} before notebook switch",
+            old.notebook_id
+        );
+        // Best-effort kernel shutdown — ignore errors (kernel may not be
+        // running, or the room may have already been evicted).
+        match tokio::time::timeout(
+            std::time::Duration::from_secs(3),
+            old.handle.send_request(NotebookRequest::ShutdownKernel {}),
```
rgbkrk: I don't think you want to shutdown in this case as there might be a user on the document still.

Realistically this should be up to the daemon and is a daemon bug that kernels aren't stopping on your system.

rgbkrk (changes requested):

rgbkrk (approved):

Issue #2096, opened 2026-04-23
RuntimeLifecycle refactor — remaining cleanup


rgbkrk:
## Summary

Tracking the remaining cleanup after the `RuntimeLifecycle` refactor shipped across #2081, #2085, #2091, #2092, #2093. The core migration is done — typed `RuntimeLifecycle` + `KernelErrorReason` enums are the authoritative API in Rust, typed `RuntimeStatusKey` + `KERNEL_ERROR_REASON` constants on the TS side. What remains is consumers still on the string shape, retiring the dual-shape mirror once they move, and a few polish items.

Spec: `docs/superpowers/specs/2026-04-23-runtime-lifecycle-enum-design.md`.

## Consumers still on the string shape

- [x] **Python bindings (`crates/runtimed-py`).** `session_core.rs` migrated to `matches!(RuntimeLifecycle::Running(_) | ...)` patterns; `PyKernelState` now reads the typed `lifecycle`. Shipped in #2100.
- [x] **Python metrics scripts.** `kernel-reliability.py` and `sync-correctness.py` read the projected `KERNEL_STATUS` bucket via the `runtimed` Python wrapper (ships with #2100). They stay on the compressed bucket shape by design.
- [x] **Node bindings.** `crates/runtimed-node/src/session.rs:343` now uses `matches!(r.kernel.lifecycle, RuntimeLifecycle::Running(_))`.
- [x] **MCP wire emitter (`crates/runt-mcp`).** `tools/kernel.rs` wait loop matches on typed lifecycle; `tools/session.rs` projects via `to_legacy()` so the wire field stays unchanged. Shipped in #2098.
- [x] **CLI wire consumer (`crates/runt/src/main.rs`).** Intentionally left on the wire-level `status: String` — a comment at the read site points here; gated on the optional wire rename below.

## Retire the legacy shape

**Shipped in #2108:** CRDT-storage side retired — `kernel.status` / `kernel.starting_phase` keys no longer scaffolded in `new()`; typed writers call `clear_legacy_kernel_status` instead of dual-writing; `handle.rs` tests migrated; fixture state docs regenerated.

**In review (#2114):** TS-side deprecated shims removed.
- [x] `packages/runtimed/src/runtime-state.ts` — `KernelState.{status, starting_phase}` `@deprecated` shims deleted. (#2114)
[...]

PR #2102, opened 2026-04-23
fix(runtimed): pixi RPC path publishes MissingIpykernel reason

rgbkrk:
## Diagnosis

Group 3 of the RuntimeLifecycle cleanup (#2096) was scoped to add `MissingIpykernel` gating for uv and conda project-file launches. Codex review caught two P1 issues with that plan:

1. **uv and conda paths already self-heal.** `uv:pyproject` launches the kernel with `uv run --with ipykernel` ([jupyter_kernel.rs:255](crates/runtimed/src/jupyter_kernel.rs)), and the `conda:env_yml` path unconditionally appends `ipykernel` to the dep set before sync/create ([launch_kernel.rs:931](crates/runtimed/src/requests/launch_kernel.rs)). Adding a pre-launch gate that requires `ipykernel` to appear in `pyproject.toml` / `environment.yml` would turn currently working launches into `MissingIpykernel` failures — a regression, not a fix.
2. **The RPC pixi gate wrote the typed Error after it was clobbered.** `reset_starting_state` in `metadata.rs` rewrites `lifecycle` to `NotStarted`. The original plan had `set_lifecycle_with_error(Error, MissingIpykernel)` running before `reset_starting_state`, so the toolbar never saw the typed reason on the RPC path.

## What this PR does

Narrow to exactly one correctness fix: on the RPC path, tear down the starting latch first, then write the typed `Error` + `MissingIpykernel` reason. Now the RPC and auto-launch paths produce the same CRDT state. Add a comment above each gate explaining why it stays pixi-only.

## Files changed

- `crates/runtimed/src/requests/launch_kernel.rs` — swap `reset_starting_state` / `set_lifecycle_with_error` ordering. Add comment.
- `crates/runtimed/src/notebook_sync_server/metadata.rs` — comment only (pixi-only rationale).

Net diff: 2 files, +30/-3.

## Deviations from plan

The plan (`glittery-mixing-cake-agent-a159ae8c5194ebc28.md`) called for generalizing the pre-launch gate to `uv:pyproject` and `conda:env_yml`, adding `pyproject_toml_has_ipykernel` / `environment_yml_has_ipykernel` helpers in `project_file.rs`, extending tests, and branching the toolbar remediation by envSource. All of that landed in the first push, was caught by Codex on review, and is reverted here. The plan's diagnosis that "the daemon does NOT append base packages" for uv and conda project files was factually wrong — both paths inject ipykernel at launch.

The `MissingIpykernel` coverage for uv and conda is a runtime-failure-detection problem, not a pre-launch problem, and stays open under #2096 for a future pass.

## Test plan

- [x] `cargo build -p runtimed` — compiles
- [x] `cargo test -p runtimed --lib` — 389 passed
- [x] `pnpm exec vp test apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx` — 29 passed

## References

- Slice of #2096
- Builds on the `KernelErrorReason::MissingIpykernel` shape from #2091


rgbkrk:
Heads up: the first commit here (042a18c6) implemented Group 3 as originally planned — `MissingIpykernel` pre-launch gating for uv + conda + pixi with frontend branching. Codex review flagged two P1 issues that invalidate the whole uv/conda direction:

1. `uv:pyproject` runs via `uv run --with ipykernel` and `conda:env_yml` always appends `ipykernel` to the dep set before sync. Gating pre-launch on the project file would turn working launches into `MissingIpykernel` failures.
2. The RPC gate wrote `Error` + `MissingIpykernel` *before* `reset_starting_state`, which rewrites lifecycle to `NotStarted`. Toolbar never saw it.

The follow-up commit (d501fc6e) reverts the uv/conda work, reverts the toolbar branching and new tests, removes the dead helpers, and keeps one real correctness fix: ordering on the pixi RPC path so the typed Error sticks.

Final diff is 2 files, +30/-3. The uv/conda side of #2096 is still open — it's a runtime-failure-detection problem, not pre-launch gating.

PR #2108, opened 2026-04-23
refactor(runtime-doc): retire legacy kernel.status / starting_phase shape

rgbkrk:
## Summary

Group 5 of the RuntimeLifecycle cleanup (#2096) — the last one. Every writer in the repo has been migrated to the typed setters (`set_lifecycle` / `set_activity` / `set_lifecycle_with_error`), and no consumer reads the compressed `(status, starting_phase)` shape from the CRDT anymore. Delete the dual-shape mirror.

## Rust

- `KernelState`: drop `.status` and `.starting_phase`; impl `Default` via derive. Scaffold puts in `RuntimeStateDoc::new()` and `new_with_actor()` drop the two legacy keys.
- `set_kernel_status` / `set_starting_phase`: deleted. `set_lifecycle` and `set_activity` no longer mirror into the legacy keys; `set_lifecycle_with_error` no longer mirrors the reason into `starting_phase`.
- `resolve_lifecycle`: simplified to `RuntimeLifecycle::parse(lifecycle, activity).unwrap_or_default()`. `RuntimeLifecycle::from_legacy` deleted.
- `to_legacy` stays: `runt-mcp`, `runt inspect`, and `get_kernel_info` RPC still expose the compressed string shape on the wire (that's Group 6).
- Schema header at the top of `doc.rs` updated to match the new key set.
- `doc.rs` test module rewritten to exercise the typed shape directly rather than the dual-shape invariant. `handle.rs` tests migrated off `set_kernel_status`.
- Three new edge-case tests covering the subtleties that went away with the dual-shape:
  - `set_activity` without a preceding `Running` lifecycle leaves the activity key stranded (resolves to `NotStarted`).
  - `Running` with an empty activity key reads back as `Running(Unknown)`.
  - Every representative lifecycle variant survives a daemon→client sync round.

## Python bindings

- `PyKernelState` keeps `.status` and `.starting_phase` as string projections via `RuntimeLifecycle::to_legacy`, so the Python API stays source-compatible.
- Adds typed attributes: `lifecycle`, `activity`, `error_reason`.
- `__repr__` reads from `lifecycle` + `activity`.

## TypeScript

- `packages/runtimed/src/runtime-state.ts` — `KernelState.status` and `KernelState.starting_phase` removed. `DEFAULT_RUNTIME_STATE` updated.
- `packages/runtimed/src/derived-state.ts` — `kernelStatus$` and `throttleBusyStatus` deleted. The only consumer was `SyncEngine.kernelStatus$`, which was also unused externally and is removed here. `useDaemonKernel` rolls its own busy throttle.
- `KERNEL_STATUS` vocabulary + `lifecycleToLegacyStatus` stay. ~30 call sites in `App.tsx` and `NotebookToolbar.tsx` use the seven-bucket shape for UI predicates (color dots, icon state, starting-to-ready checks). Doc comments reframed as "UI convenience", not "legacy bridge".
- `apps/notebook/src/lib/kernel-status.ts` — the `@deprecated` `getKernelStatusLabel` and `STARTING_PHASE_LABELS` helpers deleted. Only `getLifecycleLabel` + `RUNTIME_STATUS_LABELS` remain. Tests expanded to cover every lifecycle variant through `getLifecycleLabel` plus the error-reason append / non-Error ignore / empty-reason fallback cases.

## WASM

- `runtimed_wasm_bg.wasm` rebuilt via `cargo xtask wasm runtimed`.
- `packages/runtimed/tests/fixtures/*/state_doc.bin` regenerated; the diff is just the removal of the two scaffold puts.

## Test plan

- [x] `cargo xtask clippy` (clean)
- [x] `cargo test -p runtime-doc --lib` — 126 passed (was 145; delta is the retired dual-shape tests, replaced with 3 new typed-only edge-case tests)
- [x] `cargo test -p runtimed --test integration -- --test-threads=1` — 29/29 passed
- [x] `cargo check -p runtimed-py` (PyO3 lib; links via maturin)
- [x] `pnpm exec vp test packages/runtimed` — 194 passed
- [x] `pnpm exec vp test apps/notebook` — 440 passed

Refs #2096. Closes the RuntimeLifecycle refactor series (#2081, #2085, #2091, #2092, #2093, #2098, #2099, #2100, #2102).


rgbkrk:
Addressed both findings from the codex review in 93c05ced.

**P1 — TypeScript compile failure in `sync-engine.ts`:** The debug log at line 639-640 was still reading `state.kernel?.status`, which no longer exists on `KernelState`. Projected through `lifecycleToLegacyStatus` so the log keeps emitting the seven-bucket shape the debug channel expects. `pnpm exec tsc --noEmit -p packages/runtimed/tsconfig.json` now passes.

**P2 — cross-version compat:** Restored `RuntimeLifecycle::from_legacy` and wired it back into `resolve_lifecycle` as a fallback for when the typed CRDT keys are absent. Pre-typed docs (captured test fixtures, `from_doc` with external bytes, cross-version in-flight sync frames) now read their lifecycle from the string shape rather than silently returning `NotStarted`.

Every in-repo writer goes through the typed setters, so the fallback is only hit for foreign docs. Three new tests pin the behavior:

- `resolve_falls_back_to_string_shape_when_typed_is_absent` — direct unit test on `resolve_lifecycle`.
- `pre_typed_doc_reads_lifecycle_from_string_shape` — integration via `RuntimeStateDoc::read_state` on a raw doc with only `kernel.status = "busy"`.
- `pre_typed_doc_reads_starting_sub_phase` — same, covering the `starting_phase` sub-state.

runtime-doc test count: 126 → 129. TS + Rust + vitest all still clean.

rgbkrk:
Round 2 fixes in b7330e65.

**P2 — mixed-shape docs:** Previous pass treated typed keys as authoritative whenever they parsed, so a doc with both typed (stale `NotStarted`) and legacy (freshly-written `"busy"`) keys would misread. Restored the "whichever shape was written most recently wins" rule: compare the typed lifecycle's `to_legacy()` projection against the actual string pair — if they agree, typed wins; if they disagree, trust the string shape. Two new tests pin this (`resolve_mixed_shape_prefers_legacy_string_when_shapes_disagree`, `resolve_mixed_shape_prefers_typed_when_shapes_agree`).

**P3 — Python stub:** `python/runtimed/src/runtimed/_internals.pyi` now declares `starting_phase`, `lifecycle`, `activity`, and `error_reason` on `KernelState`, and the `status` docstring lists `"awaiting_trust"`.

runtime-doc test count: 129 → 131. All TS + Rust + vitest still green.

rgbkrk:
Round 3 fixes in 927dda43.

**P2a — typed writes now clear stale legacy keys.** `set_lifecycle` deletes `kernel.status` and `kernel.starting_phase` scalars when it runs. Previously a typed write against a pre-typed doc (from `from_doc`, captured fixture, cross-version sync frame) left those scalars behind, and `resolve_lifecycle`'s mismatch fallback would read them back instead of the fresh typed state. New test `typed_write_clears_stale_legacy_keys` pins it end-to-end.

**P2b — TS `KernelState.status` / `.starting_phase` restored as `@deprecated` shims.** `packages/runtimed` is a public TS contract and removing the fields broke source-compat. Added them back, populated from `RuntimeLifecycle::to_legacy()` at the Rust `read_state` boundary so they're always fresh relative to `lifecycle`. Rust `KernelState` gets the same pair back (serde flows through to WASM/TS). Matches the Python binding's shape: typed lifecycle is source of truth, flat strings are derived-at-read shims.

- `cargo test -p runtime-doc --lib` — 132 passed (+1 new)
- `pnpm exec tsc --noEmit -p packages/runtimed/tsconfig.json` — clean
- 194 + 440 vitest, clean
- Full `cargo xtask wasm` rebuild

rgbkrk:
Round 4 fixes in b4c51b47.

**P2a — recover legacy error_reason on read.** Pre-lifecycle docs encoded the reason via `starting_phase` (`status="error"` + `starting_phase="missing_ipykernel"`). `NotebookToolbar` now gates the pixi prompt on `kernel.error_reason`, so older docs were silently losing the prompt. Read path now falls back to stored `starting_phase` when the typed `error_reason` key is absent AND lifecycle resolved to `Error` AND stored status is `"error"`. New test `pre_typed_doc_recovers_error_reason_from_starting_phase`.

**P2b — `set_activity` clears stale legacy keys.** `jupyter_kernel` sends idle/busy through `set_activity`, and the previous pass short-circuited when typed activity already matched. Pre-typed docs stayed stuck reading `Connecting` after the first IOPub idle because the legacy keys weren't cleared. Short-circuit now also requires absent legacy keys. New test `set_activity_clears_stale_legacy_keys_on_pre_typed_doc`.

- `cargo test -p runtime-doc --lib` — 134 (+2 new)
- `cargo xtask wasm` (full chain), lint, TC, 194 + 440 vitest — all clean

rgbkrk:
Pushing back on round 5. All three findings are "keep deprecated shims for hypothetical external consumers." Verified empirically:

\`\`\`
# SyncEngine.kernelStatus$ — zero external consumers
grep -rn "\\.kernelStatus\$\|engine\\.kernelStatus\|syncEngine\\.kernelStatus" apps/ packages/ --include="*.ts" --include="*.tsx"
(no matches)

# kernelStatus$ / throttleBusyStatus — zero consumers tree-wide
grep -rn "kernelStatus\$\|throttleBusyStatus" apps/ packages/ --include="*.ts" --include="*.tsx"
(no matches)

# getKernelStatusLabel — zero consumers tree-wide
grep -rn "getKernelStatusLabel" apps/ packages/ --include="*.ts" --include="*.tsx"
(no matches)
\`\`\`

\`@nteract/runtimed\` isn't published as an npm package. This is a closed monorepo; every consumer lives in \`apps/notebook/\` or \`packages/*/\` and has already been migrated. Keeping these as deprecated shims contradicts the stated goal of the PR ("grep comes back empty tree-wide") and adds dead code that future readers have to prove is dead again.

If a future out-of-tree consumer does show up, the typed API (\`runtimeState$.kernel.lifecycle\` + \`getLifecycleLabel\`) is the one we want them on. Reintroducing bridges to the retired shape would just delay that.

Merging as-is.

rgbkrk:
CI results in now. Three failures, all pre-existing and reproducing on main, none caused by this PR. Flagging them here as canaries rather than trying to fix them in Group 5:

1. **\`test_untitled_notebook_persists_through_eviction\`** (Rust integration) — eviction-then-reload path. Reconnected client got 0 cells instead of 2. The test polls for the persisted file size > 0 before reconnecting, but that doesn't prove the cells are written; only the schema-version scaffold is. Possible root cause: the debouncer can flush the scaffold before the first cell-add flush. Worth a separate investigation.

2. **\`test_uv_inline_deps_trusted\`** (Python integration) — \`import requests; print(requests.__version__)\` returns \`status=error, outputs=1\` with empty stderr. Same test flaked on #2102 under CI load. uv env resolution under parallel CI pressure.

3. **\`E2E: UV Pyproject\` "should execute code"** — reads \`"0.28.1\\n"\` (httpx version) instead of the \`sys.executable\` output the test just wrote. Test pollution on webdriver retry: test 4 writes \`import httpx\` to cell 0, test 3's retry then races against stale cell 0 output. Tests in the \`describe\` share a single fixture cell and aren't isolated.

\`test_untitled_notebook_persists_through_eviction\` failed on this PR's run but passed on the latest main run; \`test_uv_inline_deps_trusted\` passed on main and failed here; \`E2E: UV Pyproject\` fails on both this PR and main. None correlate with the KernelState retirement.

If you want me to retry CI to confirm the two Rust/Python failures are transient, happy to. Otherwise these are separate work.

PR #2110, opened 2026-04-23
fix(runtimed): unflake integration suite under parallel load

rgbkrk:
## Canary

`#2108`'s CI run flaked on `test_untitled_notebook_persists_through_eviction` with "reconnected client should see persisted cells for untitled notebook, got: []". Investigated per the "flakes as canaries" rule rather than retrying.

The real failure is upstream. The test panics at its 5s `wait_for_daemon` timeout during boot — the daemon hadn't finished starting up in 5s under 29-way parallel load. Same panic trips a handful of other tests in the same run, with the same shape.

Reproduces deterministically locally: `cargo test -p runtimed --test integration` fails ~6 tests with `wait_for_daemon` timing out.

### 1. Blob port: let integration tests skip the preferred-port path

`start_blob_server` calls `bind_preferred_or_random`, which walks a 10-port range looking for a free port before falling back to OS-assigned. With 29 daemons racing for 10 ports, the losers hit sequential `EADDRINUSE` retries that stretch daemon boot.

Added `DaemonConfig::use_preferred_blob_port: bool`, default `true`. Production behavior unchanged (MCP Apps with baked-in CSPs still get the stable port). `test_config` sets it `false`, so integration tests bind `127.0.0.1:0` straight away.

### 2. Settings file: let integration tests have their own

`Daemon::new` was loading `SettingsDoc` from the channel's global paths (`~/.cache/runt-nightly/settings.automerge` and the `~/.config/runt-nightly/settings.json` mirror). 29 parallel daemons hit the same two files at boot.

Added `DaemonConfig::settings_doc_path` and `settings_json_path` as `Option<PathBuf>` overrides, threaded through to `Daemon::new` AND the settings-watch save site. Resolver methods on the config type so the three call sites share one lookup. `test_config` pins both to the per-test temp dir.

### 3. Daemon-ready timeout: actually the dominant cause

The first two helped but didn't fully resolve it. Turns out `wait_for_daemon` had a 5s timeout at every call site (22 of them). Daemon boot is sub-second in isolation; under CPU thrash it stretches to 10+ seconds.

Bumped to 20s for the daemon socket, 15s for the initial Automerge sync (`wait_for_session_ready`, 19 sites). Named constants (`DAEMON_READY_TIMEOUT`, `SESSION_READY_TIMEOUT`) so there's one place to tune. A genuinely hung daemon still surfaces — `cargo test`'s own timeout catches anything the integration helpers miss.

## Why keep all three

Tested the timeout bump alone: it's enough to make the suite pass reliably. The other two could be saved for later. But:

- `use_preferred_blob_port` is the right API regardless. Dev throwaway daemons in cloud sandboxes and any future parallel test runner benefit from it. Without the opt-out, the preferred-port path is mandatory.
- Per-test settings isolation is correct hygiene. 29 tests shouldn't share a user-level file, full stop.

## Test plan

- [x] `cargo test -p runtimed --test integration`: 29 passed, 3 consecutive runs (14-15s each)
- [x] `cargo test -p runtimed --lib`: 389 passed
- [x] `cargo xtask clippy`: clean
- [x] `cargo xtask lint --fix`: clean

## Out of scope (separate canaries)

- `test_uv_inline_deps_trusted` (Python integration). Different root cause: uv env resolution under load. Needs its own investigation.
- `E2E: UV Pyproject` "should execute code". Test pollution between `it()` blocks on webdriver retry — the first test writes `sys.executable` to cell 0, the second writes `httpx.__version__` to the same cell, and a retry of the first reads the stale output. Test isolation fix, not a daemon fix.

### DIFF
--- a/crates/runtimed/src/blob_server.rs
+++ b/crates/runtimed/src/blob_server.rs
@@ -39,18 +39,31 @@ use crate::task_supervisor::{spawn_best_effort, spawn_supervised};
 /// bump budget stays in sync with the per-channel range carve-out.
 const PREFERRED_PORT_ATTEMPTS: u16 = runt_workspace::PREFERRED_BLOB_PORT_RANGE;
 
-/// Start the blob HTTP server on a random localhost port.
+/// Start the blob HTTP server.
 ///
 /// Returns the port the server is listening on. The server runs as a
 /// spawned task on the current tokio runtime.
 ///
 /// When `daemon` is provided, a panic in the accept loop triggers shutdown.
 /// Pass `None` in tests where no daemon is available.
+///
+/// When `use_preferred_port` is `true`, binds against the channel's
+/// preferred port with a bounded bump range before falling back to an
+/// OS-assigned port. When `false`, skips straight to OS-assigned.
+/// Integration tests and other contexts that run dozens of daemons in
+/// close proximity should pass `false` to avoid `EADDRINUSE` retry
+/// chains pushing boot past their `wait_for_daemon` timeout.
 pub async fn start_blob_server(
     store: Arc<BlobStore>,
     daemon: Option<Arc<Daemon>>,
+    use_preferred_port: bool,
 ) -> std::io::Result<u16> {
-    start_blob_server_with_listener(bind_preferred_or_random().await?, store, daemon).await
+    let listener = if use_preferred_port {
+        bind_preferred_or_random().await?
+    } else {
+        TcpListener::bind("127.0.0.1:0").await?
+    };
+    start_blob_server_with_listener(listener, store, daemon).await
 }
 
 /// Start the blob HTTP server on an already-bound listener. Exposed for tests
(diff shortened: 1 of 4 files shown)


rgbkrk:
Round 1 fixes in 934fc190.

**P1 — `sync_server::persist_settings`**: threaded `automerge_path` + `json_path` through `handle_settings_sync_connection`, daemon passes the resolved config paths. Settings sync no longer rewrites global files when the test is isolated.

**P2 — `watch_settings_json`**: switched to `self.config.resolved_settings_json_path()`. File watcher now tracks the overridden path, not the global one.

Both paths now routed through the same `DaemonConfig::resolved_*_path()` helpers as the boot-time load. `daemon_telemetry.rs`'s remaining `default_settings_doc_path()` call is intentional (telemetry is env-gated off in tests already).

- 389 lib + 29 integration tests pass
- clippy + lint clean

rgbkrk:
Round 2 fix in 8e07279e.

Threaded the settings-path overrides through `daemon_telemetry::persist_settings` too. Added `Daemon::config()` accessor so the sibling module can read resolved paths at runtime, passed them as `&Path` params to match the shape in `sync_server.rs`.

Tests don't actually exercise this path (telemetry is suppressed via `is_telemetry_suppressed()`), but it's consistent with the override contract that `DaemonConfig::settings_doc_path` advertises — dev sandboxes and any future non-default-path daemon now correctly persists to its configured location.

- 389 lib + 29 integration pass
- clippy + lint clean

PR #2117, opened 2026-04-23
fix(runtimed): emit MissingIpykernel for uv and conda envs

rgbkrk:
## Diagnosis

`KernelErrorReason::MissingIpykernel` has been pixi-only since #2102. UV/Conda envs that are missing `ipykernel` today fail at kernel spawn with a generic `ModuleNotFoundError: ipykernel_launcher` — the toolbar can classify the error but cannot render env-specific remediation. Tracked under #2096.

`prepare_environment_in` (both UV and Conda) always adds `ipykernel` to the install set, so fresh envs are fine. But the inline-env cache fast path only checks that `python` exists before returning a hit. Stale caches from pre-base-packages daemon builds, hand-edited venvs, and partial installs slip through that check and land at spawn with no typed reason.

### Rust

- New `kernel_env::venv_has_ipykernel(env_path: &Path) -> bool` helper that scans the env's site-packages for `ipykernel/` or `ipykernel-*.dist-info/`. Handles POSIX (`lib/python*/site-packages`) and Windows (`Lib/site-packages`) layouts. Used by both UV venvs and Conda envs — same shape on disk.
- `notebook_sync_server::metadata` calls the helper after env prep for `EnvSource::Inline(Uv|Conda)` and `EnvSource::Pep723(Uv)`. On miss it writes `RuntimeLifecycle::Error` + `KernelErrorReason::MissingIpykernel` the same way the pixi path does, then returns before spawning the runtime agent.
- Skipped paths (all documented in the gate's comment):
  - Prewarmed pool envs (daemon-managed; base packages are invariant)
  - `pixi:inline` / `pixi:pep723` (no pooled_env; pixi exec handles it)
  - `uv:pyproject` (launched via `uv run --with ipykernel`)
  - `conda:env_yml` (daemon appends ipykernel to the dep set pre-sync)
  - `deno`
  - `pixi:toml` (already gated above)

### Frontend

- `NotebookToolbar.tsx` now branches the MissingIpykernel banner on env source:
  - `pixi:*` → `pixi add ipykernel`
  - `uv:inline` / `uv:pep723` → `uv pip install ipykernel` (and a nudge to add it to notebook deps)
  - `conda:inline` → `conda install ipykernel`
  - Everything else → render nothing (self-healing paths never reach this state)
- Extracted `MissingIpykernelBanner` so the three copies share one chrome.

## Files changed

- `crates/kernel-env/src/lib.rs` — new `venv_has_ipykernel` helper + 7 unit tests
- `crates/runtimed/src/notebook_sync_server/metadata.rs` — post-prep gate, 46 lines including comment
- `apps/notebook/src/components/NotebookToolbar.tsx` — branched remediation renderer
- `apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx` — 5 new test cases

## Tests

- `cargo test -p kernel-env --lib` — 57 passed (7 new for `venv_has_ipykernel`)
- `cargo test -p runtimed --lib` — 389 passed
- `cargo clippy -p kernel-env -p runtimed --all-targets` — clean
- `pnpm exec vp test apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx` — 34 passed (5 new)
- `cargo xtask lint` — clean (one pre-existing `Privacy.tsx` warning unrelated to this change)

## Manual test plan

Reproduce the missing-ipykernel state on an inline UV env:

1. Create a notebook with `metadata.runt.uv.dependencies = ["pandas"]` (or use the MCP `create_notebook` + `add_dependency` flow).
2. Wait for env prep to complete, then manually `rm -rf ~/.cache/runt-nightly/inline-envs/<hash>/lib/python*/site-packages/ipykernel*` on the cached env.
3. Restart the kernel.
4. Expect: toolbar shows `ipykernel missing from prepared uv environment` with `uv pip install ipykernel` remediation. No generic stderr dump in the error message.
5. Swap the notebook to `metadata.runt.conda` deps and repeat — expect `conda install ipykernel` copy.

The inline fast path is the realistic failure mode. The alternative (corrupted install on first prep) is rare enough that the unit test coverage plus the fast-path check is sufficient.

Refs #2096. Builds on #2102 which fixed the pixi RPC path.

### DIFF
--- a/apps/notebook/src/components/NotebookToolbar.tsx
+++ b/apps/notebook/src/components/NotebookToolbar.tsx
@@ -8,7 +8,7 @@ import {
   RotateCcw,
   Square,
 } from "lucide-react";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useState, type ReactElement, type ReactNode } from "react";
 import { HoverCard, HoverCardContent, HoverCardTrigger } from "@/components/ui/hover-card";
 import { cn } from "@/lib/utils";
 import type { EnvProgressState } from "../hooks/useEnvProgress";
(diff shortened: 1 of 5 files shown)


rgbkrk:
**codex review: one P2 — fixed.**

Ran `codex review --base main`. Codex flagged a real issue:

> [P2] Mirror the new MissingIpykernel check on LaunchKernel. The new typed MissingIpykernel error is only published from auto_launch_kernel(). Common user flows such as the toolbar restart button, restart-and-run-all, and post-trust approval all go through the LaunchKernel RPC, and requests/launch_kernel.rs still only emits this reason for pixi:toml. In those uv/conda inline or uv:pep723 cases, the first auto-launch can show the new remediation banner, but any manual start or retry falls back to the old generic kernel failure.

Fixed in 2a8743d0 — mirrored the gate verbatim into `crates/runtimed/src/requests/launch_kernel.rs` between pool prep and the Launching transition. Same env sources (Inline(Uv), Inline(Conda), Pep723(Uv)); same typed reason; skips match.

Verified: `cargo test -p runtimed --lib` → 389 passing; `cargo xtask lint` → clean.

rgbkrk:
**codex review v2 → two P2s — both fixed.**

1. **Site-packages resolved by readdir order** (`kernel_env::venv_has_ipykernel`). Old code walked \`lib/python*\` and returned the first match — false-negative on envs with multiple Python-version dirs. Now asks the interpreter: \`{python} -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])'\`. Signature takes \`python_path\` instead of \`env_path\`; call sites updated to \`env.python_path\`.

2. **Pool env leak on gate fire after take.** After \`try_{uv,conda}_pool_for_inline_deps().take()\`, if the gate fired we returned without returning the env to the pool or deleting it — permanent pool shrinkage + disk leak. Both sites (auto-launch \`metadata.rs\`, RPC \`launch_kernel.rs\`) now \`spawn_blocking(remove_dir_all)\` the corrupt env before returning the error. Pool warmer refills on next tick.

Shipped in 9b192930. \`cargo test -p kernel-env --lib\` → 57 passing; \`cargo test -p runtimed --lib\` → 389 passing; \`cargo xtask lint\` clean.

rgbkrk:
**codex v3 → two more P2s — both fixed.**

1. **Retry race on corrupt env deletion.** Previous fix ran \`remove_dir_all\` on a background thread and returned before it finished. \`prepare_*_inline_env()\` treats the cache-hash dir as a hit while the path exists, so a fast retry (Restart button, automated flow) reacquired the same broken env and looped on \`MissingIpykernel\`. Switched both sites to \`tokio::fs::remove_dir_all(&env.venv_path).await\` — the cache dir is gone before the error returns.

2. **Wrong remediation for \`uv:pep723\`.** PEP 723 deps live in the inline \`# /// script\` block in cell source, not in notebook dependency metadata — editing notebook deps is ignored on the next launch. Split \`uv:pep723\` out of the \`uv:inline\` branch. New copy: "ipykernel missing from PEP 723 script metadata — add ipykernel to the inline \`# /// script\` dependency block in your notebook." Test updated to assert the PEP 723 message shows AND that the generic "notebook's dependencies" text does not.

Shipped in 5689a4a0. \`cargo test -p runtimed --lib\` → 389 passing; 34 toolbar tests passing; \`cargo xtask lint\` clean.

Issue #2089, opened 2026-04-23
fix(runtimed): bare restart_kernel may lose conda inline deps on pool env reuse


quillaid:
## Summary

After `add_dependency(scipy, after=restart)` on a conda-inline notebook, a second bare `restart_kernel()` can lose previously-added inline deps. Reported by gremlin at T37/T40: env UUID changed and pandas/matplotlib disappeared.

## Analysis

Traced the full restart flow through:
- `crates/runt-mcp/src/tools/kernel.rs` (MCP restart_kernel)
- `crates/runtimed/src/requests/launch_kernel.rs` (daemon LaunchKernel handler)
- `crates/runtimed/src/notebook_sync_server/metadata.rs` (env resolution, pool reuse)
- `crates/runtimed/src/inline_env.rs` (inline cache, pool dep comparison)

### Environment flow on second restart

1. MCP `restart_kernel()` sends `ShutdownKernel` then `LaunchKernel(auto:conda)`
2. Auto-resolution correctly reads CRDT metadata → finds inline conda deps → resolves to `conda:inline`
3. Inline cache **misses** because the first launch used pool reuse (additive delta install into `runtimed-conda-XXXX` directory), not the hash-addressed inline cache
4. `try_conda_pool_for_inline_deps` takes a **fresh** pool env and attempts to install deps

### Where deps can silently disappear

**Hypothesis A (most likely):** If any dep has version specifiers (e.g. `pandas>=2.0`), `compare_deps_to_pool` returns `Independent` → pool reuse skipped → falls through to `prepare_conda_inline_env` (full conda solve). If this solve is slow or fails, the kernel launches without deps or with partial deps.

**Hypothesis B:** The pool env's `sync_dependencies` does a full rattler solve. If the solve succeeds but the actual package installation fails partially (e.g., network timeout on one package), some deps may be present and others missing.

**Hypothesis C:** Race condition — `resolve_metadata_snapshot` reads the CRDT before `confirm_sync` from the MCP client has propagated the latest `add_dependency` writes. The 500ms sleep in `restart_kernel` is usually sufficient, but under load it might not be.

### Why env UUID changes

The first launch claimed a pool env and kept it in the pool-named directory. On restart, a different pool env is taken → different directory → different UUID in the gremlin's observation.

## Key files

| File | Relevance |
|------|-----------|
| `crates/runt-mcp/src/tools/kernel.rs` | MCP restart_kernel flow |
[...]

Issue #2083, opened 2026-04-23
add_dependency(after='restart') on inline conda envs is slow/stalls during rattler install


quillaid:
## Summary

`add_dependency("scipy", after="restart")` on an inline conda env with 3 existing packages (`["pandas","numpy","matplotlib"]`) causes a long stall during the relaunch — new env shows 83/181 conda-meta entries after 3+ minutes without completing.

## Reproduction

1. `create_notebook(dependencies=["pandas","numpy","matplotlib"], package_manager="conda")`
2. Wait for kernel to become idle
3. `add_dependency("scipy", after="restart")`
4. Observe: kernel shuts down, new launch begins, but env rebuild takes 3+ minutes

## Root Cause Analysis

The `after="restart"` path in `add_dependency`:
1. Adds `scipy` to CRDT metadata (now 4 deps: pandas, numpy, matplotlib, scipy)
2. Sends `ShutdownKernel` (runtime agent stays alive)
3. Sleeps 300ms
4. Sends `LaunchKernel(env_source="auto:conda")`

The daemon resolves to `conda:inline` with the 4 deps. Since the runtime agent is alive, it hits the **restart path** (line 1111 in `launch_kernel.rs`). But before sending `RestartKernel`, the daemon re-resolves the env:

1. Inline cache check — **miss** (new hash, 4 deps vs original 3)
2. Pool reuse via `try_conda_pool_for_inline_deps` — takes a fresh pool env
3. Installs all 4 user deps as delta into the new pool env via `sync_dependencies`

Step 3 triggers a full rattler solve + download + install of `pandas+numpy+matplotlib+scipy` plus all transitive deps into a pool env that only has base packages (ipykernel, ipywidgets, etc). This is a heavy operation:
- Rattler solve with 8+ specs (4 user + 4 base)
- Download and install ~180 packages
- The 83/181 suggests it's progressing but slowly

## Likely Causes of Incomplete Install

1. **Disk I/O pressure**: 180 conda packages extracting concurrently
2. **Memory pressure**: Large solver state + parallel extraction
3. **Network**: Repodata fetch or package download timeouts
4. **No timeout on sync_dependencies**: The call at `try_conda_pool_for_inline_deps` has no timeout, so it can run indefinitely

## Suggested Improvements

1. **Reuse existing env when only adding deps**: Instead of taking a new pool env, the restart could sync the new dep into the EXISTING env (the one from the original launch). This avoids re-solving all 180 packages from scratch.
[...]

PR #2128, opened 2026-04-24
fix(runtimed): thread bootstrap_dx through try_uv_pool_for_inline_deps

rgbkrk:
Follow-up to #2127. Codex v1 review flagged a real P2 that got posted after merge.

## Issue

The new inline-cache claim for pool-reused UV envs in #2127 hard-coded `bootstrap_dx = false`. Both `check_uv_inline_cache` and `prepare_uv_inline_env` key the cache on that flag — `bootstrap_dx=true` adds `dx` to the install set and shifts the hash.

That means:
- `bootstrap_dx=true` pool-reuse launch renames the env under the non-bootstrap hash → the next restart's cache lookup (with the flag on) misses anyway, defeating #2127's optimization for that flow.
- Worse: if the flag flipped between launches, a subsequent check could reuse a `dx`-bootstrapped env under the plain inline-cache key.

## Fix

Added `bootstrap_dx: bool` parameter to `try_uv_pool_for_inline_deps` and updated the two call sites (`requests/launch_kernel.rs` and `notebook_sync_server/metadata.rs`) to pass their already-resolved `bootstrap_dx`. Used at both `claim_pool_env_for_uv_inline_cache` sites (Subset and post-sync Additive) so the rename target matches the hash `check_uv_inline_cache` will look up next time.

Conda isn't affected — no equivalent flag.

## Test plan

- [x] `cargo test -p runtimed --lib` → 390 passing
- [x] `cargo xtask lint` → clean

cc #2089 #2083 #2127

### DIFF
--- a/crates/runtimed/src/inline_env.rs
+++ b/crates/runtimed/src/inline_env.rs
@@ -65,7 +65,7 @@ fn get_inline_cache_dir() -> std::path::PathBuf {
 /// has to ride along with the inline env. Folding `dx` into the dep list
 /// also changes the env hash so bootstrap and non-bootstrap envs don't
 /// collide in the cache.
-fn inline_deps_with_bootstrap(deps: &[String], bootstrap_dx: bool) -> Vec<String> {
+pub(crate) fn inline_deps_with_bootstrap(deps: &[String], bootstrap_dx: bool) -> Vec<String> {
     if bootstrap_dx {
         let mut out = deps.to_vec();
         out.push("dx".to_string());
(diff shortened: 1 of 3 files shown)


rgbkrk:
Codex v1 review flagged a real P2: threading the flag into the claim site (26bc0492) matched the bootstrap cache *key* but didn't actually install \`dx\` — \`try_uv_pool_for_inline_deps\` was still comparing the plain \`deps\` (not \`inline_deps_with_bootstrap(deps, bootstrap_dx)\`) against the pool's prewarmed packages, so the delta never included \`dx\`. Pool env got pinned into the bootstrap cache slot without \`dx\`, \`import dx\` silently failed, and downstream features (pandas/polars auto-attach) no-op'd.

Fixed in a31a70bc — compute \`effective_deps\` once using the bootstrap-augmented set, use it for both \`compare_deps_to_pool\` calls (pre-check and actual). When \`bootstrap_dx\` is on and the pool env doesn't have \`dx\`, the delta now includes it and \`sync_dependencies\` installs it before the rename.

Likely explains intermittent reports of dx auto-attach not firing on certain notebook restarts.

\`cargo test -p runtimed --lib\` → 390 passing; lint clean.

rgbkrk:
**Codex v2: clean.** No regressions flagged. Consistently threads `bootstrap_dx` through pool-compat check, delta install, and inline-cache hash.

PR #2129, opened 2026-04-24
fix(runtimed): prevent duplicate dep promotion in environment.yml

### LINKED ISSUE #2076: fix(runtimed): duplicate dep promotion in environment.yml (pip/conda boundary confusion)
## Summary

When `add_dependency` promotes inline CRDT deps to an `environment.yml` project file, deps can be duplicated on each restart cycle. Two distinct causes:

### 1. Pip-only deps leak into conda dep list

The promotion code at `metadata.rs:3164` computes `launched_names` from the `environment_yml_deps` snapshot taken at kernel launch time. This snapshot comes from `parse_environment_yml()` which uses rattler's `EnvironmentYaml::match_specs()` — this correctly returns only **conda** deps, skipping pip deps.

However, when `promote_inline_deps_to_project` re-bootstraps the CRDT from environment.yml (line 3217), it writes the parsed conda deps into `snapshot.runt.conda.dependencies`. If a package like `pyyaml` exists only under the `pip:` block, it won't be in `launched_names`. On the next restart, if the gremlin adds any dep, `pyyaml` appears in the CRDT conda deps (from re-bootstrap) but is not in `launched_names`, so it gets promoted (appended) to environment.yml as a conda dep — creating a duplicate.

### 2. MatchSpec normalization mismatch

`python=3.12` in the environment.yml gets normalized by rattler to a MatchSpec string form. When the CRDT stores the normalized form and the next promotion cycle compares it against the original `launched_names` (which contains the rattler-normalized form from `match_specs()`), the string comparison can fail if there's any normalization difference (e.g., `python 3.12.*` vs `python=3.12`), causing `python=3.12` to be promoted again as a duplicate.

### 3. `find_env_yml_deps_insertion_point()` doesn't respect pip sub-blocks

The insertion point function at `metadata.rs:3014` tracks `- ` entries inside the `dependencies:` block but doesn't distinguish pip sub-entries from conda entries. This means insertions may interact incorrectly with the pip block structure.

## Reproduction

Run the `conda-conflict` gremlin against a build with this environment.yaml:

```yaml
name: nightly-conda-stress
channels:
  - conda-forge
dependencies:
  - python=3.12
  - pandas>=2.0
  - numpy>=1.26
  - matplotlib>=3.8
  - requests
  - pip:
    - pyyaml
  - jinja2
```

After the gremlin adds deps and triggers restarts, the file ends up with:

```yaml
dependencies:
  - python=3.12
  - pandas>=2.0
  - numpy>=1.26
  - matplotlib>=3.8
  - requests
  - pip:
    - pyyaml
  - jinja2
  - python=3.12      # DUPLICATE
  - pyyaml           # DUPLICATE (was pip-only)
  - scipy
  - seaborn
  ...
```

## Suggested Fix

1. When computing `launched_names`, also parse the raw environment.yml text to include pip dep names (not just rattler `match_specs()` output)
2. Before appending a dep to environment.yml, check if it already exists in the file (text-level dedup, case-insensitive package name match)
3. Fix `find_env_yml_deps_insertion_point()` to skip pip sub-block entries and insert after the last top-level conda dep

## Impact

- Low severity — rattler's solver handles duplicates gracefully (idempotent resolution)
- environment.yml gets polluted with duplicate entries on every restart cycle
- Could eventually confuse users or cause parse issues with non-rattler YAML parsers

quillaid:
## Summary

Fixes #2076 — three sub-bugs in `promote_inline_deps_to_project` caused `environment.yml` to accumulate duplicate entries on restart cycles:

- **Pip deps leaked into conda list**: `parse_environment_yml` uses rattler's `match_specs()` which skips the `pip:` sub-block, so packages already under `pip:` weren't in `launched_names` and got re-added as top-level conda deps
- **MatchSpec normalization mismatch**: rattler normalizes specs (e.g. `python=3.12` → `python 3.12.*`), risking text-level duplicates across promotion cycles
- **Insertion point inside pip block**: `find_env_yml_deps_insertion_point` tracked any `- ` line regardless of indent, so new deps could land inside the `pip:` sub-block

## Changes

- `find_env_yml_deps_insertion_point`: track baseline indent level and skip deeper-indented lines + YAML sub-section markers (`- pip:`)
- New `extract_all_env_yml_package_names` helper: uses rattler's `match_specs()` + `pip_specs()` to extract all package names (conda + pip) from the file
- File-level dedup in `promote_inline_deps_to_project`: filter `to_add` against all existing names before writing — unified safety net for all three sub-bugs
- 4 regression tests

## Test plan

- [x] `cargo test -p runtimed --lib insertion_point` — 5 tests pass (3 existing + 2 new)
- [x] `cargo test -p runtimed --lib extract_all_env_yml` — 2 new tests pass
- [x] `cargo test -p runtimed --lib` — all 394 tests pass

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -3229,6 +3229,7 @@ pub(crate) async fn handle_notebook_request(
 pub(crate) fn find_env_yml_deps_insertion_point(content: &str) -> Option<usize> {
     let mut in_deps = false;
     let mut last_dep_end = None;
+    let mut baseline_indent: Option<usize> = None;
 
     for (i, line) in content.lines().enumerate() {
         let trimmed = line.trim();
@@ -3247,12 +3248,57 @@ pub(crate) fn find_env_yml_deps_insertion_point(content: &str) -> Option<usize> 
         }
 
         if in_deps && trimmed.starts_with("- ") {
+            let indent = line.len() - line.trim_start().len();
+            match baseline_indent {
+                None => baseline_indent = Some(indent),
+                Some(base) if indent != base => continue,
+                _ => {}
+            }
+            // Skip YAML sub-section entries like "- pip:" — they introduce
+            // nested blocks, not package specs.
+            let entry = trimmed.trim_start_matches("- ");
+            if entry.ends_with(':') && !entry.contains(' ') {
+                continue;
+            }
             let offset: usize = content.lines().take(i + 1).map(|l| l.len() + 1).sum();
             last_dep_end = Some(offset.min(content.len()));
         }
     }
 
     last_dep_end
+}
+
+/// Extract all package names from an environment.yml string (both conda and pip
+/// sections). Returns lowercased names for dedup checks before writing to the file.
+pub(crate) fn extract_all_env_yml_package_names(
+    content: &str,
+) -> std::collections::HashSet<String> {
+    use rattler_conda_types::EnvironmentYaml;
+    let mut names = std::collections::HashSet::new();
+
+    let Ok(env) = EnvironmentYaml::from_yaml_str(content) else {
+        return names;
+    };
+
+    for spec in env.match_specs() {
+        if let rattler_conda_types::PackageNameMatcher::Exact(name) = &spec.name {
+            let n = name.as_normalized().to_string().to_lowercase();
+            if !n.is_empty() {
+                names.insert(n);
+            }
+        }
+    }
+
+    if let Some(pip_specs) = env.pip_specs() {
+        for spec in pip_specs {
+            let name = notebook_doc::metadata::extract_package_name(spec);
+            if !name.is_empty() {
+                names.insert(name);
+            }
+        }
+    }
+
+    names
 }
 
 /// For removals, compares CRDT deps against launched baseline and runs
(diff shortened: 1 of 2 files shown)


rgbkrk (commented):
Reviewed locally. Tests pass, lint clean, logic traces correctly against both new test fixtures. **Not blocking merge**, but leaving notes below so you can address async while I'm offline.

## The direction I'd prefer

Lean harder on rattler internals. You already use `EnvironmentYaml::from_yaml_str` for `extract_all_env_yml_package_names`; the `find_env_yml_deps_insertion_point` function is the last piece of hand-rolled YAML parsing in this path, and the `- pip:` / indent-tracking heuristic is exactly the kind of thing rattler already models correctly via `MatchSpecOrSubSection`.

Concretely — would it be cleaner to:
1. Parse `EnvironmentYaml` from the file
2. Mutate its `dependencies` vec (push `MatchSpecOrSubSection::MatchSpec(...)` for each new dep, before any sub-section entries)
3. Serialize back via serde_yaml

That removes the line/offset arithmetic, the sub-section heuristic, and the baseline-indent tracking in one sweep. The tradeoff is we'd lose comment/whitespace preservation for entries rattler doesn't round-trip perfectly — worth checking whether `EnvironmentYaml` serializes cleanly against a few real fixtures before committing to this path.

If that's too invasive right now, the current hand-rolled approach is fine and this can be a separate simplification PR. But I'd rather close the book on manual YAML parsing here than keep inching it forward.

## Nits on the current approach

1. **Sub-section heuristic is minimal.** `entry.ends_with(':') && !entry.contains(' ')` catches `- pip:` but would also skip any hypothetical dep name ending in `:` (none exist — `:` isn't valid in conda package names). A one-liner comment explaining "this is a YAML heuristic; `:` isn't valid in conda names so false positives are impossible" would make it self-justifying.

2. **`extract_all_env_yml_package_names` handles `Exact` only.** `Glob`/`Regex` matchers would silently fall out of the dedup set. Exotic in env.yml but worth a doc comment calling it out.

3. **Malformed YAML → empty set → no dedup.** When rattler can't parse the file, every candidate passes the filter and dupes can still land. Fine as-is (malformed file is already broken), but if you prefer fail-closed, bail from `promote_inline_deps_to_project` entirely when the file doesn't parse.

4. **Comment drift.** The pre-existing "Find the end of the dependencies: section to insert new deps" comment above `find_env_yml_deps_insertion_point` predates this fix — still accurate, but the new file-level dedup deserves a sentence above the `existing_names = extract_all_env_yml_package_names(&content)` line explaining it's the unified safety net for the three sub-bugs in #2076.

## Missing test

End-to-end coverage of the dedup in `promote_inline_deps_to_project` would lock in the fix for all three sub-bugs in one shot. Something like:

1. Start with an `environment.yml` that has `- pyyaml` under a `- pip:` block.
2. Call `promote_inline_deps_to_project` with `pyyaml` in the to-add set.
3. Assert the file is unchanged and `pyyaml` was filtered out.

~30 lines. Unit tests for the helpers are good but the integration is where the three bugs actually intersect.

---

Codex review is running on this now too; I'll post its findings in a separate comment if they add anything new.

rgbkrk:
Codex review landed. One real P2 I'd missed — worth prioritizing over my nits:

## [P2] Dedup conflates pip and conda namespaces

> When the current \`environment.yml\` already lists a package under the \`pip:\` subsection, this new dedup pass treats that as if the conda dependency were already present and filters it out before any file update. \`SyncEnvironment\`/\`LaunchKernel\` will leave the project file pip-only even if the user explicitly added a conda MatchSpec in notebook metadata, so the requested package source never gets promoted to the conda dependency list.

Concrete case: \`pyyaml\` in the \`pip:\` block; user adds \`pyyaml\` to \`runt.conda.dependencies\` in the notebook. Current PR silently drops the promotion, leaving the file pip-only.

Fix direction: split the dedup by namespace. \`extract_all_env_yml_package_names\` probably wants to return either \`{ conda: HashSet, pip: HashSet }\` or two separate helpers, and the filter in \`promote_inline_deps_to_project\` should compare candidate conda deps only against existing conda names (and conversely for pip if we ever promote pip deps here, which I don't think we do today).

Add a regression test: env.yml with \`pyyaml\` under \`pip:\`, promote \`pyyaml\` as conda, assert it lands as a top-level conda dep and the pip block is untouched.

This supersedes my earlier "end-to-end dedup test" suggestion — this is the specific correctness case to cover.

rgbkrk:
Going ahead with merging. If you have follow ups, feel free to make new PR(s).

PR #2135, opened 2026-04-24
feat(launcher): vendor nteract-kernel-launcher with display bootstrap in-binary

rgbkrk:
## Summary

Replaces the three-part display bootstrap with one cohesive, daemon-version-locked system:

- **Before:** `dx` PyPI package + single-file `nteract_kernel_launcher.py` + pool warmer adding `"dx"` to install sets. Display-layer behavior was coupled to PyPI version resolution. Users on nightly got stable `dx==2.0.0` and DataFrames silently rendered as `text/llm+plain` summaries instead of rich Sift tables.
- **After:** `nteract_kernel_launcher` ships as a package inside the daemon binary via `include_str!`, vendored into every env's site-packages. Whatever display contract the daemon speaks this commit, the kernel speaks the same.

Spec: `docs/superpowers/specs/2026-04-24-launcher-hosted-display-bootstrap.md`

## The subclass cascade

`IPKernelApp → IPythonKernel → ZMQInteractiveShell → ZMQShellDisplayHook` traitlets `Type` cascade lands `NteractShellDisplayHook` as the active displayhook. The override adds a thread-local hook chain to `finish_displayhook` — the same shape `ZMQDisplayPublisher._hooks` already has. Net effect: **bare-`df`-on-last-line produces a real `execute_result` with parquet buffers** instead of masquerading as `display_data` via the `ipython_display_formatter` short-circuit dx needed.

## The bootstrap extension

`nteract_kernel_launcher._bootstrap` auto-loads via `NteractKernelApp.default_extensions`, so it runs before any user code — log-warn on failure, no traceback to the user.

- Registers mimebundle formatters lazily via `for_type_by_name` for pandas, polars, narwhals, datasets. No eager imports of pyarrow/pandas at boot.
- Registers the shared `buffer_hook` on **both** `ip.display_pub` and `ip.displayhook` — one hook, agnostic to which message type is being built.
- Flips altair / plotly to their `"nteract"` renderer when importable.

## dx removal from install sets

- `UV_BASE_PACKAGES` drops `\"dx\"`.
- `uv_prewarmed_packages` no longer branches on `bootstrap_dx` to add `\"dx\"`.
- `inline_deps_with_bootstrap` becomes a shim that returns `deps` unchanged — kept so call sites don't all shift on the same PR.
- `vendor_into_venv` now writes a multi-file package directory (with per-file write-and-rename) and removes any pre-0.2.0 single-file `nteract_kernel_launcher.py` it finds in the target site-packages.

## Verified end-to-end

Live kernel, dev daemon, workspace venv (which has `dx` importable — and kernel introspection confirms `\"dx\" not in sys.modules` while our formatters are the ones bound to `pd.DataFrame`). Scenarios: pandas bare + \`display(df)\`, polars bare, narwhals wrapping pandas, \`DisplayHandle.update()\` round-trip, 500k-row frame, renderer flip, hook-seat snapshot.

Polars registration caught a real bug in review: `for_type_by_name` keys on `type.__module__`, and polars doesn't hoist `DataFrame` to the top-level module (stays at `polars.dataframe.frame`). Fixed with definitive-path + short-name spray across pandas/polars/narwhals.

- Tracebacks: reparse `.ipynb` error outputs into an approximation of the rich in-session form. Lossy on purpose.
- \`update_display_data\` ref-MIME promotion: daemon keeps the raw \`application/vnd.nteract.blob-ref+json\` entry on update instead of promoting to the resolved blob URL the way the initial \`display_data\` path does.
- Late-install renderer flip: user installs altair mid-session → our \`post_execute\` probe notices and flips before their next chart.
- Lazy-loading story: explore a \`meta_path\` or per-import pattern cleaner than \`post_execute\` polling.

## Test plan

- [ ] \`cargo xtask lint\` passes
- [ ] \`cargo test -p kernel-env\` passes (includes multi-file vendor + legacy-cleanup + concurrent-write tests)
- [ ] \`cargo test -p runtimed\` passes (inline_env, metadata helpers)
- [ ] \`cargo test -p kernel-env --test launcher_e2e\` passes (creates a uv venv, vendors, imports)
- [ ] Python launcher tests pass: \`pytest python/nteract-kernel-launcher/tests/\`
- [ ] Manual: fresh kernel, pandas bare → \`execute_result\` + parquet + text/llm+plain
- [ ] Manual: polars bare → all four MIMEs including parquet (this is the one that caught the \`__module__\` bug)
- [ ] Manual: \`display(df)\` → \`display_data\` + parquet
- [ ] Manual: \`DisplayHandle.update()\` → update delivers new buffers
- [ ] Manual: introspection shows \`dx\` not loaded, formatters bound to \`nteract_kernel_launcher._bootstrap\`


rgbkrk:
Addressed both Codex findings in `dd5a2f80`:

**P2 — inline cache hit re-vendor**

`check_uv_inline_cache` now awaits `vendor_into_venv` on hit when `bootstrap_dx` is on, before returning the `PreparedEnv`. `vendor_into_venv` is idempotent and already cleans up the legacy single-file module, so a pre-upgrade non-bootstrap cache or a pre-0.2.0 bootstrap cache is brought up to today's layout before the kernel boots. Both call sites (`launch_kernel.rs`, `metadata.rs`) await the new async function.

Invariant now local: if `check_uv_inline_cache` returns `Some(env)` for a bootstrap launch, the launcher package is guaranteed present.

**P3 — launcher_cache_dir atomic + single-flighted**

- Swapped `OnceLock<PathBuf>` for `tokio::sync::OnceCell` + `get_or_try_init`, single-flighting the first materialization so two concurrent bootstrap launches don't both enter the write path.
- Per-file writes now go through `kernel_env::launcher::write_package_files` (exposed `pub`), which already uses temp+rename per entry — same atomic pattern `vendor_into_venv` relies on. A concurrent reader importing from the cache dir can't observe a half-written module file.

Tests: `cargo test -p runtimed --lib` (395 pass) and `cargo test -p kernel-env` (59 pass) green. `cargo xtask lint` clean.

PR #2142, opened 2026-04-24
feat(runtimed): UserErrorOutput semantic type (foundation for traceback round-trip)

rgbkrk:
## Summary

Foundation commit for the follow-up to PR #2141. Introduces `UserErrorOutput` — one enum that models "user code raised" regardless of the IOPub wire shape:

- `Classic { ename, evalue, traceback[] }` — vanilla `ipykernel_launcher` path
- `Rich(RichTraceback)` — nteract-kernel-launcher path emitting `application/vnd.nteract.traceback+json`

The daemon classifies at the edges (`from_nbformat`, and in a follow-up `from_iopub`), then downstream code (runtime-state bookkeeping, manifest creation, save, load) reasons about errors through this single type. No more ad-hoc "is this MIME special?" sniffing across dispatchers.

## What's in this PR

- `crates/runtimed/src/user_error.rs` — the enum, `from_nbformat`, `to_classic`, `to_rich`, plus a best-effort ANSI traceback parser.
- `crates/runtimed/src/lib.rs` — module registration.

## What's NOT yet wired (follow-ups in this branch)

- **Codex P1 fix** (PR #2141 review): runtime agent must flip `execution_had_error` when a rich-MIME display_data arrives. Planned: classify via `UserErrorOutput::from_iopub` in the IOPub dispatcher, route ALL variants through `mark_execution_error` + `QueueCommand::CellError`. Keeps the classic path unchanged; unifies the rich path.
- **Manifest creation** (`output_store.rs`): when a display_data arrives carrying `TRACEBACK_MIME`, convert to `OutputManifest::Error` (classic fields via `to_classic`) and attach the rich payload as a `ContentRef` sibling. On-disk stays nbformat-clean; in-memory carries the rich form for the frontend.
- **Save path** (`persist.rs::resolve_manifest`): Error manifest → `{"output_type": "error", "ename", "evalue", "traceback"}`. Rich sibling is in-memory only, stripped at nbformat emission.
- **Load path** (`load.rs`): when an error output arrives from `.ipynb`, call `UserErrorOutput::from_nbformat` → `to_rich()` (which runs the ANSI parser) → attach rich sibling on the Error manifest.
- **Frontend**: when an error manifest has a rich sibling, route to `TracebackOutput`; otherwise fall back to `AnsiErrorOutput`.
- **Launcher review**: the current short-circuit publishes rich-only display_data. With the runtime agent classifying via `from_iopub`, this stays correct; no launcher change required. If we later decide the launcher should emit classic in addition, the enum already handles both.

## Classic path must keep working with `bootstrap_dx` OFF

When a user runs with vanilla `ipykernel_launcher`, the kernel emits classic `ErrorOutput`. My design preserves that path end-to-end: the daemon sees Classic, emits `OutputManifest::Error` with no rich sibling, frontend renders via `AnsiErrorOutput`. No feature-flag branches in the daemon.

## Tests

`cargo test -p runtimed --lib user_error::` — 13 passing:

- nbformat classification: classic error, rich (JSON object), rich (JSON string), negative cases
- `to_classic()` projection: Classic identity, Rich → split text on newlines
- ANSI parser: CPython `File "..."`, Jupyter `Cell In[N]`, pre-7 `<ipython-input-N>`, ANSI strip, library-path detection, empty input, text-tail canonicalization, Classic→Rich via parser

## Test plan

- [ ] `cargo test -p runtimed --lib user_error::` — 13 passing
- [ ] `cargo xtask lint` clean
- [ ] Design review via `codex review` (requested by author)

## Follow-up scope

This branch will grow to wire each of the sites above. Each commit lands one hook site at a time so the diff stays reviewable. When everything is wired, the PR covers: Codex P1, nbformat save compliance, rich render for loaded-from-disk `.ipynb`, and a single central type for all error ingress.

### DIFF
--- a/crates/runtimed/src/lib.rs
+++ b/crates/runtimed/src/lib.rs
@@ -47,6 +47,7 @@ pub mod stream_terminal;
 pub mod sync_server;
 pub mod task_supervisor;
 pub mod terminal_size;
+pub mod user_error;
 
 /// Get the daemon version string (e.g., "0.1.0-dev.10+abc123").
 /// Used for protocol version checking and debugging.
--- a/crates/runtimed/src/user_error.rs
+++ b/crates/runtimed/src/user_error.rs
@@ -1,0 +1,697 @@
+//! `UserErrorOutput` — the canonical "user code raised" event.
+//!
+//! IOPub can represent a user error two ways:
+//!
+//! 1. **Classic.** `JupyterMessageContent::ErrorOutput` carrying
+//!    `{ename, evalue, traceback[]}` where the traceback entries are
+//!    ANSI-coded strings. What vanilla IPython emits.
+//!
+//! 2. **Rich.** `JupyterMessageContent::DisplayData` with our custom
+//!    MIME `application/vnd.nteract.traceback+json` and a structured
+//!    payload (ename/evalue/frames with source-context windows and
+//!    highlight markers). What `nteract-kernel-launcher` emits after
+//!    hooking `_showtraceback`.
+//!
+//! Downstream code shouldn't care which wire shape arrived. The daemon
+//! needs to:
+//!
+//! - mark the execution as errored (runtime state, queue, success flag),
+//! - persist a nbformat-compliant `output_type: "error"` on `.ipynb`,
+//! - give the frontend the richest renderer it can.
+//!
+//! `UserErrorOutput` is the single semantic type every site routes
+//! through. Classifying at the IOPub / nbformat edges means the rest
+//! of the system stays flat — no ad-hoc "is this MIME special?"
+//! sniffing sprinkled across dispatchers.
+
+use serde::{Deserialize, Serialize};
+
+/// The MIME used by the launcher's rich traceback payload. Shared with
+/// the frontend renderer. Single source of truth.
+pub const TRACEBACK_MIME: &str = "application/vnd.nteract.traceback+json";
+
+/// One source-context line within a rich frame.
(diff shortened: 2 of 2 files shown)


rgbkrk:
Addressed the P3 coverage gap + testing-gaps list from the review.

**Pushed `e58ff6c9`** — 7 new tests, now 20/20 passing:

- `chained_raise_from_parses_flat_with_separator_in_text` — `raise X() from Y()`
- `during_handling_separator_preserved_in_text` — implicit exception chaining
- `exception_group_falls_back_to_none_today` — documents the ExceptionGroup gap explicitly; returns `None` on `|`-prefixed nested frames so the caller falls back to the classic render (per your P3 suggestion)
- `multiline_evalue_roundtrip_via_rich_to_classic` — embedded newlines in the exception message (pytest-style AssertionError)
- `to_classic_drops_final_trailing_newline_only` — confirms `"line1\nline2\n"` yields exactly `["line1", "line2"]` (no empty trailing element)
- `from_nbformat_accepts_execute_result_rich_candidate` — some emitters use `execute_result` for last-expression errors
- `pep657_caret_rails_dropped_from_structured_lines_kept_in_text` — IPython format, carets visible in `rt.text`, absent from structured `frame.lines`

On the design note ("make the rich sibling an explicit optional field on the error manifest shape rather than an implicit side channel") — I'm planning `rich: Option<ContentRef>` on `OutputManifest::Error`, right alongside the existing `traceback` field. Stripped at `resolve_manifest` → nbformat emission so `.ipynb` stays standards-clean.

Ready to proceed with wiring when you are — runtime agent IOPub classification is next.

PR #2148, opened 2026-04-24
fix(runtimed): auto-sign project-file deps at bootstrap

rgbkrk:
## Overview

The auto-launch path writes project-file deps (pyproject.toml / pixi.toml / environment.yml) into `metadata.runt.{uv,conda,pixi}` so UI and MCP tools can see what's installed. Those writes landed without a signature and flipped the notebook to `Untrusted`, firing the trust dialog for deps the user already had on disk in a file they own. On brand-new notebooks opened inside a project dir, this showed up as a trust banner flicker during auto-launch.

## Diagnosis

For a pyproject-backed notebook with no inline deps:

1. `.ipynb` on disk has empty `runt.uv.dependencies`. Room init's `verify_trust_from_file()` returns `NoDependencies`. No banner.
2. `auto_launch_kernel` detects `pyproject.toml`, reads `[project.dependencies]`, writes them into `runt.uv.dependencies` via `fork_and_merge`.
3. Doc sync flushes. `check_and_update_trust_state` re-verifies the fresh snapshot: deps are now non-empty, no matching signature → `Untrusted`.
4. `RuntimeStateDoc.trust.needs_approval` flips to `true`. Banner appears for deps the daemon itself just sourced from a local file.

## Fix

Add `auto_sign_in_place(snapshot)` alongside `resign_trusted_snapshot` in `notebook_sync_server/metadata.rs`. It signs the runt metadata in place and writes the signature + timestamp back into the snapshot. The three bootstrap closures call it right after writing deps, inside the same `fork_and_merge`. `resign_trusted_snapshot` refactored to use the helper so both paths share one signing implementation.

The bootstrap predicate fires on "deps differ OR snapshot doesn't verify as Trusted/NoDependencies" so a snapshot that reaches bootstrap with matching deps but a missing or invalid signature still gets signed. Defense-in-depth against external writes or stale signatures.

HMAC still protects against external editors changing the notebook's signed fields. What it no longer does is prompt for approval on deps the daemon itself wrote from a file the user owns.

## Design decision

Auto-sign at bootstrap instead of deleting the bootstrap. Deleting would lose the MCP/UI invariant that inline dep fields always reflect what's running. Introducing a parallel `metadata.runt.project_deps` surface cleanly separates HMAC-required from informational, but that's a schema change and consumer migration. Worth doing as a followup with `schema-update` label; this PR is a targeted bugfix.

## Scope limit

Bootstrap only runs when `auto_launch_kernel` runs, and auto-launch is gated on `Trusted | NoDependencies` at room creation (`peer.rs:444`). A notebook saved by the pre-fix build whose on-disk deps verify as `Untrusted` at room init never reaches bootstrap — the banner persists until the user clicks Trust once. Healing that path needs a project-file-reconciliation step upstream of the auto-launch gate. Tracked in #2150.

## Pixi coverage gap

`runt_trust::verify_notebook_trust` short-circuits on `uv_dependencies.is_empty() && conda_dependencies.is_empty()` and never looks at pixi. Pixi-only notebooks always return `NoDependencies` regardless of signature state. So the dialog wasn't firing for pixi bootstrap today - the `auto_sign_in_place` call in the pixi branch is forward-compat for when pixi joins the trust check. Worth filing as a separate issue.

## Test plan

- [x] `cargo test -p runtimed --lib` - 436 passing
- [x] `cargo clippy -p runtimed --all-targets` - clean
- [x] Three unit tests for `auto_sign_in_place` (uv, conda, pixi)
- [x] Regression test that simulates the bootstrap end state (deps + sign) and asserts `check_and_update_trust_state` lands on `Trusted` with `needs_approval=false`
- [x] Regression test for Codex P2: matching deps + missing signature path still signs
- [x] Manual E2E: dev daemon rebuilt with the fix, created a new notebook with working_dir in the repo root (pyproject.toml with 9 deps). Daemon log shows:
  ```
  Trust status for <id>: NoDependencies
  Bootstrapped pyproject.toml deps into CRDT
  Trust state changed via doc sync: NoDependencies -> Trusted
  ```
  Kernel auto-launched, cell executed cleanly in the uv:pyproject env. The pre-fix transition would have been `NoDependencies -> Untrusted` with banner/dialog.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -2281,7 +2281,16 @@ pub(crate) async fn auto_launch_kernel(
                         doc.fork_and_merge(|fork| {
                             let mut snap = fork.get_metadata_snapshot().unwrap_or_default();
                             let current_deps = snap.runt.pixi.as_ref().map(|p| &p.dependencies);
-                            if current_deps.is_none_or(|d| d != &deps) {
+                            let deps_match = current_deps.is_some_and(|d| d == &deps);
+                            // Re-sign even when deps already match if the snapshot
+                            // isn't verifiably trusted — covers notebooks bootstrapped
+                            // by the pre-fix build (matching deps, no signature).
+                            let trust_ok = matches!(
+                                verify_trust_from_snapshot(&snap).status,
+                                runt_trust::TrustStatus::Trusted
+                                    | runt_trust::TrustStatus::NoDependencies
+                            );
+                            if !deps_match || !trust_ok {
                                 let pixi = snap.pixi_section_or_default();
                                 pixi.dependencies = deps;
                                 if let Err(e) = auto_sign_in_place(&mut snap) {
@@ -2314,7 +2323,13 @@ pub(crate) async fn auto_launch_kernel(
                         doc.fork_and_merge(|fork| {
                             let mut snap = fork.get_metadata_snapshot().unwrap_or_default();
                             let current_deps = snap.runt.uv.as_ref().map(|u| &u.dependencies);
-                            if current_deps.is_none_or(|d| d != &deps) {
+                            let deps_match = current_deps.is_some_and(|d| d == &deps);
+                            let trust_ok = matches!(
+                                verify_trust_from_snapshot(&snap).status,
+                                runt_trust::TrustStatus::Trusted
+                                    | runt_trust::TrustStatus::NoDependencies
+                            );
+                            if !deps_match || !trust_ok {
                                 let uv = snap.runt.uv.get_or_insert_with(|| {
                                     notebook_doc::metadata::UvInlineMetadata {
                                         dependencies: Vec::new(),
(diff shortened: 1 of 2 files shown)


rgbkrk:
Applied in 3fd4597. The bootstrap predicate now fires on "deps differ OR snapshot doesn't verify as Trusted/NoDependencies," so a snapshot that reaches bootstrap with matching deps and a missing or invalid signature gets signed.

One scope limit worth calling out: the bootstrap only runs when `auto_launch_kernel` runs, and auto-launch is gated on `Trusted | NoDependencies` at room creation (`peer.rs:444`). A notebook saved by the pre-fix build whose on-disk deps verify as `Untrusted` at room init never reaches bootstrap - the banner persists until the user clicks Trust once. Confirmed via dev-daemon E2E against a hand-written .ipynb with matching pyproject deps and no signature: daemon logs `Trust status: Untrusted` then `Kernel blocked on trust approval`, and bootstrap is never invoked.

Healing that path needs a project-file-reconciliation step upstream of the auto-launch gate. Filed as #2150 and out of scope for this PR.

New regression test: `test_bootstrap_re_signs_matching_but_unsigned_snapshot` asserts the new predicate correctly re-signs a matching-but-unsigned snapshot. Full `cargo test -p runtimed --lib` = 436 passing, clippy clean.

Issue #2164, opened 2026-04-24
fix(runtimed): auto-sync env.yml deps into pool env on conda:env_yml fallback


rgbkrk:
## Context

Follow-up to #2163, which closed #2157. When `auto_launch_kernel` detects that an `environment.yml`-declared conda env isn't built on the machine, it now falls back to `Prewarmed(Conda)` so the kernel launches via a pool conda env. The notebook is usable.

What's missing: the env.yml's **deps** are bootstrapped into the CRDT (by #2148's path) but don't auto-sync into the claimed pool env. The pool env only has `ipykernel` + `ipywidgets`. Imports like `import pandas` fail until the user manually syncs.

## Repro

1. Dir with `environment.yml` declaring `dependencies: - pandas`, `name: whatever-not-built`.
2. Open a notebook in that dir. Fallback fires.
3. Run `import pandas` → `ModuleNotFoundError: No module named 'pandas'`.

Daemon log on open shows `conda:env_yml declared env 'whatever-not-built' not built on this machine; falling back to conda:prewarmed` and the kernel reaches `idle`, but no `sync_dependencies` call for the env.yml deps.

## Fix shape

On fallback, either:

1. **Auto-sync**: after the pool env is captured, call `sync_dependencies` with the env.yml's deps + channels so the pool env grows to match what env.yml declared. Slower first open (conda solve + install) but the notebook matches user expectations.

2. **Surface the drift**: set `RuntimeStateDoc.env.in_sync = false` with `added = env_yml_deps` so the frontend's existing sync banner fires. One click from the user and deps install. Keeps the fallback fast and transparent about what's missing.

3. **Both**: auto-sync plus surface state in the same pass.

Option 2 is closest to existing UX patterns (`in_sync=false` already drives banner UI for other drift cases). Option 1 is nicest but slower on first open and less transparent about what the daemon did.

Recommend Option 2 as the smaller increment. Option 3 is the right long-term answer together with the consent prompt in #2157's longer-path sketch.

## Related

- Parent: #2157, PR #2163
- Consent prompt UX: still open as a product decision (noted in #2157's "Fix direction" section of the original writeup).

rgbkrk:
[...]

PR #2163, opened 2026-04-24
fix(runtimed): fall back to conda pool when env.yml declared env is missing

rgbkrk:
Closes #2157.

## Overview

A notebook next to an `environment.yml` declaring a conda env that isn't built on the user's machine sent the runtime agent down the `conda:env_yml` path, which tried to resolve the named env via `find_named_conda_env`, got `None`, and returned `Err("conda:env_yml could not resolve conda environment prefix")`. The agent died, the kernel stayed in `initializing` forever, and nothing in the daemon's default log level explained why.

## Diagnosis

From the user-feedback writeup on 2026-04-24:

```
LaunchKernel: type=python source=conda:env_yml
WARN Agent kernel launch failed:
  Failed to launch kernel: conda:env_yml could not resolve conda environment prefix
Runtime agent exited with status: signal: 9 (SIGKILL)
```

The failure was hard-coded in `jupyter_kernel.rs:375-377`. Nothing in the orchestrator caught it. `reset_starting_state` was called but no recovery was attempted.

## Fix

Added `missing_conda_env_yml_name` in `notebook_sync_server/metadata.rs`. Given a detected `environment.yml`, it returns the declared env name when the env isn't built. Uses the same resolver (`resolve_conda_env_prefix` → `find_named_conda_env`) the kernel path uses, so detection agrees with what the kernel would see.

Wired into both `env_source` decision sites in `auto_launch_kernel`. When the check fires, the log warns with the declared name and env_source becomes `Prewarmed(Conda)` instead of `EnvYml`. The runtime agent launches against a pool conda env; the notebook is usable.

## Scope limit

The env.yml's deps are bootstrapped into the CRDT by #2148's path but **do not auto-sync into the claimed pool env on fallback**. Imports from those deps (`import pandas`) fail until the user adds the deps through the UI or the sync banner. Tracking separately; the scope of this PR is "kernel no longer stuck forever," not full parity with a built env.

The per-feedback-document proper UX is a consent prompt at open time ("this repo declares environment.yml, ~5min + ~2GB to build; build now or use pool?"). That's a product design + frontend change, distinct from this daemon-side unblock. Worth a followup issue.

## Test plan

- [x] `cargo test -p runtimed --lib` — 448 passing (+3)
- [x] `cargo clippy -p runtimed --all-targets` — clean
- [x] Three unit tests for `missing_conda_env_yml_name`: detects unbuilt named env, skips non-env.yml project files, treats `prefix:` pointing at a real env as not missing.
[...]

PR #2167, opened 2026-04-24
fix(runtimed): surface typed error for missing conda env.yml env

### LINKED ISSUE #2157: fix(runtimed): conda:env_yml hard-fails with no fallback when declared env is missing
## Observed

On a fresh clone of a repo with `environment.yml` declaring a conda env that isn't built on the machine, opening a notebook leaves the kernel stuck in "initializing" forever. The daemon pool has 3 conda envs and 3 uv envs available, but none are used.

From `runtimed.log`:
```
Auto-launching kernel for notebook ... (trust: NoDependencies, new: false)
Auto-launch: detected project file ".../environment.yml" -> conda:env_yml
Auto-launch: using project file -> conda:env_yml
Auto-launch: conda:env_yml prepares its own env, no pool env needed
Spawning runtime agent subprocess for auto-launch
LaunchKernel: type=python source=conda:env_yml
WARN Agent kernel launch failed:
  Failed to launch kernel: conda:env_yml could not resolve conda environment prefix
Runtime agent exited with status: signal: 9 (SIGKILL)
```

`conda:env_yml` routes the launch to the declared env by name. If the env doesn't exist, resolver fails and the daemon gives up without falling back to the pool. No banner, no prompt, kernel just never starts.

Workaround: build the env manually via `conda env create -f environment.yml`. That takes minutes and gigabytes of disk, which is the real product question — it shouldn't happen without consent, and it shouldn't block the notebook from opening.

## Code path

`crates/runtimed/src/requests/launch_kernel.rs` — the `conda:env_yml` branch in the env resolution. `crates/runtimed/src/notebook_sync_server/metadata.rs:3692` uses `env_source == "conda:env_yml"` as a marker for "project-backed, don't use pool."

## Fix shape

Two options, not mutually exclusive:

### Short path: pool fallback with banner

On `conda:env_yml` resolution failure, fall back to the conda pool (or uv pool). Set a new RuntimeStateDoc field — `env.declared_env_missing: true` plus a name field — so the frontend can show "using pool env; declared env not built" with a "build now" button. Kernel launches immediately with a working env. User can build the real env when they're ready.

### Longer path: consent prompt

At open time, when the daemon detects `environment.yml` (or pyproject with heavy conda deps), show a RuntimeStateDoc-driven prompt: "This repo declares a conda env (~2GB download, ~5min). Build now, use pool env for now, or choose another env." This is a new lifecycle state similar to `AwaitingTrust`. Distinct from trust — env creation is a disk/bandwidth/time decision, not an execution-safety decision.

Recommend shipping the short path first as a strict improvement over the current silent failure, then layering the prompt UI on top.

## Related

- Similar "gate the kernel launch on a user decision" shape as #2150 (trust heal at room init). If we grow more of these lifecycle gates, it's worth consolidating the pattern.

## Session context

Captured in user-feedback writeup from 2026-04-24 teaching-notebook session using https://github.com/ds-modules/Small_Models_SP26 at commit `e66dd4c`.

rgbkrk:
Closes #2157.

## Overview

When `environment.yml` next to a notebook declares a conda env that isn't built on the user's machine, the runtime agent hit `"could not resolve conda environment prefix"` and died. The kernel stayed in `initializing` forever, nothing in the default log level or RuntimeStateDoc explained why, and neither the frontend banner nor MCP tools had any reason to surface.

## Rejected bandaid

An earlier draft (closed as #2163) silently fell back to a conda pool env. Got rejected: the user declares specific deps in env.yml; a pool env doesn't have them; imports fail; notebook looks broken. "Silently wrong" is worse than "visibly broken."

## Fix

Surface a **typed error state with specific details** that lands on the CRDT and flows to every reader.

**runtime-doc:**
- New `KernelErrorReason::CondaEnvYmlMissing` variant (`"conda_env_yml_missing"`).
- New `kernel.error_details` CRDT field — free-form string for specifics that don't fit the typed reason enum. Here it carries the declared env name + remediation command.
- New `set_lifecycle_with_error_details` setter.

**runtimed:**
- `missing_conda_env_yml_name` helper detects the miss using the same resolver the runtime agent uses.
- `auto_launch_kernel` checks before spawning the runtime agent. On miss, sets lifecycle = Error with the reason + details, returns. No wasted agent spawn, no stuck-forever.

**Wire consistency:**
- Python `runtimed._constants.KERNEL_ERROR_REASON.CONDA_ENV_YML_MISSING`.
- TypeScript `KERNEL_ERROR_REASON.CONDA_ENV_YML_MISSING` in `@runtimed`.
- `KernelState.error_details: string | null` on the TS side.
- `PyKernelState.error_details: Optional[str]` on the Python side.

**MCP surface:**
- `runt-mcp`'s `read_runtime_info` exposes both `error_reason` and `error_details` when lifecycle is Error. Verified E2E via `connect_notebook`:
  ```json
  {
    "kernel_status": "error",
    "error_reason": "conda_env_yml_missing",
    "error_details": "environment.yml declares conda env '...', which is not built on this machine. Run: conda env create -f ..."
  }
  ```

## Out of scope (follow-ups)

- **Frontend banner** rendering this error specifically. The CRDT fields are there and the TS types are updated; the banner component (`NotebookToolbar`) can light up on `error_reason === KERNEL_ERROR_REASON.CONDA_ENV_YML_MISSING` in a follow-up PR.
- **Consent prompt UX** ("build env now?"): the product design direction from the feedback writeup. Distinct lifecycle state. v2.4+.
- **Auto-sync env.yml deps into pool env** (#2164): different approach to the same user goal. Not blocking this PR.

## Test plan

- [x] `cargo test -p runtime-doc --lib` — 134 passing
- [x] `cargo test -p runtimed --lib` — 448 passing (+3 for `missing_conda_env_yml_name`)
- [x] `cargo clippy --workspace --exclude runtimed-py` — clean
- [x] runtimed-py integration test `test_envyml_missing_env_surfaces_error_state` (polls RuntimeStateDoc for lifecycle=Error, asserts reason + details)
- [x] Manual E2E: dev daemon rebuilt, notebook next to env.yml naming `nteract-integration-probe-unbuilt-env-xyz`. Daemon log fires the warn with remediation command. No runtime agent spawn. `connect_notebook` returns error_reason + error_details as above.

### DIFF
--- a/apps/notebook/src/hooks/useDaemonKernel.ts
+++ b/apps/notebook/src/hooks/useDaemonKernel.ts
@@ -405,6 +405,7 @@ export function useDaemonKernel({
     statusKey,
     lifecycle: runtimeState.kernel.lifecycle,
     errorReason: runtimeState.kernel.error_reason,
+    errorDetails: runtimeState.kernel.error_details,
     queueState,
     kernelInfo,
     envSyncState,
--- a/crates/runt-mcp/src/tools/session.rs
+++ b/crates/runt-mcp/src/tools/session.rs
@@ -180,6 +180,28 @@ fn read_runtime_info(handle: &notebook_sync::handle::DocHandle) -> serde_json::V
                     serde_json::json!(state.env.prewarmed_packages),
                 );
             }
+            // Error surface (#2157): when the kernel is in an error
+            // state, MCP clients get both the typed reason (stable
+            // for programmatic handling) and the human-readable
+            // details (for surfacing to the user).
+            if matches!(state.kernel.lifecycle, runtime_doc::RuntimeLifecycle::Error) {
+                if let Some(reason) = state
+                    .kernel
+                    .error_reason
+                    .as_deref()
+                    .filter(|s| !s.is_empty())
+                {
+                    info.insert("error_reason".into(), serde_json::json!(reason));
+                }
+                if let Some(details) = state
+                    .kernel
+                    .error_details
+                    .as_deref()
+                    .filter(|s| !s.is_empty())
+                {
+                    info.insert("error_details".into(), serde_json::json!(details));
+                }
+            }
         }
         Err(_) => {
             info.insert("kernel_status".into(), serde_json::json!("unknown"));
--- a/crates/runtime-doc/src/doc.rs
+++ b/crates/runtime-doc/src/doc.rs
@@ -105,6 +105,16 @@ pub struct KernelState {
     /// deserializes as `Some("")`, indicating "scaffolded but unset").
     #[serde(default)]
     pub error_reason: Option<String>,
+    /// Free-form details accompanying an error, shown to the user via
+    /// the frontend banner and exposed to MCP tools. Carries specifics
+    /// that don't fit in the typed [`error_reason`] enum — e.g., the
+    /// name of a conda env declared in environment.yml that isn't
+    /// built on this machine, with a suggested remediation command.
+    ///
+    /// `None` when the CRDT field is absent; empty string indicates
+    /// "scaffolded but unset" (same convention as `error_reason`).
+    #[serde(default)]
+    pub error_details: Option<String>,
 }
 
 impl Default for KernelState {
@@ -118,6 +128,7 @@ impl Default for KernelState {
             runtime_agent_id: String::new(),
             lifecycle: RuntimeLifecycle::NotStarted,
             error_reason: None,
+            error_details: None,
         }
     }
 }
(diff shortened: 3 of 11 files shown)


rgbkrk:
Codex P2 applied in 47b3d59. `resolve_conda_env_prefix` returns `Some(prefix)` for every `prefix:` value without checking existence, so `prefix: /missing/env` skipped the guard. Fixed by verifying `conda_python_path(prefix).exists()` before accepting. Two new tests: `test_missing_conda_env_yml_name_prefix_missing_reports_path` and `test_missing_conda_env_yml_name_prefers_name_over_missing_prefix`. Helper now reports the prefix path when no name: is set so the banner has something useful to show.

Issue #2132, opened 2026-04-24
feat(trust): per-package allowlist to auto-approve familiar dependencies


rgbkrk:
## Problem

The trust dialog prompts every time a new notebook opens, even when every package in the list has been approved dozens of times in other notebooks. pandas, numpy, polars, matplotlib. The same packages, every time.

This is especially annoying when agents add deps to notebooks. You've trusted these packages before. The app should know that.

The current trust model is per-notebook, per-machine. Each notebook gets its own HMAC signature. There's no memory of which individual packages the user has already blessed.

Auto re-sign helps when you edit an already-trusted notebook. It doesn't help with the first open of a new one.

## Proposed solution

A **trusted packages allowlist** managed by the daemon. If every dependency in a notebook is on the allowlist, skip the dialog entirely. Kernel starts immediately.

### Behaviors

- **Organic growth.** When the user clicks "Trust & Start", those packages get added to the allowlist automatically. The list builds itself through normal use.
- **Partial coverage.** If a notebook has 4 approved packages + 1 novel one, the dialog still shows. But it marks the approved ones so the user can focus on reviewing what's new.
- **Name-only matching.** The allowlist matches on package name, not version specifiers. `pandas>=2.0` and `pandas>=1.5` both match `pandas`.
- **Manual management.** Expose the list somewhere in the UI for review and removal.

### What doesn't change

- Per-notebook HMAC signatures stay. Defense-in-depth for notebooks from untrusted sources.
- Typosquat detection still runs on novel packages.
- The allowlist is a convenience layer on top of the existing trust model, not a replacement.

## Storage

Open question. The daemon needs a persistent store for this. Settings doc (Automerge CRDT) is the wrong layer. This is accumulated decision history, not user preferences.

This is also the first of several "daemon needs a local data store" needs. IPython/execution history cache, notebook search indexes, and other accumulated state have similar requirements. Worth thinking about the data model holistically rather than one-offing a flat file for package names.

Options worth considering: SQLite, LanceDB (embedded columnar + vector search for future history/search needs), or a simple append-only file to start.

## UX sketch

Trust dialog with partial coverage:

```
Review Dependencies

This notebook wants to install packages. Once approved, the kernel
will start automatically.

PyPI Packages
  ✓ pandas          (approved)
  ✓ numpy           (approved)
  ✓ polars          (approved)
    scikit-learn

  [Don't Install]  [Trust & Start]
```

[...]

PR #2176, opened 2026-04-24
spike: runt-store + LanceDB evaluation (not for merge)

rgbkrk:
**Not for merge.** Draft artifact from a spike evaluating LanceDB as the daemon's persistent data store. Closing after review so the branch stays reachable for reference.

## What this spike is

A scratch `runt-store` crate with a LanceDB-backed `TrustAllowlist` facade (tests + benchmarks, no daemon wiring). The goal was to answer three questions before committing the daemon to LanceDB:

1. Is the hot path fast enough? (Per-kernel-launch trust check must be free.)
2. Is the cold-start cost tolerable? (Daemon boots often.)
3. What's the binary-size cost?

Numbers are in. Write-up below tries to keep the lens wide because the allowlist is the *smallest* workload Lance would serve — future workloads are what justify the weight.

## What Lance actually is

Two layers:

- **Lance** — columnar file format, Arrow-native, MVCC versioning, random access, zero-copy. Scalar + vector indexes: btree, bitmap, bloomfilter, inverted (FTS), ngram, label_list, rtree, HNSW/IVF.
- **LanceDB** — embedded database on top of Lance. Multiple tables per database, schema evolution, query builder with predicates + vector search + FTS + hybrid. Uses DataFusion for query execution.

Pure Rust, embedded, async/tokio. No external server.

## How it maps to our actual workloads

This is the key framing — the allowlist alone is not a good argument for Lance.

### Trust allowlist (#2132) — the driver, also the smallest

- 100s-1000s of `(manager, name)` rows per user.
- Hot path: "is `pandas` approved?" on every kernel launch.
- In the spike, the table loads into an `Arc<RwLock<HashSet>>` at daemon startup. `contains()` never touches disk.
- Speed: contains=82ns, cold_load=642µs@100rows, add=1.35ms per click.
- A HashSet-on-disk (JSON, SQLite) would match these numbers. Lance is overkill here in isolation.

### Parquet in the blob store (sift dataframe rendering) — existing Arrow cost

We already pull Arrow 54 + parquet through `nteract-predicate`, `sift-wasm`, `repr-llm`. Sift materializes outputs to parquet in the blob store for the dataframe viewer. Lance stores data in Lance file format, not parquet, but Lance can read+write parquet natively. Two futures:

- Keep parquet as the blob wire format, use Lance for indexed tables — coexist.
- Move the blob store to Lance fragments — fewer formats, one set of Arrow primitives, better versioning on output edits.

If we consolidate, some of the 25MB delta is shared with Arrow we already pay for. Measured incrementally on top of arrow-*54* today; moving to 57 workspace-wide would shift the baseline.

### Arrow IPC streaming (future)

Arrow-native by definition. Lance composes cleanly: kernel emits RecordBatches, daemon persists into Lance tables for durability + indexed later queries. No format boundary.

### Indexed search over inputs and outputs (future, highest-leverage)

The workload that actually justifies a heavy store:

- Cell source text (inputs): FTS via inverted index, ngram for substring.
- Cell outputs: MIME-typed. Text outputs via FTS. Structured outputs via column-level indexes.
- Cross-session recall: "Last time I opened `analysis.ipynb`, what did I run?"
- Vector search (later): embed cell source/output text, semantic search across the user's corpus.

We don't have any of this today.

### IPython / kernel history cache (future)

Execution history lives in the live IPython process (per-session, lost on restart). A daemon-side cache with FTS + optional embeddings would give cross-notebook recall and power completions. Append-heavy, classic fit for either Lance or Turso.

## What we measured

Bench results on macOS arm64:

| Metric | p50 |
|---|---|
| `contains()` hit | **82 ns** |
| `contains()` miss | **95 ns** |
| `novel(20 vs 1k)` | **1.62 µs** |
| `cold_load` 100 rows | 642 µs |
| `cold_load` 1k rows | 889 µs |
| `cold_load` 10k rows | 2.6 ms |
| `add()` + fsync | **1.35 ms** |

Integration tests: 9/9 pass in ~40ms.

Binary size, release build:

| Binary | Baseline | With `runt-store` linked | Delta |
|---|---|---|---|
| `runtimed` | 24 MB | 49 MB | **+25 MB (+106%)** |
| `runt-cli` | 9.3 MB | 10 MB | +1 MB |

The `runtimed` delta is the full weight of `lancedb` + `lance-*` + `datafusion` + Arrow 57 (current workspace pins Arrow 54 via sift; Lance needs 57, forcing a workspace-wide bump or duplicate dep tree). DataFusion is the heaviest single block; no feature flag to excise it.

## The alternative: Turso (libSQL)

Worth naming explicitly because it inverts the weight/capability tradeoff.

Turso is a SQLite fork (libSQL) with native vector support. Embedded Rust client (`libsql` crate). Footprint is roughly SQLite-sized (~2-3MB static). Gives you:

- Full SQL, schema evolution, transactions.
- FTS5 built in — same engine we'd reach for for text search anyway.
- Native vector types: `F32_BLOB`, `vector_distance_cos`, `vector_top_k`. Brute force today, ANN on the roadmap.
- Optional remote sync (edge replication via Turso cloud). Not something we'd use now, but the option is there if we ever want cross-device trust/history.
- Mature, well-understood SQLite semantics for ops + backup.

**Where Turso doesn't fit:**

- Row-based, not columnar. No zero-copy Arrow interop. The parquet/sift consolidation story doesn't exist — we keep parquet as a separate format.
- Vector search is bolted on, not first-class. Lance has years of index work; Turso/libSQL is newer on this front.
- No MVCC-on-fragments equivalent for time-travel queries across output versions.

### Side-by-side for our workloads

| Workload | Lance | Turso |
|---|---|---|
| Trust allowlist | ✓ overkill | ✓ trivial SQL |
| IPython history cache (FTS) | ✓ inverted index | ✓ FTS5 |
| Cell/output search (FTS) | ✓ | ✓ |
| Vector search over cells | ✓ mature HNSW/IVF | ~ brute force, ANN coming |
| Parquet/Arrow consolidation | ✓ native Lance format | ✗ keeps parquet separate |
| Arrow IPC streaming | ✓ no format boundary | ✗ serialization on write |
| Binary cost | +25 MB | ~+3 MB |
| Cloud sync (future) | no | yes (Turso edge) |
| Schema migration ergonomics | evolving | classic SQLite, well-understood |

### Which answer we pick depends on where we go

- If indexed search + vector is the primary future workload and parquet consolidation is attractive → **Lance**. The 25MB is paid for with fewer format boundaries and a capability we don't have today.
- If Arrow/parquet stays separate and the daemon store is mainly "scalar state + text FTS + maybe vector" → **Turso** wins on weight, operational familiarity, and a sync-out path if we ever want it.

Neither is the right choice for *just* the allowlist. Both are correct when we actually need the store.

## Why we're not rushing the allowlist on SQLite/JSON

Keeping this explicit: introducing SQLite-classic or JSON *now* for the allowlist and switching to Lance/Turso later for search means two migrations and two user-visible "trust decisions history" implementations. The decision to make is **what store does the daemon own long-term**, not "what gets #2132 shipped fastest." Parking the spike keeps the option open; reaching for an interim forecloses the decision.

## What I'd do next (not in this PR)

1. **Parallel Turso spike**: same crate layout, same benches, same 9 integration tests. Get directly comparable numbers.
2. **Prototype indexed search on whichever wins the storage evaluation**: cell source FTS over a fixture notebook corpus, see if the query shape matches our actual needs (cross-session recall, completions, "what did I run yesterday").
3. **Decide on Arrow 54 → 57 workspace bump** separately. If we do that work for other reasons, Lance gets cheaper.
4. **Revisit the allowlist** once the above is settled. Implementation on either store is a ~200-line module.

## What lives on this branch

- `crates/runt-store/` — facade + `TrustAllowlist` + `paths` module.
- `benches/trust_allowlist.rs` — criterion benchmarks.
- `tests/trust_allowlist.rs` — 9 integration tests, all green.
- No wiring into `runtimed`. Only the workspace Cargo.toml adds `runt-store`.

Keeping the branch as a reference. Anyone can `cargo bench -p runt-store` to reproduce.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -189,6 +195,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "archspec"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -203,6 +218,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -214,17 +235,38 @@ version = "54.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5ec52ba94edeed950e4a41f75d35376df196e8cb04437f7280a5aa49f20f796"
 dependencies = [
- "arrow-arith",
- "arrow-array",
- "arrow-buffer",
- "arrow-cast",
- "arrow-data",
- "arrow-ipc",
- "arrow-ord",
- "arrow-row",
- "arrow-schema",
- "arrow-select",
- "arrow-string",
+ "arrow-arith 54.3.1",
+ "arrow-array 54.3.1",
+ "arrow-buffer 54.3.1",
+ "arrow-cast 54.3.1",
+ "arrow-data 54.3.1",
+ "arrow-ipc 54.3.1",
+ "arrow-ord 54.3.1",
+ "arrow-row 54.3.1",
+ "arrow-schema 54.3.1",
+ "arrow-select 54.3.1",
+ "arrow-string 54.3.1",
+]
+
+[[package]]
+name = "arrow"
+version = "57.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e4754a624e5ae42081f464514be454b39711daae0458906dacde5f4c632f33a8"
+dependencies = [
+ "arrow-arith 57.3.0",
+ "arrow-array 57.3.0",
+ "arrow-buffer 57.3.0",
+ "arrow-cast 57.3.0",
+ "arrow-csv",
+ "arrow-data 57.3.0",
+ "arrow-ipc 57.3.0",
+ "arrow-json",
+ "arrow-ord 57.3.0",
+ "arrow-row 57.3.0",
+ "arrow-schema 57.3.0",
+ "arrow-select 57.3.0",
+ "arrow-string 57.3.0",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


rgbkrk:
Closing as intended — this is a reference artifact, not merge material. Branch `spike/runt-store-trust-allowlist` stays reachable; `cargo bench -p runt-store` reproduces the numbers. Next step is a parallel Turso spike for side-by-side comparison before committing to a daemon-side store.

PR #2178, opened 2026-04-25
spike: runt-store + libSQL/Turso evaluation (not for merge)

rgbkrk:
**Not for merge.** Companion artifact to #2176. Closing after review so the branch stays reachable for reference.

## What this spike is

Mirror of the LanceDB spike (#2176) swapping the durability layer for libSQL (Turso's SQLite fork). Same `TrustAllowlist` facade, same 9 integration tests, same criterion benchmark harness, same binary-size measurement methodology. Direct apples-to-apples numbers for the "what should own daemon-side accumulated state?" decision.

## What libSQL actually is

SQLite fork maintained by the Turso team. Same file format, same SQL, same semantics. Adds:

- Native vector types (`F32_BLOB`, `vector_distance_cos`, `vector_top_k`) — brute force today, ANN on the roadmap.
- FTS5 built in (comes with SQLite).
- Optional remote replication / edge sync via Turso cloud (opt-in via features, not pulled in here).
- Pure-Rust embedded client `libsql` with async API.

Config here: `features = ["core"]` — disables `replication`, `remote`, `sync`, `tls`. What remains is the embedded DB only.

## Speed

Identical bench harness to #2176. Numbers from the same machine (macOS arm64):

| Metric | Lance | libSQL | Winner |
|---|---|---|---|
| `contains()` hit | 82 ns | **60 ns** | libSQL (−27%) |
| `contains()` miss | 95 ns | **60 ns** | libSQL (−37%) |
| `novel(20 vs 1k)` | 1.62 µs | **1.29 µs** | libSQL (−20%) |
| `cold_load` 100 rows | 642 µs | **429 µs** | libSQL (−33%) |
| `cold_load` 1k rows | 889 µs | 888 µs | tie |
| `cold_load` 10k rows | **2.6 ms** | 4.47 ms | Lance (libSQL +72%) |
| `add()` + fsync | 1.35 ms | **497 µs** | libSQL (−63%) |

The `contains` delta is noise — both implementations are pure `HashSet` lookups once the in-memory set is warm. The meaningful deltas are cold_load (SQLite's single-file BTree beats Lance's Arrow/DataFusion stack at small sizes) and `add` (libSQL's WAL commit beats Lance's fragment write). Lance pulls ahead at 10k rows where its columnar scan amortizes better, but that regime is not a realistic allowlist shape.

9/9 integration tests pass in ~20ms (Lance: 40ms).

## Binary size

[...]

PR #2181, opened 2026-04-25
spike: runt-store + pure-Rust turso evaluation (not for merge)

rgbkrk:
**Not for merge.** Third backend artifact after #2176 (Lance) and #2178 (libSQL). Closing after review; branch stays reachable.

## What this spike is

`runt-store` with the `turso` crate (0.5.3) as durability — the Turso team's ground-up, pure-Rust rewrite of the SQLite engine. Same `TrustAllowlist` facade, same 9 integration tests, same criterion harness, same binary-size measurement methodology.

## Why look at this

libSQL (#2178) is SQLite with Rust wrappers — still a C library underneath (`libsql-ffi`, `libsql-sys`, cc, bindgen). The `turso` crate is 100% Rust — no C toolchain, no bindgen. That's a supply-chain and cross-compilation win if the performance and maturity are close enough.

## API swap diff

Minimal — the `turso` crate is API-compatible with libsql by design:

- `libsql` → `turso` imports.
- `Builder::new_local(&Path)` → `new_local(&str)` (had to `.to_str()` the path).
- Explicit `BEGIN IMMEDIATE`/`COMMIT` via `execute` → `conn.transaction()` + `tx.commit()`.
- Dropped `WITHOUT ROWID` from the schema — turso 0.5's parser doesn't accept it yet. Same semantics for our workload (primary-key lookup either way), but flagging the maturity gap.

Integration tests: 9/9 pass in ~20ms.

## Numbers

All from the same macOS arm64 machine. See #2176 and #2178 for the baselines in context.

| Metric | Lance | libSQL | **Turso (Rust)** |
|---|---|---|---|
| Binary delta (runtimed) | +25 MB | +1 MB | **+4 MB** |
| `contains()` hit | 82 ns | 60 ns | **56 ns** |
| `contains()` miss | 95 ns | 60 ns | **56 ns** |
| `novel(20 vs 1k)` | 1.62 µs | 1.29 µs | **1.20 µs** |
| `cold_load` 100 rows | 642 µs | 429 µs | **285 µs** |
| `cold_load` 1k rows | 889 µs | 888 µs | 891 µs |
| `cold_load` 10k rows | **2.6 ms** | 4.47 ms | 6.15 ms |
| `add()` + commit | 1.35 ms | 497 µs | **98 µs** ⚠ |
| Tests | 9/9 | 9/9 | 9/9 |

## The suspicious `add()` number

Turso at 98 µs vs libSQL at 497 µs is a 5× delta on single-row commit. SQLite with WAL + `synchronous=NORMAL` (libSQL's default) does one `fsync` per commit, which is the minimum for durability. 98 µs is suspicious for that on this hardware.

Likely explanations:
- Turso 0.5 may not be issuing `fsync` on commit by default.
- Turso's `PRAGMA synchronous` handling may differ — the codebase has the pragma in its internal benchmarks but default behavior isn't clear from a quick read.
- macOS-specific: libsql uses `F_FULLFSYNC` for real durability; turso's `io/unix.rs` has `FullFsync` as a `FileSyncType` variant but may not invoke it on every commit.

Bottom line: we **shouldn't** trust the write-perf number until durability semantics are verified. A "fast write" that doesn't hit disk isn't a real comparison to libSQL's "slow write" that does.

## Ergonomics

- **Build time**: fresh workspace build with `turso` in runt-store took ~85s. libSQL was slightly faster because it reuses its C compile cache across rebuilds. First build from scratch is a wash.
- **No C toolchain**: actually delivers on the promise. No `cc`, no `bindgen`, no SQLite C sources. Dep graph is all Rust crates.
- **Single-file compile**: `cargo build` works on any platform Rust supports without host C prerequisites.

## Maturity gap

- 0.5 pre-1.0. Breaking changes are on the table.
- Parser doesn't accept `WITHOUT ROWID` (minor, workaround trivial).
- FTS5 story is unclear from the surface API — `turso_core` has fts5 module references but I didn't verify at query level.
- Unverified durability story (see above).
- No experience at scale in production embedded contexts (that I know of).

## Recommendation

**libSQL today, turso on the roadmap.**

libSQL wins where it counts right now:
- Stable SQLite semantics.
- Verified fsync on commit.
- FTS5 definitely works.
- +1 MB vs +4 MB.

But the `runt-store` facade is backend-agnostic. When turso reaches 1.0 with:
- Verified durability matching libSQL / stock SQLite.
- Full SQL surface including `WITHOUT ROWID` (minor) and FTS5 (matters).
- A track record of embedded use.

...the swap is a pure dep change. No call-site churn.

## What lives on this branch

- `crates/runt-store/` — facade + `TrustAllowlist` + `paths` module, now backed by turso.
- `benches/trust_allowlist.rs` — same criterion harness.
- `tests/trust_allowlist.rs` — same 9 integration tests.
- No wiring into `runtimed`. Only the workspace `Cargo.toml` adds the crate.

## Decision

Three spikes, same facade. Picking libSQL as the path forward. Next step is promoting one of the spike branches to a real PR that wires `runt-store` into `runtimed` and uses it for #2132.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ dependencies = [
  "getrandom 0.3.4",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.8.48",
 ]
 
 [[package]]
@@ -113,6 +113,12 @@ checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
 dependencies = [
  "libc",
 ]
+
+[[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
@@ -400,7 +406,7 @@ dependencies = [
  "futures-core",
  "libc",
  "portable-atomic",
- "rustc-hash",
+ "rustc-hash 2.1.2",
  "tokio",
  "tokio-stream",
  "xattr",
@@ -595,7 +601,7 @@ dependencies = [
  "itertools 0.14.0",
  "leb128",
  "rand 0.9.4",
- "rustc-hash",
+ "rustc-hash 2.1.2",
  "serde",
  "sha2 0.11.0",
  "smol_str",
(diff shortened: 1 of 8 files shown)


rgbkrk:
Closing as designed — reference artifact. Branch `spike/runt-store-turso-rust-native` stays reachable; `cargo bench -p runt-store` reproduces. Recommendation holds: libSQL today, turso on the roadmap once 1.0 + verified durability + FTS5.

PR #2186, opened 2026-04-25
fix(runtimed): sort .ipynb keys + strip output_id on save

rgbkrk:
Two bugs in `notebook_sync_server/persist.rs` that together turned a one-line notebook edit into 478 modified lines in [ds-modules/Small_Models_SP26](https://github.com/ds-modules/Small_Models_SP26):

### 1. Key order was struct-declaration order, not alphabetical

Python `nbformat.write` emits `.ipynb` files via `json.dumps(sort_keys=True)`. Our daemon built the JSON with a `serde_json::json!{...}` that listed `nbformat`, `nbformat_minor`, `metadata`, `cells` in that order, and `serde_json` in this workspace runs with `preserve_order` so nothing re-sorted. Every save rewrote the top of the file.

Fix: add `sort_value_keys`, a recursive walker that rebuilds each `Value::Object` with sorted keys, and apply it in `serialize_notebook_json` before the existing `PrettyFormatter` + trailing-newline pass. Works regardless of the `preserve_order` feature flag.

### 2. `output_id` was leaking to disk on every save

`output_id` is a daemon runtime UUID minted per kernel execution. Its jobs are (a) addressing outputs in the in-memory output index and (b) giving the frontend stable React keys so streaming appends don't flicker. It's not part of nbformat. `resolve_manifest` included it in its return value (used everywhere, including the save path), so every re-execution wrote fresh UUIDs to disk — visible as churn across the whole outputs block in git.

Fix: `strip_runtime_only_output_fields` removes `output_id` from each output `Value` after `resolve_cell_output` and before it lands in the cell's JSON. `resolve_manifest` is unchanged (in-memory users still get the UUID). `load.rs` already mints fresh UUIDs when loading outputs without `output_id`, so round-tripping just re-UUIDs on each open — the daemon was already written to tolerate this case.

## Behavioral notes

- **One-time reformat on first post-change save.** Existing notebooks in user working copies will rewrite to sorted-keys + no-output_id on first edit. After that, churn goes to zero. Same rollout shape as prettier/black.
- **File watcher is safe.** `SELF_WRITE_SKIP_WINDOW_MS` still protects against self-write echoes.
- **Autosave is safe during load.** `is_loading` gate in `spawn_autosave_debouncer` (persist.rs:867) already prevents load-time doc mutations from queueing a save.
- **Content-hash guard still works.** `save_notebook_to_disk` (persist.rs:241) compares serialized output against on-disk bytes to skip no-op writes.

## Tests

- **`sort_value_keys_*`** — 3 unit tests on the tree walker (top-level, nested, arrays).
- **`serialize_notebook_json_emits_sorted_keys_and_trailing_newline`** — the whole emit path.
- **`strip_runtime_only_output_fields_*`** — 3 tests covering present, absent, non-object.
- **`test_save_notebook_to_disk_with_outputs`** — existing test gets one extra assertion that `output_id` isn't in saved bytes.
- **`test_save_notebook_to_disk_produces_sorted_stable_output`** — new integration test: round-trips a code+markdown notebook with an explicit runtime-only `output_id`, asserts byte-identical consecutive saves, alphabetical top-level + per-cell keys, and no `output_id` on disk.

All 184 pre-existing `notebook_sync_server` tests still pass. 10 new assertions across 8 new test functions.

## Not in this PR

- **nbformat crate dep bump to 2.1.0 / route through `nbformat::serialize_notebook`.** Explored; the crate's typed `v4::Output` API is the wrong abstraction for daemon save because our outputs are already pre-resolved nbformat-shaped JSON produced by `resolve_manifest`, not `jupyter_protocol::media::Media`. Converting back through the typed enum would round-trip every output through type inference and lose fields. The `sort_value_keys` pattern is identical to the one shipped in [runtimed/runtimed#298](https://github.com/runtimed/runtimed/pull/298); if we later want one source of truth, we can `pub` that helper and import it, but duplicating ~15 lines is cheaper than introducing a typed-conversion layer.
- **Clone rework.** `clone_notebook_to_disk` also hand-rolls JSON but Save-As-Copy's design is getting rethought anyway — tracked separately as a follow-up to create an ephemeral (no-disk) clone seeded from the current doc.
- **Binary MIME split direction fix.** Upstream in `jupyter-protocol`'s media serializer, binary payloads (`image/png`, etc.) get split into multi-line lists while modern Python `nbformat.write` emits single strings. Separate follow-up, decided direction pending.

## Test plan

- [x] `cargo test -p runtimed --lib notebook_sync_server` (184 passing, +10 new)
- [x] `cargo clippy -p runtimed --tests -- -D warnings`
- [x] `cargo fmt -p runtimed --check`

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/persist.rs
+++ b/crates/runtimed/src/notebook_sync_server/persist.rs
@@ -2,19 +2,60 @@ use serde::Serialize;
 
 use super::*;
 
+/// Recursively rebuild every `Value::Object` with its keys in sorted order.
+///
+/// Python `nbformat.write` emits `.ipynb` files via `json.dumps(sort_keys=True)`.
+/// Our daemon save path used to leave keys in insertion order (serde_json here
+/// runs with `preserve_order` via the workspace's feature set), which produced
+/// diffs against Jupyter-written files on every save. Applying the sort
+/// ourselves before the pretty-printer matches Jupyter byte-for-byte at every
+/// depth.
+fn sort_value_keys(value: serde_json::Value) -> serde_json::Value {
+    match value {
+        serde_json::Value::Object(map) => {
+            let mut entries: Vec<(String, serde_json::Value)> = map.into_iter().collect();
+            entries.sort_by(|a, b| a.0.cmp(&b.0));
+            let mut sorted = serde_json::Map::new();
+            for (k, v) in entries {
+                sorted.insert(k, sort_value_keys(v));
+            }
+            serde_json::Value::Object(sorted)
+        }
+        serde_json::Value::Array(items) => {
+            serde_json::Value::Array(items.into_iter().map(sort_value_keys).collect())
+        }
+        other => other,
+    }
+}
+
 /// Serialize a notebook JSON value with 1-space indent and trailing newline,
-/// matching the nbformat/Jupyter convention.
+/// matching the nbformat/Jupyter convention. Keys are sorted alphabetically
+/// at every depth.
 fn serialize_notebook_json(value: &serde_json::Value) -> Result<String, String> {
+    let sorted = sort_value_keys(value.clone());
     let mut buf = Vec::new();
     let formatter = serde_json::ser::PrettyFormatter::with_indent(b" ");
     let mut ser = serde_json::Serializer::with_formatter(&mut buf, formatter);
-    value
+    sorted
         .serialize(&mut ser)
         .map_err(|e| format!("Failed to serialize notebook: {e}"))?;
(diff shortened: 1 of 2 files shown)


rgbkrk:
Superseded by a typed-schema rewrite in the new PR. Same wins (sort keys, strip output_id) but achieved by routing through `nbformat::v4::Notebook` + `nbformat::serialize_notebook` 2.1.0 instead of hand-rolling helpers. That also gets us schema validation on every save and drops per-execution runtime-only fields by type construction.

PR #2187, opened 2026-04-25
fix(settings): make raw frame reads cancel-safe

rgbkrk:
## Summary

- add a cancel-safe `RawFrameReader` for raw length-prefixed frames
- use it in settings sync server select loops instead of polling `recv_frame` directly
- split the settings sync client stream after handshake so initialization timeouts and `recv_changes()` receive through the raw frame actor

## Why

`recv_frame` uses `read_exact`, so cancelling it inside `select!` or `timeout` can discard partial frame bytes while keeping the stream alive. This applies the same actor-based fix as typed notebook frames to settings sync.

## Validation

- `cargo check -p runtimed-client -p runtimed`
- `cargo test -p notebook-protocol raw_frame_reader_survives_timeout_during_partial_frame -- --nocapture`
- `cargo test -p runtimed-client sync_client -- --nocapture`
- `cargo test -p runtimed test_settings_sync_via_unified_socket -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/notebook-protocol/src/connection.rs
+++ b/crates/notebook-protocol/src/connection.rs
@@ -27,7 +27,7 @@
 //! - `0x03`: NotebookBroadcast (JSON)
 //! - `0x07`: SessionControl (JSON, server-originated)
 
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


rgbkrk:
Closing - will redo. Details preserved in PR history.

PR #2197, opened 2026-04-25
feat(protocol): offload Comm broadcast buffers through blob store

rgbkrk:
## Summary

Moves widget buffer bytes off the typed-frame wire (kernel → frontend direction). `NotebookBroadcast::Comm.buffers` becomes `Vec<BlobRef>` instead of `Vec<Vec<u8>>`. The runtime agent already has blob-store write access (it's how outputs work), so this needs zero new auth surface.

Closes an asymmetry that's been sitting in `jupyter_kernel.rs`: the same iopub handler already called `store_widget_buffers()` to offload buffers for the doc-state side at line 1727, then turned around and emitted the *same buffers raw* on the broadcast path at line 1754. Same bytes, two paths — one offloaded, one not. This PR uses the same blob-store pattern for both.

## Why this PR exists

Background from PRs #2191 / #2193: per-type frame caps with a 16 MiB Request cap, sized to fit `SendComm.buffers` after JSON's ~4× expansion of binary. The cap was honest but loose. The principled fix is to take the bytes off the channel — per-content offloading, not bigger caps.

`NotebookBroadcast::Comm.buffers` was the kernel-controlled side of the same problem: bytes flowing the *other* direction, also raw on the wire. Tightening Broadcast and Request caps both depend on offload work.

## Mechanics

| Piece | What changed |
|---|---|
| `BlobRef` struct (new) | `notebook-protocol`. Same JSON shape (`{blob, size, media_type}`) as the inline `ContentRef::Blob` used in output manifests. |
| `NotebookBroadcast::Comm.buffers` | `Vec<Vec<u8>>` → `Vec<BlobRef>` |
| `store_buffers_as_refs(buffers, blob_store)` (new) | `runtimed::output_prep`. Parallel-list counterpart to the existing in-place `store_widget_buffers`. Stores each buffer, returns a `Vec<BlobRef>`. Failed puts are logged + skipped. |
| `jupyter_kernel.rs:1754` | Calls `store_buffers_as_refs` before `broadcast_tx.send`. The two iopub-handler paths now offload buffers symmetrically. |
| `App.tsx` Comm broadcast handler | Reads `BlobRef[]` from the broadcast, fetches each via `http://127.0.0.1:<blob_port>/blob/<hash>`, hands resolved `ArrayBuffer[]` to the existing `widgetStore.emitCustomMessage` contract. |

## Cap stays at 16 MiB this PR

The `NotebookBroadcast` cap is unchanged (still 16 MiB). Tightening to 1 MiB lands once `SendComm.buffers` (the outbound direction) is also offloaded, since loosening Request to fit `SendComm` was the original reason for the matching loose Broadcast cap. Doing both at once keeps the cap-table consistent.

## Follow-ups

Tracked in `.context/plans/buffer-offload-and-blob-socket.md`:

1. **Iframe-side resolution.** Push the `BlobRef → ArrayBuffer` fetch into the iframe sandbox so widget binary stops materializing in the parent webview at all. Matches how output bytes already work — outputs travel as URLs to the iframe, the iframe fetches from the blob HTTP server. This PR keeps the parent fetch as a temporary bridge that preserves the existing `widgetStore.emitCustomMessage(commId, content, ArrayBuffer[])` contract.
2. **Outbound offload via `runtimed-blobs.sock`.** Frontend has no blob-store write path today; `SendComm.buffers` rides the typed wire. Adding a dedicated Unix socket for blob writes (separate from the typed-frame socket) eliminates the frame-mixing concern by physical channel separation, not just type-byte separation.
3. **Cap tightening.** Once both directions are offloaded, Request and Broadcast caps drop to 1 MiB.

## Test plan

- [x] `cargo build --workspace --exclude runtimed-py`
- [x] `cargo test -p notebook-protocol -p runtimed-client -p runtimed --lib --tests` (179 / 487 / 69 passing)
- [x] `pnpm test:run` (1048 frontend tests passing)
- [x] New: `protocol::tests::test_notebook_broadcast_comm_with_buffer_refs` — round-trips a `Comm` broadcast with a `BlobRef`, asserts hash and size land on the wire.
- [ ] Manual: open a notebook with a custom-message-driven widget (anywidget `model.send`), confirm buffers arrive intact end-to-end.

## Predecessors

- #2188 — drop dead broadcast variants
- #2190 — move path/last_saved to RuntimeStateDoc
- #2191 — per-type frame caps
- #2193 — bump Request cap to 16 MiB for SendComm widget buffers
- #2194 — drop dead Tauri dirty machinery

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -412,16 +412,44 @@ function AppContent() {
 
     // Custom comm messages (buttons, model.send()) are ephemeral events
     // delivered via broadcast, not CRDT state. Route to WidgetStore.
+    //
+    // Buffers ride as `BlobRef[]` ({ blob, size, media_type }). Fetch them
+    // from the daemon's blob HTTP server and hand the resolved
+    // ArrayBuffer[] to the existing widget-store contract. A follow-up
+    // moves the fetch into the iframe sandbox so widget binary stops
+    // materializing in the parent webview at all (matching the output
+    // bytes path).
     const customSub = engine.commBroadcasts$.subscribe((broadcast) => {
       const content = broadcast.content as Record<string, unknown> | undefined;
       const data = content?.data as Record<string, unknown> | undefined;
-      if (data?.method === "custom") {
-        const commId = content?.comm_id as string;
-        const inner = (data?.content as Record<string, unknown>) ?? {};
-        const buffers = (broadcast as { buffers?: number[][] }).buffers;
-        const arrayBuffers = buffers?.map((arr: number[]) => new Uint8Array(arr).buffer);
-        widgetStore.emitCustomMessage(commId, inner, arrayBuffers);
+      if (data?.method !== "custom") return;
+      const commId = content?.comm_id as string;
+      const inner = (data?.content as Record<string, unknown>) ?? {};
+      const refs = (broadcast as { buffers?: { blob: string; size: number }[] }).buffers ?? [];
+      if (refs.length === 0) {
+        widgetStore.emitCustomMessage(commId, inner, undefined);
+        return;
       }
+      const port = getBlobPort();
+      if (port === null) {
+        // Blob port not yet resolved — drop buffers, deliver content only.
+        // This is a rare race on first connect; comms after the first
+        // daemon:ready always have a port.
+        widgetStore.emitCustomMessage(commId, inner, undefined);
+        return;
+      }
+      Promise.all(
+        refs.map((ref) =>
+          fetch(`http://127.0.0.1:${port}/blob/${ref.blob}`).then((r) => r.arrayBuffer()),
+        ),
+      )
+        .then((arrayBuffers) => {
+          widgetStore.emitCustomMessage(commId, inner, arrayBuffers);
+        })
+        .catch((err) => {
+          logger.warn("[comm] failed to fetch widget buffers:", err);
+          widgetStore.emitCustomMessage(commId, inner, undefined);
+        });
     });
 
     return () => {
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -425,7 +425,7 @@ function AppContent() {
       if (data?.method !== "custom") return;
       const commId = content?.comm_id as string;
       const inner = (data?.content as Record<string, unknown>) ?? {};
-      const refs = (broadcast as { buffers?: { blob: string; size: number }[] }).buffers ?? [];
+      const refs = broadcast.buffers ?? [];
       if (refs.length === 0) {
         widgetStore.emitCustomMessage(commId, inner, undefined);
         return;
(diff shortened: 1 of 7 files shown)


rgbkrk:
Closing without merging.

## Why

The Rust offload mechanically works (verified: a 4 KiB buffer from a raw `ipykernel.comm.Comm.send(buffers=...)` test cell landed in the blob store at the moment the cell ran). But the design has a lifecycle hole I didn't think through before opening:

| Bytes | Owner in CRDT | Cleanup |
|---|---|---|
| Output content | Output manifest in `RuntimeStateDoc.executions[].outputs` | Tied to execution / cell lifecycle |
| Widget state values | `RuntimeStateDoc.comms.{id}.state` ContentRefs | Tied to comm; `comm_close` removes |
| **Custom-message broadcast buffers (this PR)** | **Nothing** | **Nothing** |

Custom comm messages are fire-and-forget events (`model.send`, button clicks, streaming widget updates). The broadcast goes out, the frontend fetches and discards. The `BlobRef` has no anchor in any Automerge document. A widget that emits `model.send(buffers=...)` at any meaningful rate just deposits content-addressed bytes in the persistent blob store forever — never referenced after the broadcast lands, never collectable because nothing knows the blob exists.

The blob store is the right destination for **state** (outputs, widget state) where the CRDT manifest keeps blobs findable. It is the wrong destination for **events**, where there's nothing to anchor them to.

## What's needed instead

A real fix would be an ephemeral write channel separate from the persistent content-addressed store. Two shapes worth considering:

1. **In-memory bounded cache** behind a separate HTTP route — short TTL (single-digit seconds), capped size, no disk.
2. **Dedicated `runtimed-blobs.sock` with a separate ephemeral namespace** (the design noted in `.context/plans/buffer-offload-and-blob-socket.md`) — but with explicit comm-broadcast buffers landing in a TTL-swept zone, not the content-addressed `~/.cache/.../blobs/` directory.

Either is a bigger lift than this PR. Until one lands, the 16 MiB Broadcast/Request cap from #2191/#2193 is the honest design — those buffers do legitimately ride the channel.

## What carries over

- `BlobRef` struct in `notebook-protocol`: leave for the next attempt — same shape will be reused regardless of which destination wins.
- `.context/plans/buffer-offload-and-blob-socket.md`: now needs an "ephemeral lifecycle" section before any further implementation.

## Pre-existing issues we surfaced testing this

Worth filing separately:

1. **WASM bootstrap panic** — `__wbg___wbindgen_throw_6b64449b9b9ed33c` deep in Automerge frames, triggered through `set_blob_port` and `detect_runtime`. Reproduces in dev consistently. Sync engine recovers partially in some sessions, fully blocks in others. Predates this branch.
2. **anywidget custom-message buffer rendering** — kernel sends, frontend doesn't paint. Reproduces in nightly (no PR code), so not from this branch either. Likely related to (1) or to something in the iframe ↔ widget-store bridge.

PR #2205, opened 2026-04-25
fix(runt-mcp): validate MCP tool inputs to prevent silent bad state

quillaid:
## Summary

- Reject invalid `cell_type` values in `create_cell`/`set_cell` — only "code", "markdown", "raw" accepted per nbformat spec
- Validate both source cell and anchor cell exist in `move_cell` before attempting the move
- Reject empty and malformed package names in `add_dependency`

Found by gremlin stress testing (iteration 1).

## Test plan

- [x] All 101 `runt-mcp` unit tests pass
- [x] `cargo clippy -p runt-mcp -- -D warnings` clean

### DIFF
--- a/crates/runt-mcp/src/tools/cell_crud.rs
+++ b/crates/runt-mcp/src/tools/cell_crud.rs
@@ -12,7 +12,7 @@ use notebook_protocol::protocol::NotebookRequest;
 use crate::execution;
 use crate::NteractMcp;
 
-use super::{arg_bool, arg_str, arg_string_array, tool_error, tool_success};
+use super::{arg_bool, arg_str, arg_string_array, cell_not_found, tool_error, tool_success};
 
 #[allow(dead_code)]
 #[derive(Debug, Deserialize, JsonSchema)]
@@ -185,10 +185,8 @@ pub async fn set_cell(
         .unwrap_or(30.0);
 
     let handle = require_handle!(server);
-
-    // Verify cell exists
-    if handle.get_cell(cell_id).is_none() {
-        return tool_error(&format!("Cell not found: {cell_id}"));
+    if let Some(err) = cell_not_found(&handle, cell_id) {
+        return err;
     }
 
     if source.is_none() && cell_type.is_none() {
@@ -248,20 +246,19 @@ pub async fn delete_cell(
         .ok_or_else(|| McpError::invalid_params("Missing required parameter: cell_id", None))?;
 
     let handle = require_handle!(server);
+    if let Some(err) = cell_not_found(&handle, cell_id) {
+        return err;
+    }
 
     let peer_label = server.get_peer_label().await;
     crate::presence::emit_focus(&handle, cell_id, &peer_label).await;
 
-    let deleted = handle
+    handle
         .delete_cell(cell_id)
         .map_err(|e| McpError::internal_error(format!("Failed to delete cell: {e}"), None))?;
 
-    if deleted {
-        let result = serde_json::json!({ "cell_id": cell_id, "deleted": true });
-        tool_success(&serde_json::to_string_pretty(&result).unwrap_or_default())
-    } else {
-        tool_error(&format!("Cell not found: {cell_id}"))
-    }
+    let result = serde_json::json!({ "cell_id": cell_id, "deleted": true });
+    tool_success(&serde_json::to_string_pretty(&result).unwrap_or_default())
 }
 
 /// Move a cell to a new position.
@@ -276,12 +273,9 @@ pub async fn move_cell(
 
     let after_cell_id = arg_str(request, "after_cell_id");
 
-    // Validate the cell being moved exists
-    if handle.get_cell(cell_id).is_none() {
-        return tool_error(&format!("Cell not found: {cell_id}"));
+    if let Some(err) = cell_not_found(&handle, cell_id) {
+        return err;
     }
-
-    // Validate the anchor cell exists (if specified)
     if let Some(anchor) = after_cell_id {
         if handle.get_cell(anchor).is_none() {
             return tool_error(&format!(
--- a/crates/runt-mcp/src/tools/cell_crud.rs
+++ b/crates/runt-mcp/src/tools/cell_crud.rs
@@ -78,6 +78,10 @@ pub struct ClearOutputsParams {
 }
 
 /// Valid cell types per the nbformat spec.
+///
+/// TODO: promote to a shared `CellType` enum in `notebook-doc` so
+/// deserialization rejects invalid types at the boundary rather than
+/// requiring runtime checks in each consumer.
 const VALID_CELL_TYPES: &[&str] = &["code", "markdown", "raw"];
 
 /// Create a new cell, optionally executing it.
--- a/crates/runt-mcp/src/tools/cell_crud.rs
+++ b/crates/runt-mcp/src/tools/cell_crud.rs
@@ -77,6 +77,9 @@ pub struct ClearOutputsParams {
     pub cell_ids: Option<Vec<String>>,
 }
 
+/// Valid cell types per the nbformat spec.
+const VALID_CELL_TYPES: &[&str] = &["code", "markdown", "raw"];
+
 /// Create a new cell, optionally executing it.
 pub async fn create_cell(
     server: &NteractMcp,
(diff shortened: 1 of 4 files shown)


crates/runt-mcp/src/tools/cell_crud.rs
```diff
@@ -77,13 +77,23 @@ pub struct ClearOutputsParams {
     pub cell_ids: Option<Vec<String>>,
 }
 
+/// Valid cell types per the nbformat spec.
+const VALID_CELL_TYPES: &[&str] = &["code", "markdown", "raw"];
```
rgbkrk, line 85: Can we be rejecting closer up in the spec? This should be part of an enum right?

Also these could be static strs right?

crates/runt-mcp/src/tools/cell_crud.rs
```diff
@@ -256,6 +272,20 @@ pub async fn move_cell(
 
     let after_cell_id = arg_str(request, "after_cell_id");
 
+    // Validate the cell being moved exists
+    if handle.get_cell(cell_id).is_none() {
+        return tool_error(&format!("Cell not found: {cell_id}"));
+    }
```
rgbkrk: Probably a good idea.

crates/runt-mcp/src/tools/cell_crud.rs
```diff
@@ -256,6 +272,20 @@ pub async fn move_cell(
...
+
+    // Validate the anchor cell exists (if specified)
+    if let Some(anchor) = after_cell_id {
+        if handle.get_cell(anchor).is_none() {
+            return tool_error(&format!(
+                "Anchor cell not found: {anchor}. Pass null/omit after_cell_id to move to the start."
```
rgbkrk: Doesn't null move it to the bottom of the notebook?

crates/runt-mcp/src/tools/deps.rs
```diff
@@ -189,6 +189,22 @@ pub async fn add_dependency(
...
+        if name.is_empty() {
+            return tool_error(
+                "Empty package name. Provide a valid package specifier (e.g. \"pandas>=2.0\").",
+            );
+        }
+        // Package names must start with a letter or digit (PEP 508 / conda).
+        if !name.chars().next().is_some_and(|c| c.is_alphanumeric()) {
+            return tool_error(&format!(
+                "Invalid package specifier: \"{name}\". Package names must start with a letter or digit."
+            ));
```
rgbkrk: The daemon also does validation. Do we consume it here?

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: Agreed — it should be an enum. Added a TODO for now since promoting it to `notebook-doc::CellType` touches deserialization across multiple consumers and felt like a separate PR. Removed the explicit `'static` — clippy flags it as redundant on constants.

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: Kept as-is — the existence check + validation catches bad IDs early with a clear MCP error instead of a cryptic Automerge failure.

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: Actually no — `compute_position(None)` calls `new_before(first)`, which inserts at the **start**. The message is correct: null/omit = move to beginning. To move to the end you'd pass the last cell's ID.

Ref: `notebook-doc/src/lib.rs` line 1782.

crates/runt-mcp/src/tools/deps.rs
quillaid: Good call. Removed the MCP-layer custom validation and added `validate_package_specifier()` to the UV path in `add_dep_for_manager`, matching how conda/pixi already use `validate_conda_package_specifier()`. Now all three package manager paths go through notebook-doc's canonical validators before touching the CRDT.

crates/runt-mcp/src/tools/deps.rs
```diff
@@ -145,9 +145,12 @@ pub(crate) fn add_dep_for_manager(
                 .add_pixi_dependency(package)
                 .map_err(|e| format!("Failed to add pixi dependency: {e}"))
         }
-        PackageManager::Uv | PackageManager::Unknown(_) => handle
-            .add_uv_dependency(package)
-            .map_err(|e| format!("Failed to add uv dependency: {e}")),
+        PackageManager::Uv | PackageManager::Unknown(_) => {
+            notebook_doc::metadata::validate_package_specifier(package)?;
```
rgbkrk, line 149: Does our outer handler make sure to translate errors to nice ones for LLMs to read?

crates/runt-mcp/src/tools/cell_crud.rs
rgbkrk: Oh sorry I was thinking this was create_cell for some reason. Hope we have good semantics there.

crates/runt-mcp/src/tools/cell_crud.rs
rgbkrk: Gosh are there more places we need to be doing this? We probably need a nice `cell_exists(cell_id)?` invariant we can call in more places.

crates/runt-mcp/src/tools/cell_crud.rs
rgbkrk: ok sure.

crates/runt-mcp/src/tools/deps.rs
quillaid: Good catch — the errors from `validate_package_specifier` were going through `McpError::internal_error` (JSON-RPC -32603), which gives the LLM an opaque protocol error. Changed to `tool_error` which returns `isError: true` with the validator's human-readable message text. The messages from `validate_package_specifier` are already clear ("invalid package name 'foo' ... Package names may only contain letters, digits, hyphens, underscores, and dots").

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: create_cell and move_cell have intentionally different defaults:

- **create_cell** with no `index` → appends to **end** (via `handle.last_cell_id()`)
- **move_cell** with `after_cell_id: null` → moves to **start** (via `compute_position(None)` → `new_before(first)`)

Both are the natural default for their use case: new cells go at the bottom where the user is working, while null-anchor move explicitly means "before everything."

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: Added `cell_not_found()` helper in `tools/mod.rs`:

```rust
pub fn cell_not_found(handle, cell_id) -> Option<Result<CallToolResult, McpError>>
```

Usage: `if let Some(err) = cell_not_found(&handle, cell_id) { return err; }`

Now used consistently in: `set_cell`, `delete_cell`, `move_cell`, `execute_cell`. The read-only tools (`get_cell`, editing tools, cell_meta) already had their own existence checks via `get_cell()`/`get_cell_source()`/`get_cell_metadata()` returning None — those are fine since the check is inherent in the read operation.

crates/runt-mcp/src/tools/mod.rs
```diff
@@ -473,6 +473,24 @@ pub fn arg_string_array(request: &CallToolRequestParams, key: &str) -> Option<Ve
...
+pub fn cell_not_found(
+    handle: &notebook_sync::handle::DocHandle,
+    cell_id: &str,
+) -> Option<Result<CallToolResult, McpError>> {
+    if handle.get_cell(cell_id).is_some() {
+        None
+    } else {
+        Some(tool_error(&format!("Cell not found: {cell_id}")))
+    }
+}
```
rgbkrk, line 495: I was thinking more like an `assert_cell_exists(handle, cell_id)?` kind of thing and then we handle it with some of our error handling at the tool level (?).

crates/runt-mcp/src/tools/cell_crud.rs
rgbkrk: Why'd you pick that instead of my cell_exists suggestions

crates/runt-mcp/src/tools/mod.rs
quillaid: Done — renamed to `assert_cell_exists` returning `Result<(), McpError>` using `McpError::invalid_params`, so all call sites are now just:

```rust
assert_cell_exists(&handle, cell_id)?;
```

Consistent with how missing required params already use `invalid_params`.

crates/runt-mcp/src/tools/cell_crud.rs
quillaid: Sorry about the confusion — my previous push still had the `cell_not_found` pattern when you saw it. It's now `assert_cell_exists(&handle, cell_id)?` everywhere, including the anchor cell check in move_cell:

```rust
assert_cell_exists(&handle, cell_id)?;
if let Some(anchor) = after_cell_id {
    assert_cell_exists(&handle, anchor)?;
}
```

All cell mutation tools use the same `?`-based guard now.

PR #2236, opened 2026-04-26
feat(app): banner for untyped kernel-launch failures

rgbkrk:
Visual polish on #2234. The daemon now writes `RuntimeLifecycle::Error` + `error_details` for every failed kernel launch, but the only surface was `NotebookToolbar`'s aria-label + title — stderr tails with five or six lines aren't going to read well in a `title` attribute.

## Component

`KernelLaunchErrorBanner` — destructive-red, alert icon, "Kernel failed to start" heading, a scrollable `<pre>` that preserves newlines / monospace (`max-h-32` so a pathological traceback doesn't eat the whole window), Retry + dismiss X. Icon uses the existing `AlertCircle` / `RotateCw` / `X` set for visual coherence with the other banners.

## App.tsx wiring

Renders above the toolbar, below `UntrustedBanner`. Gated on:

- `lifecycle.lifecycle === "Error"`
- `errorDetails` non-empty
- `errorReason` is **neither** `MissingIpykernel` (toolbar prompt has targeted install guidance) **nor** `CondaEnvYmlMissing` (has its own flow)

Dismissal state (`dismissedLaunchError: string | null`) is cleared in a `useEffect` whenever the kernel leaves Error, so the next failure from any retry path (banner button, toolbar restart, automatic relaunch) re-shows the banner fresh. The earlier draft had a bug where toolbar-initiated retries inherited a banner dismissal; effect-based clearing fixes it.

Retry wires to `tryStartKernel`.

## Test

`apps/notebook/src/components/__tests__/kernel-launch-error-banner.test.tsx` — four cases:

- Heading rendered
- Multi-line details preserved under the `<pre>` (RTL normalizes whitespace in its default matcher, so we read `textContent` directly)
- Retry callback
- Dismiss callback

## Test plan

- [x] `pnpm test:run apps/notebook/src/components/__tests__/kernel-launch-error-banner.test.tsx` — 4 pass
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `cargo xtask lint` — clean

Would pair naturally with manual QA once a repro is available (kill `nteract_kernel_launcher` from the launcher cache, launch any python notebook, banner should show the stderr tail).

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
-import { deriveEnvManager, deriveRuntimeKind, KERNEL_ERROR_REASON, NotebookClient } from "runtimed";
+import { deriveEnvManager, deriveRuntimeKind, NotebookClient } from "runtimed";
 import { IsolationTest } from "@/components/isolated";
 import { MediaProvider } from "@/components/outputs/media-provider";
 import { getCrdtCommWriter, setCrdtCommWriter } from "@/components/widgets/crdt-comm-writer";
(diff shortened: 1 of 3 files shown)


rgbkrk:
Good catches, both addressed in `35b8579e`.

**P2 — conda env.yml details.** Grepped the frontend: zero consumers of `CONDA_ENV_YML_MISSING` beyond the toolbar tooltip. Dropped the exclusion — env.yml missing-env failures now route through the banner and show the daemon's actionable details ("environment.yml declares conda env '<name>'. Run: conda env create -f environment.yml"). When a dedicated conda-env-yml UI lands, that reason can swap back out of the generic gate.

**P3 — Deno duplicate.** Added `runtime !== "deno"` so the toolbar's targeted amber "Deno not available" prompt stays the only Deno surface. No more stacked banners.

**Structure.** Factored the gate into `shouldShowKernelLaunchErrorBanner(params)` exported from `KernelLaunchErrorBanner.tsx`. App.tsx composes it with the dismissal check. Added 6 tests exhaustively covering the helper (lifecycle, empty details, `MissingIpykernel`, `CondaEnvYmlMissing` passing through, Deno skip) — 10 cases in the file now.

PR #2238, opened 2026-04-26
chore(notebook): kill save_notebook/has_notebook_path/verify_notebook_trust Tauri proxies

rgbkrk:
Three Tauri commands in `crates/notebook` were duplicating work the daemon already owns. Delete them and route callers to the authoritative path.

## What went away

| Command | Was | Is now |
|---|---|---|
| `save_notebook` | Tauri proxy for `NotebookRequest::SaveNotebook { format_cells: true }` | `host.transport.sendRequest` via new `NotebookClient.saveNotebook({ formatCells })` + `SaveNotebookError` |
| `has_notebook_path` | Read `context.path` from `WindowNotebookRegistry` | `runtimeState.path != null` from `useRuntimeState()` |
| `verify_notebook_trust` | Fetch metadata, re-run `runt_trust::verify_notebook_trust` in the Tauri process | `useTrust()` composes `TrustInfo` from `runtimeState.trust.status` + `useDependencies()` + `useCondaDependencies()` |

## Design decision

Trust *verification* moves to "read from daemon's CRDT" instead of "ask Tauri, which re-verifies locally against the daemon's metadata". The daemon already runs `runt_trust::verify_notebook_trust` internally on every metadata change and writes the result to `RuntimeStateDoc.trust.status`. The Tauri command was a redundant second pass. Trust *approval* (signing with the local HMAC key) stays Tauri-side for this PR — it moves to the daemon in a follow-up, at which point `runt-trust` can drop out of `crates/notebook`'s dep graph entirely.

The save-in-place path loses one round-trip (`has_notebook_path`) and reads path state from the CRDT it was already listening to.

## What stays

- `save_notebook_as` — has Tauri-side side effects (recent-menu refresh, kernel relaunch on path change).
- `clone_notebook_to_ephemeral` — opens a new window.
- `approve_notebook_trust` — HMAC signing client-side.
- `apply_path_changed` — real Tauri bookkeeping (session restore, window-title cache).

These go in phase 2 once the side effects have landing zones on the daemon side.

## Test plan

- [x] `cargo check -p notebook`
- [x] `cargo xtask lint`
- [x] `cargo xtask clippy`
- [x] `cargo test -p notebook --lib` (66/66)
- [x] `pnpm test:run` (1085/1085 + 3 skipped across 63 suites)
- [ ] Cmd+S on a saved notebook: silent save, no `has_notebook_path` round-trip
- [ ] Cmd+S on an untitled notebook: save dialog → daemon writes file → title updates
- [ ] Trust dialog: open a notebook with inline UV deps and no signature → dep lists + typosquat warnings render, approve button works
- [ ] Concurrent peer edits deps from an MCP agent → `runtimeState.trust.status` flips to `signature_invalid`, dialog re-renders without a Tauri round-trip

Design notes: `docs/superpowers/specs/2026-04-25-kill-notebook-mutation-proxies-phase-1.md`

### DIFF
--- a/apps/notebook/src/hooks/useAutomergeNotebook.ts
+++ b/apps/notebook/src/hooks/useAutomergeNotebook.ts
@@ -614,8 +614,12 @@ export function useAutomergeNotebook() {
   // ── File operations ────────────────────────────────────────────────
 
   const save = useCallback(async () => {
-    await saveNotebook(host, flushSync);
-  }, [host, flushSync]);
+    // `runtimeState.path` is the authoritative source: the daemon writes
+    // it on save / save-as / untitled promotion. Reading it here avoids
+    // a Tauri round-trip to the WindowNotebookRegistry.
+    const hasPath = runtimePath != null;
+    await saveNotebook(host, flushSync, hasPath);
+  }, [host, flushSync, runtimePath]);
 
   const openNotebook = useCallback(() => openNotebookFile(host), [host]);
 
--- a/apps/notebook/src/hooks/useTrust.ts
+++ b/apps/notebook/src/hooks/useTrust.ts
@@ -1,8 +1,10 @@
 import { useNotebookHost } from "@nteract/notebook-host";
 import type { TrustInfo, TyposquatWarning } from "@nteract/notebook-host";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { logger } from "../lib/logger";
-import { useRuntimeState } from "../lib/runtime-state";
+import { useRuntimeState, useRuntimeStateLoaded } from "../lib/runtime-state";
+import { useCondaDependencies } from "./useCondaDependencies";
+import { useDependencies } from "./useDependencies";
 
 export type { TrustInfo, TyposquatWarning };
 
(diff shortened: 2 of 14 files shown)


rgbkrk:
## Fixed P1: default trust state passing the launch gate

Verified the race the reviewer flagged: `DEFAULT_RUNTIME_STATE.trust.status` is `"no_dependencies"` (`packages/runtimed/src/runtime-state.ts:255`), which `tryStartKernel` reads as trusted (`App.tsx:596`). Before the first `RuntimeStateDoc` frame lands, a fast Start/Run would fire `LaunchKernel`, and the daemon marks trust trusted inside that handler (`crates/runtimed/src/requests/launch_kernel.rs:75`) — bypassing the approval dialog.

The old `verify_notebook_trust` Tauri command fetched metadata from the daemon async; when the daemon wasn't connected yet it threw, `checkTrust()` returned `null`, and every caller had an `if (!info) return false;` guard. Fail-closed.

**Fix:** track a `loaded` flag on the runtime-state store that flips true only when the daemon has pushed at least one snapshot, and reset on disconnect. `useTrust` holds `trustInfo` at `null` until then; `checkTrust()` returns that same null; `needsApproval`/`hasDependencies`/`totalDependencies` default to the safe side. The existing `if (!info) return false;` guards in `tryStartKernel`/`handleSyncDeps` restore fail-closed semantics.

Amended into the PR commit, push is `bd8bd450` → `71e24a3c`. Tests still green (1085/1085).

PR #2254, opened 2026-04-26
perf(sift): batch viewport cell reads

rgbkrk:
## Summary

- Add a `store_viewport_cells(handle, rows)` WASM export that returns a flat batch of display strings, numeric raw values, and null flags for requested viewport rows.
- Route Sift viewport prefetch through the batch export so visible rows are cached with one WASM call instead of per-cell `is_null`, `get_cell_string`, and `get_cell_f64` calls.
- Keep the existing per-cell methods as fallback paths for uncached access.
- Rebuild the WASM bindings and embedded Sift renderer plugin artifacts.

## Why

The table render path already prefetched visible rows, but the implementation still crossed the JS/WASM boundary for every visible cell. With overscan and wide dataframes, that means thousands of calls per render frame. The new export keeps the data formatting in Rust/WASM and amortizes the boundary crossing over the whole viewport window.

## Validation

- `cargo test -p sift-wasm`
- `pnpm --filter @nteract/sift exec vp test run`
- `pnpm --filter @nteract/sift exec playwright test e2e/notebook.spec.ts`
- `cargo xtask wasm`
- `cargo xtask verify-plugins`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook/src/renderer-plugins/sift.js
+++ b/apps/notebook/src/renderer-plugins/sift.js
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:cacc957f8557861b763e58b54e631a8d125694aaed8e033086ba8928b3c0c507
-size 388013
+oid sha256:e8ab7f94f7bd39379cbb2ca2a09b2244c2a21774f02abfa0fad70e79842680e7
+size 388218
--- a/crates/runt-mcp/assets/plugins/sift.js
+++ b/crates/runt-mcp/assets/plugins/sift.js
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:bf72b369d8c0abe3dfe42f8c29884a81cf967e30e4e9eebd6a30b9594c2edcd9
-size 388206
+oid sha256:29ae16c6391e878ba5fea0c2ca62ca1e33be721b65c05862096e4ad2ad89e4d3
+size 388411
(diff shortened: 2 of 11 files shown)


rgbkrk:
Perf measurement for the viewport-read change:

Method: local Node/WASM microbench on `packages/sift/public/data.arrow` (100,000 rows x 12 columns). It compares the old `prefetchViewport` behavior (`is_null` + `get_cell_string` + `get_cell_f64` per visible cell) against the new `store_viewport_cells` batch path, including the JS cache-shaping work both implementations do. 200 measured iterations per viewport size after warmup.

| viewport | cells | legacy p50 | batched p50 | p50 speedup | legacy p95 | batched p95 |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 30 rows | 360 | 0.200 ms | 0.134 ms | 1.49x | 0.260 ms | 0.203 ms |
| 60 rows | 720 | 0.381 ms | 0.239 ms | 1.59x | 0.515 ms | 0.316 ms |
| 100 rows | 1,200 | 0.596 ms | 0.373 ms | 1.60x | 0.749 ms | 0.462 ms |
| 160 rows | 1,920 | 0.999 ms | 0.617 ms | 1.62x | 1.129 ms | 0.703 ms |

This is intentionally scoped to the boundary we changed. It is not a full browser scroll trace, but it demonstrates the expected reduction in WASM boundary overhead for visible-cell hydration.

PR #2256, opened 2026-04-26
feat(notebook): gate kernel actions on first RuntimeStateSync

rgbkrk:
Follow-up to #2238. The trust-verification refactor left a small window where kernel-launch actions could fire against default (unloaded) trust state. #2238's `trustInfo === null` guard in `useTrust` closes the hole when handlers check trust — this PR closes it at the action-handler entry point so keyboard shortcuts and menu routes can't bypass it either.

## Shape

No new state, no new named observable. The engine already exposes `sessionStatus$` as a `ReplaySubject(1)` carrying `{ notebook_doc, runtime_state, initial_load }`. App subscribes to it via a new generic `useObservable(obs$, initial)` helper and derives:

```ts
const sessionStatus = useObservable<SessionStatus | null>(sessionStatus$, null);
const sessionReady = sessionStatus?.runtime_state === "ready";
```

`runtime_state === "ready"` means the daemon received and acked the first RuntimeStateSync round-trip. Before that, `runtimeState.trust.status` is the static default and must not be used to gate trust-relevant actions.

## Gated handlers

- `tryStartKernel` — launches or routes to trust dialog.
- `handleSyncDeps` — hot-sync or restart with package install.
- `handleStartKernelWithPyproject` — direct `launchKernel("python", "uv:pyproject")` path.
- `handleTrustApprove` post-success launch now routes through `tryStartKernel` rather than calling `launchKernel` directly.

All fail-closed with a debug log when `sessionReady` is false. Not gated: `handleExecuteCell` against an already-running kernel (no install, no trust implication), and `save`/`openNotebook`/`cloneNotebook` (unrelated to trust).

## Design choice

Exposed `sessionStatus$` as a stable proxy `Observable` off `useAutomergeNotebook` rather than adding a named `runtimeStateReady$: Observable<boolean>` on the engine. The state machine already exists — collapsing it to a bool at the engine boundary loses information (e.g., a future caller that also wants `initial_load === "ready"` would need another named observable). Per-consumer derivation keeps the engine surface lean and stays aligned with the "React components subscribe to engine observables" direction.

The `useObservable` helper is the payload: any future piece of UI state driven by an engine observable can be read with one line, no module-level store, and the component is testable by pushing values through a mock subject.

## Test plan

- [x] `cargo check -p notebook`
- [x] `cargo xtask lint` (clean after `--fix`)
- [x] `pnpm test:run` (1103/1103 + 3 skipped across 67 suites)
- [x] Added `use-observable.test.tsx` covering initial value, emission, unmount-unsubscribe
- [ ] Cmd+Return on a cell before first sync completes: handler no-ops with debug log (in Nightly logs, not visible)
- [ ] Cmd+Return after sync completes: executes normally
- [ ] Click "Start Kernel" button before sync: no launch fires
- [ ] Click "Sync Deps" before sync: no install fires
- [ ] Trust dialog approve → kernel launches

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
-import { deriveEnvManager, deriveRuntimeKind, NotebookClient } from "runtimed";
+import { deriveEnvManager, deriveRuntimeKind, NotebookClient, type SessionStatus } from "runtimed";
 import { IsolationTest } from "@/components/isolated";
 import { MediaProvider } from "@/components/outputs/media-provider";
 import { getCrdtCommWriter, setCrdtCommWriter } from "@/components/widgets/crdt-comm-writer";
@@ -55,6 +55,7 @@ import {
 } from "./lib/cell-ui-state";
 import { startCursorDispatch } from "./lib/cursor-registry";
 import { KERNEL_STATUS } from "./lib/kernel-status";
+import { useObservable } from "./lib/use-observable";
 import { logger } from "./lib/logger";
 import { getNotebookCellsSnapshot } from "./lib/notebook-cells";
 import { useDetectRuntime } from "./lib/notebook-metadata";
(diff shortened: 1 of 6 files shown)


rgbkrk:
## Fixed P1 + P2

**P1 — execute bypass.** `handleExecuteCell` and `handleRunAllCells` both called through to the daemon's `execute_cell` / `run_all_cells` even if `tryStartKernel` returned false without a trust-dialog reason. Against a runtime agent that survived from a prior session, that would queue into RuntimeStateDoc before first sync. Both handlers now fail-closed at entry on `!sessionReady`.

**P2 — stale ready across rebootstrap.** `SyncEngine.resetForBootstrap()` cleared only its private `latestSessionStatus` field; the `ReplaySubject(1)` kept its old cached value. Added an emission of `{ notebook_doc: "pending", runtime_state: "pending", initial_load: { phase: "not_needed" } }` at the end of `resetForBootstrap` so downstream consumers flip back to non-ready immediately.

New test covers the rebootstrap case: `resetForBootstrap emits a pending status so stale ready doesn't leak across reconnect`. Asserts both live subscribers and fresh subscribers see the pending value after the reset.

Amended into the PR commit, push is `de232d7a` → `01eab071`. 1104/1104 frontend tests pass.

rgbkrk:
## Fixed P2 — restart paths

`handleRestartKernel` and `restartAndRunAll` both called `shutdownKernel` before any `sessionReady` check. That hits the runtime-agent request path and writes `RuntimeLifecycle::Shutdown` even though we haven't confirmed trust. Both now fail-closed at entry on `!sessionReady`.

`handleSyncDeps`'s fallback `shutdownKernel` is already inside its own `sessionReady` gate, so it's safe.

Amended, push `01eab071` → `d318201d`. 1104/1104 tests still green.

PR #2259, opened 2026-04-26
fix(sift): trap wheel scroll in table viewport

rgbkrk:
## Summary
- Trap wheel scrolling inside the Sift viewport when the table can scroll on that axis
- Manually clamp vertical and horizontal wheel deltas so leftover momentum does not chain into notebook/page scroll
- Let wheel events pass through for non-scrollable tables
- Add unit and notebook-gallery E2E coverage for boundary behavior
- Rebuild renderer plugin artifacts with `cargo xtask wasm`

## Validation
- `pnpm --filter @nteract/sift exec vp test run`
- `pnpm --filter @nteract/sift exec playwright test e2e/notebook.spec.ts`
- `pnpm --filter @nteract/sift exec playwright test e2e/table.spec.ts e2e/header-overlap.spec.ts`
- `cargo xtask wasm`
- `cargo xtask verify-plugins`
- `cargo xtask lint --fix`

## Audit note
This is draft because the behavior still needs product feel testing in the notebook/Tauri path, and the renderer plugin dev loop is still too easy to get stale. I am auditing that separately in-thread.


rgbkrk:
## Plugin/dev-surface audit

This PR is staying draft because the wheel-trap experiment still does not fully solve fling behavior. While looking at the surrounding build/runtime paths, I found a few plugin-management issues that are probably worth pulling out into follow-up work.

### Current architecture

- The notebook app uses Vite virtual modules for isolated renderer assets. In dev, `apps/notebook/vite-plugin-isolated-renderer.ts` rebuilds the isolated renderer and plugins in-memory from source, then full-reloads Vite. Production falls back to checked-in artifacts under `apps/notebook/src/renderer-plugins/`.
- MCP Apps are different. `apps/mcp-app` ships a self-contained `output.html`, then lazy-loads renderer plugins from the daemon using `<script src="{blob_base_url}/plugins/{name}.js">` and optional CSS links. Those JS/CSS/wasm files are embedded into the daemon binary via `include_bytes!` from `crates/runt-mcp/assets/plugins/`.
- Sift is the sharpest edge: source lives in `packages/sift`, wasm glue comes from `crates/sift-wasm/pkg`, notebook plugin artifacts are raw CJS, MCP plugin artifacts are wrapped for `window.__nteract`, and MCP wasm is copied separately to `/plugins/sift_wasm.wasm`.

### Concrete stale-build risk

`crates/runtimed/build.rs` only declares `rerun-if-changed` for markdown/plotly/vega/leaflet plugin assets. It omits `sift.js`, `sift.css`, and `sift_wasm.wasm`, even though `crates/runtimed/src/embedded_plugins.rs` embeds all three. That means a Sift-only plugin rebuild can leave cargo believing `runtimed` is fresh, so the daemon can keep serving stale embedded Sift assets until some other Rust/build input invalidates it.

That should be fixed independently of the scroll experiment.

### Dev ergonomics gaps

- `cargo xtask wasm` correctly rebuilds sift-wasm, copies the MCP wasm, and chains `renderer-plugins`; `verify-plugins` checks wasm-bindgen import/export coherence. That catches symbol drift, but it does not guarantee the running daemon is serving the rebuilt assets.
- The daemon serves `/plugins/*` from embedded bytes with `Cache-Control: public, max-age=86400`. That is right-ish for release, but hostile for dev unless URLs are versioned or dev mode sends no-store/cache-busted URLs.
- `apps/mcp-app/build-plugins.ts` is a second plugin build path and explicitly excludes Sift. The canonical path is `scripts/build-renderer-plugins.ts`, but the split makes it easy to update “MCP plugins” and accidentally not update the one plugin we are actively debugging.
- MCP app dev preview uses the real MCP Apps bridge in a browser, which is good, but the current fixture set does not exercise daemon-served plugins. `makeToolResult` supports `blob_base_url`, but the preview calls do not pass one, so Sift/Parquet does not run through the same `/plugins/*` path agents/users see in Claude Desktop.
- The notebook gallery exists as a real-browser surface, but today it is mostly component/settings coverage. It is valuable precisely because it avoids Tauri hooks, so it should stay Tauri-free and grow an output/plugin gallery surface rather than making agents drive the Tauri shell for renderer iteration.

### Suggested follow-ups

1. Add the missing `sift.js`, `sift.css`, and `sift_wasm.wasm` watch entries to `crates/runtimed/build.rs`.
2. Add a dev-mode plugin asset overlay in the daemon: under `RUNTIMED_DEV`, serve `/plugins/*` from the workspace filesystem instead of embedded bytes when available. Keep embedded bytes for release.
3. In dev mode, either send `Cache-Control: no-store` for `/plugins/*` or append a content hash/mtime query when MCP Apps loads plugin URLs.
4. Collapse plugin builds onto one xtask-owned entrypoint. Ideally `cargo xtask renderer-plugins --target notebook|mcp|all --plugin sift` or similar, with `apps/mcp-app/build-plugins.ts` removed or reduced to a thin wrapper.
5. Add an MCP App dev fixture for Parquet/Sift that passes a real or local fixture `blob_base_url`. Then add Playwright coverage against `apps/mcp-app` in a normal browser for Sift popover CSS, wasm loading, scroll behavior, and plugin failure states.
6. Expand `apps/notebook/gallery/` from component gallery into a renderer/plugin gallery: a browser-only page that can host isolated output frames, Sift fixtures, and mocked blob/plugin endpoints. Add a boundary check so gallery code cannot transitively import Tauri APIs.
7. Add a high-level `cargo xtask plugin-dev sift` command that runs the right sequence: build wasm when needed, rebuild renderer plugins, verify plugin/wasm coherence, and either refresh the filesystem-served dev assets or tell the user exactly which daemon process must restart.

The goal I would aim for: notebook app and MCP Apps can still have different runtime wrappers, but source-to-visible-output should be a single, inspectable loop. Agents should be able to open a browser gallery/dev preview, see the same plugin bundle path used by MCP Apps, and know whether the daemon is serving embedded release assets or live dev assets.

rgbkrk:
Closing this direction. The Sift-local wheel trap still does not fully prevent notebook scroll fling, so we are moving to a notebook-level output well / click-to-interact model on a fresh branch.

Issue #2263, opened 2026-04-26
feat(runt-mcp): get_results should survive room reconnects


quillaid:
## Problem

`get_results(execution_id)` fails with "Execution not found" after a session drop. The execution IDs returned by `execute_cell` become dangling references when the room is evicted (30s after last peer disconnects). Agents waste turns retrying before falling back to `get_cell`.

## Evidence

Gremlin testing surfaced this in two independent suites:
- **Refiner** (DS suite): Hit 3 consecutive eviction errors, wasted 6 turns + a major replan cycle.
- **Breaker** (AX suite): Also hit it during reconnect-heavy scenarios.

As MCP adoption grows and agents rely more heavily on `execution_id` tracking (especially after #2152), this is projected to cause ~14% efficiency loss in multi-step agent workflows.

## Why it matters

`get_results` is the only tool that ties output to a **specific execution** — critical for:
- **Ghost-output detection** (#2086): distinguishing stale outputs from fresh ones requires execution-level tracking.
- **Async execution polling**: agents that `execute_cell` then poll `get_results` later need the ID to remain valid across reconnects.

`get_cell` works as a fallback but loses execution-level tracking — it returns the cell's *current* outputs regardless of which execution produced them.

## Current architecture

The execution lifecycle has a persistence gap:

| State | Location | Survives room eviction? |
|-------|----------|------------------------|
| `ExecutionState` (status, outputs, source, seq) | `RuntimeStateDoc` (in-memory Automerge doc) | **No** — dropped entirely |
| Cell's `execution_id` pointer | `NotebookDoc` CRDT (persisted) | **Yes** — saved to `.ipynb` and `.automerge` |
| Output content (blobs) | Content-addressed blob store on disk | **Yes** |
| Execution count | Written to `.ipynb` on save | **Yes** |

**The key disconnect:** the cell's `execution_id` pointer survives, but the `ExecutionState` it points to doesn't — creating a dangling reference after room eviction.

**On reload** (`crates/runtimed/src/notebook_sync_server/load.rs:493+`): synthetic execution IDs are generated and overwrite the originals, so even the persisted pointer is lost.

Key files:
- `crates/runtime-doc/src/doc.rs:1060-1165` — `ExecutionState` CRUD (`create_execution`, `set_execution_running/done/count`)
- `crates/notebook-doc/src/lib.rs:1371-1395` — `set_execution_id` / `get_execution_id` (persisted CRDT pointer)
[...]

PR #2267, opened 2026-04-26
feat(runt-mcp): persist terminal execution results

rgbkrk:
## Summary
- add a daemon-owned durable execution store for terminal execution snapshots
- expose execution_store_dir through daemon info and wire runt-mcp to read it
- make get_results(execution_id) fall back to durable records after room eviction or reconnect
- reuse matching durable execution IDs when loading saved notebook outputs
- mark durable execution blob refs during GC and prune records on the existing 30-day retention window

Refs #2263

## Verification
- cargo fmt
- git diff --check
- cargo test -p runtimed-client execution_store --lib
- cargo test -p runt-mcp get_results_ --lib
- cargo test -p runtimed durable --lib
- cargo test -p runtimed execution_store_gc --lib
- cargo check -p runt

### DIFF
--- a/crates/runt-mcp/src/tools/execution.rs
+++ b/crates/runt-mcp/src/tools/execution.rs
@@ -316,22 +316,60 @@ pub async fn get_results(
     })?;
     let full_output = arg_bool(request, "full_output").unwrap_or(false);
 
-    let handle = require_handle!(server);
-
-    let runtime_state = handle.get_runtime_state().map_err(|_| {
-        McpError::internal_error("Failed to read RuntimeStateDoc".to_string(), None)
-    })?;
-
-    let exec = match runtime_state.executions.get(execution_id) {
-        Some(e) => e,
-        None => {
-            return tool_error(&format!(
-                "Execution not found: {execution_id}. \
-                 It may have been evicted when the notebook room closed."
-            ));
-        }
+    let handle = {
+        let guard = server.session.read().await;
+        guard.as_ref().map(|session| session.handle.clone())
     };
 
+    if let Some(handle) = handle.as_ref() {
+        let runtime_state = handle.get_runtime_state().map_err(|_| {
+            McpError::internal_error("Failed to read RuntimeStateDoc".to_string(), None)
+        })?;
+        if let Some(exec) = runtime_state.executions.get(execution_id) {
+            let cell = handle.get_cell(&exec.cell_id);
+            return render_execution_result(
+                server,
+                execution_id,
+                exec,
+                Some(&runtime_state.comms),
+                cell,
+                full_output,
+            )
+            .await;
+        }
+    }
+
+    let store =
+        runtimed_client::execution_store::ExecutionStore::new(server.execution_store_path.clone());
+    if let Some(record) = store.read_record(execution_id).await {
+        let exec = runtime_doc::ExecutionState {
+            cell_id: record.cell_id.clone().unwrap_or_default(),
+            status: record.status,
+            execution_count: record.execution_count,
+            success: record.success,
+            outputs: record.outputs,
+            source: record.source,
+            seq: record.seq,
+        };
+        let cell = handle
+            .as_ref()
+            .and_then(|handle| handle.get_cell(&exec.cell_id));
+        return render_execution_result(server, execution_id, &exec, None, cell, full_output).await;
+    }
+
+    tool_error(&format!(
+        "Execution not found: {execution_id}. It may have been evicted and no durable result record was found."
+    ))
+}
+
+async fn render_execution_result(
+    server: &NteractMcp,
+    execution_id: &str,
+    exec: &runtime_doc::ExecutionState,
+    comms: Option<&std::collections::HashMap<String, runtime_doc::CommDocEntry>>,
+    cell: Option<notebook_doc::CellSnapshot>,
+    full_output: bool,
+) -> Result<CallToolResult, McpError> {
     // Determine display status with clear indication of completeness
     let (display_status, is_terminal) = match exec.status.as_str() {
         "done" => ("done", true),
@@ -354,7 +392,6 @@ pub async fn get_results(
     );
 
     // Resolve outputs from the execution's manifests
-    let comms = Some(&runtime_state.comms);
     let outputs = if !exec.outputs.is_empty() {
         output_resolver::resolve_cell_outputs_for_llm(
             &exec.outputs,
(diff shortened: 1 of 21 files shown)


crates/runtimed/src/notebook_sync_server/load.rs
```diff
@@ -490,17 +491,37 @@ where
...
-                    let _ = sd.set_execution_done(&synthetic_eid, true);
-                    cell_eids.insert(cell.id.clone(), synthetic_eid);
+                if let Some(eid) = cell_eids.get(&cell.id) {
+                    let _ = sd.create_execution(eid, &cell.id);
+                    let _ = sd.set_outputs(eid, output_refs);
+                    if let Ok(ec) = cell.execution_count.parse::<i64>() {
+                        let _ = sd.set_execution_count(eid, ec);
+                    }
+                    let _ = sd.set_execution_done(eid, true);
                 }
```
rgbkrk, line 532: **[normal] Reload discards durable record's error status, forcing reused executions to status=done**

Reload paths in `streaming_load_cells` and `load_notebook_from_disk_with_state_doc_and_execution_store` unconditionally call `sd.set_execution_done(eid, true)` after reusing a durable record's execution_id, clobbering `status="error"`/`success=false` records back to `status="done"`/`success=true` in RuntimeStateDoc. The result is that the same `get_results(execution_id)` call returns opposite answers depending on whether an MCP session is connected: the live-state path sees the clobbered "done" while the durable-store fallback still sees "error". Fix by passing the durable record's success bit through (e.g. `set_execution_done(eid, record.success.unwrap_or(true))`) or skipping `set_execution_done` entirely when the eid was reused from a durable record.

crates/runt-mcp/src/tools/execution.rs
```diff
@@ -316,22 +316,60 @@ pub async fn get_results(
...
+            execution_count: record.execution_count,
+            success: record.success,
+            outputs: record.outputs,
+            source: record.source,
+            seq: record.seq,
+        };
+        let cell = handle
+            .as_ref()
+            .and_then(|handle| handle.get_cell(&exec.cell_id));
+        return render_execution_result(server, execution_id, &exec, None, cell, full_output).await;
```
rgbkrk: **[normal] get_results renders durable record with cell snapshot from a different notebook**

In `get_results`'s durable-store fallback, the cell snapshot is looked up via the *active session's* handle (`handle.get_cell(&exec.cell_id)`), but the durable record came from a potentially different notebook. Cell IDs are only unique within a notebook room, not across them. If the active session contains a cell with a colliding ID (hand-edited .ipynb files, copy-pasted cells, descriptive IDs like `cell-1` / `imports`), the response will attach the **wrong source code** from notebook B to the **correct outputs / execution_id** from notebook A. The fix is a one-liner: pass `None` in the durable branch so `render_execution_result` uses `fallback_cell`, which carries the trustworthy `exec.source` from the durable record.

crates/runtimed/src/notebook_sync_server/load.rs
```diff
@@ -490,17 +491,37 @@ where
...
+                execution_store,
+                &context_id,
+                Some(&context_id),
+                &cell.id,
+                &cell.source,
+                parsed_ec,
+                output_refs,
+            )
+            .await;
+            cell_eids.insert(cell.id.clone(), eid);
```
rgbkrk: **[normal] Streaming load uses raw user path for context_id, mismatching the canonical path persist uses**

Streaming reload mints synthetic execution_ids on every non-canonical path open, silently defeating this PR's durable execution-id feature. `streaming_load_cells` derives `context_id` from the raw `path` parameter (load.rs:498), but `persist_terminal_execution_records` derives it from `room.identity.path` (peer.rs:2158-2165), which the daemon stores as the canonical path. Since `ExecutionStore::find_matching_notebook_cell` requires exact string equality on both `context_id` and `notebook_path`, every reload via a non-canonical path (`./foo.ipynb`, `/abs/./foo.ipynb`, symlinks like `/tmp` vs `/private/tmp`, unexpanded `~/`) silently misses every persisted record and falls through to `uuid::Uuid::new_v4()`. Fix by canonicalizing `path_buf` before passing it into `needs_load` (daemon.rs:2137), or by reading `room.identity.path` inside `streaming_load_cells` to build the context_id.

rgbkrk: **[normal] Notebook load does O(cells x records) directory scans of the execution store**

Opening a notebook now triggers C x R disk reads, where C is the cell count and R is the number of records in the durable execution store. `streaming_load_cells` calls `durable_or_synthetic_execution_id` once per cell with outputs, which routes through `find_matching_notebook_cell` -> `list_context` -> `list_records`, and `list_records` does a full `read_dir` plus `read` + `serde_json::from_slice` of every `.json` in the store on every call. Fix by listing the store once per load (filtered to `("notebook", context_id)`) and matching in memory, or by indexing records on disk by `(context_id, cell_id)`.

crates/runtimed/src/notebook_sync_server/peer.rs
```diff
@@ -2123,3 +2150,58 @@ async fn send_doc_sync<W: tokio::io::AsyncWrite + Unpin>(
...
+    }
+}
+
+pub(crate) fn notebook_execution_context_id(
+    room: &NotebookRoom,
+    notebook_path: Option<&str>,
+) -> String {
+    notebook_path
+        .map(str::to_string)
+        .unwrap_or_else(|| room.id.to_string())
```
rgbkrk, line 2231: **[normal] Persist rewrites all terminal executions on every RuntimeStateSync frame**

On every state-changing RuntimeStateSync frame (and on every kernel runtime-state phase transition), `persist_terminal_execution_records` walks every terminal execution in `RuntimeStateDoc` and unconditionally calls `store.write_record` on each one. `write_record` does a full read+serialize+tmp-write+rename per record, so for a notebook with N completed cells every state sync performs ~3N filesystem operations regardless of which execution(s) actually changed. Sync frames are produced frequently during active execution (status changes, output streaming, comm activity), and N grows monotonically across a session, so the cost grows superlinearly with notebook activity. The unconditional `record.updated_at = Utc::now()` on each write also defeats `prune_older_than` GC because every record's timestamp is refreshed on every sync. Fix by tracking a per-execution-id last-persisted head (or hashing the record / diffing against on-disk state) and only writing records that actually changed since the previous persist.

crates/runtimed-client/src/execution_store.rs
```diff
@@ -0,0 +1,456 @@
...
+        let path = self.path_for_id(&record.execution_id)?;
+        let tmp = self.root.join(format!(
+            "{}.{}.tmp",
+            record.execution_id,
+            uuid::Uuid::new_v4()
+        ));
+        let bytes = serde_json::to_vec_pretty(&record).map_err(std::io::Error::other)?;
+        tokio::fs::write(&tmp, bytes).await?;
+        tokio::fs::rename(&tmp, &path).await?;
+        Ok(())
```
rgbkrk, line 145: **[nit] ExecutionStore tmp files leak permanently if rename fails or daemon crashes**

Tmp files written by `ExecutionStore::write_record` (`{execution_id}.{uuid}.tmp`) can leak permanently if the daemon is killed between the `tokio::fs::write` and `tokio::fs::rename` calls, or if the rename itself fails. `list_records` filters by `.json` extension so reads aren't polluted, but `prune_older_than` also iterates the .json list, so leaks accumulate indefinitely with no recovery path. Worth a hygiene cleanup: either sweep `*.tmp` on `ExecutionStore::new`, or drop tmp files into a sibling `tmp/` subdirectory that is wiped at startup.

crates/runtimed-client/src/execution_store.rs
```diff
@@ -0,0 +1,456 @@
...
+                            "[execution-store] Failed to prune execution record {:?}: {}",
+                            path,
+                            e
+                        );
+                    }
+                }
+            }
+        }
+        removed
+    }
```
rgbkrk, line 267: **[nit] prune_older_than has TOCTOU race that can delete just-refreshed execution records**

`prune_older_than` checks `record.updated_at` against the cutoff using a snapshot from `list_records()`, then deletes by path without re-stat'ing. Between the two, a concurrent `persist_terminal_execution_records` can atomically rewrite that same file via tmp+rename, and prune ends up unlinking the freshly-written record. The window is narrow (microseconds, and only reachable when a room with records older than `RUNTIMED_BLOB_GC_GRACE_SECS` receives a state-sync mid-prune), and the impact is self-healing (next state sync re-persists), so this is a defensive nit rather than a release blocker. Cheap fix: re-read `updated_at` after stat before unlink, or skip records whose execution_id is in any active room's `RuntimeStateDoc`.

rgbkrk (commented):
Pre-merge review from a cloud ultrareview pass. Seven findings, five normal severity and two nit.

Probably-want-before-merge (correctness):
- `bug_009` (load.rs:515): reload clobbers `status="error"` durable records back to `status="done"`. Same `get_results` call returns opposite answers depending on whether an MCP session is connected.
- `bug_006` (execution.rs:354): durable fallback can attach source from notebook B to outputs from notebook A when cell IDs collide.
- `bug_008` (load.rs:494): canonical-path mismatch between write side and read side; reload via `./foo.ipynb` or macOS `/tmp` symlinks silently mints fresh execution_ids and orphans every durable record.

Fine-to-fix-as-followups (perf, GC, hygiene):
- `bug_002` (load.rs:494): O(cells x records) disk scans on every OpenNotebook.
- `bug_001` (peer.rs:2185): every state-sync frame rewrites all terminal records and bumps their `updated_at`, defeating `prune_older_than` GC.
- `bug_004` (execution_store.rs:110): `*.tmp` files leak on SIGKILL/rename failure with no reaper.
- `bug_010` (execution_store.rs:218): TOCTOU between prune snapshot and unlink can delete a just-refreshed record (microsecond window, self-healing on next sync).

rgbkrk:
All seven findings from the previous review look addressed in `a26b8a77`. Quick rundown of what landed:

**Correctness (the load-bearing ones):**
- `bug_009`: `LoadedExecution { execution_id, success }` carries the durable record's `terminal_success()` through; `set_execution_done(eid, execution.success)` replaces the unconditional `true`. Status preserved on reload.
- `bug_006`: `execution.rs:354` now passes `None` for the cell snapshot, so the durable branch uses `fallback_cell` from `exec.source` rather than cross-notebook lookup.
- `bug_008`: `streaming_load_cells` derives `context_id` and `notebook_path` from `room.identity.path` (canonical), matching the persist side. The other path that still uses raw `path` is `#[cfg(test)]`, so production is clean.

**Perf / GC / hygiene:**
- `bug_002`: `durable_execution_records()` lists once per load and matching happens in memory. Option 1 from the comment.
- `bug_001`: two-layer fix. Per-connection `persisted_execution_records` HashMap short-circuits unchanged executions before any disk hit, and `write_record` itself bails early via `payload_matches` so `updated_at` doesn't tick on a no-op write. `unchanged_write_preserves_updated_at` test locks it in.
- `bug_004`: `cleanup_tmp_files_sync` runs in `ExecutionStore::new`. `new_removes_stale_tmp_files` test added.
- `bug_010`: `prune_older_than` re-reads the record before unlinking and bails if `updated_at` is no longer past the cutoff. `prune_rereads_before_delete` test added.

Nothing else flagged. LGTM.

PR #2277, opened 2026-04-27
fix(notebook-sync): replace blocking confirm_sync with passive waiters

quillaid:
## Summary

Parallel MCP tool calls (8+ concurrent `create_cell`) deterministically dropped notebook sessions. Root cause: `confirm_sync_impl` monopolized the single-threaded sync task select loop for up to 10s (5 rounds × 2s), starving inbound frame processing. The daemon's write buffer filled → `send_typed_frame` error → peer disconnected.

Three changes fix the bottleneck:

- **Replace blocking `confirm_sync_impl` with passive `SyncWaiter`s.** Each `ConfirmSync` command registers target heads + reply oneshot. Waiters resolve when the Frame arm processes `AutomergeSync` frames that advance `shared_heads`. N concurrent calls → N waiters resolved by O(1) incoming frames.
- **Flip biased select priority: Frame > Command > Changed.** Incoming daemon frames are consumed first, preventing socket backpressure. The old `Changed > Command > Frame` order starved frame processing during mutation bursts.
- **Remove the 500ms ack wait after local changes.** The daemon's ack arrives in the Frame arm. The inline wait was a secondary bottleneck during mutation bursts.

Also drops unnecessary `confirm_sync` calls in `create_cell` and `set_cell` — non-executing CRDT mutations don't need sync confirmation (the sync task propagates changes asynchronously; `execute_and_wait` calls `confirm_sync` before execution requests where the daemon actually needs the cell source).

## Evidence

| Metric | Before | After |
|--------|--------|-------|
| Session drops (notebook-stress, 8-10 concurrent create_cell) | 33 (100% repro) | 0 expected |
| Session drops (30-gremlin bug bash) | 84 total (2.8/gremlin avg) | Near-zero expected |
| confirm_sync worst-case latency | 10s (5 rounds × 2s) | ~1 round-trip (ms) |
| Select loop availability during confirm_sync | 0% (blocked) | 100% (non-blocking) |

## Files changed

- `crates/notebook-sync/src/sync_task.rs` — Core fix: `SyncWaiter`, `resolve_sync_waiters`, select priority flip, 500ms wait removal
- `crates/notebook-sync/src/handle.rs` — `confirm_sync()` timeout (5s best-effort, matching old 5×2s budget)
- `crates/runt-mcp/src/tools/cell_crud.rs` — Drop redundant `confirm_sync` in `create_cell`/`set_cell`

## Test plan

- [x] All 41 existing `notebook-sync` unit tests pass
- [x] Release build compiles and installs as nightly
- [x] Nightly daemon starts and reports correct version
- [ ] Gremlin replay: notebook-stress suite with 8-10 concurrent `create_cell` batches
- [ ] Gremlin replay: full 30-gremlin bug bash — session drop count should be near zero
- [ ] Manual: `execute_cell` after `create_cell` still works (confirm_sync in execute path preserved)
- [ ] Manual: `add_dependency` + `sync_environment` still works (confirm_sync in deps path preserved)

### DIFF
--- a/crates/notebook-sync/src/handle.rs
+++ b/crates/notebook-sync/src/handle.rs
@@ -575,16 +575,33 @@ impl DocHandle {
 
     /// Confirm that the daemon has merged all our local changes.
     ///
-    /// Performs up to 5 sync round-trips, checking that the daemon's
-    /// shared_heads include our local heads. Call this before executing
-    /// a cell to ensure the daemon has the cell's source.
+    /// Registers a non-blocking waiter in the sync task. The waiter resolves
+    /// when incoming sync frames advance `shared_heads` past our current doc
+    /// heads. This is typically a single round-trip (milliseconds).
+    ///
+    /// Times out after 5s — if the daemon hasn't acked by then, returns Ok
+    /// anyway (best-effort, matching the old behavior of continuing after
+    /// 5 rounds of 2s each).
     pub async fn confirm_sync(&self) -> Result<(), SyncError> {
         let (reply_tx, reply_rx) = oneshot::channel();
         self.cmd_tx
             .send(SyncCommand::ConfirmSync { reply: reply_tx })
             .await
             .map_err(|_| SyncError::Disconnected)?;
-        reply_rx.await.map_err(|_| SyncError::Disconnected)?
+        match tokio::time::timeout(std::time::Duration::from_secs(5), reply_rx).await {
+            Ok(Ok(result)) => result,
+            Ok(Err(_)) => Err(SyncError::Disconnected),
+            Err(_) => {
+                // Timeout — best-effort, likely synced. The waiter will be
+                // pruned from the sync task's list on the next check (its
+                // receiver is dropped).
+                debug!(
+                    "[doc-handle] confirm_sync timed out for {} (best-effort OK)",
+                    self.notebook_id
+                );
+                Ok(())
+            }
+        }
     }
 
     /// Flush pending RuntimeStateDoc sync frames from the daemon.
--- a/crates/notebook-sync/src/sync_task.rs
+++ b/crates/notebook-sync/src/sync_task.rs
@@ -115,6 +115,14 @@ pub enum SyncCommand {
     },
 }
 
+/// Interval for the maintenance tick that re-drives sync for pending waiters.
+///
+/// This is the safety-net retry: after an inbound `AutomergeSync` frame clears
+/// Automerge's in_flight state, `generate_sync_message()` may now succeed for
+/// changes that were previously blocked. The primary re-drive happens inline in
+/// the Frame arm; this tick catches edge cases (e.g. no further inbound frames).
+const MAINTENANCE_TICK: Duration = Duration::from_millis(200);
+
 /// A pending `ConfirmSync` waiter.
 ///
 /// Instead of blocking the select loop for up to 10s (5 rounds × 2s),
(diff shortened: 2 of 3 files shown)


quillaid (commented):
## AX / Gremlin Testing Review

I run the automated gremlin test suite against nightly builds. Here's the AX perspective on this fix, grounded in concrete measurement data.

### The problem this fixes — measured impact

The `notebook-stress` gremlin creates 50 cells in batches of 8-10 concurrent `create_cell` calls. On the pre-fix build (2.3.4+25d8ef1), **every single batch** dropped the MCP session. 100% repro rate:

- **33 session drops** in one gremlin run
- **126 turns** consumed (90 budget, went over) — 58 extra turns spent on reconnect/recovery
- **$1.08 cost** for what should be a $0.68 test
- **611s elapsed** vs 216s expected

This wasn't just notebook-stress. Across a 35-gremlin bug bash (4 waves), we measured **88 total session drops** (2.5/gremlin avg). The deterministic parallel-call drops from notebook-stress accounted for 37.5% of all drops.

### Validation of this fix

I validated the exact build produced by this PR branch:

**Solo notebook-stress:** 33 → **0 session drops**. 126→68 turns, $1.08→$0.68, 611s→216s. Every batch of 8-10 concurrent `create_cell` calls completed without a single session drop. All 50 cells created, executed via `run_all_cells`, edited, saved — zero errors.

**7-gremlin AX suite (concurrent, batch-size 6):** Session drops across the suite went from **50 → 22** (56% reduction). Key improvements:
- `save-roundtrip`: 3 drops → 0, now achieves VERDICT: PASS (previously couldn't complete)
- `data-scientist`: 5→2 drops, 48→34 turns ($0.58 saved per run)
- `notebook-stress` concurrent: 33→14 (deterministic drops gone; residual 14 are ambient drops from 6 other gremlins sharing the daemon)

### What this PR covers well from an AX perspective

1. **The passive `SyncWaiter` pattern** eliminates the head-of-line blocking that caused deterministic drops. N concurrent confirm_sync calls → N waiters resolved by O(1) incoming frames. This directly maps to the agent's parallel tool call pattern.

2. **Flipped select priority (Frame > Command > Changed)** keeps the daemon's write buffer drained. This is the correct priority for agent workloads where the daemon is chatty (streaming outputs, state sync, broadcasts).

3. **Removing `confirm_sync` from `create_cell`/`set_cell`** is a significant AX win. Non-executing CRDT mutations don't need sync confirmation — the sync task propagates changes asynchronously. This reduces per-cell latency for the most common operations.

4. **The 5s best-effort timeout** on `confirm_sync` preserves the existing contract while being non-blocking. Agents already handle stale-doc scenarios via `get_cell` verification.

### What this PR doesn't cover (not blockers)

1. **Ambient session drops (~2-3/gremlin)** persist under concurrent daemon load. These are a separate issue from the parallel-tool-call bottleneck. The fix correctly targets the deterministic trigger without claiming to solve all session drops.

2. **Request-ID routing for overlapping daemon requests** isn't in scope here. This matters for long-running requests (LaunchKernel) where broadcasts interleave with the response. PR #2279 addresses this, but it's orthogonal to the session-drop fix.

3. **The `confirm_sync` → `confirm_state_sync` fix in `cell_read.rs`** (present in #2279 but not here). When `get_cell` has an empty output but a non-empty execution count, it calls `confirm_sync()` to flush — but it actually needs RuntimeStateDoc frames, not notebook-doc sync. Minor but worth cherry-picking.

### Verdict

The gremlin data is unambiguous: this fix eliminates the deterministic session drops that were the single biggest AX cost in the nightly build. The change surface is minimal (3 files, focused on sync_task.rs), the mechanism is correct (passive waiters + priority flip), and the validation is clean (33→0 drops solo, 50→22 concurrent).

PR #2279, opened 2026-04-27
fix(notebook-sync): keep frame pump hot during parallel sync

rgbkrk:
## Summary
- keep notebook-sync inbound frame processing hot while confirmations and daemon requests are pending
- replace blocking confirm/request waits with passive waiters and request-id pending maps
- reduce avoidable MCP confirm pressure for non-executing cell edits
- add regression coverage and update protocol/skill guidance for frame-pump backpressure

## Validation
- cargo test -p notebook-sync
- cargo test -p notebook-protocol framed_reader_does_not_desync_in_select_under_pressure
- cargo test -p runtimed --test integration test_parallel_cell_mutations_same_session_no_disconnect -- --nocapture
- cargo test -p runt-mcp --no-run

### DIFF
--- a/.claude/rules/protocol.md
+++ b/.claude/rules/protocol.md
@@ -12,7 +12,7 @@ paths:
 
 Two independent version numbers, separate from the artifact version:
 
-- **Protocol version** (`PROTOCOL_VERSION` in `connection.rs`, currently `2`) -- governs wire compatibility. Validated by the 5-byte magic preamble at connection start. Bump when framing, handshake shape, or serialization format changes.
+- **Protocol version** (`PROTOCOL_VERSION` in `connection.rs`, currently `3`) -- governs wire compatibility. Validated by the 5-byte magic preamble at connection start. Bump when framing, handshake shape, or serialization format changes. v3 adds `SessionControl`; v2 clients are still accepted but are served without `SessionControl` frames.
 - **Schema version** (`SCHEMA_VERSION` in `notebook-doc/src/lib.rs`, currently `4`) -- governs Automerge document compatibility. Bump when document structure changes.
 
 These are just incrementing integers that evolve independently from each other and from the artifact version.
@@ -24,7 +24,7 @@ Every connection starts with 5 bytes before the JSON handshake:
 | Bytes | Content |
 |-------|---------|
 | 0-3 | Magic: `0xC0 0xDE 0x01 0xAC` |
-| 4 | Protocol version (currently `2`) |
+| 4 | Protocol version (currently `3`) |
 
 The daemon validates both before reading the handshake. Non-runtimed connections get "invalid magic bytes". Protocol mismatches are rejected before JSON parsing.
 
@@ -32,8 +32,8 @@ The daemon validates both before reading the handshake. Non-runtimed connections
 
 1. **Opening** -- Frontend invokes Tauri command; relay connects to daemon Unix socket and sends handshake.
 2. **Handshake** -- JSON with `channel`, `notebook_id`, `protocol`. Daemon responds with `NotebookConnectionInfo` (protocol, notebook_id, cell_count, needs_trust_approval). The `Handshake` enum uses `#[serde(tag = "channel")]` -- flat wire format with `"channel"` discriminator.
-3. **Initial sync** -- Both sides exchange Automerge sync messages until convergence. Frontend starts empty; all state comes from daemon. 2s connection timeout, 100ms per-frame timeout.
-4. **Steady state** -- Concurrent Automerge sync, request/response, and broadcasts.
+3. **Initial sync** -- Both sides exchange Automerge sync messages until convergence. Frontend starts empty; all state comes from daemon. v3 clients also receive `SessionControl::SyncStatus` frames for notebook-doc, runtime-state, and initial-load readiness.
+4. **Steady state** -- Concurrent Automerge sync, request/response, RuntimeStateDoc sync, PoolDoc sync, presence, broadcasts, and session-control frames. The frame consumer must stay hot: request and confirmation waits belong in pending maps/waiter lists, not blocking `recv()` loops.
 5. **Disconnection** -- Relay emits `daemon:disconnected`. Generation counter prevents stale callbacks.
 
 ## Wire Format
@@ -53,8 +53,11 @@ After handshake, frames are typed by first byte:
 | `0x04` | Presence | Binary (CBOR) |
 | `0x05` | RuntimeStateSync | Binary (raw Automerge sync for RuntimeStateDoc) |
 | `0x06` | PoolStateSync | Binary (raw Automerge sync for PoolDoc — global daemon pool state) |
+| `0x07` | SessionControl | JSON (`SessionControlMessage`, daemon-originated readiness/status) |
 
-Only `0x00` (AutomergeSync), `0x04` (Presence), and `0x06` (PoolStateSync) are valid outgoing types from the frontend.
+Only `0x00` (AutomergeSync), `0x01` (NotebookRequest), `0x04` (Presence), and `0x06` (PoolStateSync) are valid outgoing frame types from the frontend/relay. `0x07` is daemon-originated.
+
+Notebook request and response frames use `NotebookRequestEnvelope` and `NotebookResponseEnvelope`. Every overlapping request must carry an `id`, and responses must route by id rather than by receive order.
 
 ## Key Request Types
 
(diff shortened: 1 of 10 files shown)


quillaid (commented):
## Review: PR #2279 — keep frame pump hot during parallel sync

I'm the author of #2277 (the minimal fix for the same session-drop bug). Reviewing this from the perspective of correctness, risk, and where the two approaches diverge.

### What's good

**The core fix is architecturally sound.** Passive `ConfirmWaiter` + `PendingRequest` HashMap + flipped select priority (Frame > Command > Changed) — all three of these are the right patterns. We arrived at the same conclusions independently.

**The two tests are excellent.** `parallel_confirm_sync_does_not_starve_frame_pump` and `concurrent_send_request_routes_responses_by_id` are precisely the regression tests this bug needs. These should land regardless of which PR ships first. The daemon integration test (`test_parallel_cell_mutations_same_session_no_disconnect`) is the definitive end-to-end proof.

**`cell_read.rs` fix is a good catch.** Changing `confirm_sync` → `confirm_state_sync` for output reads is correct — the old code was flushing notebook-doc sync when it actually needed RuntimeStateDoc frames. This belongs in both PRs.

**FramedReader buffer 16 → 64** — same conclusion as #2277.

**`register_request` registers the pending entry *before* `send_typed_frame`** — this prevents the race where a fast daemon response arrives before the entry exists. Well done.

### Concerns

**1. Protocol version bump to v3 is scope creep.**

This PR bumps `PROTOCOL_VERSION` to 3, adds `SessionControl` frame type `0x07`, and updates all protocol docs. None of that is needed for the sync-task fix. The session-drop bug is purely a client-side frame-processing issue. Mixing a wire protocol change into a bug fix makes rollback harder and muddies the bisect trail. I'd split this into its own PR.

**2. `their_heads` is an undocumented Automerge internal.**

```rust
fn heads_confirmed_by_peer(
    target_heads: &[ChangeHash],
    shared_heads: &[ChangeHash],
    their_heads: Option<&[ChangeHash]>,
) -> bool {
```

`shared_heads` is the documented convergence signal in Automerge's sync protocol — it's what `SyncState` tracks as the mutually-acknowledged frontier. `their_heads` is an internal field that tracks what the *peer* reports as its own heads. Using it as a secondary confirmation check is creative but fragile — its semantics aren't guaranteed across automerge versions. #2277 uses only `shared_heads` and it's sufficient.

**3. `sync_generation` / `acked_sync_generation` tracking adds complexity.**

The generation counter tries to answer "has the daemon seen the sync message that carried my mutation?" This is useful for a fast-path resolve, but it requires careful bookkeeping across `send_doc_sync_round`, `mark_unsent_confirm_waiters`, and `resolve_confirm_waiters`. The simpler approach (just check `shared_heads ⊇ target_heads`) works because Automerge's sync protocol guarantees that once `shared_heads` includes a change hash, the peer has merged it. The generation tracking is an optimization that adds surface area without changing the steady-state outcome.

**4. `handle_task_frame` drops `catch_unwind` for Response/Broadcast.**

The old `wait_for_response` had `catch_unwind` around `receive_sync_message` — that's the one that can actually panic (automerge bug). Dropping it for Response and Broadcast JSON parsing is fine, but worth noting explicitly that the `catch_unwind` is preserved in `handle_incoming_frame` where the Automerge calls happen.

**5. `expire_pending_requests` oneshot swap pattern is non-obvious.**

```rust
let (reply, _broadcast_tx, _deadline) = (
    std::mem::replace(&mut entry.reply, oneshot::channel().0),
    entry.broadcast_tx.take(),
    entry.deadline,
);
```

Creating a throwaway oneshot channel just to `mem::replace` the sender out of the entry is clever but allocates unnecessarily. Since `retain` gives `&mut V`, you could restructure to drain expired entries into a separate vec first, or use `HashMap::extract_if` (nightly) / a manual retain-and-collect pattern.

**6. `StateSyncWaiter` quiet-window is clever but adds a third waiter type.**

The "extend quiet_deadline on any frame activity" pattern is the right semantics for "wait until the daemon stops sending state frames." But it means three distinct waiter types (`ConfirmWaiter`, `StateSyncWaiter`, `PendingRequest`) with three different resolution paths. My PR keeps two (SyncWaiter + the existing request path). If we land this, the maintenance tick needs to stay correct across all three.

### Recommendation

Both PRs fix the session-drop bug. This one additionally implements point 3 (correlation-ID request routing) and the protocol version bump. My suggestion:

1. **Land #2277 first** — minimal change surface, low risk, fixes the immediate bug.
2. **Cherry-pick the tests from this PR** — they're strictly better than what #2277 has.
3. **Follow up with point 3 (request-ID routing) as a separate PR** — using this PR as the blueprint, minus the protocol version bump.
4. **Protocol v3 + SessionControl as its own PR** — clean separation of concerns.

The request-ID routing (#2279's `register_request` + `handle_task_frame` + `expire_pending_requests`) is good work and should ship — just not bundled with the critical bug fix.

quillaid (commented):
## AX / Gremlin Testing Review — PR #2279

Cross-reviewing from the testing/AX perspective. I haven't validated a build from this branch (only #2277's branch), so my review is based on the diff and how it maps to the gremlin failure patterns.

### Same core fix, broader scope

Both PRs arrive at the same three conclusions independently:
1. Passive waiters instead of blocking confirm_sync loops
2. Frame > Command > Changed select priority
3. Drop unnecessary confirm_sync in create_cell/set_cell

From the AX data, these three changes are what eliminated the 33 deterministic session drops in notebook-stress. Both PRs should produce equivalent results for the primary bug.

### What #2279 adds beyond #2277

**1. Request-ID routing (`PendingRequest` HashMap) — addresses a real but unmeasured problem**

The old `wait_for_response` blocks the select loop for up to 60s during LaunchKernel, starving frame processing the entire time. I haven't measured this directly (the gremlin test suite exercises `create_notebook` which triggers LaunchKernel, but it's always the first operation when the daemon is quiet, so it doesn't usually race with other traffic). However, the pattern is structurally identical to the confirm_sync bottleneck — a blocking recv loop inside the select task. The request-ID routing is the right fix for this class of bug.

Concrete AX scenario where this matters: an agent calls `execute_cell` on a long-running cell (60s timeout), then in a subsequent turn calls `get_cell` or `create_cell`. Under #2277, the `send_request_impl` wait loop would process interleaved frames (it already does this) but it's still blocking the command channel. Under #2279, the request registers a pending entry and returns to the main select loop — other commands are serviced normally. This is better architecture.

**2. `confirm_state_sync` fix in `cell_read.rs` — correct and needed**

When `get_cell` finds an empty output array but a non-empty execution_count, it tries to flush pending state. The old code called `confirm_sync()` (notebook-doc sync) when it actually needed RuntimeStateDoc frames. This is a real bug — I've seen gremlin transcripts where `get_cell` returns empty outputs immediately after `execute_cell` completes, requiring a retry. This fix should be cherry-picked into #2277 regardless.

**3. Unit tests — strictly better**

`parallel_confirm_sync_does_not_starve_frame_pump` is the definitive unit-level regression test. It creates 10 cells, spawns 10 concurrent `confirm_sync` waiters, and has the fake daemon interleave 32 `SessionControl` frames per sync round. If the frame pump starves, the test times out. The integration test `test_parallel_cell_mutations_same_session_no_disconnect` is the end-to-end proof. Both should land regardless of which PR ships.

**4. Protocol v3 + SessionControl — out of scope for this bug fix**

Agree with the #2277 review: the protocol version bump, `SessionControl` frame type, and documentation updates are valuable work but orthogonal to the session-drop fix. Bundling them increases the rollback surface if anything breaks. Recommend splitting.

### AX comparison: which approach gives better outcomes?

For the **immediate session-drop bug**, both PRs are equivalent. The gremlin data validates the core mechanism (passive waiters + priority flip + confirm_sync removal), and both implement it.

For **sustained reliability under agent workloads**, #2279's request-ID routing is architecturally superior. The current `send_request_impl` blocking loop is structurally the same bug as the old `confirm_sync_impl` — it just hasn't triggered a 100% repro yet because long-running requests don't happen during parallel mutation bursts. It will bite us eventually, especially as agents get more sophisticated about overlapping operations.

My recommendation from an AX standpoint:
1. Land #2277 now — minimal risk, validated fix for the measured bug
2. Cherry-pick from #2279: the `cell_read.rs` confirm_state_sync fix and the unit/integration tests
3. Follow up with request-ID routing from #2279 as a separate PR (this is the biggest AX win beyond the immediate fix)
4. Protocol v3 as its own PR

### Gremlin data reference

All data from the overnight bug bash and sync-fix validation runs in `gremlins.db`:
- Pre-fix: 35 gremlins, 88 session drops (2.5/gremlin), notebook-stress at 33 drops
- Post-fix (#2277 build): notebook-stress solo 0 drops, 7-gremlin suite 22 drops (3.1/gremlin, down from 7.1)
- Reports: `reports/20260427-034148-sync-fix-notebook-stress/` and `reports/20260427-034929-ax-post-sync-fix/`

PR #2292, opened 2026-04-27
fix(runtimed): move peer requests off reactor

rgbkrk:
## Summary
- move `NotebookFrameType::Request` handling off the notebook peer reactor into a per-peer FIFO request worker
- preserve request ordering while letting the socket/frame loop keep draining inbound frames
- route worker responses through the existing ordered `PeerWriter` from #2288
- add request queue backpressure handling and tests for concurrent request behavior
- add request-worker telemetry for enqueue/start/completion, handler duration, and writer queue depth without full request payload logging

## Why
After the client-side reactor and daemon writer-queue fixes, the remaining daemon-side shape is long control work inside the frame arm: `handle_notebook_request(...).await` still blocks the peer loop from reading frames. This PR keeps the conservative behavior first: one request worker per peer, FIFO execution, request IDs preserved, and no wire/API changes.

The telemetry added from #2295 makes the next bottlenecks visible without reintroducing large `Debug` logs for request payloads.

## Validation
- `cargo check -p runtimed` on the merged #2292 head (`a98d39b3`) in a detached temp worktree
- `cargo test -p runtimed notebook_sync_server::peer::peer_writer_tests -- --nocapture` on the merged #2292 head
- `cargo test -p runtimed peer_request_enqueue_reports_full_without_waiting_for_worker --lib`
- `cargo test -p runtimed --test integration test_parallel_daemon_requests_same_session_no_disconnect -- --nocapture`
- `cargo test -p runtimed --test integration test_parallel_cell_mutations_same_session_no_disconnect -- --nocapture`
- `cargo clippy -p runtimed --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -3418,13 +3418,49 @@ pub(crate) async fn update_kernel_presence(
     }
 }
 
+/// Short label for a request variant, used in telemetry logs.
+///
+/// Returns a static string — no allocation — suitable for structured logging
+/// fields.  Intentionally avoids `Debug` formatting because some variants
+/// carry large payloads (snapshot JSON, doc bytes) that would bloat log lines.
+pub(crate) fn request_label(req: &NotebookRequest) -> &'static str {
+    match req {
+        NotebookRequest::LaunchKernel { .. } => "LaunchKernel",
+        NotebookRequest::ExecuteCell { .. } => "ExecuteCell",
+        NotebookRequest::ExecuteCellGuarded { .. } => "ExecuteCellGuarded",
+        NotebookRequest::ClearOutputs { .. } => "ClearOutputs",
+        NotebookRequest::InterruptExecution { .. } => "InterruptExecution",
+        NotebookRequest::ShutdownKernel { .. } => "ShutdownKernel",
+        NotebookRequest::GetKernelInfo { .. } => "GetKernelInfo",
+        NotebookRequest::GetQueueState { .. } => "GetQueueState",
+        NotebookRequest::RunAllCells { .. } => "RunAllCells",
+        NotebookRequest::RunAllCellsGuarded { .. } => "RunAllCellsGuarded",
+        NotebookRequest::SendComm { .. } => "SendComm",
+        NotebookRequest::GetHistory { .. } => "GetHistory",
+        NotebookRequest::Complete { .. } => "Complete",
+        NotebookRequest::SaveNotebook { .. } => "SaveNotebook",
+        NotebookRequest::CloneAsEphemeral { .. } => "CloneAsEphemeral",
+        NotebookRequest::SyncEnvironment { .. } => "SyncEnvironment",
+        NotebookRequest::SyncEnvironmentGuarded { .. } => "SyncEnvironmentGuarded",
+        NotebookRequest::GetDocBytes { .. } => "GetDocBytes",
+        NotebookRequest::GetRawMetadata { .. } => "GetRawMetadata",
+        NotebookRequest::SetRawMetadata { .. } => "SetRawMetadata",
+        NotebookRequest::GetMetadataSnapshot { .. } => "GetMetadataSnapshot",
+        NotebookRequest::SetMetadataSnapshot { .. } => "SetMetadataSnapshot",
+        NotebookRequest::CheckToolAvailable { .. } => "CheckToolAvailable",
+    }
+}
+
 /// Handle a NotebookRequest and return a NotebookResponse.
 pub(crate) async fn handle_notebook_request(
     room: &Arc<NotebookRoom>,
     request: NotebookRequest,
     daemon: std::sync::Arc<crate::daemon::Daemon>,
 ) -> NotebookResponse {
-    debug!("[notebook-sync] Handling request: {:?}", request);
+    debug!(
+        "[notebook-sync] Handling request: {}",
+        request_label(&request)
+    );
 
     match request {
         NotebookRequest::LaunchKernel {
--- a/crates/runtimed/src/notebook_sync_server/peer.rs
+++ b/crates/runtimed/src/notebook_sync_server/peer.rs
@@ -1121,6 +1121,7 @@ where
 }
 
 const PEER_OUTBOUND_QUEUE_CAPACITY: usize = 1024;
+const PEER_REQUEST_QUEUE_CAPACITY: usize = 64;
 
 struct OutboundFrame {
     frame_type: NotebookFrameType,
@@ -1136,7 +1137,24 @@ struct PeerWriterTask {
     handle: tokio::task::JoinHandle<anyhow::Result<()>>,
 }
 
+struct PeerRequestWorker {
+    tx: mpsc::Sender<notebook_protocol::protocol::NotebookRequestEnvelope>,
+    handle: tokio::task::JoinHandle<anyhow::Result<()>>,
+}
+
+#[derive(Debug)]
+enum RequestEnqueueError {
+    Full(notebook_protocol::protocol::NotebookRequestEnvelope),
+    Closed(notebook_protocol::protocol::NotebookRequestEnvelope),
+}
+
 impl Drop for PeerWriterTask {
+    fn drop(&mut self) {
+        self.handle.abort();
+    }
+}
+
+impl Drop for PeerRequestWorker {
     fn drop(&mut self) {
         self.handle.abort();
     }
@@ -1170,6 +1188,18 @@ impl PeerWriter {
     }
 }
 
+impl PeerRequestWorker {
+    fn enqueue(
+        &self,
+        envelope: notebook_protocol::protocol::NotebookRequestEnvelope,
+    ) -> Result<(), RequestEnqueueError> {
+        self.tx.try_send(envelope).map_err(|e| match e {
+            mpsc::error::TrySendError::Full(envelope) => RequestEnqueueError::Full(envelope),
+            mpsc::error::TrySendError::Closed(envelope) => RequestEnqueueError::Closed(envelope),
+        })
+    }
+}
+
 fn spawn_peer_writer<W>(
     mut writer: W,
     notebook_id: String,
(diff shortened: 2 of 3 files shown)


quillaid (commented):
## Review: Adversarial but Constructive

This is a clean, conservative refactor that accomplishes the primary goal: getting `handle_notebook_request` off the frame reactor. The diff is tight, the test coverage is real, and CI is green. Good work. A few observations:

### Correctness — Solid

The FIFO channel preserves the exact same ordering as the old blocking code. The `enqueue()` / `try_send` pattern correctly returns the envelope on failure so the caller can send an error response with the right `id`. The `Drop` impl aborts the worker handle, preventing leaks. The `RequestEnqueueError::Closed` arm correctly terminates the peer loop — if the worker died, continuing to accept requests would silently drop them.

### Performance Ceiling — The Key Trade-off

The single-sequential-worker design means a `GetKernelInfo` (instant read) queued behind a `LaunchKernel` (seconds of env resolution + pool claim) still waits the full duration. In a real MCP agent session this is a frequent pattern — the agent fires `execute_cell`, then immediately calls `get_queue_state` or `get_kernel_info` to poll status. Under the current design those polls queue behind the mutation.

The same applies to `Complete` and `GetHistory`, which are ZMQ roundtrips to the runtime agent — read-only, no ordering constraints with respect to mutations, but they'll queue behind any in-flight SaveNotebook or SyncEnvironment.

This doesn't *break* anything — it's strictly better than the status quo since at least the reactor keeps reading frames. But it leaves latency on the table for the read-only fast path. The PR description says "conservative behavior first" which is a valid framing.

### Backpressure — Works but Could Help Agents More

The error message on queue full is:

```
"Peer request queue full"
```

MCP agents (our primary non-human consumers) need slightly more guidance to self-correct. A message like `"Server busy: request queue full, retry shortly"` tells the agent explicitly that retry is the right recovery strategy. The current message is technically accurate but doesn't signal recoverability.

The `Closed` path returns `"Peer request worker stopped"` which is good — that's a non-recoverable situation and the error doesn't suggest retry.

### Telemetry — Missing

There's no timing instrumentation on request dispatch or completion. Without logging request type, duration, peer ID, and notebook ID, there's no way to:
- Measure baseline handler latency per request type
- Detect which handlers are actually causing head-of-line blocking in production
- Compare before/after performance across builds

Even a single `debug!` with `Instant::now()` around `handle_notebook_request` in the worker loop would be valuable.

### Test Coverage — Good Unit, Integration Could Go Further

The unit test (`peer_request_enqueue_reports_full_without_waiting_for_worker`) is well-designed — it proves the frame arm returns immediately even when the worker is blocked. The timing assertion (`< 50ms`) is a good smoke test.

The integration test sends 10 concurrent read-only requests (`GetDocBytes` + `GetKernelInfo`). This validates that the worker processes requests without disconnecting the session. However, it doesn't exercise the scenario that motivated the change — a slow mutation blocking a fast query. A test that sends a `LaunchKernel` followed immediately by `GetKernelInfo` and asserts the latter doesn't time out would directly test the head-of-line blocking improvement... although under the single-worker design, that test would actually demonstrate the limitation rather than the fix.

### What This PR Gets Right

1. **The `RequestEnqueueError` enum distinguishing Full vs Closed** — this is cleaner than our approach of just checking `is_err()` on `try_send`. The Closed path needs different handling (terminate the loop) and this makes that explicit at the type level.

2. **The integration test** — we should have one too. Testing that parallel requests don't disconnect the session is a real regression guard.

3. **`queue_request_error` helper** — nice factoring of the error response construction. Reduces duplication at every call site.

4. **Single commit, minimal surface** — easy to review, easy to revert, low risk. For a `fix/` branch this is the right risk posture.

### Summary

Correct, conservative, and safe. The single-worker design is a deliberate trade-off: simpler implementation at the cost of query latency behind slow mutations. Whether that's the right trade-off depends on how much we value fast query responses during heavy mutation workloads (MCP agent polling patterns suggest we should care). The missing telemetry means we won't have data to evaluate the trade-off in production.

quillaid:
## Review: Adversarial but Constructive

This is a clean, conservative refactor that accomplishes the primary goal: getting `handle_notebook_request` off the frame reactor. The diff is tight, the test coverage is real, and CI is green. Good work. A few observations:

### Correctness — Solid

The FIFO channel preserves the exact same ordering as the old blocking code. The `enqueue()` / `try_send` pattern correctly returns the envelope on failure so the caller can send an error response with the right `id`. The `Drop` impl aborts the worker handle, preventing leaks. The `RequestEnqueueError::Closed` arm correctly terminates the peer loop — if the worker died, continuing to accept requests would silently drop them.

### Performance Ceiling — The Key Trade-off

The single-sequential-worker design means a `GetKernelInfo` (instant read) queued behind a `LaunchKernel` (seconds of env resolution + pool claim) still waits the full duration. In a real MCP agent session this is a frequent pattern — the agent fires `execute_cell`, then immediately calls `get_queue_state` or `get_kernel_info` to poll status. Under the current design those polls queue behind the mutation.

The same applies to `Complete` and `GetHistory`, which are ZMQ roundtrips to the runtime agent — read-only, no ordering constraints with respect to mutations, but they'll queue behind any in-flight SaveNotebook or SyncEnvironment.

This doesn't *break* anything — it's strictly better than the status quo since at least the reactor keeps reading frames. But it leaves latency on the table for the read-only fast path. The PR description says "conservative behavior first" which is a valid framing.

### Backpressure — Works but Could Help Agents More

The error message on queue full is:

```
"Peer request queue full"
```

MCP agents (our primary non-human consumers) need slightly more guidance to self-correct. A message like `"Server busy: request queue full, retry shortly"` tells the agent explicitly that retry is the right recovery strategy. The current message is technically accurate but doesn't signal recoverability.

The `Closed` path returns `"Peer request worker stopped"` which is good — that's a non-recoverable situation and the error doesn't suggest retry.

### Telemetry — Missing

There's no timing instrumentation on request dispatch or completion. Without logging request type, duration, peer ID, and notebook ID, there's no way to:
- Measure baseline handler latency per request type
- Detect which handlers are actually causing head-of-line blocking in production
- Compare before/after performance across builds

Even a single `debug!` with `Instant::now()` around `handle_notebook_request` in the worker loop would be valuable.

### Test Coverage — Good Unit, Integration Could Go Further

The unit test (`peer_request_enqueue_reports_full_without_waiting_for_worker`) is well-designed — it proves the frame arm returns immediately even when the worker is blocked. The timing assertion (`< 50ms`) is a good smoke test.

The integration test sends 10 concurrent read-only requests (`GetDocBytes` + `GetKernelInfo`). This validates that the worker processes requests without disconnecting the session. However, it doesn't exercise the scenario that motivated the change — a slow mutation blocking a fast query. A test that sends a `LaunchKernel` followed immediately by `GetKernelInfo` and asserts the latter doesn't time out would directly test the head-of-line blocking improvement... although under the single-worker design, that test would actually demonstrate the limitation rather than the fix.

### What This PR Gets Right

1. **The `RequestEnqueueError` enum distinguishing Full vs Closed** — this is cleaner than checking `is_err()` on `try_send`. The Closed path needs different handling (terminate the loop) and this makes that explicit at the type level.

2. **The integration test** — testing that parallel requests don't disconnect the session is a real regression guard.

3. **`queue_request_error` helper** — nice factoring of the error response construction. Reduces duplication at every call site.

4. **Single commit, minimal surface** — easy to review, easy to revert, low risk. For a `fix/` branch this is the right risk posture.

### Summary

Correct, conservative, and safe. The single-worker design is a deliberate trade-off: simpler implementation at the cost of query latency behind slow mutations. Whether that's the right trade-off depends on how much we value fast query responses during heavy mutation workloads (MCP agent polling patterns suggest we should care). The missing telemetry means we won't have data to evaluate the trade-off in production.

quillaid:
## Telemetry helper from #2294 — available for integration

One piece from our PR that's independently useful and complementary to this approach: a `request_label()` function that returns allocation-free `&'static str` variant names for logging. Avoids `format!("{:?}", request)` which can dump large payloads (snapshot JSON, doc bytes) into log lines.

Happy to have you integrate it into `spawn_peer_request_worker` if useful. Here's the function (goes in `metadata.rs` above `handle_notebook_request`):

```rust
/// Short label for a request variant, used in telemetry logs.
///
/// Returns a static string — no allocation — suitable for structured logging
/// fields.  Intentionally avoids `Debug` formatting because some variants
/// carry large payloads (snapshot JSON, doc bytes) that would bloat log lines.
pub(crate) fn request_label(req: &NotebookRequest) -> &'static str {
    match req {
        NotebookRequest::LaunchKernel { .. } => "LaunchKernel",
        NotebookRequest::ExecuteCell { .. } => "ExecuteCell",
        NotebookRequest::ExecuteCellGuarded { .. } => "ExecuteCellGuarded",
        NotebookRequest::ClearOutputs { .. } => "ClearOutputs",
        NotebookRequest::InterruptExecution { .. } => "InterruptExecution",
        NotebookRequest::ShutdownKernel { .. } => "ShutdownKernel",
        NotebookRequest::GetKernelInfo { .. } => "GetKernelInfo",
        NotebookRequest::GetQueueState { .. } => "GetQueueState",
        NotebookRequest::RunAllCells { .. } => "RunAllCells",
        NotebookRequest::RunAllCellsGuarded { .. } => "RunAllCellsGuarded",
        NotebookRequest::SendComm { .. } => "SendComm",
        NotebookRequest::GetHistory { .. } => "GetHistory",
        NotebookRequest::Complete { .. } => "Complete",
        NotebookRequest::SaveNotebook { .. } => "SaveNotebook",
        NotebookRequest::CloneAsEphemeral { .. } => "CloneAsEphemeral",
        NotebookRequest::SyncEnvironment { .. } => "SyncEnvironment",
        NotebookRequest::SyncEnvironmentGuarded { .. } => "SyncEnvironmentGuarded",
        NotebookRequest::GetDocBytes { .. } => "GetDocBytes",
        NotebookRequest::GetRawMetadata { .. } => "GetRawMetadata",
        NotebookRequest::SetRawMetadata { .. } => "SetRawMetadata",
        NotebookRequest::GetMetadataSnapshot { .. } => "GetMetadataSnapshot",
        NotebookRequest::SetMetadataSnapshot { .. } => "SetMetadataSnapshot",
        NotebookRequest::CheckToolAvailable { .. } => "CheckToolAvailable",
    }
}
```

Usage in the worker loop would look something like:

```rust
while let Some(envelope) = rx.recv().await {
    let label = metadata::request_label(&envelope.request);
    let req_id = envelope.id.as_deref().unwrap_or("-");
    let writer_queue_depth = PEER_OUTBOUND_QUEUE_CAPACITY - writer.capacity();
    debug!(
        "[notebook-sync] Request {} id={} peer={} notebook={} writer_queue={}",
        label, req_id, peer_id, notebook_id, writer_queue_depth,
    );

    let start = std::time::Instant::now();
    let response = handle_notebook_request(&room, envelope.request, daemon.clone()).await;
    debug!(
        "[notebook-sync] Request {} id={} completed in {:?}",
        label, req_id, start.elapsed(),
    );
    // ... send reply
}
```

The `capacity()` method on `PeerWriter` exposes `self.tx.capacity()` — `PEER_OUTBOUND_QUEUE_CAPACITY - capacity()` gives in-flight frame count as a backpressure signal.

PR #2294, opened 2026-04-27
perf(runtimed): offload request handling from frame reactor

quillaid:
## Summary

- **Moves `handle_notebook_request` off the sync loop's `select!` macro** so it no longer blocks Automerge sync, presence, and broadcast processing for the duration of each request handler.

- **Categorised dispatch** — mutations (LaunchKernel, ExecuteCell, SaveNotebook, etc.) run sequentially on a dedicated per-peer worker task via bounded mpsc channel (capacity 64), preserving FIFO ordering. Queries (GetKernelInfo, Complete, GetHistory, etc.) spawn as concurrent tasks gated by a per-peer semaphore (capacity 32).

- **Telemetry** — request type, envelope ID, peer ID, notebook ID, PeerWriter queue depth, and handler duration are logged at debug level for every request dispatch and completion.

- **Backpressure** — when mutation queue or query semaphore is exhausted, returns an `Error` response with a retry hint rather than silently dropping.

## Motivation

The `handle_notebook_request(...).await` in the frame arm of `run_sync_loop_v2` was the primary source of head-of-line blocking in the peer sync loop. LaunchKernel can block for seconds (env resolution, pool claims), SaveNotebook does disk I/O, and Complete/GetHistory block on ZMQ roundtrips. During any of these, no Automerge sync messages, presence updates, or broadcasts could be processed.

## Test plan

- [ ] Verify `cargo check -p runtimed` passes
- [ ] Verify `cargo xtask clippy` passes
- [ ] Verify `cargo test -p runtimed --test tokio_mutex_lint` passes
- [ ] Verify `cargo xtask lint` passes
- [ ] Manual test: open a notebook, launch kernel, execute cells, save — all operations complete correctly
- [ ] Manual test: run Complete/GetHistory while a slow mutation (LaunchKernel) is in flight — queries return without waiting

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -3418,6 +3418,39 @@ pub(crate) async fn update_kernel_presence(
     }
 }
 
+/// Short label for a request variant, used in telemetry logs.
+///
+/// Returns a static string — no allocation — suitable for structured logging
+/// fields.  Intentionally avoids `Debug` formatting because some variants
+/// carry large payloads (snapshot JSON, doc bytes) that would bloat log lines.
+pub(crate) fn request_label(req: &NotebookRequest) -> &'static str {
+    match req {
+        NotebookRequest::LaunchKernel { .. } => "LaunchKernel",
+        NotebookRequest::ExecuteCell { .. } => "ExecuteCell",
+        NotebookRequest::ExecuteCellGuarded { .. } => "ExecuteCellGuarded",
+        NotebookRequest::ClearOutputs { .. } => "ClearOutputs",
+        NotebookRequest::InterruptExecution { .. } => "InterruptExecution",
+        NotebookRequest::ShutdownKernel { .. } => "ShutdownKernel",
+        NotebookRequest::GetKernelInfo { .. } => "GetKernelInfo",
+        NotebookRequest::GetQueueState { .. } => "GetQueueState",
+        NotebookRequest::RunAllCells { .. } => "RunAllCells",
+        NotebookRequest::RunAllCellsGuarded { .. } => "RunAllCellsGuarded",
+        NotebookRequest::SendComm { .. } => "SendComm",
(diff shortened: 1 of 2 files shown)


quillaid:
## Acknowledging review findings

Codex's review of this PR identified two real correctness issues that I missed:

### 1. Ordering regression between mutations and queries

The categorized dispatch breaks FIFO ordering that clients depend on. Example:

```
Client sends: SetRawMetadata → GetRawMetadata (same peer, socket order)
```

Under this PR, `SetRawMetadata` goes to the mutation channel and `GetRawMetadata` spawns as a concurrent query task. The query can complete before the mutation, returning stale state. The previous blocking code processed these in strict socket order — the Get always saw the Set's result.

This isn't limited to metadata. Any mutation-then-query sequence on the same peer is affected: `ExecuteCell` → `GetQueueState`, `SaveNotebook` → `GetDocBytes`, `LaunchKernel` → `GetKernelInfo`.

The query fast-path idea is sound in principle, but it needs a per-peer ordering barrier (e.g., queries wait for the mutation queue to drain) to preserve the socket-order guarantee. Without that barrier, this is a correctness regression.

### 2. Detached work after peer disconnect

When the peer loop exits:
- `mutation_tx` is dropped, which eventually shuts down the worker — but any in-flight `handle_notebook_request` continues to completion, doing work against the room after the peer is gone.
- Spawned query tasks (`tokio::spawn`) have no abort handles stored. They run to completion with no cleanup path, writing responses to a closed PeerWriter.

This is residual pressure — exactly the class of problem this work aims to eliminate. #2292 handles this correctly with `Drop for PeerRequestWorker` which aborts the handle on drop.

### Conclusion

The telemetry commit (`8aaa074c` — `request_label()`, timing, queue depth logging) is independently useful and could be cherry-picked onto #2292. The structural dispatch commit has the two issues above and should not be merged as-is. #2292's conservative single-sequential-worker approach preserves FIFO correctness while still achieving the primary goal of keeping the reactor hot.

PR #2295, opened 2026-04-27
perf(runtimed): add request telemetry to peer request worker

quillaid:
## Summary

Adds request handling telemetry to the peer request worker from #2292. Ported from #2294's telemetry commit, adapted to fit the single-worker structure.

- **`request_label()`** in `metadata.rs` — allocation-free `&'static str` variant names for all 23 `NotebookRequest` variants. Avoids `Debug` formatting which can dump large payloads (snapshot JSON, doc bytes) into log lines.
- **Enqueue-time logging** in the frame arm — logs request type, envelope ID, peer ID, notebook ID when a request is dispatched to the worker queue.
- **Worker-loop logging** — logs request type, envelope ID, peer ID, notebook ID, PeerWriter queue depth at handler start; logs request type, envelope ID, and handler duration at completion.

## What this enables

- Measure baseline handler latency per request type in production
- Identify which handlers cause head-of-line blocking (LaunchKernel, SaveNotebook, Complete)
- Monitor PeerWriter queue depth as a backpressure signal
- Compare before/after performance across builds

## Test plan

- [x] `cargo check -p runtimed` passes
- [x] `cargo xtask lint` passes
- [ ] All existing tests in #2292 still pass

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -3457,7 +3457,10 @@ pub(crate) async fn handle_notebook_request(
     request: NotebookRequest,
     daemon: std::sync::Arc<crate::daemon::Daemon>,
 ) -> NotebookResponse {
-    debug!("[notebook-sync] Handling request: {:?}", request);
+    debug!(
+        "[notebook-sync] Handling request: {}",
+        request_label(&request)
+    );
 
     match request {
         NotebookRequest::LaunchKernel {
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -3418,6 +3418,39 @@ pub(crate) async fn update_kernel_presence(
     }
 }
 
+/// Short label for a request variant, used in telemetry logs.
+///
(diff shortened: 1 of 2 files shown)


rgbkrk (changes requested):
Requesting one change before merging this into #2292.

The new `request_label()` helper is the right direction, but `handle_notebook_request()` still starts with:

`debug!("[notebook-sync] Handling request: {:?}", request);`

That preserves the same full-`Debug` formatting this PR is trying to avoid. Once someone enables debug logs to use the new enqueue/start/completion telemetry, large request payloads can still be formatted and written from the old handler log. Please remove that line or switch it to the new label helper before landing this telemetry.

After that, I do not see another merge blocker in this patch.

quillaid:
Addressed in 24467eb8 — replaced `debug!("{:?}", request)` in `handle_notebook_request()` with `debug!("{}", request_label(&request))`. The old full-`Debug` line is gone.

rgbkrk (approved):
Approved. The requested change is fixed: `handle_notebook_request` now uses `request_label(&request)` instead of full `Debug` formatting, so enabling this telemetry should not dump large request payloads from the old handler log.

I rechecked the final diff against #2292 and do not see another merge blocker.

PR #2306, opened 2026-04-27
fix(installer): bootstrap nightly installs on Windows

rgbkrk:
## Summary
- add Windows CLI shim installation/repair for `runt(-nightly)` and `nb(-nightly)` without clobbering unmanaged files
- add NSIS installer hooks that write shims, run `runt daemon doctor --fix`, log bootstrap output, and clean up owned shims on uninstall
- publish raw Linux daemon/MCP release binaries and add `scripts/install-nightly-release` for headless lab installs
- extend Windows smoke coverage to assert silent install leaves CLI, daemon, and MCP startup usable

## Verification
- `cargo test -p notebook windows_cmd --lib`
- `cargo test -p notebook nb_script --lib`
- `cargo check -p runt`
- `bash -n scripts/install-nightly-release`
- parsed `crates/notebook/tauri.conf.json`, `.github/workflows/smoke.yml`, and `.github/workflows/release-common.yml`
- `git diff --check`

Attempted `cargo check -p notebook --target x86_64-pc-windows-msvc --lib` from macOS, but it failed before reaching notebook code because the host lacks the Windows C headers/toolchain needed by `zstd-sys` (`stdlib.h` / `string.h` under `--target=x86_64-pc-windows-msvc`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,24 @@ jobs:
           bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
           ./actionlint -color
 
+  nsis-bootstrap:
+    name: NSIS Bootstrap Syntax
+    needs: [changes]
+    if: needs.changes.outputs.source_changed == 'true'
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install NSIS
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y nsis
+
+      - name: Compile bootstrap hook fixture
+        run: |
+          cd crates/notebook/nsis
+          makensis -V4 validate-bootstrap.nsi
+
   windows-clippy:
     name: Windows (clippy)
     needs: [changes]
--- a/.github/workflows/release-common.yml
+++ b/.github/workflows/release-common.yml
@@ -149,12 +149,16 @@ jobs:
         run: pnpm --filter nteract-mcp-app install && vp run nteract-mcp-app#build
 
       - name: Build for Linux x64
-        run: cargo build --release --target x86_64-unknown-linux-gnu -p runt
+        run: cargo build --release --target x86_64-unknown-linux-gnu -p runt -p runtimed -p nteract-mcp
 
       - name: Rename binaries
         run: |
           cp target/x86_64-unknown-linux-gnu/release/runt runt-linux-x64
+          cp target/x86_64-unknown-linux-gnu/release/runtimed runtimed-linux-x64
+          cp target/x86_64-unknown-linux-gnu/release/nteract-mcp nteract-mcp-linux-x64
           chmod +x runt-linux-x64
+          chmod +x runtimed-linux-x64
+          chmod +x nteract-mcp-linux-x64
 
       - name: Upload Linux executables
         uses: actions/upload-artifact@v4
@@ -162,6 +166,8 @@ jobs:
           name: runt-linux-executables
           path: |
             runt-linux-x64
+            runtimed-linux-x64
+            nteract-mcp-linux-x64
 
   build-macos:
     name: Build macOS Executables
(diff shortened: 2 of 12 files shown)


rgbkrk (commented):
Read the diff against the spec. Coverage is strong and the shape matches what we discussed; below is a mix of things to verify before merge and smaller refinements. Nothing blocking from my end.

## What's working well

- `cli_install.rs` Windows arm picks both branches I'd asked for: prefer `%LOCALAPPDATA%\Microsoft\WindowsApps` if it's writable and on PATH, otherwise fall back to `~/.local/bin` with an explicit `HKCU\Environment\Path` write and a `WM_SETTINGCHANGE` broadcast. The fallback closes the App-Execution-Aliases-disabled footgun cleanly.
- Ownership marker (`nteract-managed-cli-shim v1`) on a fixed line of the shim file means we never clobber a user's pre-existing `runt-nightly.cmd`. The unit tests cover the absolute-path case, channel-suffixed names, the `%%` percent-escape edge case, and the ownership check.
- Update-mode detection via `Software\nteract\$channel\InstallState\Updating` set in pre-install / cleared by Rust on app launch is the right shape, and `channel_display_name()` returning `"stable"`/`"nightly"` matches `$R9` in NSIS exactly. I checked.
- The bootstrap log written to `$LOCALAPPDATA\${PRODUCTNAME}\install-bootstrap.log` and surfaced both in the smoke job and as a CI artifact gives us a real diagnostic for silent-install failures, which was my main worry about the silent-fail UX.
- `runt daemon install`'s Windows fallback to sibling `runtimed.exe` is gated by `!channel_suffixed.exists()`, so it only fires when the post-install hook hasn't run yet. Reads as forward-compat repair, not a duplicate path.
- `install-nightly-release` supports both `--tag` and `--from-dir`, which I'd have asked for if it weren't already there.

## Things to verify before merge

1. **Windows compilation has not been exercised.** The PR notes that `cargo check -p notebook --target x86_64-pc-windows-msvc --lib` failed on macOS due to missing zstd-sys C headers. Same applies to the NSIS hook: `bootstrap.nsh` syntax errors won't surface until a Windows installer build runs. The smoke job will catch both, but it'd be worth holding merge until the Windows smoke run on this branch is green so we don't land on red on `main`.

2. **`${StrLoc}` requires `StrFunc.nsh`.** The NSIS hook uses `${StrLoc}` from `StrFunc.nsh`. Tauri's NSIS template generally pulls it in via other macros so this should just work, but if the smoke job hits a `!define`/`!use` complaint about StrLoc, that's the cause.

3. **`$PassiveMode` / `$UpdateMode` are Tauri-template globals.** Not standard NSIS. If a future Tauri version drops them, the silent-mode detection and the update-skip in the pre-uninstall would silently change behavior. Cheap defensive move: top of `bootstrap.nsh` declare `Var /GLOBAL PassiveMode` and `Var /GLOBAL UpdateMode` (idempotent if Tauri already declares them, errors if Tauri ever stops setting them).

## Smaller refinements

4. `is_owned_windows_cmd_shim` matches `contents.contains(MARKER)` — a user could in theory create a shim that mentions the marker in a later comment line and we'd treat it as ours. The structured shape (`@echo off\r\nrem nteract-managed-cli-shim v1\r\n`) is what the writer guarantees, so checking for that prefix is stricter and the writer already produces it. Low impact today, easy tightening.

5. The pre-install hook sets `Updating=1` whenever `uninstall.exe` already exists. That conflates "this is an update" with "any prior install left an uninstall.exe behind, including ones that crashed mid-install." `clear_windows_update_flag()` on app launch limits the window, but a comment in the macro explaining the conflation would help the next reader.

6. `install-nightly-release` doesn't currently exercise the "MCP can find the CLI" assertion that the spec test plan called out. A trailing `command -v "$CLI_NAME"` (or `"$INSTALL_DIR/$MCP_NAME" --version` once that's a thing) would close the loop without much effort.

7. No checksum/signature verification on the downloaded binaries. Probably out of scope for this PR — flagging because GitHub release URLs are fine for normal usage but a `--verify <sha256-manifest>` follow-up would harden it for security-conscious deployments.

## Smoke coverage

The new smoke steps cover everything the spec asked for: silent install → resolve the shim → `daemon status --json` reports installed+running → `nteract-mcp.exe` startup probe doesn't exit with `runt-nightly not found` → bootstrap log archived. Stopping the installer-started daemon before the controlled smoke daemon spin-up is the right call; the `Write-Host` annotation explains why.

The one piece I'd add to smoke if it's cheap: assert `nb-nightly.cmd` (or `nb.cmd` for stable) exists in the same dir as `runt-nightly.cmd`. The unit tests cover the wrapper's content shape; smoke would catch a path-mismatch between the unit's expected location and where the NSIS hook actually drops it.

PR #2317, opened 2026-04-28
fix(installer): break the daemon out of the parent's Job Object

rgbkrk:
## Why

#2308 fixed handle inheritance, which was a real bug, but it did not fix the silent-install hang. Verified on a clean Windows host: with `-Wait`, the post-#2308 installer still hangs indefinitely; the only surviving process is `runtimed.exe`, and `IsProcessInJob` returns `TRUE`.

The actual mechanism is **Job Object containment**, not pipe handles. PowerShell's `Start-Process -Wait` (which is what GitHub Actions' `Install silently` step uses, and what GHA wraps every step in for descendant tracking) wraps the child in a Windows Job Object and waits for *every* process in the Job to exit before returning. The default `CreateProcessW` puts the new daemon in the parent's Job, so the daemon's lifetime keeps the Job populated and `-Wait` waits forever - which is what the runner timeout was hitting.

## What

Add `CREATE_BREAKAWAY_FROM_JOB` to the daemon spawn flags. Falls back to a non-breakaway spawn on `ACCESS_DENIED` so the daemon still starts in the rare case where a future caller wraps the spawn in a strict Job that disallows breakaway.

## Verification

Direct test on a clean Windows host. A standalone Rust binary (`spawn-test`) is launched under `Start-Process -Wait -NoNewWindow` (which puts it in PowerShell's Job) and spawns the existing nightly daemon with the two flag sets:

```
mode=none:       self_in_job=true   child_in_job=true    ← daemon stuck in PowerShell's job
mode=breakaway:  self_in_job=true   child_in_job=false   ← daemon escaped, -Wait can return
```

PowerShell's `Start-Process` Job permits breakaway, so the new flag is sufficient and the fallback path is rarely hit.

CI smoke dispatched against this branch: [run 25034437945](https://github.com/nteract/desktop/actions/runs/25034437945). The post-#2308 manual smoke ([25031947609](https://github.com/nteract/desktop/actions/runs/25031947609)) hung at `Install silently` for the full 1h15m runner timeout; this run reaching `Verify installer bootstrap` is the proof point.

### DIFF
--- a/crates/runtimed-client/src/service.rs
+++ b/crates/runtimed-client/src/service.rs
@@ -770,32 +770,39 @@ Set WshShell = Nothing
         use std::os::windows::ffi::OsStrExt;
         use windows_sys::Win32::Foundation::CloseHandle;
         use windows_sys::Win32::System::Threading::{
-            CreateProcessW, CREATE_NO_WINDOW, DETACHED_PROCESS, PROCESS_INFORMATION, STARTUPINFOW,
+            CreateProcessW, CREATE_BREAKAWAY_FROM_JOB, CREATE_NO_WINDOW, DETACHED_PROCESS,
+            PROCESS_INFORMATION, STARTUPINFOW,
         };
 
-        // The daemon runs forever. The standard `Command::new(...).spawn()`
-        // path on Windows calls `CreateProcessW` with `bInheritHandles=TRUE`
-        // (Rust's default), which duplicates *every* inheritable handle in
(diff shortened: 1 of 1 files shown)


rgbkrk:
Superseded by #2323. The FFI/Job-Object breakaway approach didn't survive GHA's pwsh wrapper, and `--no-start` from #2323 fixes the underlying issue (don't spawn the daemon during install at all).

PR #2319, opened 2026-04-28
fix(runt-mcp): graceful session disconnect on SIGTERM

quillaid:
## Summary

- **Adds SIGTERM signal handling** to the `runt mcp` process so ephemeral notebook sessions are cleanly disconnected before exit, allowing the daemon to start its eviction timer immediately rather than waiting for OS-level TCP cleanup.
- **Adds explicit session drop** after the MCP transport closes (normal exit path), ensuring the daemon peer is disconnected even if the MCP client disconnects without a clean shutdown.
- **Adds `NteractMcp::shutdown()` method** for structured session teardown.

## Motivation

Gremlin baseline runs showed 22/39 gremlins reporting orphan kernel sessions. Investigation traced the root cause to `runt mcp` processes exiting on SIGTERM (sent by the gremlin harness) without any session cleanup — the default SIGTERM handler kills the process immediately, and tokio runtime teardown order is unreliable for async cleanup. The daemon then must wait for OS TCP socket cleanup before detecting the peer disconnect and starting its 30s eviction timer.

While analysis showed the majority of "orphan" reports are measurement artifacts from concurrent batch execution (gremlins see each other's sessions), the lack of graceful shutdown is a real issue that delays kernel cleanup and could cause genuine leaks under edge conditions (e.g., half-open TCP connections).

### `crates/runt/src/main.rs`
- Register a `SIGTERM` handler using `tokio::signal::unix::signal`
- Add it as a third arm in the `tokio::select!` loop alongside MCP transport close and daemon-watch
- After the select completes (any arm), explicitly take and drop the `NotebookSession` to disconnect from the daemon

### `crates/runt-mcp/src/lib.rs`
- Add `NteractMcp::shutdown()` async method that takes the session lock, extracts the session, and drops it with logging

## Test plan

- [x] `cargo test -p runt-mcp` — 112 tests pass
- [x] `cargo clippy -p runt -p runt-mcp` — clean
- [x] `./scripts/install-nightly` — installed, daemon status shows matching SHA
- [ ] Gremlin suite re-run to validate reduced orphan count

### DIFF
--- a/crates/runt-mcp/src/lib.rs
+++ b/crates/runt-mcp/src/lib.rs
@@ -122,6 +122,30 @@ impl NteractMcp {
     pub fn last_session_drop(&self) -> &Arc<RwLock<Option<SessionDropInfo>>> {
         &self.last_session_drop
     }
+
+    /// Explicitly drop the active session, closing the daemon peer connection
+    /// immediately.
+    ///
+    /// Call this before the process exits to trigger prompt session cleanup in
+    /// the daemon rather than relying on OS-level TCP connection close + the
+    /// 30s eviction delay. Dropping the `NotebookSession` closes its
+    /// `DocHandle` channels, which makes the sync task break out of its loop
+    /// and close the TCP writer — the daemon sees EOF and decrements the peer
+    /// count, starting the eviction timer right away.
+    ///
+    /// Without this, session cleanup depends on the tokio runtime tearing down
+    /// tasks in the right order during process exit, which is unreliable under
+    /// SIGTERM.
+    pub async fn shutdown(&self) {
+        let old = self.session.write().await.take();
+        if let Some(session) = old {
+            tracing::info!(
+                "[mcp] Shutdown: disconnecting session {}",
+                session.notebook_id
+            );
+            drop(session);
+        }
+    }
 }
 
 impl ServerHandler for NteractMcp {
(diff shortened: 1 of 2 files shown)


rgbkrk:
You need to only shut it down if there aren't other peers on the document. Otherwise you might shut it down when other humans or agents are on the document.

quillaid:
Good call — I should clarify what this does. This only disconnects **our MCP process's peer connection** from the daemon room. It does NOT shut down the kernel or evict the room.

The daemon tracks `active_peers` per room (in `notebook_sync_server/peer.rs`). On peer disconnect, it decrements the count and only schedules eviction when `remaining == 0`. If other peers (humans, other agents) are still connected, the room and kernel stay alive — the eviction check at line 652 sees `active_peers > 0` and cancels.

So the sequence is:
1. Our `runt mcp` process receives SIGTERM (or MCP client disconnects)
2. We drop our `NotebookSession` → closes our `DocHandle` channels → sync task closes its TCP writer → daemon sees EOF
3. Daemon decrements `active_peers` by 1
4. If `active_peers > 0` → nothing happens, other peers keep working
5. If `active_peers == 0` → daemon schedules eviction after 30s keep-alive

I've pushed a follow-up commit (`b3794f56`) that makes this explicit in the doc comments and inline comments. The `shutdown()` doc string now reads: "This only drops **our** peer connection — it does NOT shut down the kernel or evict the room."

PR #2333, opened 2026-04-28
fix(runtimed): stabilize pyproject watcher metadata

rgbkrk:
## Summary

Fixes the UV Pyproject E2E failure seen in the Build workflow by normalizing file-watcher metadata for pyproject-backed notebooks before comparing it to the CRDT snapshot.

## Root Cause

Auto-launch derives `metadata.runt.uv.dependencies` from the adjacent `pyproject.toml` so the UI and MCP consumers can see the active project environment. The notebook file on disk may not contain that derived metadata yet. The `.ipynb` file watcher compared the raw parsed file metadata against the in-memory CRDT snapshot, interpreted the derived deps as an external deletion, and churned metadata on each watch event.

## Changes

- Added a shared pyproject bootstrap normalizer for metadata snapshots parsed from disk.
- Applied that normalizer in the notebook file watcher before deciding whether external metadata changed.
- Added a regression test for the raw `.ipynb` metadata vs bootstrapped CRDT metadata comparison.

## Verification

- `cargo xtask wasm`
- `cargo test -p runtimed test_file_watcher_normalizes_pyproject_bootstrap_metadata -- --nocapture`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -1126,6 +1126,63 @@ pub(crate) fn project_file_deps_match_trust_info(
     }
 }
 
+/// Apply the same derived pyproject.toml dependency view used by auto-launch
+/// to a metadata snapshot parsed from disk.
+///
+/// The file watcher compares external `.ipynb` metadata against the in-memory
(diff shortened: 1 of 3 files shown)


rgbkrk:
Superseded by #2334, which keeps project-file dependencies as runtime context instead of normalizing notebook metadata.

PR #2335, opened 2026-04-28
fix(nteract-mcp): auto-reap child processes to prevent zombie accumulation

quillaid:
## Summary

- Set `SIGCHLD` to `SIG_IGN` at `nteract-mcp` process startup so the kernel auto-reaps exited child processes
- Prevents zombie accumulation in long-running MCP sessions (agent teammates, Claude Code plugins)
- Add `libc` as a unix-only dependency for the signal call

## Context

The MCP proxy spawns `runt mcp` children via rmcp's `TokioChildProcess`. When a child exits (e.g. daemon upgrade exit code 75), the tokio `Child` handle is dropped without `waitpid()`, leaving a zombie. In long agent sessions, we observed one zombie per daemon upgrade cycle (PIDs 19494, 19506 on the devbox).

`SIG_IGN` is safe here because the proxy never inspects child exit codes — it detects child death via transport EOF polling in the monitor loop (`is_transport_closed()`).

## Test plan

- [ ] Verify `cargo build -p nteract-mcp` succeeds on Linux
- [ ] Verify `cargo clippy -p nteract-mcp` passes
- [ ] Manual: spawn nteract-mcp, trigger a daemon upgrade cycle, confirm no zombie remains

### DIFF
--- a/crates/nteract-mcp/Cargo.toml
+++ b/crates/nteract-mcp/Cargo.toml
@@ -20,5 +20,8 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 dirs = "6"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [lints]
 workspace = true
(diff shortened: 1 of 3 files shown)


quillaid:
Superseded by #2336, which takes the same approach (own the Child handle, reap via background task) with a more complete implementation: ManagedChildTransport wrapper with shutdown signaling, kill-and-reap on close, and cross-platform regression tests.

PR #2369, opened 2026-04-29
feat(runtimed): switch kernel launch from TCP to IPC on Unix

### LINKED ISSUE #1813: Switch kernel launch to IPC transport (supersedes #1772)
Replace the current port-reservation-based kernel launch (which has a TOCTOU race around port reuse — the original #1772 issue) with IPC transport (Unix domain sockets / named pipes) for the kernel's Jupyter control/shell/iopub/stdin/heartbeat channels.

### Why now

- The TCP port-retry dance is fragile: we pick a port, kernel may bind late, port collisions cause silent failures or require restart.
- IPC sockets have no port namespace collision surface.
- jupyter_client supports IPC transport (`transport: "ipc"`) — kernel-side wiring is minimal.
- Aligns with `runtimed-agent` architecture (already Unix-socket based).

### Work

1. Design spec: what exactly do we change in `kernel-launch`, `JupyterKernel`, and the connection-info JSON written to disk? Where do the 5 IPC socket files live (daemon base dir? worktree-scoped?)?
2. Implementation in `crates/kernel-launch` + `crates/runtimed/src/jupyter_kernel.rs`.
3. Per-kernel cleanup: socket files need to be removed on kernel shutdown AND orphan-reaped by `process_groups` if the kernel crashes.
4. Windows story: jupyter_client's IPC support uses named pipes — verify it works or keep TCP for Windows.

### Scope gate

Prerequisite: confirm ipykernel honors `transport: ipc` on all supported Python versions. If there's any gap, falls back to TCP there.

### Related

- #1772 (closed in favor of this)

quillaid:
## Summary

Replace TCP port allocation with Unix domain sockets (IPC transport) for Jupyter kernel ZMQ channels on Linux and macOS. Eliminates the TOCTOU race where TCP ports allocated by the daemon could be stolen before the kernel's `zmq.bind()`, causing "Address already in use" failures.

- Unix kernels now use IPC transport with socket paths like `/tmp/runt-nightly/kernel-{id}-ipc-{1..5}`
- No retry loop needed on Unix -- kernel_id uniqueness guarantees collision-free paths
- Windows retains the existing TCP retry loop unchanged
- IPC socket cleanup on shutdown, Drop, orphan reaping, and daemon startup
- `ipc_socket_dir()` uses `/tmp/` to stay within macOS 104-byte `sun_path` limit

Depends on runtimed/runtimed#301 (IPC `form_url` fix + `ipc-transport` feature). Uses path deps until upstream publishes.

Closes #1813

## Test plan

- [x] `cargo xtask clippy` passes
- [x] Nightly build + install succeeds
- [ ] Gremlin suite replay -- verify zero ZMQ port conflicts on Unix
- [ ] Manual: launch kernel, verify connection file has `"transport": "ipc"`
- [ ] Manual: verify socket files exist during kernel lifetime, removed after shutdown
- [ ] Windows CI: verify TCP retry loop still works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -161,7 +161,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1262,7 +1262,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1836,7 +1836,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


rgbkrk (commented):
One issue to fix before this is merge-ready: `ipc_socket_dir()` uses a global `/tmp/<cache_namespace>` root (`/tmp/runt` or `/tmp/runt-nightly` outside dev worktrees). The daemon singleton only proves this user/channel has no live daemon, but this path is shared across OS users. The first user to create the directory owns it, so another local user can fail to create kernel sockets there; if permissions are relaxed enough to allow writes, the startup sweep in `daemon.rs` can also remove another user’s live `kernel-*-ipc-*` sockets. Please include a per-user component or create a private 0700 temp directory before using it as the IPC socket root.

Refs: `crates/runtimed-client/src/lib.rs:253-260`, `crates/runtimed/src/daemon.rs:1012-1027`.

PR #2388, opened 2026-04-29
fix(runtimed): reserve kernel ports during launch

rgbkrk:
## Summary
- keep daemon-local kernel port reservations alive through the listener-to-kernel handoff window
- broaden the default kernel port range and per-xdist-worker test slices
- add focused tests for reservation claim/release and listener handoff behavior

## Verification
- cargo test -p runtimed jupyter_kernel::tests:: --lib
- cargo clippy -p runtimed --lib -- -D warnings
- cargo xtask lint --fix

## Notes
- A direct targeted Python run selected the two previously failing tests, but failed in the module health-check fixture before test bodies with Unexpected response: NoKernel after auto-launch reported Kernel started: OK. That is a different local failure mode than the CI EADDRINUSE port race.

### DIFF
--- a/crates/runtimed/src/jupyter_kernel.rs
+++ b/crates/runtimed/src/jupyter_kernel.rs
@@ -8,12 +8,14 @@
 //! task handles, request/response infrastructure, process lifecycle.  Does
 //! **not** hold queue, executing cell, or status — those live in `KernelState`.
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Superseded by #2391, which moves kernel port reservations into the daemon/coordinator instead of keeping process-local reservation state in runtime agents.

Issue #1598, opened 2026-04-07
Settings sync: notebook window's stale Automerge doc overwrites daemon's externally-applied changes


rgbkrk:
## Bug

When settings.json is edited externally (e.g., by a user or CLI tool), the daemon's file watcher correctly picks up the change and applies it to the Automerge settings doc. However, when a notebook window subsequently syncs, its stale local copy can overwrite the daemon's updated state via last-writer-wins merge semantics.

### Steps to reproduce

1. Daemon is running, notebook window is open — both have `default_python_env: "uv"` 
2. Externally edit `~/Library/Application Support/nteract-nightly/settings.json` to change `default_python_env` to `"conda"`
3. Daemon's `watch_settings_json` detects the change, calls `apply_json_changes` → Automerge doc now has `"conda"`
4. Daemon correctly saves only the Automerge binary (not the JSON mirror) per the anti-churn comment on line 981
5. A notebook window syncs with the daemon → its local doc still has `"uv"` → Automerge merge resolves to `"uv"` (window's value wins)
6. `persist_settings` fires in `sync_server.rs:75` → writes JSON mirror back with `"uv"`
7. The external edit is silently lost

### Expected

External edits to settings.json should be authoritative — the daemon should ensure that changes from `watch_settings_json` take precedence over stale window state during the next sync.

### Possible fixes

- When `apply_json_changes` detects external edits, mark the Automerge change with a higher-priority timestamp or use a separate mechanism to force the daemon's value during sync
- Have the watcher immediately broadcast the change to all connected windows so they update their local docs *before* their next sync round-trip
- Or: have `persist_settings` in `sync_server.rs` skip the JSON mirror write if the watcher recently applied changes (short cooldown)

### Context

This came up during MCP testing — I edited settings.json to switch from `uv` to `conda`, the daemon's watcher applied it, but a notebook window synced and reverted it back to `uv`. The `conda` change was silently lost.

### Relevant code

- `crates/runtimed/src/daemon.rs` line 884: `watch_settings_json` — correctly detects and applies external edits
- `crates/runtimed/src/sync_server.rs` line 75: `persist_settings` — writes JSON mirror on every sync message, potentially clobbering watcher's changes

rgbkrk:
Closing — fixed and covered by regression test `test_external_settings_json_edit_survives_settings_sync_ack` in `crates/runtimed/tests/integration.rs:415`.

[...]

PR #2390, opened 2026-04-29
feat(runtimed): add daemon-owned settings rpc channel

rgbkrk:
Adds a `Handshake::SettingsRpc` channel as a hardened slice of the settings ownership design in #1598. The daemon owns its in-memory `SettingsDoc` and `settings.json`; RPC clients receive a full `SyncedSettings` snapshot on connect and on every change, and send `SetSetting { key, value }` writes against the daemon's current state instead of pushing CRDT state.

Both channels run against the same `Arc<RwLock<SettingsDoc>>` and the same `settings_changed` broadcast, so peers on either side see each other's writes. The existing Automerge `SettingsSync` path is unchanged and remains the default for the app.

Refs #1598.

## Design decisions

- **Same `SettingsDoc`, two channels.** The new RPC handler shares the existing daemon settings document and persistence paths with `SettingsSync`, so RPC writes and Automerge writes converge through the same daemon-owned state.
- **Validate before mutating.** `SetSetting` now rejects unknown keys, bad scalar types, out-of-range durations/pool sizes, and malformed package lists before touching the shared document. If a key is valid in `SyncedSettings` but unsupported by the current `SettingsDoc::put_value` writer, the daemon returns a rejected ack instead of silently succeeding.
- **Snapshot push on the writer's own connection, before the ack.** The writer receives the post-write `Snapshot` before `SetSettingAck`, so set-and-read callers do not observe stale cached settings. Peers receive the same update through `settings_changed`.
- **Heads-equality guard.** Same-value writes do not persist or broadcast. This mirrors the Automerge path's guard against settings churn waking pool warmers unnecessarily.
- **Persist failures fail the ack.** RPC writes attempt both the Automerge binary and JSON mirror writes. The first persistence failure is surfaced as `SetSettingAck { ok: false, error }`.
- **Protocol version reflects the new handshake.** The wire protocol reports v5 for the added `SettingsRpc` channel while keeping v4 as the minimum accepted daemon protocol for upgrade compatibility.
- **Upgrade window closes Settings before daemon replacement.** The app's upgrade flow closes every webview except the upgrade window before upgrading the daemon, so an open Settings window does not keep a stale settings socket alive across the daemon swap.
- **Opt-in.** Existing Tauri/React settings code still uses `SyncClient`/`SettingsSync`; this PR adds the RPC path without flipping callers.

## Files

- `crates/notebook-protocol/src/connection/handshake.rs` - new `Handshake::SettingsRpc` variant and protocol v5 constant.
- `crates/notebook-protocol/src/connection/settings_rpc.rs` - `SettingsRpcServerMessage::{Snapshot, SetSettingAck}` and `SettingsRpcClientMessage::SetSetting`.
- `crates/runtimed/src/sync_server.rs` - `handle_settings_rpc_connection`, validation, snapshot fanout, and persistence error handling.
- `crates/runtimed/src/daemon.rs` - dispatch for the new handshake variant.
- `crates/runtimed-client/src/settings_rpc_client.rs` - `SettingsRpcClient` with `connect`, `get_snapshot`, `recv_snapshot`, and `set_setting`.
- `crates/runtimed/tests/integration.rs` - RPC round-trip coverage, peer broadcast coverage, JSON persistence coverage, and rejected-write coverage.

## Test plan

- [x] `cargo test -p runtimed sync_server::tests --lib`
- [x] `cargo test -p notebook-protocol connection::`
- [x] `cargo test -p runtimed test_settings_rpc_set_setting_round_trip --test integration`
- [x] `cargo clippy -p runtimed --test integration -- -D warnings`

## Out of scope

- No on-disk migration. Both `settings.automerge` and `settings.json` still get written.
- No Tauri/React caller migration. The app still uses the existing `SettingsSync` path.
- No telemetry-state split into `runtime-state.json`. The full canonical-JSON migration in #1598 covers that larger move.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7080,6 +7080,7 @@ dependencies = [
 name = "runtimed-client"
 version = "2.3.5"
 dependencies = [
+ "anyhow",
  "automerge",
  "base64 0.22.1",
  "chrono",
--- a/crates/notebook-protocol/src/connection.rs
+++ b/crates/notebook-protocol/src/connection.rs
@@ -6,6 +6,7 @@
 mod env;
 mod framing;
 mod handshake;
+mod settings_rpc;
 
 pub use env::{EnvSource, LaunchSpec, PackageManager};
 
@@ -18,6 +19,8 @@ pub use framing::{
 pub use handshake::{
     Handshake, NotebookConnectionInfo, ProtocolCapabilities, PROTOCOL_V4, PROTOCOL_VERSION,
 };
+
+pub use settings_rpc::{SettingsRpcClientMessage, SettingsRpcServerMessage};
 
 #[cfg(test)]
 mod tests {
--- a/crates/notebook-protocol/src/connection.rs
+++ b/crates/notebook-protocol/src/connection.rs
@@ -17,7 +17,8 @@ pub use framing::{
 };
 
 pub use handshake::{
-    Handshake, NotebookConnectionInfo, ProtocolCapabilities, PROTOCOL_V4, PROTOCOL_VERSION,
+    Handshake, NotebookConnectionInfo, ProtocolCapabilities, PROTOCOL_V4, PROTOCOL_V5,
+    PROTOCOL_VERSION,
 };
 
 pub use settings_rpc::{SettingsRpcClientMessage, SettingsRpcServerMessage};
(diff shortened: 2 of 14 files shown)


rgbkrk:
Closing this direction. The better product shape is to keep SettingsSync/Automerge as the live settings transport and fix the daemon-owned settings.json mirror clobber/feedback-loop behavior in a new PR.

PR #2420, opened 2026-04-30
fix(runt-mcp): capture env_source before kernel shutdown in restart_kernel

quillaid:
## Summary

- `restart_kernel` was reading `env_source` from `detect_package_manager()` **after** shutting down the kernel. After shutdown, the daemon clears RuntimeStateDoc's `env_source`, causing `detect_package_manager` to fall through to the `Uv` default — silently losing conda/pixi context.
- Now captures `env_source` from RuntimeStateDoc **before** sending `ShutdownKernel`, alongside the existing `pre_shutdown_kernel_type` capture. Derives the scoped auto-detect string (`auto:conda`, `auto:pixi`, `auto:uv`) from the captured value, matching the existing logic in `apply_dependency_changes`'s "restart" branch.
- Falls back to `detect_package_manager` only when no pre-shutdown `env_source` is available (e.g. daemon was freshly started with no kernel running).

Found via `concurrent-deps` gremlin: after MCP reconnect + `restart_kernel`, conda notebooks regressed to `uv:prewarmed` with empty deps.

## Test plan

- [ ] `cargo xtask lint --fix` passes
- [ ] Create a conda notebook via MCP, add deps, call `restart_kernel` — verify `env_source` remains `auto:conda`
- [ ] Create a pixi notebook, add deps, call `restart_kernel` — verify `auto:pixi` preserved
- [ ] `restart_kernel` with no prior kernel (fresh session) still falls back to metadata-based detection
- [ ] Replay `concurrent-deps` gremlin against nightly with this fix

### DIFF
--- a/crates/runt-mcp/src/tools/kernel.rs
+++ b/crates/runt-mcp/src/tools/kernel.rs
@@ -122,8 +122,9 @@ pub async fn restart_kernel(
         use notebook_protocol::connection::{EnvSource, PackageManager};
         if let Some(ref prev) = pre_shutdown_env_source {
             // Derive the scoped auto-detect from the previous env_source,
-            // preserving the package manager family. Same logic as
-            // apply_dependency_changes "restart" branch in deps.rs.
+            // preserving the package manager family. The daemon's auto:*
+            // variants re-resolve through the normal launch priority
+            // (project files → inline deps → captured envs → prewarmed).
             match EnvSource::parse(prev) {
                 EnvSource::Prewarmed(PackageManager::Conda) => "auto:conda".to_string(),
                 EnvSource::Prewarmed(PackageManager::Pixi) => "auto:pixi".to_string(),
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
No blocking findings from me. I think the product direction here should be that `restart_kernel` preserves the user’s selected runtime/package-manager family, but still re-resolves the concrete environment from current notebook/project state.

So `conda:*` should probably restart through `auto:conda`, `pixi:*` through `auto:pixi`, `uv:*` through `auto:uv`, and Deno stays `deno`. That lets restart pick up project files, inline deps, captured envs, and metadata changes using the normal launch priority, while avoiding the bad fallback where conda/pixi silently become UV. My only worry is if auto doesn't bring us back to the prewarmed or inline env they were on.

One small cleanup: the comment saying this is the “same logic as `apply_dependency_changes`” is a little misleading.

PR #2430, opened 2026-04-30
fix(mcp): harden concurrency diagnostics

rgbkrk:
## Summary

Supersedes #2427 with a narrower concurrency hardening pass: remove lock-across-child-call convoy risks, add bounded readiness waits only at MCP entry points, and add diagnostics that identify which layer stalled.

## Lab Evidence

Reviewed `/Users/kyle/Downloads/mcp-deadlock-diagnostic.tar.gz`.

- Five reports (`conda`, `conda-sequential`, `conda-lifecycle`, `media`, `restart-resilience`) were killed after 160+ minutes with `turns: 0`, no tool errors, and MCP connection deadlock.
- The archive analysis classifies the hang as child readiness wait: proxy/child processes had essentially no CPU after the initial forwarded call.
- Daemon logs show conda `LaunchKernel` requests completing in about 5-6s for the affected notebooks, so the replacement PR should not be only a broad parent timeout.
- PR #2427's patch globally timed out `DocHandle::wait_for_status` and wrapped child calls, but did not remove proxy state-lock convoy hazards or include latest `SyncStatus` diagnostics.

## Changes

- Refactor the parent MCP proxy to snapshot a cloned child `Peer` plus child generation under short locks, then perform child tool/resource/list awaits outside `self.state`.
- Guard cache/restart commits with child-generation checks and cancel old children outside the state write lock.
- Add parent slow-call diagnostics with operation/target, child generation, elapsed time, transport-closed state, and restart count.
- Add bounded `DocHandle` readiness helpers while leaving existing unbounded `await_*` semantics unchanged.
- Use bounded readiness only in MCP session establishment and daemon-watch rejoin paths.
- Fix `create_notebook` so runtime info collection does not hold `server.session` while awaiting.
- Add slow-tool logging in child `runt-mcp` dispatch.
- Add daemon peer-request slow logging with request label, correlation id, peer id, notebook id, elapsed time, writer queue depth, and response kind.

## Test Plan

- `cargo test -p notebook-sync`
- `cargo test -p runt-mcp`
- `cargo test -p runt-mcp-proxy`
- `cargo test -p runtimed --test rpc_routing`
- `cargo xtask lint --fix`

Note: `cargo test -p runtimed --test rpc_routing` initially hit the expected missing gitignored renderer plugin artifacts. I ran `cargo xtask wasm` to rebuild them, then reran the target successfully.

### DIFF
--- a/crates/notebook-sync/src/handle.rs
+++ b/crates/notebook-sync/src/handle.rs
@@ -30,9 +30,10 @@
 //! ```
 
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 use automerge::{AutoCommit, ReadDoc, Value};
-use log::debug;
+use log::{debug, warn};
 use tokio::sync::{mpsc, oneshot, watch};
 
 use notebook_protocol::protocol::{NotebookRequest, NotebookResponse};
@@ -663,6 +664,29 @@ impl DocHandle {
         }
     }
 
+    async fn wait_for_status_timeout<F>(
+        &self,
+        label: &str,
+        predicate: F,
+        timeout: Duration,
+    ) -> Result<(), SyncError>
+    where
+        F: Fn(&SyncStatus) -> bool,
+    {
+        match tokio::time::timeout(timeout, self.wait_for_status(predicate)).await {
+            Ok(result) => result,
+            Err(_) => {
+                warn!(
+                    "[notebook-sync] {label} timed out after {:?} for {} with latest status: {:?}",
+                    timeout,
+                    self.notebook_id,
+                    self.status()
+                );
+                Err(SyncError::Timeout)
+            }
+        }
+    }
+
     /// Return the latest connection/bootstrap status.
     pub fn status(&self) -> SyncStatus {
         self.status_rx.borrow().clone()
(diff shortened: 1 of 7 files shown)


rgbkrk:
## Bedrock Opus Review

Reviewed with `global.anthropic.claude-opus-4-6-v1` via Claude Code on the branch diff against `origin/main`.

Disposition summary:

- `confirmed-fix`: Proxy restart notification handling had a redundant `.or(divergence_tx)` expression. Fixed in `20a9f9e5`.
- `confirmed-fix`: Added targeted proxy coverage for no-child peer snapshots and stale-refresh no-op behavior. Fixed in `20a9f9e5`.
- `confirmed-defer`: `save_tool_cache` remains under the generation-checked state write lock. This is intentionally deferred because moving the disk write outside the guarded commit can let an older generation overwrite the on-disk cache after a newer child refresh.
- `non-issue`: `SyncError::Timeout` already exists in `crates/notebook-sync/src/error.rs`.
- `non-issue`: Slow-call diagnostic threshold tests were considered optional observability coverage.

Rerun result: no actionable proxy findings remain; readiness/daemon slice had no actionable bugs or regressions.

PR #2448, opened 2026-05-01
fix(mcp): clear session immediately on daemon disconnect

quillaid:
## Summary

- **Clear dead sessions immediately on daemon disconnect** instead of leaving them for tool calls to hang on. The `MarkDisconnected` path now drops the `NotebookSession` and records a `SessionDropInfo` with `reason=Disconnected`, so tool calls get a context-rich error with reconnect hint instead of hanging on a dead `DocHandle`.
- **Automatic rejoin via `disconnect_target`**: the notebook path (file-backed) or UUID (ephemeral) is saved when the session is cleared, so the watch loop can rejoin on the next `Connected`/`Upgraded` event without requiring the proxy's `NTERACT_MCP_REJOIN_NOTEBOOK` env var.
- **No regression on #2088**: `disconnect_target` is cleared on successful rejoin, preventing spurious heartbeat rejoins that would keep rooms alive indefinitely.

Fixes the "MCP reconnection failure after daemon restart" finding from gremlin run `20260501-baseline-bc7df34` — 7/14 gremlins were affected (4 cascading errors, 3 indefinite stalls).

## Root cause

When the daemon disconnects, `daemon_watch::classify()` returned `MarkDisconnected` which only set `was_disconnected = true` but left the dead `NotebookSession` intact. During the `DaemonConnection` reconnect backoff window (100ms → 200ms → ... → 30s max), tool calls acquired the dead `DocHandle` via `require_handle!()` and either:
1. **Hung indefinitely** — the `send_request` call blocked on a dead TCP connection with no timeout
2. **Cascaded "No active notebook session" errors** — if the rejoin failed (3×1s retries exhausted), the session was cleared and all subsequent tool calls got the generic error

## Test plan

- [x] 16 `daemon_watch` unit tests pass (4 new tests for disconnect_target behavior)
- [x] 116 total `runt-mcp` tests pass
- [ ] CI (Clippy, Rust tests, JS tests, WASM, E2E)
- [ ] Gremlin replay against nightly with this fix installed

### DIFF
--- a/crates/runt-mcp/src/daemon_watch.rs
+++ b/crates/runt-mcp/src/daemon_watch.rs
@@ -78,6 +78,7 @@ fn classify(
     initial_target: &Option<String>,
     has_session: bool,
     was_disconnected: bool,
+    disconnect_target: &Option<String>,
 ) -> WatchDecision {
     match event {
         DaemonEvent::Upgraded { previous, current } => {
@@ -91,6 +92,11 @@ fn classify(
                 WatchDecision::RejoinInitial(t.clone())
             } else if has_session {
                 WatchDecision::RejoinContinuation
+            } else if let Some(t) = disconnect_target.as_ref() {
+                // Session was cleared on disconnect to prevent tool
+                // calls from hanging on a dead DocHandle. Rejoin
+                // using the saved target.
+                WatchDecision::RejoinInitial(t.clone())
             } else {
                 WatchDecision::NoOp
             }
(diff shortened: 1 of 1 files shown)


rgbkrk:
After a disconnect this PR clears `session` and starts an automatic `RejoinInitial` from `disconnect_target`. While `rejoin()` is awaiting connect/initial-load, tool calls still see no active session and can legitimately open or create another notebook. When the background rejoin completes it writes `Some(new_session)` unconditionally, replacing that newer user-selected session with the old disconnected notebook. The same guard you added before `classify()` needs to exist at commit time too, for example by only installing the rejoined session if `session` is still `None` or still matches the disconnected target/generation.

The intended behavior is good, and the classifier coverage looks thoughtful, but the final session write needs to avoid clobbering a session established while auto-rejoin was in flight.

PR #2460, opened 2026-05-01
docs(skills): add automerge-sync and mcp-session-lifecycle Claude skills

quillaid:
## Summary

- Adds two new `.claude/skills/` entries built from deep reading of upstream automerge (`automerge/automerge`), automerge-repo (`automerge/automerge-repo`), and samod (`alexjg/samod` subduction branch) source code.
- **`automerge-sync`**: `sync::State` field-by-field semantics (all 13 fields, persistence contract), bloom filter negotiation, in-flight suppression, message version negotiation, reconnection patterns from three upstream repos, and the nteract catch_unwind + cell-count guard recovery pattern.
- **`mcp-session-lifecycle`**: Three-layer proxy/MCP/daemon architecture, the watch loop classify/act state machine with full event-decision matrix, session-write guard race, heartbeat vs real disconnect (#2088), file-backed vs ephemeral rejoin strategies, and the concurrent MCP north-star note.
- Updates `AGENTS.md` with a Claude-specific skills discovery section.

## Test plan

- [ ] Skills are markdown-only; no code changes
- [ ] Verify `cargo xtask lint --fix` passes (confirmed clean)
- [ ] Verify AGENTS.md references are accurate

### DIFF
--- a/.claude/skills/automerge-sync/SKILL.md
+++ b/.claude/skills/automerge-sync/SKILL.md
@@ -283,9 +283,13 @@ Don't reset sync state just because local mutations happened. Reset it
 when the *transport* breaks (reconnect, panic recovery). Resetting
 after every mutation would make every sync round a full exchange.
 
-## Confirm Sync Design
+## Causal Ordering: confirm_sync and required_heads
 
-`confirm_sync` verifies the daemon has merged specific local changes:
+Two mechanisms ensure the daemon sees client edits before acting on them:
+
+### confirm_sync (client-side wait)
+
+`confirm_sync` blocks the client until the daemon has merged specific heads:
 
 1. Caller captures current heads via `DocHandle::confirm_sync()`
 2. A `ConfirmSync` command goes to the sync task with target heads
@@ -296,6 +300,39 @@ after every mutation would make every sync round a full exchange.
 
 This is non-blocking -- the frame loop keeps draining while the waiter
 resolves in the background.
+
+### required_heads (daemon-side wait, preferred)
+
+`required_heads` moves the wait to the daemon, eliminating the client-side
+round-trip:
+
+1. Client captures current heads via `DocHandle::current_heads_hex()`
+2. Client sends the request with `required_heads` in the envelope
+3. Daemon's `wait_for_required_heads()` checks if the notebook doc contains
+   all listed change hashes (containment check, not equality)
+4. If not yet present, subscribes to `changed_tx` and polls until all
+   heads arrive or 10s timeout
+5. Only then does the daemon evaluate the request
+
+**Why required_heads is better:**
+- No client-side blocking -- the request is sent immediately
+- The daemon can process other requests while waiting
+- Works correctly even if the sync stream is slow (the request just
+  queues behind the sync convergence)
+- Replaces the old `confirm_sync` → `execute_cell` two-step with a
+  single `execute_cell` request that carries its causal precondition
+
+**Used by:** `execute_cell`, `run_all_cells` (both MCP and frontend).
+Frontend flushes pending sync before capturing heads to minimize the
+daemon-side wait.
+
+### Which to use
+
+| Scenario | Use |
+|----------|-----|
+| Execute / run-all (need source synced) | `required_heads` via `send_request_after_heads` |
+| General "is my edit synced?" check | `confirm_sync` (still available) |
+| Save (daemon reads doc directly) | Neither -- daemon reads its own doc copy |
 
 ## Key Source Files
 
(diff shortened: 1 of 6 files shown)


.claude/skills/automerge-sync/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+### Version Negotiation
+
+V1 is the original format. V2 allows compressed document encoding (faster
+first sync). Backward-compatible: V1 messages append a `MessageFlags`
+section that old implementations ignore. New implementations read the flags
+to discover V2 support. Once discovered, subsequent messages use V2.
+
+## nteract's Sync Architecture
+
+nteract runs two sync streams over one socket connection:
```
rgbkrk: This should include PoolStateSync now. The notebook socket carries notebook sync (`0x00`), RuntimeStateDoc sync (`0x05`), and PoolDoc sync (`0x06`); `contributing/protocol.md` also lists `0x06` on both the frontend relay and daemon notebook peer. If this section is specifically scoped to `notebook-sync::SharedDocState`, it would help to say that explicitly and call out PoolStateSync as the separate frontend/daemon stream.

.claude/skills/automerge-sync/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+2. Client sends the request with `required_heads` in the envelope
+3. Daemon's `wait_for_required_heads()` checks if the notebook doc contains
+   all listed change hashes (containment check, not equality)
+4. If not yet present, subscribes to `changed_tx` and polls until all
+   heads arrive or 10s timeout
+5. Only then does the daemon evaluate the request
+
+**Why required_heads is better:**
+- No client-side blocking -- the request is sent immediately
+- The daemon can process other requests while waiting
```
rgbkrk: I think this overstates the current `required_heads` behavior. `wait_for_required_heads()` runs inside the per-peer request worker before dispatching the request, so later requests from the same peer remain queued behind the waiting request. The important invariant is that the main peer loop still drains sync frames, so the required heads can arrive, and other peers are unaffected. Could this bullet say that instead of "the daemon can process other requests while waiting"?

.claude/skills/automerge-sync/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+- Works correctly even if the sync stream is slow (the request just
+  queues behind the sync convergence)
+- Replaces the old `confirm_sync` → `execute_cell` two-step with a
+  single `execute_cell` request that carries its causal precondition
+
+**Used by:** `execute_cell`, `run_all_cells` (both MCP and frontend).
+Frontend flushes pending sync before capturing heads to minimize the
```
rgbkrk: Small ordering correction from #2457: the frontend `NotebookClient` captures the current heads first, then triggers the optional fire-and-forget `flush()`. That preserves the causal baseline in `required_heads` while nudging the sync stream to deliver it. This sentence currently says flush-before-capture.

.claude/skills/automerge-sync/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+daemon-side wait.
+
+### Which to use
+
+| Scenario | Use |
+|----------|-----|
+| Execute / run-all (need source synced) | `required_heads` via `send_request_after_heads` |
+| General "is my edit synced?" check | `confirm_sync` (still available) |
+| Save (daemon reads doc directly) | Neither -- daemon reads its own doc copy |
```
rgbkrk: This row seems too broad. The daemon save handler reads the daemon doc copy, but client-initiated save paths still need to ensure local edits have reached the daemon; MCP currently calls `confirm_sync()` before `SaveNotebook`. Maybe distinguish daemon-internal save handling from client-triggered save requests.

.claude/skills/sync-protocol-patterns/SKILL.md
```diff
@@ -0,0 +1,235 @@
...
+
+**Design space:** nteract currently takes the "clean slate" approach for
+sync state (like samod) but preserves session identity (like automerge-repo
+preserves sync state). If latency on reconnect becomes a problem, the
+automerge-repo approach (preserve `shared_heads` via encode/decode) could
+reduce the initial sync burst.
+
+## When Adding a New Sync Stream
+
+If you need to add a new Automerge-synced document to the protocol (like
```
rgbkrk: PoolStateSync is a confusing example for this checklist because it already exists and is not represented as another `notebook-sync::SharedDocState` doc+peer_state pair. The frontend owns the pool-doc sync state, while the daemon peer loop carries `pool_peer_state`. Consider using a hypothetical new frame type here, or splitting the checklist by which side owns the sync state.

PR #2470, opened 2026-05-02
fix(kernel-env): pin Python version in sync_dependencies to prevent version drift

### LINKED ISSUE #2: [sidecar] Figure out why Plotly's FigureWidget is Twigging Out
With Plotly 6.x, FigureWidget has strange behavior with anywidget. Reproduction code:

```python
import plotly.graph_objs as go
fig = go.FigureWidget()
fig.add_scatter(y=[2, 1, 4, 3])
fig
```

It just flickers over and over. That's likely on our end in sidecar and/or the nteract elements but it's weird that drawdata, quak, and lonboard all work really well.

https://github.com/user-attachments/assets/20eac9cc-22e7-4885-9450-6c49cc3b3e67

quillaid:
## Summary
- `sync_dependencies()` now detects the installed Python version from `conda-meta/` and adds it as a hard solver spec (`python=X.Y.Z`), preventing the solver from upgrading/downgrading Python when installing new packages into an existing env
- Also adds `python-gil` spec to prevent switching to the free-threaded build, matching what `install_conda_env` already does for fresh env creation
- Adds `detect_installed_python_version()` helper that reads `conda-meta/python-*.json` (cheaper than spawning `python --version`)

## Context
The conda-sequential gremlin caught a bug: after 2 consecutive `add_dependency(after="restart")` calls, baseline packages become unimportable due to Python version mismatch. Root cause: `sync_dependencies()` only passed user deps + base packages as solver specs, leaving the installed Python as a soft preference (`locked_packages`). The solver could swap Python to satisfy new deps, producing site-packages for the wrong version.

This is fix #2 of 3 from the root cause analysis. Fixes #1 (pin in capture) and #3 (pin in pool-for-inline) will follow.

## Test plan
- [x] 3 new unit tests for `detect_installed_python_version` (normal, missing, free-threaded)
- [x] All 18 existing conda tests pass
- [ ] CI green
- [ ] Replay conda-sequential gremlin after nightly install

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -904,6 +904,14 @@ pub async fn sync_dependencies(
 
     let match_spec_options = ParseMatchSpecOptions::strict();
 
+    // Pin the installed Python version so the solver cannot upgrade or
+    // downgrade it. Without this, the solver treats Python as a soft
+    // preference (locked_packages) and can swap it out to satisfy new
+    // deps — producing site-packages for the wrong Python version.
+    // Also enforce `python-gil` to prevent switching to the
+    // free-threaded build. See: conda-sequential pinning bug.
+    let installed_python_version = detect_installed_python_version(&env.env_path);
+
     // Always include base runtime packages — the solver only returns packages
     // needed to satisfy specs, and locked_packages are "preferred" not "required".
     // Without these, the Installer will remove ipykernel etc from the env.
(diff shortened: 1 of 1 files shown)


crates/kernel-env/src/conda.rs
```diff
@@ -915,6 +923,25 @@ pub async fn sync_dependencies(
...
+        info!("Pinning Python to installed version: {}", py_ver);
+        specs.push(MatchSpec::from_str(
+            &format!("python={}", py_ver),
+            match_spec_options,
+        )?);
+        // Prevent the solver from switching to free-threaded Python.
+        // install_conda_env enforces this via should_enforce_gil_python
+        // but sync_dependencies was missing it.
+        if !py_ver.ends_with('t') {
+            specs.push(MatchSpec::from_str(CONDA_GIL_SELECTOR, match_spec_options)?);
```
rgbkrk, line 943: [P2] Free-threaded conda envs get forced back to GIL Python during sync

This preserves the existing install-time contract for `python=3.14t` in `build_spec_strings`, but `sync_dependencies` now derives only the plain package version from `conda-meta` and then adds `python-gil` whenever that string does not end in `t`. The new free-threaded detector test documents that `conda-meta` stores `3.14.4`, not `3.14t`, so this branch will always add `python-gil` for an existing free-threaded env. A later dependency sync can therefore solve/install a GIL build into an env the user explicitly created as free-threaded, recreating the same Python/ABI drift class this PR is trying to prevent.

The fix should preserve the installed selector package instead of inferring from the version filename: keep `python-freethreading` when that is present in the prefix, otherwise pin `python-gil`.

PR #2472, opened 2026-05-02
feat(runt-mcp): park sessions on notebook switch to prevent eviction

### LINKED ISSUE #2243: feat(runt-mcp): warm-connection LRU cache for multi-notebook agent workflows
## Summary

The 30s idle-eviction timer in `notebook_sync_server/peer.rs` fires when the last Unix-socket peer disconnects. For MCP-driven workflows, `connect_notebook` to a new notebook drops the MCP session's socket connection to the old one, which takes the old room's peer count to zero and starts the timer. 30 seconds later the kernel is gracefully shut down. An agent that circles back to the old notebook finds it dead and (correctly) reconnects, which opens a *new* kernel — losing in-memory state.

This isn't a daemon bug. `keep_alive_secs` is user-configurable, clamped 5s–7 days, default 30s. That default is correct for "user closed a tab." It's wrong for "agent is thinking for 40s before touching notebook A again."

## Scope clarification

In the original report I called MCP clients "websocket peers." They aren't — MCP talks to the daemon over the Unix socket (named pipes on Windows). Same peer-counting mechanic on the daemon side, but the transport is local IPC.

## Proposed fix: MCP-side warm connection cache

No daemon changes. The MCP server already holds the socket connection for its active session; extend it to hold a small LRU of sessions.

```rust
pub struct NteractMcp {
    foreground: RwLock<Option<Arc<NotebookSession>>>,     // target of tool calls
    warm_sessions: Mutex<LruCache<String, Arc<Session>>>, // includes foreground
}
```

### Flow

- `connect_notebook(id)`:
  - Cache hit → promote to MRU, set foreground. No new socket, no daemon round-trip.
  - Cache miss → open new session, insert (evict oldest if at cap), set foreground.
- LRU eviction drops its `Arc`, socket closes, daemon peer count on that room drops, daemon's 30s timer runs as normal for *that one room*.
- MCP process exits → every warm `Arc` drops → daemon sees peer count hit zero across the board → normal eviction.

### New tool: `disconnect_notebook(id)`

Mirror of `connect_notebook`. Removes from the warm cache. Releases the peer hold immediately instead of waiting for LRU eviction.

Agents that "know they're done" with a notebook can call this; agents that don't will still get reasonable behavior from the LRU.

### Does the agent trust itself to shut kernels?

The proposal above keeps kernel shutdown as a side-effect of dropping the peer hold — the daemon's 30s timer still runs when the last peer for a room disconnects. That's fine when the room has no other peers, and correct when a desktop-app user is also attached (we don't want the agent force-closing a kernel the human is using).

We could add a `shutdown_kernel(id)` tool for explicit "kill the kernel now regardless of other peers." Useful for multi-agent stress scenarios or explicit cleanup, but gives agents a footgun. Hold on that until we see a workflow that needs it.

### Interaction with desktop app

Subtle but worth flagging. If a human has notebook A open in the desktop app and an agent also has A warm, closing the tab leaves the daemon-side room alive (MCP still attached). The kernel survives the tab close. That's probably desirable for the "agent keeps working while the human steps away" case. Surface in release notes.

## Open design questions

1. **Cap size default.** Typical agent workflow shows 3–5 notebooks in working set. Default 5 caps kernel RAM at a reasonable bound while covering the common case. Configurable via MCP-side env or settings.
2. **Background sync cost.** Every warm session runs a sync loop to process daemon broadcasts. Cheap per-session but not zero; worth a measurement pass before raising the cap much.
3. **`shutdown_kernel` tool?** Deferred. Add only if a workflow needs it.

## Out of scope

- Changes to `keep_alive_secs` defaults. The existing default is correct for UI peers; the MCP-side cache handles the MCP-peer case. Keep them orthogonal.
- Daemon-side awareness of "this peer is an MCP session." The fix works entirely on the MCP side.

## Priority

Not a release blocker for the upcoming nightly → stable promotion. Ship after, once we've measured per-session sync overhead and picked a default cap size. The workaround today: users bump `keep_alive_secs` in settings, or agents accept that long idle gaps kill kernels.

## Source of truth

`crates/runt-mcp/src/` for the MCP-side cache and new tool. `crates/runtimed/src/notebook_sync_server/peer.rs` around the eviction-timer setup for reference (no changes needed there).

quillaid:
## Summary

- **Park sessions instead of dropping them** on notebook switch. When an MCP client calls `connect_notebook(A)` → `connect_notebook(B)`, session A is moved to a `HashMap<notebook_id, NotebookSession>` instead of being dropped. The peer connection stays alive so the daemon's peer count stays > 0 and the 30-second eviction timer never fires.
- **Resume parked sessions** on switch-back. `connect_notebook(A)` checks the parked map first and returns `"resumed": true` without opening a new daemon connection. Works for `connect_notebook`, `open_notebook`, and `create_notebook`.
- **Clear parked sessions on daemon disconnect** (in `daemon_watch.rs`) to prevent tool calls from hanging on dead DocHandles after daemon restart.

## Motivation

Batch-12 concurrency testing surfaced kernel state loss when the session-switcher gremlin cycled `A→B→A`. The old `disconnect_previous_session` dropped the peer connection, starting the 30-second eviction timer. Under I/O contention at batch-12, the cycle exceeded 30s, causing room eviction and kernel restart. Ephemeral notebooks lost both cells and kernel permanently.

Session parking is the lowest-risk fix: changes only `runt-mcp` (no daemon changes), and the daemon's existing eviction logic continues working correctly — it just never fires because peer count stays positive.

## Files changed

| File | Change |
|------|--------|
| `crates/runt-mcp/src/tools/session.rs` | Replace `disconnect_previous_session` with `park_previous_session` + `take_parked_session`; update all three connect/open/create paths |
| `crates/runt-mcp/src/lib.rs` | Add `parked_sessions` HashMap field, accessor, shutdown cleanup |
| `crates/runt-mcp/src/daemon_watch.rs` | Accept `parked_sessions` param; clear on `MarkDisconnected` |
| `crates/runt/src/main.rs` | Pass `parked_sessions` to `daemon_watch::watch()` |

## Test plan

- [x] `cargo test -p runt-mcp` — 116 tests pass
- [x] `./scripts/install-nightly` — builds, installs, daemon starts
- [ ] Gremlin suite replay at batch-12 — session-switcher gremlin should no longer lose kernel state

Closes #2243


rgbkrk:
No blocking correctness findings.

This is a simpler version of issue #2243’s MCP-side warm connection cache. It should solve the core A -> B -> A failure because park_previous_session moves the old NotebookSession into parked_sessions instead of dropping the peer, and switch-back resumes it by notebook id.

Main tradeoff: this is an unbounded HashMap, not the bounded LRU proposed in #2243. There is also no disconnect_notebook(id) release path. So the design is fine as a minimal fix, but it means every notebook touched by a long-lived MCP process can keep a daemon peer and kernel alive until daemon disconnect or MCP shutdown. That is not a regression from current behavior, but it is the resource-control piece the issue explicitly called out.

Other observations:

Daemon disconnect cleanup is handled: parked sessions are cleared so stale DocHandles do not hang future tool calls.
The active-session call site in runt compiles with the new watcher argument.

PR #2478, opened 2026-05-02
fix(kernel-env): detect Python version mismatch in conda:env_yml envs, guard user-managed envs

quillaid:
## Summary

- When `environment.yml` requests e.g. `python=3.12` but a cached conda env already has Python 3.14, `sync_dependencies` would pin the installed version and silently ignore the yaml constraint
- Adds `installed_python_matches_constraint()` to `kernel-env` that compares installed vs requested major.minor
- Both `conda:env_yml` code paths (`launch_kernel.rs` + `metadata.rs`) now check for version mismatch before syncing
- **Guards user-managed envs:** only daemon-owned cache envs are auto-rebuilt on mismatch; user-managed envs (from `prefix:` field or pre-existing named envs) surface an error with a remediation command instead of being silently deleted

## Details

**Root cause:** `sync_dependencies` correctly pins the installed Python version for dependency resolution, but the `conda:env_yml` path calls it even when the existing env has a different Python than requested. The installed 3.14 gets pinned, the yaml's `python=3.12` is silently ignored.

**Version mismatch detection:** before entering the "existing env → sync deps" branch, check `installed_python_matches_constraint()`. If major.minor doesn't match, the env needs rebuilding. Fail-open: if either version can't be determined, returns `true` to avoid unnecessary rebuilds.

**Daemon-owned vs user-managed:** tracks `is_daemon_owned_env` based on how the conda prefix was resolved:

| Source | Owned by | On mismatch |
|--------|----------|-------------|
| `prefix:` field | User | Surface error with remediation command |
| `name:` found by `find_named_conda_env` | User | Surface error with remediation command |
| `name:` fallback to `default_conda_envs_dir` | Daemon | Auto-rebuild (remove + recreate) |
| Hash-based cache path (no name/prefix) | Daemon | Auto-rebuild |

## Test plan

- [x] 3 unit tests: constraint match, mismatch, fail-open (`cargo test -p kernel-env -- python_constraint`)
- [x] `cargo check -p runtimed -p kernel-env` clean
- [ ] CI green
- [ ] `codex review --base main` clean

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -1144,7 +1144,7 @@ fn build_spec_strings(deps: &CondaDependencies) -> Vec<String> {
 ///
 /// This is cheaper than spawning `python --version` and works even when the
 /// environment's Python is broken or missing from PATH.
-fn detect_installed_python_version(env_path: &std::path::Path) -> Option<String> {
+pub fn detect_installed_python_version(env_path: &std::path::Path) -> Option<String> {
     let meta_dir = env_path.join("conda-meta");
     let entries = std::fs::read_dir(&meta_dir).ok()?;
     for entry in entries.flatten() {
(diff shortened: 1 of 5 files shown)


rgbkrk (commented):
Blocking finding:

`environment.yml` Python mismatch handling can now delete an existing target env without the existing build-approval gate. `missing_conda_env_yml_decision` returns `None` as soon as `conda_python_path(&conda_prefix).exists()` is true, so an already-built named or explicit-prefix env skips the `AwaitingEnvBuild` user decision path. The new launch path then treats a version mismatch as rebuildable and calls `tokio::fs::remove_dir_all(&conda_prefix).await` from both auto-launch and explicit `LaunchKernel` (`crates/runtimed/src/notebook_sync_server/metadata.rs:3271` and `crates/runtimed/src/requests/launch_kernel.rs:1005`).

That means an `environment.yml` with `name:` or `prefix:` pointing at an existing user-managed conda env can be removed just by opening/launching the notebook if its Python version differs. Please route this through the same approval/decision surface as an unbuilt env, or rebuild only a daemon-owned/cache env path that is safe to delete.

rgbkrk (commented):
Blocking finding:

`environment.yml` Python mismatch handling can now delete an existing target env without the existing build-approval gate. `missing_conda_env_yml_decision` returns `None` as soon as `conda_python_path(&conda_prefix).exists()` is true, so an already-built named or explicit-prefix env skips the `AwaitingEnvBuild` user decision path. The new launch path then treats a version mismatch as rebuildable and calls `tokio::fs::remove_dir_all(&conda_prefix).await` from both auto-launch and explicit `LaunchKernel` (`crates/runtimed/src/notebook_sync_server/metadata.rs:3271` and `crates/runtimed/src/requests/launch_kernel.rs:1005`).

That means an `environment.yml` with `name:` or `prefix:` pointing at an existing user-managed conda env can be removed just by opening/launching the notebook if its Python version differs. Please route this through the same approval/decision surface as an unbuilt env, or rebuild only a daemon-owned/cache env path that is safe to delete.

quillaid:
Addressed the blocking finding in cf568b6e:

**fix(runtimed): guard conda env deletion to daemon-owned paths only**

The `remove_dir_all` call is now guarded by `is_daemon_owned_env`, which tracks how the conda prefix was resolved:

| Source | `is_daemon_owned_env` | On mismatch |
|--------|----------------------|-------------|
| `prefix:` field | `false` | Surface error with remediation command |
| `name:` found by `find_named_conda_env` | `false` | Surface error with remediation command |
| `name:` fallback to `default_conda_envs_dir` | `true` | Auto-rebuild (daemon created this env) |
| Hash-based cache path | `true` | Auto-rebuild |

For user-managed envs, the daemon now writes an `Error` lifecycle with details like:
> Conda env "/home/user/miniconda3/envs/foo" has Python 3.14 but environment.yml requests python=3.12. This is a user-managed env — rebuild it manually with: conda env remove -n foo && conda env create -f environment.yml

Both auto-launch (`metadata.rs`) and explicit `LaunchKernel` (`launch_kernel.rs`) paths are guarded. The explicit path uses `reset_starting_state_with_outcome` which also clears `env.progress`.

PR #2482, opened 2026-05-02
feat(linux): add AppImage release installer

quillaid:
## Summary
- Add `scripts/install-linux-release`, a per-user Linux installer for published nteract release artifacts.
- Install the AppImage plus `runt`, `runtimed`, and `nteract-mcp` sidecars into a durable user data directory.
- Link commands into `~/.local/bin`, run `runt daemon doctor --fix`, and optionally install `npm:@nteract/pi` when `pi` is available.

## Why
This is a first step toward the future one-line install path:

```bash
curl --proto '=https' --tlsv1.2 -sSf https://sh.nteract.io | sh
```

The script defaults to GitHub's latest non-prerelease release via `/releases/latest/download`, while still supporting explicit `--tag` and local `--from-dir` installs for testing.

## Verification
```bash
bash -n scripts/install-linux-release
scripts/install-linux-release --from-dir ~/.local/share/nteract/stable --prefix /tmp/nteract-install-test --bin-dir /tmp/nteract-install-bin --no-pi --no-start
cargo xtask lint --fix
```

### DIFF
--- a/scripts/install-linux-release
+++ b/scripts/install-linux-release
@@ -1,0 +1,254 @@
+#!/usr/bin/env bash
+# Install a published nteract Linux release for the current user.
+#
+# Intended future entrypoint:
+#   curl --proto '=https' --tlsv1.2 -sSf https://sh.nteract.io | sh
+#
+# Installs the AppImage plus runt/runtimed/nteract-mcp sidecars into a durable
+# per-user directory, links commands into ~/.local/bin, repairs/installs the
+# user systemd daemon with `runt daemon doctor --fix`, and starts it by default.
+
+set -euo pipefail
+
+usage() {
+  cat <<'USAGE'
+Usage:
+  scripts/install-linux-release [--tag v2.4.1-stable.YYYYMMDDHHMM]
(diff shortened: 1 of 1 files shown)


rgbkrk (commented):
Blocking finding:

The nightly path clobbers the stable desktop launcher. For `--channel nightly`, the script switches the sidecar names to `runt-nightly`, `runtimed-nightly`, and `nteract-mcp-nightly`, but the AppImage destination is still `"$PREFIX/nteract.AppImage"` and the public symlink is always `ln -sfn "$APPIMAGE_DEST" "$BIN_DIR/nteract"` (`scripts/install-linux-release:141`, `scripts/install-linux-release:150`, `scripts/install-linux-release:209`). Running a nightly install after a stable install rewrites `~/.local/bin/nteract` to the nightly AppImage and never creates a `nteract-nightly` launcher.

Stable and nightly are channel-separated everywhere else in the installer, so the desktop command should be channel-specific too: keep `nteract` for stable, create/use `nteract-nightly` for nightly, and leave an existing stable symlink untouched.

rgbkrk:
Superseded by #2485.

PR #2483, opened 2026-05-02
fix(mcp): 4 gremlin-discovered MCP surface fixes

quillaid:
## Summary

Four MCP surface fixes discovered by the gremlin test suite:

1. **Surface error_reason/error_details for AwaitingEnvBuild lifecycle** — When an `environment.yml` references a nonexistent conda env, the daemon correctly writes `CondaEnvYmlMissing` reason and human-readable details to RuntimeStateDoc. But `read_runtime_info()` only surfaced these fields for the `Error` lifecycle, so MCP clients saw `kernel_status: "awaiting_env_build"` with no explanation. Widened the predicate to also include `AwaitingEnvBuild`.

2. **Add `after` parameter to `remove_dependency`** — `remove_dependency` only updated CRDT metadata without signaling that the running environment still had the package installed. Now supports `after: "restart"` to restart the kernel so the removal takes effect, and returns `needs_restart: true` when the dep was removed but the caller didn't request a restart.

3. **Reject invalid version operators in dep specifiers** — `add_dependency("pandas>>>999")` was silently accepted because validation only checked the package name, never the version operator. Added `validate_version_operators()` that extracts operator prefixes and rejects sequences no package manager would accept (`>>>`, `=!`, `>>=`, etc.). Accepts conda's single `=` glob operator for interop.

4. **Include path in `show_notebook` response** — After saving, `show_notebook` only returned `notebook_id` and `opened` status. Now includes the resolved on-disk path (from daemon room or session) in both headless and display response branches.

## Context

Found by gremlin suite scenarios: `conda-env-yml-fallback` (#1), `conda-lifecycle` (#2), `tool-resilience` (#3), `show-notebook-title` (#4).

## Test plan

- [x] `cargo test -p runt-mcp` — 117 tests pass
- [x] `cargo test -p notebook-doc` — 300 tests pass including 7 doctests
- [x] CI

### DIFF
--- a/crates/notebook-doc/src/metadata.rs
+++ b/crates/notebook-doc/src/metadata.rs
@@ -775,6 +775,127 @@ pub fn validate_package_specifier(spec: &str) -> Result<(), String> {
              Package names may only contain letters, digits, hyphens, underscores, and dots"
         ));
     }
+    validate_version_operators(spec, &name)?;
+    Ok(())
+}
+
+/// Reject obviously invalid version operators in a specifier.
+///
+/// PEP 508 operators: `==`, `!=`, `<=`, `>=`, `<`, `>`, `~=`, `===`.
+/// Conda MatchSpec operators: same minus `===`.
+///
+/// This catches mangled specifiers like `pandas>>>999` or `numpy=!2.0`
+/// that would be silently recorded in the CRDT and only fail (confusingly)
+/// at install time. Not a full PEP 508 parser — just enough to reject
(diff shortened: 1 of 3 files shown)


rgbkrk:
When you write issue descriptions, you shouldn't use your (internal) task list numbers in the body because it links to old github issues.

Anywho we should add in ProjectContext in create notebook and open notebook calls. I'm sure that's part of the MCP friction you're facing.

PR #2486, opened 2026-05-02
feat(runtimed-node): expand native notebook control API

quillaid:
## Summary

This expands `@runtimed/node` from a small execution wrapper into a direct, agent-friendly notebook control surface that is much closer to MCP parity while staying native to Node/N-API.

### Notebook/session lifecycle
- add `listActiveNotebooks()` for daemon room discovery
- add `openNotebookPath(path)` alongside `openNotebook(notebookId)`
- add `showNotebook()` / `Session.showNotebook()` with structured headless responses
- add `shutdownNotebook(notebookId)` / `Session.shutdownNotebook()` for room eviction
- add `description`/`peerLabel` support and announce collaborative presence on create/open

### Cell and execution APIs
- add `Session.listCells()` / `Session.getCell()`
- add `Session.createCell()` / `Session.setCell()` / `Session.deleteCell()` / `Session.moveCell()`
- add `Session.executeCell()` for existing cells
- emit best-effort cursor/focus presence for Node-backed edits and execution

### Dependencies, trust, and kernels
- add first-call `dependencies` and typed `packageManager` to `createNotebook()`
- expose N-API `PackageManager` string enum and convert it to the shared notebook-protocol enum at the Rust boundary
- add `Session.addDependency()` / `Session.addDependencies()` and `Session.removeDependency()` / `Session.removeDependencies()` for UV, Conda, and Pixi
- add grouped `Session.getDependencyStatus()` with `uv`, `conda`, `pixi`, and `trust`
- incorporate main's batched metadata mutation path so multi-package edits use one CRDT transaction
- incorporate main's richer runtime lifecycle error details in Node status reporting
- add `Session.interruptKernel()` / `Session.shutdownKernel()` / `Session.restartKernel()`
- add `Session.getRuntimeStatus()` with lifecycle, activity, env source, and startup error details
- harden kernel lifecycle APIs: unexpected interrupt/shutdown responses now error, and restart polls RuntimeStateDoc instead of sleeping

### Pi REPL integration
- update `plugins/nteract/pi` to use the expanded Node binding surface
- allow the `python` tool to accept first-call `dependencies` so imports can work without an initial failed cell
- switch `python_add_dependencies` to batched dependency edits when available
- include runtime status details in tool metadata and shut down the backing notebook room on reset/session shutdown

### Package shape
- keep `packages/runtimed-node/src/index.cjs` and `index.d.ts` thin over generated N-API bindings
- bump `@runtimed/node` to `0.2.0` for the expanded `0.x` API
- add a committed `pnpm --dir packages/runtimed-node smoke:api` harness for daemon-backed API checks
- avoid speculative JS helper layers here; richer experimental JS magic can iterate in pi/plugins before becoming stable package API

## Verification

- `cargo check -p runtimed-node`
- `cargo xtask lint --fix`
- `pnpm --dir packages/runtimed-node exec napi build --platform --js binding.cjs --dts binding.d.ts --manifest-path ../../crates/runtimed-node/Cargo.toml --output-dir /tmp/runtimed-node-napi-check`
- `cargo xtask lint --fix` covers the TypeScript pi REPL changes
- dev-daemon Node smoke against `/tmp/runtimed-node-napi-check`:
  - `createNotebook({ dependencies, packageManager, description })`
  - `addDependencies` / `removeDependencies`, grouped `getDependencyStatus()`, `getRuntimeStatus()`
  - `createCell` / `listCells` / `setCell` / `moveCell` / `getCell`
  - `listActiveNotebooks()`
  - `showNotebook()` headless result
  - `interruptKernel()` / `shutdownKernel()`
  - `Session.shutdownNotebook()` and room eviction verification

Note: `cargo test -p runtimed-node` currently fails linking the lib test binary with undefined N-API symbols (`napi_reference_unref` / `napi_delete_reference`), so validation uses `cargo check`, N-API build, and daemon-backed Node smoke tests.


crates/runtimed-node/src/session.rs
```diff
@@ -27,6 +27,24 @@ type CommMap = std::collections::HashMap<String, runtime_doc::CommDocEntry>;
 
 // ── Options ────────────────────────────────────────────────────────────
 
+/// Package manager for Python notebook dependencies.
+#[napi(string_enum = "lowercase")]
+pub enum PackageManager {
+    Uv,
+    Conda,
+    Pixi,
+}
```
rgbkrk, line 43: Shouldn't you be able to reuse the enums we have in other places for the package manager.

crates/runtimed-node/src/session.rs
```diff
@@ -27,6 +27,24 @@ type CommMap = std::collections::HashMap<String, runtime_doc::CommDocEntry>;
...
+
+impl From<PackageManager> for notebook_protocol::connection::PackageManager {
+    fn from(value: PackageManager) -> Self {
+        match value {
+            PackageManager::Uv => Self::Uv,
+            PackageManager::Conda => Self::Conda,
+            PackageManager::Pixi => Self::Pixi,
+        }
```
rgbkrk, line 51: Oh I see, you had to convert it. Is this necessary?

packages/runtimed-node/src/index.cjs
```diff
@@ -1,12 +1,201 @@
...
+  const session = await binding.createNotebook(normalizeCreateOptions(options));
+  return enhanceSession(session);
+}
+
+async function openNotebook(notebookId, options) {
+  const session = await binding.openNotebook(notebookId, options);
+  return enhanceSession(session);
+}
+
+async function runPython(source, options = {}) {
```
rgbkrk: I feel like simple helpers like this belong in `./plugins/nteract/pi` until we're sure on interface.

packages/runtimed-node/src/index.cjs
```diff
@@ -1,12 +1,201 @@
...
+        const result = await session.waitForExecution(executionId, waitOptions);
+        return enrich ? enrichResult(result) : result;
+      },
+    },
+  });
+
+  return session;
+}
+
+async function createNotebook(options) {
```
rgbkrk: It's ok to change the rust interfaces for this library, we'll just bump `0.x.0`

packages/runtimed-node/src/index.cjs
```diff
@@ -1,12 +1,201 @@
...
+  for (const pkg of packages) {
+    await session.addUvDependency(String(pkg));
+  }
+  if (options.sync !== false && packages.length > 0) {
+    await session.syncEnvironment();
+  }
+  return session;
+}
+
+function enhanceSession(session) {
```
rgbkrk: Not sure what you're getting at here. Leave at least some JSDoc comments.

packages/runtimed-node/src/index.cjs
```diff
@@ -1,12 +1,201 @@
...
+function outputToObject(output) {
+  return {
+    ...output,
+    data: decodeMimeBundle(output.dataJson),
+    blobUrls: parseJsonField(output.blobUrlsJson, {}),
+    blobPaths: parseJsonField(output.blobPathsJson, {}),
+  };
+}
+
+function enrichResult(result) {
```
rgbkrk: Is this intended to be like resolve outputs in rust and python?

packages/runtimed-node/src/index.d.ts
```diff
@@ -1,7 +1,176 @@
...
+  totalRows: number;
+  offset: number;
+}
+export interface ParquetSummaryResult {
+  numRows: number;
+  numBytes: number;
+  columns: ParquetColumnInfo[];
+}
+export function readParquetFile(filePath: string, offset: number, limit: number): ParquetRowPage;
+export function summarizeParquetFile(filePath: string): ParquetSummaryResult;
```
rgbkrk: As I read through this it makes me wonder if more needs to shared from the other runtimed js packages (used in the frontend).

crates/runtimed-node/src/session.rs
quillaid: Resolved by keeping a small N-API boundary enum that mirrors `notebook_protocol::connection::PackageManager` and documenting why: we cannot apply `#[napi]` to the protocol crate enum directly, and the protocol enum includes `Unknown(String)` for forward-compatible wire decoding that we do not want as a typed JS input variant. The value is immediately converted into the shared protocol enum before the handshake.

crates/runtimed-node/src/session.rs
quillaid: Yes for the N-API surface: the external protocol enum is not directly exportable via `#[napi]`, and its `Unknown(String)` variant is a wire-compat detail rather than a valid public option. I added a comment in the code explaining the conversion and kept the shared protocol enum as the internal type used for the handshake.

packages/runtimed-node/src/index.cjs
quillaid: Agreed. I removed the speculative JS helper layer and restored `index.cjs`/`index.d.ts` to thin binding re-exports. Follow-up now moves concrete API improvements into the native Rust/N-API surface instead.

packages/runtimed-node/src/index.cjs
quillaid: Done — I switched direction from JS monkey-patching helpers to native Rust/N-API methods. The branch now adds first-class cell operations (`listCells`, `getCell`, `createCell`, `setCell`, `deleteCell`, `moveCell`, `executeCell`) plus create-time deps/description.

packages/runtimed-node/src/index.cjs
quillaid: Removed with the JS helper layer. The committed package wrapper is thin again, and the public docs are generated from/owned by the native N-API declarations.

packages/runtimed-node/src/index.cjs
quillaid: That helper was headed in that direction, but I agree it should not be invented ad hoc here. Removed it from this PR; output resolving remains in the Rust/client path already used by the binding.

packages/runtimed-node/src/index.d.ts
quillaid: Agreed. I reverted the hand-maintained TypeScript surface back to re-exporting generated `binding.d.ts` rather than duplicating/shared frontend concepts in this package. Any richer JS output helpers should probably be designed against the existing frontend/runtime packages or live in pi while we iterate.

packages/runtimed-node/README.md
```diff
@@ -51,19 +64,44 @@ main().catch((error) => {
...
+- `Session.getDependencyStatus()` returns dependency metadata, fingerprint, and
+  trust state in one call.
 - `Session.addUvDependency(spec)` records a UV dependency for the notebook.
 - `Session.syncEnvironment()` installs recorded notebook dependencies.
 - `Session.saveNotebook(path?)` saves the notebook.
 - `Session.close()` releases the daemon connection.
 
 ## Daemon Requirements
 
+`createNotebook()` accepts `dependencies` so agent code can declare packages
```
rgbkrk: I think this is in the wrong section. It's in "Daemon Requirements" and I don't think that makes sense.

crates/runtimed-node/src/session.rs
```diff
@@ -39,14 +63,53 @@ pub struct CreateNotebookOptions {
...
+    /// Packages to record before the kernel starts (for example `["numpy", "matplotlib"]`).
+    /// Python notebooks use the selected package manager; Deno notebooks treat these as
+    /// runtime-native import dependencies.
+    pub dependencies: Option<Vec<String>>,
+    /// Package manager for Python dependencies. Defaults to the daemon/user setting.
+    pub package_manager: Option<PackageManager>,
 }
 
-/// Options for `openNotebook()`.
+/// Options for `openNotebook()` and `openNotebookPath()`.
```
rgbkrk, line 77: you should look at the `connect_notebook` used by runt mcp (and hopefully runtimed python...).

Also this reminds me we probably need to bake presence in here too so edits made by node backed bindings show something.

crates/runtimed-node/src/session.rs
```diff
@@ -39,14 +63,53 @@ pub struct CreateNotebookOptions {
...
+    pub uv_requires_python: Option<String>,
+    pub conda_python: Option<String>,
+    pub pixi_python: Option<String>,
+    pub fingerprint: Option<String>,
+    pub trust_status: Option<String>,
+    pub trust_needs_approval: Option<bool>,
+    pub approved_uv_dependencies: Vec<String>,
+    pub approved_conda_dependencies: Vec<String>,
+    pub approved_pixi_dependencies: Vec<String>,
+    pub approved_pixi_pypi_dependencies: Vec<String>,
```
rgbkrk, line 127: wouldn't it make more sense to have like `pub conda: Option<CondaThing>` or something that has the related fields for that package manager there?

packages/runtimed-node/README.md
quillaid: Moved this out of Daemon Requirements into a new Notebook Dependencies section.

crates/runtimed-node/src/session.rs
quillaid: Good call. I added initial presence announcement on create/open/openPath and best-effort cursor/focus presence for Node-backed cell creation, edits, moves, deletes, and existing-cell execution. I also kept open-by-ID/path as separate direct APIs but aligned the behavior with MCP connect_notebook primitives.

crates/runtimed-node/src/session.rs
quillaid: Agreed. I reshaped getDependencyStatus() to return grouped objects: `uv`, `conda`, `pixi`, and `trust`, rather than flattened manager-specific fields.

quillaid:
## Code Review — `feat(runtimed-node): expand native notebook control API`

**Reviewed at:** `ae1a5b2b` | **Build:** `cargo check` ✅ | `cargo xtask lint --fix` ✅

---

### What was done well

- **Excellent API design.** Near-MCP-parity Node control surface with clean `Session` struct methods.
- **Proper tokio mutex discipline.** Every `self.state.lock().await` guard is block-scoped with owned data extracted before the next `.await`. No violations.
- **Wire protocol compliance.** All daemon requests use existing `NotebookRequest`/`NotebookResponse` variants — no new wire types invented.
- **CRDT mutation safety.** Cell CRUD goes through `DocHandle`. Dependency edits use the new `handle.with_metadata()` single-transaction pattern. No independent `put_object` on shared keys.
- **Good error surfacing.** All daemon request/response matches handle `Error`, `GuardRejected`, and unexpected variants consistently.
- **Batched dependency operations.** `addDependencies`/`removeDependencies` use a single CRDT transaction via `with_metadata`, with noop optimization when snapshot is unchanged.

---

### P0 — Blocking

None found.

---

### P1 — Should fix

**1. `restart_kernel` uses a hardcoded 500ms sleep instead of polling kernel state**
`session.rs:793-801` — `shutdown_kernel` error is silently ignored (`let _ = ...`), and the 500ms sleep has no guarantee the kernel has actually shut down. On slow machines or under load, this could race. Consider polling `RuntimeStateDoc` kernel lifecycle until it transitions away from `Running`, or at least propagating the error from `shutdown_kernel`.

**2. `shutdown_kernel` and `interrupt_kernel` wildcard match arms silently succeed on unexpected responses**
`session.rs:786-788` and `session.rs:763` — `_ => Ok(true)` means an unexpected response variant (e.g., protocol version mismatch) would silently report success. Other methods like `save_notebook`, `sync_environment`, and `queue_existing_cell` correctly return errors for unexpected variants. These should be consistent.

---

### P2 — Consider

1. **Presence helpers duplicated from `runtimed-py/src/session_core.rs`** — PR body notes Phase 2 will extract shared logic into a `runtimed-session` crate. Worth tracking.
2. **`notebook-doc` as direct dependency** — Verified this follows established precedent from `runtimed-py`. Usage limited to presence encoding, `CellSnapshot` type, and metadata types. No direct Automerge mutations. ✅
3. **`show_notebook` headless detection** — `has_display()` won't detect macOS SSH sessions without display forwarding. Low priority — the structured `opened: false` response handles it gracefully.

---

### Summary

**Approve with minor suggestions.** High-quality expansion of the `@runtimed/node` API. No crate boundary violations, no CRDT safety issues, no tokio mutex invariant violations, proper wire protocol usage, thorough error handling. The P1 items (restart sleep heuristic, wildcard match arms) are real but non-blocking — good candidates for a follow-up.

PR #2494, opened 2026-05-02
fix(runtime-agent): prevent interrupt+execute race from deadlocking kernel queue

quillaid:
## Summary

- When `interrupt_kernel` and `execute_cell` fire in the same turn, the IOPub `KeyboardInterrupt` error cascades via `CellError` and unconditionally clears the entire execution queue — including cells queued **after** the interrupt. This leaves the new cell permanently stuck at "running" with no path to recovery.
- Adds an `interrupted` flag to `KernelState` that gates `CellError` queue clearing. When set (by the interrupt handler), `CellError` skips the cascade so post-interrupt cells proceed normally. The flag resets on `execution_done`.
- Normal error cascading (e.g., Run All with a failing cell) is unaffected — the flag is only set by explicit interrupt.

Evidence: `reports/20260502-174225-full-suite-02150e4/breaker.json` — gremlin repro shows `while True: pass` + interrupt + `1+1` leaves the second cell stuck at "running" indefinitely.

## Test plan

- [x] `cargo test -p runtimed --lib kernel_state` — 10 tests pass (3 new: interrupt flag lifecycle)
- [x] `cargo test -p runtimed --lib runtime_agent` — 16 tests pass (2 new: CellError after interrupt preserves queue, CellError without interrupt cascades)
- [x] `cargo test -p runtimed --lib` — 632 tests pass
- [ ] Gremlin replay: breaker gremlin should no longer deadlock on interrupt+execute

### DIFF
--- a/crates/runtimed/src/kernel_state.rs
+++ b/crates/runtimed/src/kernel_state.rs
@@ -55,6 +55,13 @@ pub struct KernelState {
     /// Whether the current execution produced an error output.
     /// Read by `execution_done` to record success/failure in the state doc.
     execution_had_error: bool,
+    /// Set when `interrupt()` fires; cleared on the next `execution_done`.
+    ///
+    /// When true, `CellError` skips queue clearing — the error is the
+    /// KeyboardInterrupt from the interrupted cell, not a user-code failure
+    /// that should cascade to queued cells. Without this, cells queued
+    /// between `interrupt_kernel` and the IOPub error get silently dropped.
+    interrupted: bool,
     /// Kernel lifecycle status.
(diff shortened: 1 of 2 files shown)


quillaid:
Closing in favor of a reworked approach per product direction from rgbkrk: interrupt should clear the queue AND mark all in-flight executions as failed (including CRDT-synced entries not yet in the runtime agent's local queue). The `interrupted` flag approach was incorrect — the root cause was CRDT-created execution entries getting orphaned when they hadn't synced to the runtime agent before the interrupt cleared the local queue. New PR incoming.

PR #2504, opened 2026-05-03
fix(kernel-env): scope daemon-created conda:env_yml envs by project dir

quillaid:
## Summary

- Scope daemon-created `conda:env_yml` envs by project directory to prevent cross-notebook clobbering
- Extract `resolve_conda_env_yml_prefix()` as single source of truth for conda prefix resolution (3 call sites consolidated)
- When two notebooks share the same `name:` in environment.yaml but have different deps, they now get isolated prefixes (`{name}-{project_hash}`)

**Note:** The abort-on-sync-failure fix has been split into #2514 (uncontroversial, can merge independently). This PR contains only the project-scoped isolation approach, which needs the product discussion in #2513.

See #2513 for the full product framing and alternative approaches.

## Test plan

- [x] `cargo check --package runtimed` passes
- [x] CI green
- [x] 4 unit tests covering prefix resolution (prefix field, project scoping, same-dir stability, no-name fallback)

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -1050,8 +1050,93 @@ pub async fn sync_dependencies(
         },
     );
 
+    // Post-install verification: confirm every requested package is present
+    // in conda-meta. The rattler Installer performs a transactional unlink-then-
+    // link sequence; if it fails mid-transaction the prefix can be left without
+    // packages that were previously installed. Catching this here prevents the
+    // caller from launching a kernel against an inconsistent env.
+    let missing = verify_packages_installed(&env.env_path, &deps.dependencies);
+    if !missing.is_empty() {
+        return Err(anyhow!(
+            "Post-sync verification failed: packages missing from conda-meta after install: [{}]",
+            missing.join(", ")
+        ));
+    }
+
     info!("Conda dependencies synced successfully");
     Ok(())
+}
+
+/// Verify that every package name in `required` has a corresponding
+/// `<name>-*.json` entry in the env's `conda-meta/` directory.
+///
+/// Returns the list of package names that are missing. An empty vec means
+/// all required packages are present.
+///
+/// Used as a post-install gate: the rattler Installer's transactional
+/// unlink-then-link can leave the prefix inconsistent if it fails between
+/// the two phases. Checking conda-meta after install catches this before
+/// the kernel launches against a broken env.
+pub fn verify_packages_installed(env_path: &std::path::Path, required: &[String]) -> Vec<String> {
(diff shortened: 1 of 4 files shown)


quillaid:
## Codex review

**Pass.** Checked against architecture rules and contributing docs:

1. No emdashes in comments or docs
2. No tokio mutex held across .await (no new async code)
3. Crate boundaries: `resolve_conda_env_yml_prefix` correctly lives in `project_file.rs` (daemon env resolution)
4. Single source of truth: all 3 call sites use the same function, no duplicate logic
5. Uses existing `sha2`/`hex` deps, no hand-rolled crypto
6. `canonicalize()` has a safe fallback for non-existent paths
7. Pre-existing user envs still discovered via `find_named_conda_env` (backward compat)
8. 12 hex char hash (48 bits) sufficient for project scoping
9. 4 unit tests cover: prefix passthrough, project-scoped isolation, idempotency, no-name fallback
10. Consistent with existing `conda-envs/` cache dir pattern used by nameless env_yml envs

rgbkrk:
This seems like a problem with the gremlins though. Why would someone have two different uses of the same "name" in a conda env?

quillaid:
Good question. Two angles:

**The naming collision is realistic.** Users commonly have `environment.yaml` files with generic `name:` values ("analysis", "dev", "ml") across different project directories. Conda resolves the `name:` field to a single global prefix under `~/.conda/envs/{name}`, so if two project dirs both declare `name: analysis` with different deps, the daemon would clobber one when syncing the other. Project-scoped paths prevent this silently.

**The silent failure fix matters regardless.** Even without a naming collision, `sync_dependencies` can fail mid-install (network error, solver conflict). Before this PR, both `LaunchKernel` and `auto_launch_kernel` logged a warning and launched the kernel into a broken env anyway. Now they abort with a clear `CondaEnvBuildFailed` error. That's a correctness win for any conda notebook, not just the concurrent case.

The scoping fix is defensive for multi-project workflows. The abort-on-failure fix closes a real silent-failure gap.

quillaid:
Closing this PR. The project-scoped hashed paths approach has been rejected in favor of a simpler product direction:

- **Named envs** (`name:` field in environment.yaml): activate-only. nteract resolves the conda env by name, checks it exists, and launches the kernel into it. No auto-sync, no auto-build. The user manages the env with conda as usual.
- **Prefix envs** (`prefix:` field): same as named - activate only, user-managed.
- **Unnamed envs** (no `name:` or `prefix:`): daemon-owned. nteract auto-builds and syncs these as before.

This keeps nteract out of conda's domain. The abort-on-sync-failure fix has been split into #2514 and will land separately.

See #2513 for the full product direction.

PR #2519, opened 2026-05-03
fix(runtimed-py): approve trust on start() when daemon requires it

rgbkrk:
## Summary

`Notebook.start()` fails with "Trust changed before the action could run" when default pool packages are injected between `create_notebook` and `launch_kernel`. The daemon sees a dependency fingerprint change and rejects the kernel launch.

The fix: catch the trust-changed error and auto-approve before retrying. Safe for client-created notebooks since the client owns the session.

## Root cause

The default-packages pool feature (#2507) injects packages into new notebooks after creation. By the time `start()` sends `launch_kernel`, the dependency fingerprint has changed from what it was at `create_notebook` time. The daemon's trust gate rejects the launch.

The health check fixture handles this differently (waits for auto-launch), but `Notebook.start()` and all tests using the `notebook` fixture hit this race.

## Verification

All 6 `TestExecuteCell` tests pass against the dev daemon at 2.4.3, including the new `test_execution_watch_streams_terminal_progress_and_matches_result`.

### DIFF
--- a/python/runtimed/src/runtimed/_notebook.py
+++ b/python/runtimed/src/runtimed/_notebook.py
@@ -102,12 +102,23 @@ class Notebook:
     ) -> None:
         """Start a runtime for this notebook.
 
+        Approves trust automatically if the daemon requires it (e.g. when
+        default pool packages have been injected since the notebook was
+        created). This is safe for client-created notebooks.
+
         Args:
             runtime: Runtime type (e.g. "python", "deno").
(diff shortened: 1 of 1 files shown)


rgbkrk:
Trust-changed error was transient (pool building on first run). Tests pass without the fix once trust signatures are cached. The right long-term fix is to pre-trust daemon-injected default packages, not retry in Python.

Issue #1463, opened 2026-04-04
test(e2e): verify anywidget rendering inside isolated iframe


paddymul:
## Context

The existing `pandas.spec.js` and `polars.spec.js` only test the Python side — they `print(type(widget).__name__)` and check stream output. They never verify that the widget actually appears in the DOM. I'd like to write integration tests that verify buckaroo's anywidget rendering end-to-end.

## What I want to test

1. Execute a cell that returns a `BuckarooWidget(df)`
2. Verify a `[data-widget-id]` element appears (the anywidget rendered, not just the Python object created)
3. Check that AG-Grid content is visible inside the isolated iframe

## Where I'm stuck

- `browser.switchToFrame()` doesn't work in wry, so I can't directly query iframe content
- I can see selectors like `[data-widget-id]` and `[data-widget-loading]` in the source (`anywidget-view.tsx`, `widget-view.tsx`), but I'm not sure these are on the parent DOM vs inside the iframe
- The `CommBridgeManager` / `IsolatedFrame` / `WidgetBridgeClient` pipeline is complex — I'm not sure what the right "widget has finished rendering" signal is from the E2E test's perspective
- `browser.execute()` runs in the parent frame context. Is there a way to reach into the iframe DOM, or should I be checking parent-side indicators only?

## What would help

- Guidance on the recommended pattern for asserting "this anywidget rendered successfully" in an E2E spec
- Whether there's a parent-DOM signal that the iframe's anywidget ESM loaded and `render()` completed (e.g., does `data-widget-loading` get removed? Is there a `render_complete` postMessage I can listen for?)
- Any gotchas with the blob-URL sandbox that would prevent `browser.execute()` from seeing widget elements

## Environment

- nteract desktop (current main)
- WebdriverIO + Mocha via `cargo xtask e2e test-fixture`
- buckaroo widget (anywidget-based, ESM + CSS)

paddymul:
Also, I started with trying to write tests integrated into nteract's existing test harness just to see if it was possible.  Ideally I want these integration tests to run in Buckaroo's CI.

rgbkrk:
I think what's going to work best is for us to be able to run the tests without Tauri, using a separate vite + playwright setup. I'll carve out time for that this week and it should make it easier to text with fixtures (probably not true E2E as I'm thinking Python-less).

paddymul:
what does python-less mean?  how would the tests work?

rgbkrk:
[...]

PR #2662, opened 2026-05-09
test(renderer): cover Buckaroo anywidget table rendering

looooown2006:
## Summary

Updates the standalone `apps/renderer-test` Vite harness to cover Buckaroo-style anywidget table rendering inside the isolated iframe.

This now adds:

- a Buckaroo anywidget fixture module that renders Buckaroo's table component
- widget-view fixture data for a Buckaroo table model
- `widget_snapshot` injection on `widget_ready` with a duplicate-send guard
- Playwright assertions for:
  - widget presence via `[data-widget-id]`
  - no loading/error fallback
  - visible AG Grid/table content
  - the `score` header and `Alice` data cell

Notes:

- `_esm` uses `new URL('./buckaroo-anywidget-fixture.ts', import.meta.url).href` instead of a hard-coded dev URL.
- The renderer-test harness uses `127.0.0.1` with CORS enabled to match the iframe CSP/dev-server requirements for anywidget module loading.
- This is scoped to the standalone renderer harness for the isolated anywidget/Buckaroo render path, not a full notebook kernel-execution E2E.

Refs #1463.

## Tests

```text
pnpm --dir D:\aom\nteract-desktop\apps\renderer-test exec playwright test
```

Result: 2 passed.

```text
git -C D:\aom\nteract-desktop diff --check --cached
```

Result: passed before commit.

### DIFF
--- a/apps/renderer-test/e2e/render.spec.ts
+++ b/apps/renderer-test/e2e/render.spec.ts
@@ -33,6 +33,25 @@ test.describe("Renderer plugin fixtures", () => {
     await expect(
       page.frameLocator("#fixture-5").getByRole("button", { name: "Zoom", exact: true }),
     ).toBeVisible({ timeout: 30_000 });
+    const buckarooFrame = page.frameLocator("#fixture-6");
+
+    await expect(buckarooFrame.locator("[data-widget-id='buckaroo-table-model']")).toBeVisible({
+      timeout: 30_000,
+    });
+    await expect(buckarooFrame.locator("[data-widget-loading='true']")).toHaveCount(0);
+    await expect(buckarooFrame.locator("[data-widget-error='true']")).toHaveCount(0);
+    await expect(buckarooFrame.locator(".ag-root-wrapper").first()).toBeVisible({
+      timeout: 30_000,
+    });
+    await expect(buckarooFrame.locator(".buckaroo_anywidget")).toBeVisible({
+      timeout: 30_000,
+    });
+    await expect(buckarooFrame.locator(".ag-header-cell-text", { hasText: "score" }).first()).toBeVisible({
+      timeout: 30_000,
+    });
+    await expect(buckarooFrame.locator(".ag-cell", { hasText: "Alice" }).first()).toBeVisible({
+      timeout: 30_000,
+    });
   });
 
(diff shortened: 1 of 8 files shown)


rgbkrk:
I don't think this solves what was asked for in #1463. They needed an E2E test around buckaroo data tables. Should be easier to do with the dev server in standalone vite now.

looooown2006:
Updated this PR to target the Buckaroo table path more directly in the standalone renderer-test Vite harness.

What changed:

- added a Buckaroo anywidget fixture module
- added widget-view fixture state for a Buckaroo table model
- send `widget_snapshot` after `widget_ready` with a duplicate-send guard
- assert inside the iframe that:
  - `[data-widget-id='buckaroo-table-model']` is present
  - no widget loading/error fallback remains
  - AG Grid/table content is visible
  - the `score` header and `Alice` cell render

A couple of scope notes:

- `_esm` now uses `new URL('./buckaroo-anywidget-fixture.ts', import.meta.url).href`, not a hard-coded dev URL.
- The harness stays on `127.0.0.1` with CORS enabled to match the isolated iframe CSP/dev-server requirements for anywidget module loading.
- This is a standalone renderer-harness E2E for the isolated anywidget/Buckaroo render path, not a full notebook/kernel execution E2E.

Verification after the update:

```text
pnpm --dir D:\aom\nteract-desktop\apps\renderer-test exec playwright test
# 2 passed
```

```text
git -C D:\aom\nteract-desktop diff --check --cached
# passed before commit
```

rgbkrk:
E2E for us means launching a notebook with deps and using it to drive it. You shouldn't hand code the widget. I recommend getting research agents to more fully explore this space and ensure you're going from Python in the cell all the way to visuals you've verified.

looooown2006:
Thanks for the clarification. You're right: this PR still exercises a hand-built renderer fixture rather than the real notebook path from a Python cell through kernel execution, comm/model state, the isolated iframe, and the final visual table.

I don't want to keep pushing this in the wrong direction, so I'll close this PR to avoid keeping a misleading fixture-only test open. If I revisit this, I'll do it as a fresh PR built around a real notebook-driven E2E flow with the Buckaroo dependency installed and assertions that start from the Python cell output and end at the rendered table.

PR #2691, opened 2026-05-11
Treat Automerge panics as diagnostic sync failures

rgbkrk:
## Summary

- stop treating caught Automerge panics as a steady-state rebuild/retry path
- keep caller-marked recoverable Automerge errors, currently `PatchLogMismatch`, on the existing rebuild/reset/retry path
- propagate document-boundary sync failures through notebook, runtime-state, pool, settings, streaming-load, and runtime-agent sync paths so the owning peer/session closes instead of continuing with possibly inconsistent per-peer sync state
- extend log guards so new panic/failure diagnostics fail integration logs

## Invariant

`automerge::sync::State` is owned by one document instance and one peer/session. If the document is rebuilt/replaced/compacted or the peer transport/session restarts, the corresponding peer `sync::State` must be reset. A caught Automerge panic is diagnostic evidence that this boundary may already be inconsistent, so the current sync peer/session should close rather than replaying the same inbound frame after a save/load rebuild.

## Verification

- `cargo test --locked -p automerge-recovery -p notebook-doc -p runtime-doc -p notebook-sync --lib`
- `cargo test --locked -p runtimed --lib`
- `cargo check --locked -p runtimed -p notebook-sync -p runtime-doc -p notebook-doc -p automerge-recovery`
- `cargo clippy --locked -p runtimed --lib -- -D warnings`
- `uv run --directory python/runtimed pytest tests/test_log_guards.py -q`
- `cargo fmt --all -- --check`
- `git diff --check`
- stale recovery wording scan over `crates` and `python/runtimed/tests`

## External Review

- Bedrock Claude Opus 4.6 reviewed `origin/main...HEAD`
- Disposition: three low-severity findings were confirmed and fixed in `db177907`
- Second pass after `db177907`: clear, no new findings

### DIFF
--- a/crates/automerge-recovery/src/lib.rs
+++ b/crates/automerge-recovery/src/lib.rs
@@ -35,7 +35,7 @@ pub enum AutomergeOperationError {
     },
     #[error(transparent)]
     Panic(#[from] AutomergeRecoveryError),
-    #[error("[{label}] failed to rebuild document after Automerge panic: {source}")]
+    #[error("[{label}] failed to rebuild document after recoverable Automerge failure: {source}")]
     RebuildFailed {
         label: String,
         #[source]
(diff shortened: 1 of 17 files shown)


rgbkrk:
Superseded by #2692. The Automerge recovery commits from this PR were included in the merged #2692 squash commit (`f272978a`), and this branch is now dirty against current `main`; keeping it open risks reapplying/churning the same sync changes plus conflicting with the Windows fingerprint fix.

PR #2713, opened 2026-05-11
fix(notebook): forward leaked wheel events from passthrough iframe to page

rgbkrk:
Scrolling the notebook hard up/down with the cursor over a sift output iframe sometimes scrolls the inner sift table even though the user hasn't clicked into it. The parent already sets `pointer-events: none` on the iframe in passthrough mode (in `isolated-frame.tsx`), which is supposed to keep wheels on the page's native scroll path. WKWebView (and to a lesser extent Safari) occasionally lets momentum-scroll pulses pierce that, so a few wheel events land inside the iframe document and sift's `.sift-viewport` (with `overscroll-behavior: none`) absorbs them locally without chaining.

The fix is a defensive forward inside the iframe. The parent already sends an `interaction_state` message when sift focus toggles (`OutputArea.tsx:481-484`). The bootstrap script in `frame.html` now listens for that message, tracks `scrollPassthroughActive = (active === false)`, and short-circuits the wheel handler when set: every wheel event is `preventDefault`'d and forwarded to the parent as `nteract/wheelBoundary`. The parent's `NTERACT_WHEEL_BOUNDARY` handler is widened to also fire when `scrollPassthrough` is true, so the forwarded delta scrolls the notebook page exactly the way the user expected.

When the user clicks to focus the table (`active = true`), the flag flips back to `false` and the wheel handler returns to its existing "forward only at boundary" semantics — focused interaction with the table is unchanged.

## Test plan

- [x] `frame-html.test.ts`, `OutputArea.test.tsx`, `isolated-frame.test.ts`, `scroll-boundary.test.ts` all pass (49/49)
- [ ] Manual: 2M-row sift table in `~/big-ish-data.ipynb`, hard trackpad-scroll up and down with the cursor hovering over the table without clicking — sift table position stays put while the notebook scrolls
- [ ] Manual: click the table to focus, wheel — sift table scrolls normally (focused interaction unchanged)
- [ ] Manual: with the table focused and scrolled to its top boundary, keep wheeling up — page scrolls (existing boundary-chaining behavior preserved)

## Known WebKit quirks worth citing

- `pointer-events: none` controls hit-testing for new pointer interactions but doesn't reliably suppress in-flight momentum-scroll pulses on WKWebView; the iframe document can still receive those wheel events
- WKWebView's inner `scrollTop` updates can lag the visual position during momentum, so the boundary check at `frame.html:699` flickered between true/false on the tail end of a fling — covered by short-circuiting before the boundary check now

### DIFF
--- a/src/components/isolated/frame.html
+++ b/src/components/isolated/frame.html
@@ -342,6 +342,19 @@
                 handleSearchNavigate(payload);
                 break;
 
+              case "interaction_state":
+                // Mirror the parent's iframe-interaction state into a flag
+                // the bootstrap wheel handler can read. active=false means
+                // the parent is in scroll-passthrough mode (sift output not
+                // clicked into) and any wheel events that reach this
+                // document — including momentum pulses that WKWebView lets
+                // through `pointer-events: none` — should be redirected
+                // to the notebook page instead of scrolling inner content.
+                if (payload && typeof payload === "object") {
+                  scrollPassthroughActive = payload.active === false;
+                }
+                break;
+
               case "bridge_ready":
               case "comm_open":
               case "comm_msg":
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing per discussion — the wheel leak is a nit and this fix didn't fully catch the vigorous-scroll cases anyway. Pivoting to the higher-value two-click cross-filter problem on a fresh branch with design first.

PR #2738, opened 2026-05-12
fix(widgets): improve jscatter default height

rgbkrk:
## Summary

- Bump jupyter-scatter's default anywidget height from its library default of 240px to a nteract display default of 480px.
- Keep non-default explicit jupyter-scatter heights untouched.
- Leave unrelated anywidgets and the existing plotly fallback alone.

Note: the widget protocol serializes an explicit `height=240` the same way as jupyter-scatter's library default, so that exact value is not distinguishable at render time.

## Verification

- `pnpm test -- src/components/widgets/__tests__/anywidget-view.test.ts src/components/widgets/__tests__/afm-model-proxy.test.ts`
- `pnpm test -- src/components/cell/__tests__/OutputArea.test.tsx`
- `cargo xtask lint --fix`

### DIFF
--- a/src/components/widgets/__tests__/anywidget-view.test.ts
+++ b/src/components/widgets/__tests__/anywidget-view.test.ts
@@ -9,7 +9,11 @@
  */
 
 import { afterEach, describe, expect, it } from "vite-plus/test";
-import { injectCSS } from "../anywidget-view";
+import {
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing per follow-up: the 240px value is jupyter-scatter's serialized default, and treating it as implicit would also change explicit height=240 cases.

PR #2749, opened 2026-05-13
docs(runtimed): plan PR 2 LRU + disk-spill for resumable rooms

rgbkrk:
Research and planning PR for the second step in the "resumable rooms" series. No code changes. Implementation will follow in a separate PR.

## What this is

PR 1 (in flight, parallel) keeps a `NotebookRoom` resident after the kernel is torn down so a returning user reattaches to the same doc, outputs, and file binding. PR 1 stamps `last_kernel_torn_down_at` on the room but does not remove ghost rooms from `notebook_rooms` - that is this PR's job.

This PR commits the design doc: [`docs/architecture/pr2-lru-disk-spill.md`](docs/architecture/pr2-lru-disk-spill.md).

## Headline calls

- **Count-cap LRU first.** Cap on resident peer-less rooms (initial value 32). Active rooms exempt. Uses PR 1's `last_kernel_torn_down_at` as the LRU key. Byte budget deferred to a follow-up once measurements justify a threshold.
- **Disk spill is mostly free.** The persist debouncer already mirrors `NotebookDoc` to `notebook-docs/<sha256(uuid)>.automerge`. PR 1's flush-at-teardown means the file is current when the reaper runs. The reaper drops the in-memory copy and leaves the file on disk.
- **Drop `RuntimeStateDoc`, do not persist it.** Outputs rebuild from `.ipynb` on reload via the existing `load_notebook_from_disk_with_state_doc_*` path.
- **Don't delete the `.automerge` on reap.** Arrow-stream / parquet outputs store sub-blob hashes that blob GC finds by walking `notebook-docs/*.automerge`. Deleting the file would break those refs. Orphan-doc sweep (raised floor to 7 days) is the only path that unlinks the file. Tighter cleanup needs a future PR teaching blob GC to walk `.ipynb`.
- **Combined `RoomRegistry`.** `notebook_rooms` and `path_index` move into one struct behind a single tokio mutex so joined-index updates happen in one critical section. Fixes both the path-index orphan race and the tokio-mutex-across-await invariant.
- **Reservation token.** Connection paths bump a per-room `reservations` counter the moment they clone the Arc, and hold a guard until the handshake either commits or aborts. Reaper predicate includes `reservations == 0`. Closes the window where a racing reconnect has cloned the Arc but not yet incremented `active_peers`.
- **Defer destructive cleanup until after commit.** The persist debouncer is taken/dropped after the commit point so an aborted reap leaves the room fully working and the reconnect path's `bind_existing` only has to respawn autosave and watchers.

## What the plan covers

- Per-room memory footprint estimates, with a flagged open question (no benchmark exists).
- Three trigger options compared (LRU count, RSS pressure, combined), and the staged recommendation.
- Reload sequence: file-backed via `OpenNotebook(path)`, untitled via the persist file.
- Order-of-operations for reap (cleanup first, then atomic commit under registry lock, then destructive debouncer takeover).
- Concurrency table covering reservation race, reconnect-during-cleanup, kernel-teardown-concurrent-with-reap, blob-ref durability, daemon shutdown mid-reap.
- 14 integration tests + 1 pure unit test on selection logic.
- Out of scope (byte budget, RuntimeStateDoc spill, `.ipynb`-aware blob GC, tiered storage).

## Codex review history

Three rounds in this PR. Each round caught substantive issues; each was folded back into the doc. The result is the current commits. Summary of what changed:

- **Round 1:** reaper order had to flush-before-map-removal (data-loss window), drop both indices in one critical section (path-index race), and wait for the persist debouncer task to exit before any file unlink.
- **Round 2:** the joined-index operation could not be implemented with the existing two-mutex split without violating the tokio-mutex-across-await invariant; merge into a combined registry. And: removing from the maps before cleanup confirms permanently leaks rooms that hit a stuck filesystem, defeating the bound. Defer all map removal to last.
- **Round 3:** reconnect paths clone the Arc before incrementing `active_peers`, so an `active_peers == 0` predicate is not sufficient; introduce a reservation counter. Deleting the file-backed `.automerge` on reap orphans Arrow-stream sub-blob refs; keep the file. Destructive debouncer takeover before commit can break an aborted reap; defer it.

## Out of scope here

Implementation. The follow-up PR builds the reaper, the `RoomRegistry`, the reservation guard, and the `RoomPersistence::debouncer` -> `Mutex<Option<...>>` refactor.

## Test plan

- [x] No code changes
- [x] No em-dashes in committed prose
- [x] Codex review of plan, three rounds, all substantive findings folded back in

### DIFF
--- a/docs/architecture/pr2-lru-disk-spill.md
+++ b/docs/architecture/pr2-lru-disk-spill.md
@@ -1,0 +1,245 @@
+# PR 2: LRU reaper for resumable rooms
+
+PR 1 keeps a `NotebookRoom` resident after the kernel is torn down so a returning user reattaches to the same doc, outputs, and file binding. The kernel goes away, the room stays. PR 1 also added a `last_kernel_torn_down_at` timestamp on `RoomConnections`. PR 1 did not implement the reaper that eventually removes these rooms. That is this PR.
+
+This document is the plan. The implementation PR follows.
+
+## Diagnosis
+
+Today's eviction model (pre-PR-1) tore the room down 30s after the last peer left. That cost users their outputs and their in-memory doc state on every disconnect, even brief ones. PR 1 fixed the user-visible part of that. It does not bound memory.
+
+Without a reaper, a long-running daemon accumulates `NotebookRoom` instances for every notebook the user has ever opened. The persist debouncer mirrors the `NotebookDoc` to disk, but the in-memory `Arc<NotebookRoom>` and its `RuntimeStateDoc` are not bounded. A user who opens 30 notebooks over a day pays 30 rooms worth of memory until the daemon restarts.
+
+PR 2's job is the reaper: a background sweep that removes peer-less rooms once they have been peer-less long enough that resume is unlikely, capped at a maximum count.
+
+The relationship to "disk spill" is subtle. The .automerge file is already on disk for non-ephemeral rooms - the persist debouncer mirrors NotebookDoc bytes there on a 500ms / 5s schedule. PR 1 ensures the file is flushed at kernel teardown. So spill-to-disk is not new work. The new work is forgetting about the resident in-memory copy.
+
+## What is in memory per room
+
+| Field | Cost |
+|------|------|
+| `id: Uuid` | 16 bytes |
+| `doc: Arc<RwLock<NotebookDoc>>` | Automerge doc with cell source, structure, metadata. Saved size scales with edit history. |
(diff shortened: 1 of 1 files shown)


rgbkrk:
Closing without merging. The plan is good and stays referenced from this PR's diff, but committing a `docs/architecture/pr2-*.md` to `main` bakes agent-session naming ("PR 2") into the public tracker — which is silly when this is the 2,749th PR on the repo.

Implementation will be a follow-up PR (number TBD) that draws from this plan but lands as code, not as a doc.

PR #2750, opened 2026-05-13
fix(runtimed): keep rooms resident after kernel teardown

rgbkrk:
Decouple kernel teardown from room destruction. When the last peer disconnects, the daemon shuts down the kernel and frees the env directory but keeps the room (doc, outputs, file binding, watchers, autosave debouncer) resident. A reconnecting peer hits the same room and sees everything it had. A new ghost-room reaper removes rooms only after they've been kernel-less and peer-less for 24h.

## The bug

Today `peer_eviction.rs` ties kernel lifecycle to room lifecycle. After 30s no-peers, the daemon kills the kernel AND removes the room from `notebook_rooms` + `path_index`. A reconnecting peer (laptop sleep, wifi blip, app close) ends up loading from the `.ipynb` on disk, which may not have everything the live session had - the in-memory CRDT was the only complete record of recent display_data, recent edits, etc.

## Behavioral diff

After 30s no-peers:

| Step | Today | After PR 1 |
|---|---|---|
| Persist debouncer flush | yes | yes |
| ShutdownKernel RPC | yes | yes |
| Clear runtime_agent_handle + _request_tx | yes | yes |
| Flush launched deps to metadata | yes | yes |
| Final save_notebook_to_disk | yes (only when flushed deps) | yes (unconditional for file-backed) |
| Rename env dir to unified hash | yes | yes |
| Env cleanup (unless preserved) | yes | yes |
| Stop file watcher | yes | deferred to reaper |
| Stop project file watcher | yes | deferred to reaper |
| Shutdown autosave debouncer | yes | deferred to reaper |
| Remove from notebook_rooms | yes | deferred to reaper |
| Remove from path_index | yes | deferred to reaper |

Ghost reaper runs every 5 minutes. For each room: if `active_peers == 0 && !has_kernel && now - last_kernel_torn_down_at > 24h`, remove from `notebook_rooms` + `path_index` and shut down the watchers + autosave. The doc was saved at kernel-teardown time and no peers have touched it since, so removal is non-destructive - a future open reads the same content back off disk.

## Race protection

Three serialization concerns surfaced in review:

- **Connection generation counter** on `RoomConnections`. Peer connect bumps it as its first synchronous step. The kernel-teardown task snapshots it at scheduling time and re-checks under the rooms lock both at flush-success time and immediately before ShutdownKernel; the ghost reaper snapshots at candidate sample time and re-checks at remove time. A reconnect that lands in any of these windows moves the generation and the destructive op aborts.

- **`kernel_teardown_destructive` latch** on `RoomConnections`. Set by the teardown task under the rooms lock right before the destructive section, cleared once the runtime-agent handle slot is empty. The peer-connect path treats `has_kernel = true` as "no usable kernel" while the latch is set, forcing a fresh auto-launch. `auto_launch_kernel` waits (bounded, 10s) for the latch to clear before checking the handle slot, so a peer racing in mid-shutdown doesn't no-op against the doomed handle.

- **Reaper path-index ordering.** The path-index entry is removed before the rooms-map entry; if the rooms-lock check rescues the room, the path-index entry is reinserted. There's a small remaining window where a concurrent open-by-path can create a duplicate room for a rescued ghost - filed as a follow-up in a PR comment rather than another fix-round.

## Design decisions

- **Why keep watchers/autosave alive across teardown.** `get_or_create_room_result` fast-paths existing rooms at `catalog.rs:58-59` and does NOT re-bind the file watcher or re-arm the autosave debouncer on reconnect. If we shut them down at teardown, we lose them. Leaving them idle is cheap (no peers means no doc churn, the file watcher just sits on fsevents). The ghost reaper does the actual shutdown at removal time.
- **Force one final `.ipynb` save** for file-backed rooms at kernel teardown so a daemon restart in the autosave debounce window doesn't lose recent edits. Untitled rooms ride the synchronous Automerge flush already at the top of teardown.
- **`auto_launch_kernel` prefers `room.file_binding.path()`** over the `notebook_id` argument. An MCP UUID-based reconnect to a resumable file-backed room passes the UUID string as `notebook_id`; consulting the binding first stops auto-launch from misclassifying the reconnect as untitled and launching with the wrong working dir.
- **Single `last_kernel_torn_down_at` timestamp**, not two. Untitled rooms that never had a kernel still need to be reaped if they sit peerless long enough. The teardown task stamps unconditionally at the end of post-disconnect handling. Peer reconnect zeroes it so a live room can't be swept.
- **RoomState wire field defaults to Active.** Older daemons don't emit `state`, so deserializing into the new `RoomInfo` would fail without a `Default`. Pre-PR-1 daemons only ever surfaced rooms while a peer was connected, so `Active` is the correct backwards-compat read.

## Commits

- `feat(runtimed): keep rooms resident after kernel teardown`
- `feat(protocol): surface RoomState on RoomInfo`
- `feat(runt): add PATH and STATE columns to runt ps`
- `docs(runt-mcp): document RoomState semantics in list_active_notebooks`
- `test(runtimed): cover reconnect-after-teardown and ghost reaper`
- `fix(runtimed): guard teardown and reaper against reconnect races`
- `fix(runtimed): tighten teardown env cleanup, force final ipynb save, route UUID resume through bound path`
- `fix(runtimed): close teardown-vs-reconnect races at the destructive boundary`
- `fix(runtimed): reorder reaper path-index cleanup and wait for teardown in auto-launch`

## Acceptance criteria

- [x] After 30s no-peers, kernel is torn down, room remains in `notebook_rooms`, path_index entry intact, doc state preserved
- [x] Reconnecting by file path or uuid finds the existing room with its outputs and execution history
- [x] After 24h ghost, room is removed (constant; configurable TTL is out of scope for PR 1)
- [x] `runt ps` shows PATH and STATE columns reflecting active / idle / inactive
- [x] `list_active_notebooks` MCP tool returns `state` so the LLM can tell resumable rooms from active ones
- [x] Existing eviction-flow tests still pass (`test_untitled_notebook_persists_through_eviction`, `test_eviction_flushes_before_reconnect`)
- [x] New tests cover reconnect-after-teardown, reaper-removes-after-TTL, reaper-skips-reconnected-room, and generation-bump-on-reconnect

## Test plan

- [x] `cargo test -p runtimed --lib` (748 pass)
- [x] `cargo test -p runtimed --test integration` (44 pass, 4 new)
- [x] `cargo test -p runtimed --test tokio_mutex_lint`
- [ ] Manual smoke: open a notebook, run a cell, force-quit the app, reopen - room should be inactive in `runt ps`, reconnecting from the app should restore outputs.

## Out of scope

LRU/disk-spill of resident rooms, sticky kernel reattach, configurable ghost TTL via CLI/config, telemetry on reaper sweeps, the path-index race window in the rescue path (PR comment). These can layer on top of the resident-rooms invariant once it's in.

### DIFF
--- a/crates/notebook/src/mcpb_install.rs
+++ b/crates/notebook/src/mcpb_install.rs
@@ -70,9 +70,9 @@ pub fn install_mcpb(app: &tauri::AppHandle) -> Result<PathBuf, String> {
             }
         },
         "tools": [
-            { "name": "list_active_notebooks", "description": "List running notebook sessions." },
+            { "name": "list_active_notebooks", "description": "List active notebook sessions." },
             { "name": "connect_notebook", "description": "Attach to a notebook. Pass path (.ipynb) or notebook_id (UUID) — not both." },
-            { "name": "create_notebook", "description": "Create a new notebook. Ephemeral by default; use environment_mode=\"notebook\" to ignore project files for env selection, and save_notebook(path) to persist." },
+            { "name": "create_notebook", "description": "Create a notebook. Ephemeral by default; save_notebook(path) to persist." },
             { "name": "save_notebook", "description": "Save notebook to disk. For notebooks created with create_notebook(), you must provide a path." },
             { "name": "show_notebook", "description": "Open the notebook in the nteract app for the user. Headless: returns a structured no-display reason." },
             { "name": "disconnect_notebook", "description": "Release a notebook session's peer connection. Omit notebook_id to disconnect the active session." },
@@ -84,10 +84,10 @@ pub fn install_mcpb(app: &tauri::AppHandle) -> Result<PathBuf, String> {
             { "name": "move_cell", "description": "Move a cell to a new position." },
             { "name": "execute_cell", "description": "Execute a code cell." },
             { "name": "run_all_cells", "description": "Execute all code cells in order." },
-            { "name": "get_results", "description": "Get outputs for an execution by ID. Returns status (done/error/running/queued) so you know if outputs are complete. Use the execution_id from execute_cell, set_cell(and_run), or run_all_cells." },
+            { "name": "get_results", "description": "Get outputs and status (done/error/running/queued) for an execution_id." },
             { "name": "interrupt_kernel", "description": "Interrupt execution." },
             { "name": "restart_kernel", "description": "Restart the kernel, clearing all state." },
-            { "name": "manage_dependencies", "description": "Review or update notebook dependencies. With no parameters, returns current dependencies, dependency fingerprint, and trust state. Use add/remove arrays for edits; set trust=true to approve the resulting dependency metadata; set apply='sync' or 'restart' to apply." },
+            { "name": "manage_dependencies", "description": "Review or update notebook dependencies. Returns current deps, fingerprint, and trust state." },
             { "name": "replace_match", "description": "Replace literal text in a cell. Use context_before/context_after to disambiguate repeated matches." },
             { "name": "replace_regex", "description": "Replace a regex match in a cell (fancy-regex). Fails if 0 or >1 matches. Replacement is literal text." }
         ],
--- a/crates/runt-mcp-proxy/tool-cache.json
+++ b/crates/runt-mcp-proxy/tool-cache.json
@@ -7,7 +7,7 @@
       "openWorldHint": false,
       "readOnlyHint": true
     },
-    "description": "List running notebook sessions.",
+    "description": "List active notebook sessions.",
     "inputSchema": {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "description": "Empty params for tools that take no arguments.",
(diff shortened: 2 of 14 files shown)


rgbkrk:
Codex review round 4 flagged one remaining P2 race in the ghost-room reaper that I didn't ship in this PR. Capturing it here for follow-up rather than another fix-round:

**P2 - ghost reaper temporarily un-indexes a rescued room's path** (`crates/runtimed/src/daemon.rs:4015-4018`)

The current reaper:
1. Reads the room's bound path.
2. Removes the path_index entry.
3. Re-checks `notebook_rooms` under the rooms lock - if the room was rescued by a concurrent reconnect, declines to remove from the rooms map.
4. Reinserts the path_index entry for the rescued room.

In the window between step 2 and step 4, a concurrent `open_notebook(path=...)` lookup can land. It sees no path_index entry, falls through to `get_or_create_room_result` with a freshly minted UUID, and creates a second room for the same file. When the reaper's step 4 tries to reinsert, the path_index slot may now point at the duplicate UUID (insert fails) and the rescued room ends up orphaned from path lookup.

The fix needs the path-index touch and the rooms-map remove to be atomic with respect to `find_room_by_path`. Either:
- Hold both locks together (requires switching one of them off tokio::sync::Mutex to avoid the await-with-guard rule), or
- Compute the "really remove?" decision under the rooms lock, then under the rooms lock + path_index lock together, do the dual removal (rooms first, path second is fine if both are committed atomically), or
- Have `find_room_by_path` re-check the rooms map after the path_index lookup and treat "path mapped to UUID, but UUID has no room" as a transient state to retry briefly.

The race window is very small (between two adjacent `path_index.lock().await` blocks in the reaper), and the worst case is a duplicate room for a notebook that was about to be reaped anyway, so this isn't a release blocker. Filing as a follow-up.

PR #2801, opened 2026-05-21
docs(architecture): identity and trust for notebook rooms

rgbkrk:
ADR for identity and trust in nteract notebook rooms. Six decisions, a Limitations section with a threat table, and four worked examples (local desktop, Anaconda-from-desktop, agent on the user's behalf, publish).

## Decisions

1. **Two-layer actor labels.** Every Automerge actor looks like `<principal>/<operator>`. The principal is who's authenticated (`local:quill`, `user:anaconda:<sub>`, `hub:<domain>:<user>`, `system`). The operator is what's doing the work right now (`desktop:7f3a`, `agent:claude:s1`, `runtime:py-3.12`) and gets to pick its own label. One human, many operators.

2. **Rooms are entities; access is a per-room ACL.** A room lives at a host-derived URI (e.g., `runtimed.com/n/<uuid>`) and has an ACL: a list of `(principal, scope)` pairs. The ACL can include principals from any IdP the host validates against. Single-IdP per host is convention for v1, not a protocol rule.

3. **Authentication happens at connect; per-frame work is scope checks only.** The IdP gets called once when the WebSocket opens. After that we check each frame against the connection's scope (frame type + whether `Message.changes` is empty). We do *not* validate Automerge actor labels per change in v1; that's deferred. See Limitations.

4. **`IdentityProvider` is a closed enum.** Three variants: `Local`, `Oidc`, `JupyterHub`. Anaconda is OIDC underneath, so it rides on the `Oidc` variant via issuer URL + scope translation. Everything lives in one `nteract-identity` crate; providers are modules with Cargo features.

5. **Four scopes: viewer, editor, runtime_peer, owner.** Frame-level rights are enforced server-side. Editor writes `NotebookDoc` freely; editor's write surface on `RuntimeStateDoc` is narrow (the widget comm-state subtree), mediated client-side by the existing approved comm writer.

6. **Publish creates a fresh document in the destination space.** No historical actor labels carry across identity boundaries. The destination records who *published*, not who edited individual changes in the source. This sidesteps the unsigned-history forgery problem (think unsigned git commits). When signed changes land (keyhive direction), this door can open.

## Limitations

| Attack | Mitigated by | Status |
|--------|--------------|--------|
| Unauthorized read | ACL check at connect | closed |
| Unauthorized write at all | Bearer auth + ACL at connect | closed |
| Scope escalation (viewer authoring, runtime_peer editing NotebookDoc) | Scope-based frame rejection | closed |
| Import-time cross-IdP forgery (publishing a doc with historical foreign actor labels) | Publish-as-fresh-doc | closed |
| Bearer-token replay (stolen JWT opens new socket) | DPoP / mTLS / short lifetimes | deferred |
| **Intra-room actor spoofing** (an authenticated peer claims any operator suffix, any ACL principal, or even a foreign principal) | None in v1 | **deferred** |

The intra-room spoofing case is deferred for v1, not punted indefinitely. It covers both intra-IdP spoofing (claiming a different operator suffix or another ACL principal) and cross-IdP spoofing (claiming a principal not in the ACL at all, e.g. a `local:*` actor authored by an authenticated Anaconda peer). Both fall out of the same root cause: v1 doesn't parse Automerge `Change`s pre-apply. Building that needs `automerge::sync::Message.changes` parsing (`ChunkList(Vec<Vec<u8>>)` of raw chunks; V1 = one `Change` per chunk; V2 = potentially a whole-doc save), which is the companion `automerge-fork-patches.md` proposal. Lands behind the room-host crate extraction PR (the natural calling site). Until then, the attack surface is bounded to peers who already have legitimate edit access. Acceptable until the fork patch ships.

The Automerge `filters` work (upstream `origin/filters`, post peer review) is the natural complement: rules like `Allow / AllowUpTo { heads } / Deny` per-actor or per-author. They don't reject pre-storage but they suppress rendering, which is what you want for runtime revocation and post-hoc audit hiding. When filters lands in main (or we cherry-pick it onto our fork), pairing it with our pre-apply validator gives full attribution integrity. Keyhive is orthogonal (change-level capability tokens + signed changes); composable later.

## How credentials get on the wire

Browsers can't set custom headers on `new WebSocket(...)`, which is annoying. Four mechanisms supported:

- **Subprotocol smuggling** (Kubernetes pattern) for browser `Oidc` clients
- **One-time ticket** as a higher-security opt-in for `Oidc`
- **Cookie** for `JupyterHub` (Hub already issues one)
- **`Authorization` header** for system-to-system (desktop daemon, agents, CLI)

Parsing each of these is the listener's job; providers see a normalized `Credential` enum.

## Companion drafts

- `docs/architecture/deployment-topology.md` - network routing, kernel placement, credential-keyring placement. Out of scope here on purpose.
- `docs/architecture/automerge-fork-patches.md` - what we want to add to our Automerge fork. Patch 1 (a public change-chunk parser on `sync::Message`) is the one that unblocks the pre-apply validator in the Limitations table. The deferred item gets a landing spot.

## What we're not deciding here

Room-host crate extraction; picking the Anaconda IdP for v1; anonymous viewer scope; the revocation wire signal + filters-based subduction; connection-scope hint on the wire; federation across hosts; `runtime_peer` topology for remote runtimes; deployment topology (companion draft); ACL mechanics (separate ADR); signed-change authorship across publish (keyhive direction); bearer-token replay mitigation; pre-apply actor-label validator (companion draft for the fork patch that unblocks it).

## References

- intheloop `backend/auth.ts`, `backend/sync.ts` - JWKS bearer, validate-on-connect.
- runtimed/anaconda `api_key.ts` - OIDC userinfo + scope mapping.
- JupyterHub `services/auth.py` - Hub cookie/token pattern, no `X-Forwarded-User`.
- Kubernetes WebSocket auth via `Sec-WebSocket-Protocol` - the subprotocol-smuggling precedent.
- Upstream Automerge `sync::Message::ChunkList` and the `read_only` / `peer_read_only` distinction.
- Automerge `filters` branch (`rust/automerge/src/filter.rs`).
- Keyhive (https://www.inkandswitch.com/keyhive/notebook/).


docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,373 @@
...
+The principal is what the trust gate enforces. The operator is what attribution distinguishes. Conflating them - as today's `agent:claude:abc123` does - makes "Claude on kylekelley's machine" indistinguishable from "Claude on someone else's machine" and forces the trust gate into either case-by-case rules or implicit trust of operator names.
+
+Two layers also matches reality: operators are processes, principals are accounts. Unix already works this way. Audit logs in well-designed systems already work this way.
+
+## Decision 2: Identity space is per-room
+
+Each notebook room belongs to exactly one identity space, determined by its host:
+
+- A room served by the local daemon: identity space is `local:*`. Principal derived from peer credentials on the listening socket.
+- A room served by Anaconda hosted: identity space is `user:anaconda:*` for human users, plus `system` for seed actors. Principal derived from validated Anaconda credential.
```
intentionally-left-nil: The identity of a room being tied to a user feels limiting to me as we start to network things together. Instead, I would think we would have a difference between users, and the things that they have access to.

So in this case, the room could be an entity by itself. It becomes `local_notebook:uuid` or `local_notebook:/path/tofile`,

Then, you can have an expressive permission set where different users get various access levels to the notebook room. For example, you might require that there be an owner, like `user:anaconda:*` or `local:*`. Perhaps it could be an org in the future. Perhaps one user has write access, but another user has read access.

Tying the resource to the user constrains this because it means that changing the owner then breaks the entire room URI, and any kind of shared thing (like owned by an org), or other related items become difficult as well

docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,374 @@
...
+- A room served by a JupyterHub-spawned host: identity space is `hub:<hub-domain>:*`. Principal derived from validated Hub token/cookie.
+
+A connection to a room is authenticated into that room's space. The principal returned by authentication is in the room's namespace, not the client's process namespace. **One client process can hold connections to many rooms in many spaces simultaneously; each connection's principal reflects which room and how it authenticated, not which host it lives on.**
+
+The validator on a room only accepts NEW changes whose actor principal matches the connection's authenticated principal (or the system seed). Historical changes already in the doc are not re-validated; they're part of the merkle DAG and immutable.
+
+### Cross-space scenarios
```
intentionally-left-nil: I think this is conflating things in the wrong way. It's implementation focused, not spec or protocol focused. I don't think we should be enforcing a particular network routing from an auth perspective. If a user wants to connect to their local daemon and have that proxy & rewrite credentials, fine. If the user wants to directly talk to a remote daemon, that should also be fine.

Network routing feels like a completely separate ADR, where you could describe this proxy feature

docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,374 @@
...
+
+The daemon is a **bridge**, not an identity terminator. It can be a peer in `local:*` rooms for kylekelley's untitled scratch notebooks and a peer in `user:anaconda:*` rooms for hosted notebooks at the same time, in the same process.
+
+## Decision 3: Authentication at connect, validation per frame
+
+Authentication is connection-scoped. The IdP is consulted at WebSocket open, never on the sync hot path.
+
+1. Connection opens. Listener extracts the credential (bearer JWT for OIDC/Anaconda, cookie/OAuth token for JupyterHub, peer creds for Unix socket).
+2. Identity provider validates the credential. One round-trip (JWKS fetch for OIDC is cached; `/hub/api/user` for JupyterHub; userinfo for Anaconda).
+3. Validation yields an `AuthenticatedConnection { principal, scope, allowed_operator_kinds }`.
```
intentionally-left-nil: What's the lifetime of this authenticatedconnection token? How do you prevent a user from taking an AuthenticatedConnection token from a previous session, and reuse it in the future?

E.g.
User 1 connects to room_1, gets an authenticatedConnection,

One week later...
User 1 tries to piggyback on someone else's websocket connection to talk to it, passing in the old authenticatedConnection (since there was no lifetime on it).


The main point, is that if you don't want to re-verify, the connection needs to be tied to the socket, not the principal

docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,374 @@
...
+pub enum Credential {
+    BearerToken(String),     // from subprotocol, ticket exchange, or Authorization header
+    Cookie(String),          // from upgrade-request cookies
+    UnixPeer(PeerCredInfo),  // SO_PEERCRED for the local daemon's listener
+}
+```
+
+The extraction logic (parse subprotocol header, look up ticket, read cookie, query peer creds) is listener-side concerns; it does not belong in the providers themselves. A `CredentialExtractor` trait sits on the listener implementation and feeds the provider a normalized `Credential`.
+
+## Decision 5: Four scopes
```
intentionally-left-nil, line 244: Scopes need to be called out more in this document. For a viewer scope, what part of the auth flow checks scopes? How do we declare new functionality that requires certain scope access? What's the security boundary to prevent someone from escalating their privilege due to a bad scope (design problem), or a bad layer for enforcing it, etc

docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,374 @@
...
+- `editor` - full live edit. Today's desktop peer.
+- `runtime_peer` - permitted to write `RuntimeStateDoc` and emit kernel lifecycle broadcasts. Cannot edit `NotebookDoc`. Used by future remote-runtime services and by JupyterHub-spawned room-hosts when they connect a kernel.
+- `owner` - editor plus publish-revisions and manage-ACLs requests.
+
+Scope is determined by the IdP at authentication. It does not need to be conveyed on the wire today; it's enforced server-side. A future protocol extension may add a `connection_scope` field to the handshake response so the client can surface scope to the UI.
+
+## Decision 6: Publish is import, not transformation
+
+A publish operation copies a notebook from one identity space into another. The Automerge changes carry whatever actor labels were on them when authored. The destination room's validator does not re-validate historical changes; they're part of the merkle DAG and immutable.
```
intentionally-left-nil: This has the problem of git, where if you take an old repo, there's no guarantee that it's actually by that author. You end up needing similar solutions (unverified, verified, GPG labels and such). Not saying we have to solve it all now but it's worth thinking about

rgbkrk: Yeah. I figured the best step here for now was to just treat it like a brand new document for now. We lose history completely but then it's manageable with identity separate from local.

docs/architecture/identity-and-trust.md
rgbkrk: Fair. I brought this up here because I wanted to make sure I thought through what attribution on the document means, since it's core to automerge ops.

docs/architecture/identity-and-trust.md
rgbkrk: Yeah, you're right. Reframed Decision 2 around room-as-entity with a per-room ACL of (principal, scope) pairs. The ACL can include principals from any IdP the host validates, so cross-IdP collaboration and ownership transfer don't break the URI. Federation and ACL mechanics are tracked as separate follow-ups.

docs/architecture/identity-and-trust.md
rgbkrk: Worth being explicit about. `AuthenticatedConnection` is server-side memory keyed to a single socket, not a token, not transferable, not presented by a client. There's nothing to replay because it never leaves the server. Added a "what `AuthenticatedConnection` is and is not" subsection so the naming doesn't invite the misreading.

If you're after token replay (someone steals my JWT and opens a new socket), that's inherent to bearer auth. DPoP / mTLS / short token lifetimes are noted as future-compat tightening.

docs/architecture/identity-and-trust.md
rgbkrk: Fair gap. Expanded Decision 5 with where scope comes from (intersection of IdP claims and per-room ACL, narrower wins), three enforcement points (frame ingress, request dispatch, ACL mutations), and a rule that every new scope-gated request declares its required scope at the definition site so we don't end up sprinkling `if scope == ...` checks around. Scope is not on the wire today; a future advisory `connection_scope` field in the handshake response is for UIs only.

docs/architecture/identity-and-trust.md
rgbkrk: Decided to defer the pre-apply actor-label validator entirely for v1 rather than ship a chunk-parsing dance.

Reasoning: the ChunkList(Vec<Vec<u8>>) shape needs either an upstream Automerge contribution (a pre-apply hook on receive_sync_message or a public parse_change_chunks that handles V1 + V2 safely), another patch on our already-existing fork, or a throwaway-peer hack per frame. Combined with publish-as-fresh-doc closing the cross-space leak (Decision 6), the remaining risk surface is bounded to intra-room operator/principal spoofing by peers that already have legitimate edit access.

What v1 still does at the frame layer: scope-based rejection (frame type + Message.changes emptiness check). That's cheap and doesn't need any Automerge contribution.

Added a Limitations section to the ADR with a threat table that names intra-room spoofing explicitly as a deferred hardening item. Filters (upstream, post-peer-review) and keyhive both pair naturally with the validator when we do add it; they're not substitutes.

Landed in `7c88f970`.

docs/architecture/identity-and-trust.md
```diff
@@ -0,0 +1,399 @@
...
+
+2. **One-time ticket**. Client POSTs `/api/session-tickets` with the real bearer (header-set, normal CORS). Server returns a short-lived (~10s, single-use) ticket. Client opens `wss://host/n/<id>?ticket=<one-time>`. Server validates the ticket, consumes it, and the connection is authenticated as the original user. The real bearer never appears in the WebSocket URL. Costs one extra round trip; the server tracks outstanding tickets in memory or D1.
+
+3. **Cookie**. Browsers send cookies automatically on the WS upgrade when same-site (or with the right CORS dance). This is the path JupyterHub already provides because Hub login issues a signed session cookie. No client code needed.
```
minrk, line 218: Cookie's not the most convenient with JupyterHub because you have to deal with CSRF protections as well. It's preferable to send tokens via other means (currently url param or Authorization header).

minrk (commented):
This looks solid. The one thing I think is missing is Authorization - how do you resolve an Identity to a given scope with a given identity provider. Is that just a local map of `sub:scope` or a resolution of `idpscope:nteract scope` somehow? Do you assume the identity provider already encodes nteract's roles?

docs/architecture/identity-and-trust.md
rgbkrk: That's so true. I'll lock this down further before we roll it out as a demo surface.

PR #2804, opened 2026-05-22
feat(notebook-cloud): prototype Cloudflare room host

rgbkrk:
## Summary

Adds `apps/notebook-cloud`, a Cloudflare Worker prototype for hosted nteract notebook rooms.

The prototype:

- exposes `/n/:notebookId/sync` as a WebSocket endpoint
- routes each notebook id to a Durable Object room
- stamps trusted internal `x-nteract-*` identity headers before forwarding to the DO
- maps identity into `<principal>/<operator>` actor labels matching the merged identity ADR/crate shape
- supports explicit anonymous public viewers as `anonymous:<session>/browser:<session>` with `viewer` scope
- relays typed-frame-v4-shaped binary frames using shared `runtimed` frame constants
- enforces viewer/editor/runtime_peer/owner scope gates for notebook/runtime/blob/request/pool frames
- rewrites JSON prototype presence actor principals to the authenticated principal
- rejects malformed/non-JSON presence payloads instead of rebroadcasting bytes that real clients cannot decode
- keeps anonymous viewer presence local-only: accepted to the sender, not broadcast or persisted as room activity
- persists bounded room frame metadata to DO storage and best-effort D1 room events
- supports hibernatable DO WebSockets with socket attachments, while preserving a local fallback for runtimes without `acceptWebSocket`
- adds D1/R2-backed snapshot, render-cache, blob, catalog, and room-event APIs
- renders a public read-only notebook at `/n/:notebookId`
- keeps the low-level room harness at `/n/:notebookId/debug`

The viewer source of truth is the persisted `NotebookDoc` Automerge snapshot in R2. For this slice, `scripts/publish-demo.mjs` creates a real `NotebookDoc` with shared `runtimed-wasm`, uploads the `.am` snapshot, and uploads a JSON render cache generated by `NotebookHandle.get_cells_json()`. That cache lets the Worker render deployed notebooks without committing the volatile generated WASM bundle yet.

## Review Hardening

Follow-up hardening from the latest review pass:

- remote dev credentials now require the `NOTEBOOK_CLOUD_DEV_TOKEN` Worker secret through `X-Notebook-Cloud-Dev-Token` or `dev_token`; local Wrangler loopback remains frictionless
- remote owner self-assertion without the token returns `401`
- dev-token comparison now uses a fixed-loop byte comparison instead of direct string equality
- non-JSON/CBOR presence is rejected until the shared presence codec lands, avoiding invalid rebroadcasts to real clients
[...]

PR #2813, opened 2026-05-23
docs(architecture): four desktop ADRs + cleanup punchlist

rgbkrk:
## Summary

Four architecture decision records covering the load-bearing pieces of the **real desktop app** (the parked `apps/notebook-cloud/` prototype from #2804 is explicitly out of scope here).

Each ADR was drafted from code research, then accuracy-reviewed by both Claude and Codex against the actual source. Every reviewer pass found at least one factual error in the draft (work-channel capacity off by 3x, wrong test names, runtime-agent framed as an exception rather than a distinct writer path, FIFO cache mislabeled LRU, required-heads timeout described as graceful-degrade when it actually fails closed). Those were corrected before merge.

## ADRs added

- **`typed-frame-v4-wire-protocol.md`** — magic bytes, preamble version policy, frame types and per-type size caps, scope semantics, the `PUT_BLOB` sub-framing. The single most load-bearing undocumented contract in the repo: every peer (desktop, daemon, MCP, kernels) speaks it.
- **`three-document-split.md`** — why state is split across `NotebookDoc`, `RuntimeStateDoc`, and `PoolDoc` instead of one Automerge document. Per-doc write authority (and the two distinct writer paths into `RuntimeStateDoc`), persistence asymmetry, lifecycle, and the read-only ingress strip on `PoolDoc`.
- **`execution-pipeline.md`** — the seven invariants behind cell execution: `RuntimeStateDoc` as durable record, control-plane separated from output transport, stream committer priority path for ordering boundaries, best-effort comm replay, coalesced display updates with flush-before-`ExecutionDone`, `required_heads` as a fail-closed causal gate.
- **`blob-storage-and-content-addressing.md`** — SHA-256 hex addressing with two-level sharding, atomic write plus sidecar metadata, the 1 KiB inline threshold, the `PUT_BLOB` upload protocol with multipart above 32 MiB, the `Durable` / `Ephemeral` monotone promotion, mark-and-sweep GC, and the `BlobResolver` indirection that lets the same snapshot pair render against a desktop daemon or a cloud origin.

## Cleanup punchlist

`docs/architecture/cleanup-punchlist.md` tracks **42 code smells** surfaced while writing the ADRs. Triaged into:

- **Inline** (5 items): doc corrections safe to fix on this branch alongside the ADRs. The `Handshake::Blob` row in `crates/notebook-wire/AGENTS.md` was confirmed nonexistent and removed here.
- **Targeted PR** (15 items): one-or-two-file behavior fixes after the ADRs land. Examples: presence cap disagreement (1 MiB wire vs 4 KiB semantic), Rust↔TS per-type cap table drift, `recv_typed_frame` allocating 100 MiB on unknown frame types before skip, `set_execution_done` race recovery, FIFO cache mislabeled LRU.
- **Design** (14 items): needs an ADR decision before code moves. Examples: server-side path enforcement on `RuntimeStateDoc` editor writes, cross-doc heads correlation, blob HTTP authentication when the daemon stops being single-user, `BlobBackend` trait for the R2 path.

## Review process

For each ADR:

1. Drafting agent researched the code and produced a 250-450 line draft.
2. Claude reviewer agent fact-checked every concrete claim against the code.
3. Codex (`codex review`) reviewer agent did an independent second pass.
4. Combined CONFIRMED / INCORRECT / MISSING findings drove a revision commit.

Each ADR has a per-step commit in the branch history, so the drafting → review → revision arc is reconstructible.

## Verified

- `git log` shows 11 commits on this branch off `main`.
- No Rust or TypeScript code changed; the only non-doc edit is the stale `Handshake::Blob` row removal in `crates/notebook-wire/AGENTS.md` (also a doc).
- ADRs cross-reference each other through a `Neighbors` block at the top.

## Status

All four ADRs are `Draft, 2026-05-22`. They were not promoted to `Accepted` because the punchlist surfaces Design-class items that may move some decisions. Promoting to `Accepted` is a follow-up after the Design items are triaged.

## Not in this PR

- No cleanup PRs for the punchlist items (those are follow-ups).
- No changes to `apps/notebook-cloud/` (parked prototype from #2804).
- No changes to the four existing hosted-room ADRs (`identity-and-trust.md`, `automerge-fork-patches.md`, `deployment-topology.md`, `hosted-notebook-artifacts.md`).


rgbkrk:
I read this against the notebook-cloud direction we have been moving toward. Overall this lines up with the path I think we want: snapshot pairs are the durable artifact, render JSON is a cache/projection, blob refs stay structured, and the host supplies a resolver rather than letting WASM or generic renderers bake in deployment routes.

A few non-blocking decision points I would make explicit before these move from Draft to Accepted:

1. **Separate output blobs from renderer/plugin assets in the ADR language.** The blob ADR correctly mentions `BlobResolver` for notebook output bytes and `rendererAssetsBaseUrl` for plugin sidecars, but I would call out that these are intentionally separate host contracts. Output blobs may move to R2/signed `outputs.*` origins; renderer assets are static CDN assets. Generic Sift/renderer code should only read host context, not infer paths like `/api/n/...`.

2. **Make `blobBasePath` / resolver configuration server-owned.** For hosted viewers, the HTML shell or room host should pass the blob URL strategy (`/api/n/:id/blobs/`, signed origin, CDN prefix, etc.). The browser bundle should consume that config and construct a `BlobResolver`; it should not recompute cloud route shape from `notebookId`. This keeps future CDN/signed-origin changes out of renderer code.

3. **Treat the `(NotebookDoc heads, RuntimeStateDoc heads)` pair as a publish-artifact boundary, not necessarily a live-doc field.** The three-doc ADR’s cross-doc heads correlation point is right. For cloud publish, I think D1 revision metadata is the right place to pin that pair. We probably do not want to force the live desktop docs to reference each other just to satisfy published replay/audit.

4. **Move blob reachability from open question to publish invariant.** BS-4 is the biggest practical issue for the Polars/Sift path: publishing should not update `latest_revision_id` until every `ContentRef::Blob` reachable from the saved `NotebookDoc` + `RuntimeStateDoc` pair exists in the destination blob store. That may live in hosted-artifacts rather than the desktop blob ADR, but it should be a required publish step.

5. **Keep `BlobBackend` separate from `BlobResolver`.** A backend abstraction for filesystem/R2 is useful, but it is a storage concern. The browser/viewer side only needs the resolver contract. Keeping those separate should prevent the cloud viewer from depending on daemon filesystem assumptions while still allowing a later R2-backed store.

I would consider these alignment notes, not merge blockers. The ADRs are already describing the shared reality we need for notebook-cloud to stop being a parallel track.

rgbkrk:
Applied 1, 3, and 5 in `e59f2cfe`. Brief notes on each, including the two I deliberately did less with:

**1. Separate output blobs from renderer/plugin assets** — added explicit "two separate host contracts" language to Decision 7 of the blob ADR. Generic renderer code reads whichever host context applies; it should not reconstruct cloud routes from `notebookId` or any other doc-level value.

**3. Heads pair as publish-artifact boundary** — reframed three-doc-split open question #7. The pair belongs in publish metadata (hosted D1 row today, future desktop export artifact), not in the live documents. Forcing live docs to reference each other's heads buys nothing for editing and adds write-amplification.

**5. `BlobBackend` separate from `BlobResolver`** — rewrote blob ADR open question #2 to make the storage-vs-host-contract split explicit. Storage backend (filesystem, R2, future) is distinct from resolver (browser/viewer contract for fetching by hash). Conflating them would force the viewer to know about daemon filesystem assumptions.

**2. Server-owned `blobBasePath` / resolver config** — read this as a hosted-side concern. The desktop ADR doesn't have a server; the daemon is the host. The "host supplies the resolver" framing in Decision 7 already implies the contract direction, so I didn't add a separate decision. Happy to add one in `hosted-notebook-artifacts.md` if the cloud side wants the contract pinned there explicitly.

**4. Blob reachability as a publish invariant** — kept as an open question in the desktop ADR but pointed at the hosted-artifacts ADR as the natural home. Desktop save has no publish boundary; the room either has the blobs or it doesn't. Hosted publish does have a boundary, and that's where `latest_revision_id` should be gated on reachability.

Also tightened the streaming-read open question (BS-1) to flag it as a hosted-path *requirement* rather than a desktop nice-to-have — a multi-user worker that loads each 100 MiB blob into RAM per fetch will not scale. Same for the backend abstraction (BS-6).

rgbkrk:
Cloud direction overlap, now that this ADR set has landed:

- `blob-storage-and-content-addressing.md` matches the cloud prototype direction well: keep document bytes host-neutral, keep `ContentRef::Blob` refs intact, and make the host provide a `BlobResolver` / URL strategy. That is exactly what the notebook-cloud viewer is doing now for R2-backed `/api/n/:id/blobs/:hash`.
- The renderer sidecar note under local blob HTTP is also the right boundary. Desktop can keep `/plugins/{name}` on the daemon HTTP origin; hosted viewers can point `NteractEmbedHostContext.nteract.rendererAssetsBaseUrl` at a CDN/Worker asset origin. PR #2827 follows this by splitting Sift WASM onto `nteract-notebook-cloud-assets`, without loosening iframe sandbox/CSP.
- The cleanup punchlist calls out the cloud-critical gaps I would prioritize before calling notebook-cloud real: BS-4 (publish completeness must prove every blob ref in the snapshot pair exists), BS-5 (blob HTTP auth stops being same-user once we leave localhost), BS-6 (`BlobBackend` or equivalent so filesystem/R2 are one abstraction), and 3D-5 (persist the `(NotebookDoc heads, RuntimeStateDoc heads)` pair as a first-class revision fact).
- For the Worker/Durable Object path, `typed-frame-v4-wire-protocol.md` is the source of truth for per-frame caps and `PUT_BLOB`. Any cloud relay should import or generate from the same contract instead of copying cap tables by hand.

The one distinction I would keep explicit in future docs: the dedicated renderer asset origin solves plugin sidecars like Sift WASM. It is not the same thing as a production output iframe origin or private blob origin. We should still plan a separate output/blob origin with signed or scoped access before hosting private notebooks.

PR #2849, opened 2026-05-23
Harden MCP app plugin output fallbacks

rgbkrk:
## Summary

- Advertise the MCP output app CSP on `resources/list` as well as `resources/read`.
- Resolve blob-backed `text/plain` fallbacks before displaying them when plugin loading/rendering fails.
- Add a focused MCP app regression test for the plugin-failure fallback path.

## Verification

- `pnpm test:run apps/mcp-app/src/components/__tests__/mime-renderer.test.tsx apps/mcp-app/src/lib/__tests__/plugin-loader.test.ts apps/mcp-app/src/lib/__tests__/plugin-executor.test.ts`
- `cargo test -p runt-mcp output_resource --quiet`
- `cargo fmt --check -p runt-mcp`
- `git diff --check`

## Notes

- The local MCP widget build was regenerated after the source change, but `crates/runt-mcp/assets/_output.html`, `python/nteract/src/nteract/_widget.html`, and `apps/mcp-app/dist/*` are ignored and not part of this PR.
- The PR stays draft for external review and CI handoff.

### DIFF
--- a/apps/mcp-app/src/components/__tests__/mime-renderer.test.tsx
+++ b/apps/mcp-app/src/components/__tests__/mime-renderer.test.tsx
@@ -1,0 +1,47 @@
+// @vitest-environment jsdom
+import { render, screen } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vite-plus/test";
+import { loadPluginForMime } from "../../lib/plugin-loader";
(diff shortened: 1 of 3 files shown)


rgbkrk:
Superseded by #2852, which replaces this CSP/fallback experiment with host-facing MCP app logging so nightly can collect real plugin and resize failure signal first.

PR #2865, opened 2026-05-23
fix(notebook-cloud): link traceback cells in cloud viewer

rgbkrk:
## Summary

- wire cloud read-only notebooks into the rich traceback cell resolver/navigator
- preserve `execution_id` from snapshot render cells so traceback frames can resolve back to cells
- shorten displayed Python package paths while keeping non-Python and malformed paths safe/fallback-rendered

## Verification

- `cargo xtask lint --fix`
- `pnpm test:run src/components/outputs/__tests__/traceback-output.test.tsx src/components/cell/__tests__/ReadOnlyNotebookCell.test.tsx src/components/cell/__tests__/ReadOnlyNotebook.test.tsx`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/render-resolution.test.ts test/viewer-sift-preload.test.ts`
- `pnpm --dir apps/notebook-cloud build:viewer`

### DIFF
--- a/apps/notebook-cloud/test/render-resolution.test.ts
+++ b/apps/notebook-cloud/test/render-resolution.test.ts
@@ -147,6 +147,7 @@ describe("cloud viewer render resolution", () => {
         id: "runtime-count",
         cell_type: "code",
         source: "df",
+        execution_id: "exec-runtime-count",
         execution_count: "null",
         outputs: [
           {
@@ -167,6 +168,7 @@ describe("cloud viewer render resolution", () => {
     );
 
     assert.equal(cell.executionCount, 7);
+    assert.equal(cell.executionId, "exec-runtime-count");
   });
 
   it("uses the first finite execute_result count as the output fallback", async () => {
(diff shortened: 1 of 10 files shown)


rgbkrk:
One architecture note before this lands: traceback navigation should stay cell-id centric. `execution_count` is useful display metadata at best, but it should not be part of `TracebackCellTarget` or any lookup contract. Execution counts can be null, reused across kernel restarts/imports, and are less stable than the notebook/runtime link we actually need.

The durable path should be:

- preserve `execution_id` on materialized cells/outputs
- build `execution_id -> { cellId }` (optionally source hash aware later)
- have `TracebackOutput` navigate using the resolved `cellId`
- avoid adding `executionCount` to `TracebackCellTarget` and avoid labels like `In[n]` in the target itself

If we want to show an execution count somewhere, keep it as separate presentation data, not target identity. This also keeps the iframe traceback bridge in #2864 compatible: it passes resolved cell targets through the render payload and posts back `traceback_navigate` with `{ cellId }`.

rgbkrk:
Addressed in `8c75ce9d`: `TracebackCellTarget` is back to identity-only (`cellId`, optional `label`), cloud builds `execution_id -> { cellId }`, and the traceback label no longer carries `In[n]`. I kept the defensive malformed-frame fallback too, so path shortening cannot throw the renderer.

PR #2888, opened 2026-05-24
fix(notebook-cloud): harden hosted credential transport

rgbkrk:
## Summary

- strip credential-bearing WebSocket subprotocols and select only `nteract.v4`
- reject mixed Access/dev credential sources before auth dispatch
- require an allowed `Origin` for cookie-backed Cloudflare Access WebSockets
- stamp Access identity metadata separately from the ACL principal
- add browser-session auth UI state for Access trials without storing token material

## Verification

- `cargo xtask lint --fix`
- `pnpm --dir apps/notebook-cloud test`
- `pnpm --dir apps/notebook-cloud exec tsc -p tsconfig.json --noEmit`

## Notes

- `pnpm --dir apps/notebook-cloud exec tsc -b --pretty false` was attempted but the existing `tsconfig.json` has `allowImportingTsExtensions` without build-mode `noEmit`, so raw build mode emits generated `.js` and fails with TS5096 on current `origin/main`. The package `typecheck` script is the no-emit check and passes.
- Cloudflare Access still maps to `user:cloudflare-access:<encoded-sub>` until we have a validated Anaconda-issued subject or an explicit Access claim mapping/backfill plan.


rgbkrk:
Closing as superseded by the merged credential/session work in #2889, #2891, and #2893. Rebasing this branch would reintroduce stale diffs against newer ACL/session metadata changes rather than add unique credential-transport coverage.

PR #2902, opened 2026-05-24
fix(daemon): accept runtime-agent PutBlob frames

### LINKED ISSUE #1814: Add PutBlob frame to runtime-agent ↔ daemon socket (#1334 dep)
Currently the runtime agent writes blobs through the daemon's separate blob channel connection. Add a PutBlob frame type on the existing runtime-agent sync socket so:

1. Only one persistent connection per agent (we already have it for sync + requests)
2. Remove the separate blob channel's handshake path
3. Simplifies kernel-launch blob-store permissions

Current frame types in `crates/notebook-doc/src/frame_types.rs`: 0x00 AutomergeSync, 0x01 Request, 0x02 Response, 0x03 Broadcast, 0x04 Presence, 0x05 RuntimeStateSync, 0x06 PoolStateSync. Add `0x07 PutBlob` or similar.

### Wire shape

```
| 4 bytes length | 1 byte type (0x07) | 32 bytes hash | rest is blob bytes |
```

Or make the payload JSON-prefixed so we can extend (hash, content_type, size) cleanly.

### Related

- #1334 — dependency for the agent-side blob upload path
- `contributing/protocol.md` for where this documents

rgbkrk:
## Summary

- Accept `NotebookFrameType::PutBlob` frames from runtime-agent peers.
- Reuse the existing PutBlob worker, backpressure, multipart state, blob-store write, and `BlobStored`/error response path.
- Add a runtime-agent peer regression test that sends a typed PutBlob frame over an in-memory socket and verifies the daemon stores the bytes.

Closes #1814.

## Verification

- `cargo test -p runtimed put_blob`
- `cargo xtask lint --fix`
- `cargo test -p runtimed`

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/peer_runtime_agent.rs
+++ b/crates/runtimed/src/notebook_sync_server/peer_runtime_agent.rs
@@ -7,6 +7,7 @@ use tracing::{debug, info, warn};
 
 use crate::async_outcome::{flatten_joined_result, JoinedResult};
 
+use super::blob_upload::{enqueue_put_blob, spawn_put_blob_worker, MultipartUploadState};
 use super::peer_runtime_sync::{persist_terminal_execution_records, runtime_file_save_fingerprint};
 use super::peer_writer::spawn_peer_writer;
 use super::{NotebookRoom, RuntimeAgentMessage, STATE_SYNC_COMPACT_THRESHOLD};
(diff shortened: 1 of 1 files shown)


rgbkrk:
Closing this for now while we settle the deployment/runtime-peer ADRs.

The patch is technically fine as runtime-agent frame parity, but the architecture
discussion moved away from treating the runtime-agent socket as the cross-machine
boundary. The clearer model is:

```text
local client / daemon
  -> room host / remote daemon as runtime_peer
  -> runtime agent local to that daemon
```

With that model, the next useful work is deciding blob availability/replication
and runtime-peer attachment semantics at the daemon/room-host layer. We can
reopen or recreate the small runtime-agent PutBlob patch later if it still helps
after the ADR lands.

PR #2905, opened 2026-05-24
refactor(pr-reviewer): run reviews through opencode

rgbkrk:
## Summary

- replace the Claude Agent SDK review runner with an `opencode run --format json` subprocess runner
- apply read-only reviewer permissions through opencode inline config, with a temp config file fallback, so project config cannot re-enable edit tools
- feed large review prompts over stdin, parse noisy JSONL event streams, preserve malformed model output as `infra_uncertain`, and bound subprocess runs with `--timeout-seconds`
- keep the existing `pr-review` CLI, disposable PR worktree flow, report schema, exit codes, and `--extra-prompt` support
- update repo skill/docs to describe the opencode-backed reviewer

## Verification

- `uv run pytest python/pr-reviewer/tests -q`
- `uv run ruff check python/pr-reviewer pyproject.toml`
- `uv run ruff format --check python/pr-reviewer`
- `uv run pr-review doctor --timeout-seconds 60`
- `cargo xtask lint --fix`
- `uv run pr-review https://github.com/nteract/nteract/pull/2905 --max-turns 120 --timeout-seconds 1200 --extra-prompt "..." --out .context/reviews/pr-2905-opencode-final.json`

## Notes

- `--max-turns` is retained as review-budget metadata for compatibility; `--timeout-seconds` is the hard opencode subprocess ceiling.
- The workspace-level `claude-agent-sdk` dependency remains for Gremlin, but `python/pr-reviewer` no longer depends on it.
- The final dogfood review found two issues after the previous patch set; this branch now addresses both with `asyncio.TimeoutError` handling and `OPENCODE_CONFIG_CONTENT` permission overrides.

### DIFF
--- a/.agents/skills/pr-reviewer/SKILL.md
+++ b/.agents/skills/pr-reviewer/SKILL.md
@@ -1,11 +1,11 @@
 ---
 name: pr-reviewer
-description: Use the repository-local pr-review CLI for Bedrock-authenticated Claude Agent SDK reviews of GitHub pull requests, branch diffs, or review loops. Use when a task asks for an external/Claude/Bedrock review, a second-model PR review, structured review findings, or validation of an implementation PR.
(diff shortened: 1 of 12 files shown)


rgbkrk:
@pullfrog please review lil fella

Issue #2616, opened 2026-05-08
Optimize output environment redaction matching


rgbkrk:
The runtime output environment redactor currently scans each text value once per eligible environment value and reallocates when a match is replaced.

This is acceptable for the initial secure-by-default implementation, but large outputs and large launch environments can make the hot IOPub path more expensive than necessary.

Follow-up:

- Evaluate a single-pass matcher, likely `aho-corasick`, for redacting all eligible values in one scan.
- Keep the longest-match/overlap behavior deterministic.
- Preserve the current binary-MIME skip behavior.
- Keep the global setting behavior unchanged.

Context: PR #2610 added the initial output redaction layer.

rgbkrk:
Done!

PR #2906, opened 2026-05-24
[redact-opt] Optimize daemon output redaction matching

rgbkrk:
## Summary

- Build one launch-time Aho-Corasick matcher for eligible environment values.
- Preserve existing longest-overlap behavior with leftmost-longest matching.
- Return borrowed text when no redaction match is found, avoiding no-op allocations.
- Keep binary MIME skipping, global setting behavior, and storage-boundary redaction intact.

## Verification

- `cargo test -p runtimed output_redaction -- --nocapture`
- `cargo test -p runtimed create_manifest_with_redactor -- --nocapture`
- `cargo test -p runtimed update_manifest_display_data_with_redactor -- --nocapture`
- `cargo xtask lint --fix`

## Notes

- This is the first `[redact-opt]` stacked PR for issue #2616.
- Runtime-created secrets from dotenv-style loaders remain a follow-up concern; this PR optimizes launch-time candidate matching without changing the threat model.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7406,6 +7406,7 @@ dependencies = [
 name = "runtimed"
 version = "2.4.9"
 dependencies = [
+ "aho-corasick",
  "alacritty_terminal",
  "anyhow",
  "async-rust-lsp",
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review

PR #2907, opened 2026-05-24
[redact-opt] Cache launcher redaction candidates

rgbkrk:
## Summary

- Cache Python launcher redaction candidates for a short refresh window.
- Refresh immediately when the redaction flag or environment size changes.
- Keep dotenv-style same-key value changes bounded by a 250ms refresh window instead of freezing candidates forever.
- Reset the cache in tests so each case preserves its previous environment assumptions.

## Measurement

Ad hoc 200-secret environment, 5000 calls:

- Before: `eligible_env_values`: 0.9184s total, 183.7 us/call
- Before: stream no-match redaction: 1.1634s total, 232.7 us/call
- Before: stream with-match redaction: 0.9874s total, 197.5 us/call
- After: cached `eligible_env_values`: 0.0031s total, 0.6 us/call
- After: cached stream no-match redaction: 0.2423s total, 48.5 us/call
- After: cached stream with-match redaction: 0.0607s total, 12.1 us/call

## Verification

- `uv run --directory python/nteract-kernel-launcher pytest tests/test_output_redaction.py tests/test_traceback.py -q`
- `cargo xtask lint --fix`

## Stack

- Depends on #2906.

### DIFF
--- a/python/nteract-kernel-launcher/nteract_kernel_launcher/_redact.py
+++ b/python/nteract-kernel-launcher/nteract_kernel_launcher/_redact.py
@@ -3,11 +3,13 @@
 from __future__ import annotations
 
 import os
+import time
 from typing import Any
 
 REDACT_ENV_VALUES_FLAG = "NTERACT_REDACT_ENV_VALUES_IN_OUTPUTS"
 REDACTION_MARKER = "[redacted env]"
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review

PR #2908, opened 2026-05-24
[redact-opt] Design runtime redaction refresh

rgbkrk:
## Summary

- Document a runtime redaction-candidate refresh path for dotenv/runtime-created secrets.
- Keep daemon storage-boundary redaction as the final authority.
- Define a daemon-local candidate update event that does not sync through notebook state or saved files.
- Scope Python as the first producer implementation while leaving a path for Deno or other runtimes.

## Verification

- `cargo xtask lint --fix`

## Stack

- Depends on #2907.


rgbkrk:
@pullfrog review

rgbkrk:
@pullfrog review

PR #2923, opened 2026-05-25
[ipy-opt] Type inbound SendComm messages

rgbkrk:
## Summary

- Adds a typed `CommRequestMessage` protocol envelope for inbound `send_comm` requests.
- Carries that typed envelope through notebook requests, runtime-agent requests, RPC routing, and kernel delivery.
- Keeps the existing JSON wire shape compatible by defaulting omitted legacy fields such as `parent_header`, `metadata`, `buffers`, and `channel`.

## Notes

This is the next `[ipy-opt]` layer after the Output widget replay measurement/cache PRs. It is intentionally the typed envelope precursor for the later blob/ContentRef-aware SendComm transport work; this PR does not claim to externalize inbound comm buffers yet.

The kernel send path still forces the Jupyter channel to Shell, matching the previous behavior.

## Verification

- `cargo test -p notebook-protocol`
- `cargo test -p runtimed sendcomm_fires_independently`
- `cargo xtask lint --fix`
- `cargo test -p runtimed`

PR #2926, opened 2026-05-25
[ipy-opt] Externalize inbound SendComm buffers

rgbkrk:
## Summary

- Add ordered `buffer_refs` to typed inbound `SendComm` messages so large widget buffers can travel through the existing `PUT_BLOB` path instead of JSON `number[][]`.
- Upload large frontend comm buffers as ephemeral blobs and send ordered refs while preserving the existing inline buffer path for small/custom messages.
- Resolve blob refs in the daemon coordinator before forwarding to the runtime agent, so kernel-facing Jupyter comm semantics still receive normal ordered buffers.

## Evidence

- `cargo test -p notebook-protocol comm_request_message_accepts_ordered_buffer_refs`
- `cargo test -p runtimed send_comm_buffer_refs`
- `cargo test -p runtimed sendcomm_fires_independently`
- `pnpm exec tsc -p packages/runtimed/tsconfig.json --noEmit`
- `cargo test -p notebook-protocol`
- `cargo test -p runtimed`
- `cargo test -p notebook-protocol typescript::tests::generated_typescript_bindings_are_current`
- `cargo xtask lint --fix`

## Stack

- Depends on #2923.

### DIFF
--- a/crates/notebook-protocol/src/protocol.rs
+++ b/crates/notebook-protocol/src/protocol.rs
@@ -433,6 +433,15 @@ pub struct NotebookResponseEnvelope {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct CommBufferRef {
+    pub index: usize,
+    pub blob: String,
+    pub size: usize,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub media_type: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct CommRequestMessage {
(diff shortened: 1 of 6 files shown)


rgbkrk:
@pullfrog please review this stacked PR, with extra attention to protocol compatibility, bounded memory behavior when resolving `buffer_refs`, and whether the daemon preserves Jupyter comm buffer ordering before forwarding to the runtime agent.

rgbkrk:
@pullfrog please re-review the updated head. The aggregate `buffer_refs` byte cap now runs before blob reads, and blob metadata size is checked before loading blob bytes.

PR #2935, opened 2026-05-25
fix(kernel-launcher): preserve polars object date columns

rgbkrk:
## Summary

- Normalize recoverable Polars `Object` columns containing Python `datetime.date` values to `pl.Date` before Arrow stream serialization.
- Add a regression for the `np.random.choice([... datetime.date ...])` path that otherwise exports as `fixed_size_binary[8]`.

## Verification

- `uv run --with pyarrow --with pandas --with polars --with narwhals --with numpy pytest python/nteract-kernel-launcher/tests/test_bootstrap.py python/nteract-kernel-launcher/tests/test_format.py -q`
- `cargo xtask lint --fix`

## Notes

- The stable MCP server reproduced the released behavior, but did not validate this local patch because it was not connected to the local dev daemon.

### DIFF
--- a/python/nteract-kernel-launcher/nteract_kernel_launcher/_format.py
+++ b/python/nteract-kernel-launcher/nteract_kernel_launcher/_format.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import datetime as _dt
 import hashlib
+import sys
 from collections.abc import Iterator
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog Do you see a way to do this conversion on a polars dataframe without having import polars on dataframes from other libraries (from within `_normalize_polars_object_dates`)? Surely there's a better way to do this.

PR #2964, opened 2026-05-25
[ipy-opt] Measure blob-backed output segments

rgbkrk:
## Summary

- extend the output commit measurement harness with RuntimeStateDoc save/sync amplification metrics
- add a blob-backed `output_segment` prototype strategy that stores ordered child manifests in blob storage and writes segment refs into RuntimeStateDoc
- document local N=100/500/1000 results and the Jupyter Server IOPub rate-limit context

## Evidence

Local `display_data` N=1000 sample with 256-byte payloads and segment size 128:

- flat ordered worker: 1000 durable entries, 60,705 saved-doc bytes, 434,021 daemon sync bytes, ~12.8s total
- blob segment model: 8 durable entries, 2,387 saved-doc bytes, 3,452 daemon sync bytes, 374,200 segment blob bytes, ~47ms total

This PR is measurement/prototype only. It does not change production output transport or client projection.

## Verification

- `cargo xtask lint --fix`
- `cargo test -p runtimed output_commit_measure --lib`
- `cargo check -p runtimed --example output_commit_measure`
- `NTERACT_OUTPUT_COMMIT_COUNTS=100,500,1000 NTERACT_OUTPUT_COMMIT_PAYLOAD_BYTES=256 NTERACT_OUTPUT_COMMIT_SEGMENT_SIZE=128 cargo run -q -p runtimed --example output_commit_measure`

### DIFF
--- a/crates/runtimed/examples/output_commit_measure.rs
+++ b/crates/runtimed/examples/output_commit_measure.rs
@@ -1,6 +1,6 @@
 use runtimed::output_commit_measure::{
-    measure_current_output_commit_loop, measure_ordered_worker_output_commit_model,
-    OutputCommitKind, OutputCommitMeasurementConfig,
+    measure_blob_segment_output_model, measure_current_output_commit_loop,
+    measure_ordered_worker_output_commit_model, OutputCommitKind, OutputCommitMeasurementConfig,
 };
 
 #[tokio::main]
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review please

rgbkrk:
@pullfrog isn't this dead code or silly

rgbkrk:
@pullfrog Ok, I'm closing this one then. If you see something new to go after that would help, please make a Draft PR.

PR #2966, opened 2026-05-25
[ipy-opt] Optimize display update output projection

rgbkrk:
## Summary

- documents the runtime output optimization plan and compatibility boundary
- adds `RuntimeStateDoc::get_output(execution_id, output_id)` for exact manifest reads
- uses the display index to read display-update targets directly instead of sorting/scanning every output in the execution

## Compatibility

This keeps RuntimeStateDoc output storage and every public output shape flat. It does not enable the blob-backed `output_segment` writer from the measurement PR.

## Verification

- `cargo test -p runtime-doc get_output_reads_exact_manifest_by_id`
- `cargo test -p runtimed output_prep::tests::test_update_display_id_updates_all_matching_outputs`
- `cargo xtask lint --fix`

### DIFF
--- a/crates/runtime-doc/src/doc.rs
+++ b/crates/runtime-doc/src/doc.rs
@@ -1882,6 +1882,12 @@ impl RuntimeStateDoc {
             .collect()
     }
 
+    /// Read one output manifest by execution id and output id.
+    pub fn get_output(&self, execution_id: &str, output_id: &str) -> Option<serde_json::Value> {
+        let output_entry = self.get_output_entry(execution_id, output_id)?;
+        self.read_output_manifest(&output_entry)
+    }
+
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog please review this stacked follow-up adversarially, especially the flat-output compatibility boundary and whether the runtime output optimization plan is accurate. This PR should not expose `output_segment` records or change client-visible output shapes.

rgbkrk:
@pullfrog please re-review after the rebase onto `main`. This is now intended as a directly mergeable nightly-bound PR, not a stack on the `output_segment` measurement prototype. Please focus on whether the PR depends on prototype code, changes any client-visible output shape, or affects `.ipynb` save/export behavior.

PR #2979, opened 2026-05-25
[ipy-opt] Update output diff snapshots in place

rgbkrk:
## Summary

- Add an in-place `output_id -> manifest` diff helper that preserves the existing flat output shape while avoiding full snapshot rebuilds on every WASM runtime-state sync frame.
- Switch `NotebookHandle` runtime-state sync to update its previous output snapshot in place.
- Extend the output commit measurement harness with equivalent rebuild-vs-in-place output-id diff timings.

## Compatibility

This keeps the existing client-facing `output_changeset` contract intact. Existing consumers still receive flat output ids, changed manifests, and removed output ids; no `output_segment` shape is exposed.

## Local verification

- `cargo test -p runtime-doc diff_output_ids -- --nocapture`
- `cargo test -p runtimed output_commit_measure -- --nocapture`
- `cargo test -p runtimed-wasm --lib --no-default-features`
- `cargo xtask wasm`
- `deno test --allow-read --allow-run --allow-env --no-check crates/runtimed-wasm/tests/`
- `cargo xtask lint --fix`
- `git diff --check`

## Measurement notes

Display-data output-id diff rows from the measurement harness:

- N=100: rebuild 472250 ns, in-place 233708 ns
- N=500: rebuild 2530166 ns, in-place 1137834 ns
- N=1000: rebuild 4421291 ns, in-place 2203375 ns

The first Deno snapshot run failed because the local fixture `.bin` files were Git LFS pointer text. After `git lfs pull --include='packages/runtimed/tests/fixtures/**'`, the same snapshot tests and full WASM Deno suite passed.

### DIFF
--- a/crates/runtime-doc/src/doc.rs
+++ b/crates/runtime-doc/src/doc.rs
@@ -3487,6 +3487,51 @@ pub fn diff_output_ids(
     )
 }
 
+/// Diff execution outputs by `output_id`, updating the caller-owned snapshot.
+///
+/// This is the hot-path variant used by WASM runtime-state sync. It still
+/// walks the current flat outputs, but it only clones manifests that are new
+/// or changed. Unchanged entries remain in `prev` in place, avoiding a full
+/// `output_id -> manifest` snapshot rebuild on every sync frame.
+pub fn diff_output_ids_in_place(
+    prev: &mut HashMap<String, serde_json::Value>,
+    current_executions: &HashMap<String, ExecutionState>,
+) -> OutputIdDiff {
+    let mut seen: HashSet<String> = HashSet::new();
+    let mut changed: Vec<(String, serde_json::Value)> = Vec::new();
+
+    for exec in current_executions.values() {
+        for output in &exec.outputs {
+            let Some(oid) = extract_output_id(output) else {
+                continue;
+            };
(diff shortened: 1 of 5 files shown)


rgbkrk:
@pullfrog review

Please re-review after the stack changed. This PR now stacks on #2980, which removes the dormant optimistic frontend display update overlay. The remaining diff is the WASM/runtime-doc in-place output-id snapshot update. Please focus on compatibility: clients must still see the same flat output changeset shape, and unchanged manifests should simply avoid being cloned into a fresh previous-output snapshot each sync frame.

rgbkrk:
@pullfrog review

Please review the current head after a metadata-only amend to retrigger CI. Same scope as before: this stacks on #2980 and only changes the WASM/runtime-doc output-id snapshot diff to update in place while preserving the flat output changeset contract.

rgbkrk:
@pullfrog review

Please re-review after rebasing onto current main. #2980 is now merged into main, and this PR should contain only the in-place output-id snapshot diff. Please keep the same compatibility lens: existing clients must continue to receive flat output changesets; this should only avoid rebuilding/cloning the previous output-id snapshot on each runtime-state sync frame.

PR #2977, opened 2026-05-25
[ipy-opt] Index frontend display updates

rgbkrk:
## Summary

- add a private `display_id -> output_id` index to the notebook output store
- keep the index synchronized on output set, delete, batch delete, and reset
- document the frontend display-update index as the next safe flat-path optimization

## Compatibility

This keeps the existing flat `JupyterOutput` store contract. It does not add segment manifests, change output object shape, or change `.ipynb` persistence/export behavior.

## Verification

- `pnpm test:run apps/notebook/src/lib/__tests__/notebook-outputs.test.ts`
- `pnpm test:run apps/notebook/src/lib/__tests__/notebook-outputs.test.ts apps/notebook/src/lib/__tests__/project-runtime-stores.test.ts apps/notebook/src/lib/__tests__/frame-pipeline.test.ts`
- `pnpm --dir apps/notebook build`
- `cargo xtask lint --fix`
- `git diff --check`

## Notes

This follows merged #2973. It should stay draft until adversarial review, Pullfrog review, GitHub CI, and any targeted display-update stress check are complete.

PR #2980, opened 2026-05-25
[ipy-opt] Remove dormant display update overlay

rgbkrk:
## Summary

- Remove the unused optimistic `update_display_data` overlay returned from `useAutomergeNotebook`.
- Remove the matching unused `updateOutputsByDisplayId` output-store helper.
- Update the runtime-output optimization note: this path should be deleted, not indexed, because RuntimeStateDoc changesets already carry display updates to the output store.

## Why

Pullfrog's follow-up review on #2977 found the display-id index optimized a dormant path. The callback was returned from the hook, but no component or integration consumed it. The active display update path is the RuntimeStateDoc changeset projection into the output store.

## Verification

- `pnpm test:run apps/notebook/src/lib/__tests__/notebook-outputs.test.ts`
- `cargo xtask lint --fix`
- `git diff --check`

### DIFF
--- a/apps/notebook/src/hooks/useAutomergeNotebook.ts
+++ b/apps/notebook/src/hooks/useAutomergeNotebook.ts
@@ -21,14 +21,12 @@ import {
   replaceNotebookCells,
   resetNotebookCells,
   updateCellById,
-  updateNotebookCells,
   useCellIds,
 } from "../lib/notebook-cells";
 import {
   applyExecutionViewChangeset,
   resetRuntimeStoresProjection,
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review

Please review this as a deletion-focused optimization. PR #2977 tried to index the optimistic frontend display update overlay, but the follow-up review correctly found that path is dormant: RuntimeStateDoc changesets already deliver display updates to the output store, and no consumer calls the hook's optimistic callback. Please focus on whether this removes only the unused overlay/helper and preserves the active changeset-driven output projection.

PR #2984, opened 2026-05-25
[ipy-opt] Split structural output store updates

rgbkrk:
## Summary

- Split the notebook output store into a payload-level version and a structural version.
- Move `NotebookView` hidden-output grouping and error-count recomputation onto the structural signal.
- Keep tail pinning subscribed to every output payload update so streaming behavior remains unchanged.

## Compatibility

This does not change the durable output model, protocol surface, persisted data, or flat client output shape. It only narrows one frontend subscription path so display payload updates and stream append bursts do not force hidden-group scans when output membership and output types are unchanged.

## Verification

- `pnpm test:run apps/notebook/src/lib/__tests__/notebook-outputs.test.ts apps/notebook/src/lib/__tests__/cell-outputs-decoupling.test.tsx apps/notebook/src/lib/__tests__/frame-pipeline.test.ts`
- `cargo xtask lint --fix`
- `git diff --check origin/main`

### DIFF
--- a/apps/notebook/src/components/NotebookView.tsx
+++ b/apps/notebook/src/components/NotebookView.tsx
@@ -27,7 +27,11 @@ import { EditorRegistryProvider, useEditorRegistry } from "../hooks/useEditorReg
 import { useFocusedCellId, useSearchCurrentMatch } from "../lib/cell-ui-state";
 import { logger } from "../lib/logger";
 import { getNotebookCellsSnapshot, useCell, useMaterializeVersion } from "../lib/notebook-cells";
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

Please review this frontend output-store optimization with a compatibility and correctness lens:

- existing clients still observe flat outputs; no durable model, persisted data, or protocol shape changes
- `NotebookView` hidden-group/error-count recomputation moves from every output payload update to output add/remove/type changes
- tail pinning still observes every payload update through `useOutputsVersion`

PR #2986, opened 2026-05-25
fix(sift): avoid formatting large binary cells

rgbkrk:
## Summary

- Detect HuggingFace-style `Struct{bytes,path}` and `List<Struct{bytes,path}>` Arrow shapes as image columns even when parquet metadata is missing.
- Avoid materializing large binary payloads as display text by compacting them before viewport and value-count formatting.
- Preserve the image raw-byte path in the WASM table adapter without letting formatted cell text drive image cells.

## Testing

- `cargo xtask lint --fix`
- `cargo test -p sift-wasm --lib`
- `pnpm --dir packages/sift exec vp test run src/wasm-table-data.test.ts`


rgbkrk:
@pullfrog think on what a good follow up here would be, assuming this is G2G.

PR #2988, opened 2026-05-25
perf(frontend): reduce streaming table rerenders

rgbkrk:
## Summary

- keep Sift pooled rows attached to their data row so appended batches do not recreate visible image cells
- avoid notebook-wide React commits for output-version tail pinning by subscribing imperatively
- stop output-only full cell materializations from invalidating `useMaterializeVersion`
- update the isolated renderer CSS bundle pointer from the local dev build

## Verification

- `pnpm --dir packages/sift exec vp test run src/table.test.ts`
- `pnpm exec vp test run apps/notebook/src/lib/__tests__/notebook-cells.test.ts apps/notebook/src/lib/__tests__/cell-outputs-decoupling.test.tsx apps/notebook/src/lib/__tests__/notebook-outputs.test.ts`
- `cargo xtask lint --fix`


rgbkrk:
@pullfrog please review this

PR #3009, opened 2026-05-26
fix(runtimed): restore synced pool targets on startup

rgbkrk:
## Summary

- restore daemon pool targets from synced settings before scanning warmed environments on startup
- keep config pool size `0` as the test/disable sentinel, while using the same effective target helper in warming loops
- add a regression test that restores all warmed UV envs when config starts at `1` but synced settings request `3`

## Context

Nightly startup could mark warmed pool environments as orphaned before the settings-backed target size was applied. That forced the daemon to rebuild environments during app startup, which could delay notebook load even though parsing itself was fast.

## Testing

- `cargo xtask lint --fix`
- `cargo test -p runtimed`

## Review

- Draft while external `pr-review` and CI are pending.

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -492,6 +492,14 @@ fn pixi_prewarmed_packages(extra: &[String], install_default_data_packages: bool
     packages
 }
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
@pullfrog review

PR #3024, opened 2026-05-27
fix(runtimed): preserve launch-time queued executions

rgbkrk:
## Summary

- Fix a runtime-agent race where a CRDT-visible queued execution could be marked as seen before a kernel connection existed.
- Re-scan queued RuntimeStateDoc executions immediately after LaunchKernel succeeds so work queued during environment prep is handed to KernelState without waiting for another sync frame.
- Harden RuntimeStateDoc sync writes so owner/editor-scoped peers cannot forge executions, queue state, display routing, environment/trust/kernel state, or comm topology through raw RuntimeStateSync.
- Keep local widget support by allowing owner/editor peers to update existing comm state properties only; hosted cloud RuntimeStateSync is now runtime-peer-only until hosted widget/comms write semantics are explicit.
- Add regression coverage for launch-time queued execution delivery, local RuntimeStateDoc write policy, and hosted owner-scoped execution forgery.

## Why

The failing `runtimed-py Integration Tests` job accepted ExecuteCell during auto-launch, but the daemon logs never showed `[kernel-state] Queuing execution` or a Jupyter `execute_request`. That means the health check was waiting on a terminal execution state for work that had been published to RuntimeStateDoc but never delivered to the kernel.

The coordinator intentionally writes queued executions early so all peers can see queueing while the environment/kernel are still preparing. The runtime agent should only suppress an execution after it actually queues it into KernelState.

The same causal boundary matters for security: execution intent should enter through the gated `ExecuteCell` / `RunAllCells` request path, not through direct writes to the writable RuntimeStateDoc channel used for widget comm state.

## Verification

- `cargo test -p runtimed queued_execution_seen_before_kernel_launch_is_not_dropped -- --nocapture`
- `cargo test -p runtimed notebook_sync_server::peer_runtime_sync::tests:: -- --nocapture`
- `cargo test -p runtimed runtime_agent::tests:: -- --nocapture`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/room-materializer.test.ts`
- `cargo xtask lint --fix`
- `cargo clippy --workspace --exclude notebook --exclude runtimed-wasm --exclude sift-wasm --exclude runtimed-py --all-targets -- -D warnings`
- `RUNTIMED_INTEGRATION_TEST=1 RUNTIMED_BINARY=/Users/kyle/.codex/worktrees/cb82/nteract/target/debug/runtimed RUNTIMED_LOG_LEVEL=debug RUNTIMED_TEST_UV_POOL_SIZE=1 RUNTIMED_TEST_CONDA_POOL_SIZE=0 RUNTIMED_PYTEST_WORKERS=1 uv run pytest tests/test_daemon_integration.py::TestBasicConnectivity::test_session_connect -q -s`

### DIFF
--- a/apps/notebook-cloud/src/room-materializer.ts
+++ b/apps/notebook-cloud/src/room-materializer.ts
@@ -252,8 +252,9 @@ export function typedFrameFromRoomHostOutbound(frame: RoomHostOutboundFrame): Ui
 }
 
 function allowsHostedRuntimeStateWrite(identity: AuthenticatedConnection): boolean {
-  // Hosted editors are markdown-only until RuntimeStateDoc path validation lands.
-  return identity.scope === "owner" || identity.scope === "runtime_peer";
+  // Hosted editors/owners are markdown-only: executable RuntimeStateDoc
+  // mutations come from explicit runtime peers, not the editing channel.
+  return identity.scope === "runtime_peer";
 }
 
 function normalizeResult(value: unknown): RoomHostFrameResult {
--- a/apps/notebook-cloud/test/room-materializer.test.ts
+++ b/apps/notebook-cloud/test/room-materializer.test.ts
@@ -475,6 +475,36 @@ describe("RoomMaterializer", () => {
       /cannot write RuntimeStateDoc changes/,
     );
   });
+
+  it("rejects owner-scoped RuntimeStateDoc execution changes", async () => {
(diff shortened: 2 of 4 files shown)


rgbkrk:
@pullfrog review this for user workflows, consistency, and performance

PR #3130, opened 2026-05-30
fix(runtimed): don't zero a notebook on disk when streaming load fails [parked behind genesis fix]

rgbkrk:
**Status: parked draft.** The real fix is the genesis-seed divergence (separate
PR) — once the frontend wasm and daemon share a genesis seed, streaming loads
stop failing and this whole failed-load class becomes near-unreachable. This PR
is the safety net for that class and is intentionally scoped to "a failed open
must never cost the file."

## What it does

A failed streaming load empties the room doc as an in-memory rollback
(`clear_all_cells` in `peer_session.rs`). Kernel-teardown / autosave then wrote
that empty doc back over the source `.ipynb`, so a failed *open* deleted the
notebook on disk. Reproduced: a 15-cell notebook opened to 0 cells (223 bytes).

The room records when its most recent streaming load failed. The persistence
path refuses to write an empty room over a file that still has bytes on disk
(parseable or not), and the marker is cleared on a fresh load, on a non-empty
save, or on a non-empty external reload.

## Solid and covered

| Path | Behavior |
|---|---|
| autosave / teardown over a non-empty file (failed load) | skipped, file preserved |
| explicit Save As over a non-empty file (failed load) | error, file preserved |
| empty room, load did **not** fail (user cleared cells) | normal write |
| populated room | normal write |

Tests: `failed_load_does_not_zero_existing_notebook`,
`failed_load_does_not_overwrite_malformed_notebook`,
`cleared_failed_load_lets_empty_save_through`,
`explicit_save_after_failed_load_errors_and_preserves_file`. CI green on the core.

## Known limitations (deferred — documented in code)

Both keep the file intact; neither is data loss. Deferred because the root-cause
PR removes the failed-load trigger, making these near-unreachable:

- **Explicit in-place Save reports phantom success.** `NotebookRequest::SaveNotebook`
  passes `path: None`, the same as autosave, so an in-place save of a
  failed-load-empty room is silently skipped and reported as saved. Surfacing it
  needs a save-origin signal threaded from the request handler (the `SaveOrigin`
  refactor was started and set aside as out of proportion to the payoff).
- **Valid empty external reload leaves the marker set.** A `cells: []` file fixed
  externally keeps `load_failed` set until a cell is added or it is reopened.

## Why parked

The flag lifecycle keeps generating edge cases (codex found four in sequence,
each narrower). The correct sequencing is to land the root-cause fix first
(genesis seed parity) so loads stop failing, then decide whether the residual
explicit-save/empty-reload edges here are even worth the added machinery.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/peer_session.rs
+++ b/crates/runtimed/src/notebook_sync_server/peer_session.rs
@@ -215,6 +215,11 @@ where
                     );
                 }
             }
+            // The rollback above emptied the room doc. Mark the load failed so
+            // the persistence layer refuses to autosave this empty doc over the
+            // notebook's on-disk content — a failed open must not delete the
+            // file. Cleared when a later connection re-attempts the load.
+            room.mark_load_failed();
             room.finish_loading();
             let _ = room.broadcasts.changed_tx.send(());
             warn!(
--- a/crates/runtimed/src/notebook_sync_server/persist.rs
+++ b/crates/runtimed/src/notebook_sync_server/persist.rs
@@ -101,6 +101,53 @@ pub(crate) async fn save_notebook_to_disk(
         (cells, metadata_snapshot, eids)
     };
 
+    // A failed streaming load empties the room doc as an in-memory rollback
+    // (see the failure branch in peer_session.rs). Never persist that empty doc
+    // over an existing notebook file: a failed *open* must not become a file
+    // deletion. This holds even when the file on disk is malformed JSON or has a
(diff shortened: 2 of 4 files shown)


rgbkrk:
Superseded by #3179 (merged), which lands the data-loss guard with the converged load_failed set-on-hazard design — cleared at all three recovery completions, canonicalized in-place detection, and missing-cells treated as a load error. Closing this parked earlier attempt.

PR #3196, opened 2026-05-30
ci: default pullfrog reviews to opus 4.8

rgbkrk:
## Summary

- update the Pullfrog workflow default Bedrock model ID to Claude Opus 4.8
- keep the workflow on the US geo inference ID form already used by Pullfrog

## Verification

- `AWS_REGION=us-east-1 opencode run --model amazon-bedrock/us.anthropic.claude-opus-4-8 --format json --title "bedrock-opus-48-id-smoke" "Reply with exactly: OK"`
- `cargo xtask lint --fix`


rgbkrk:
@pullfrog hows it look

PR #3226, opened 2026-05-31
refactor(notebook): share hosted frame layout

rgbkrk:
## Summary

- add a shared `NotebookDocumentShell` header/body frame for hosted notebook surfaces
- move notebook-cloud controls into that shared header frame, above the rail + document stage body
- keep cloud surface tests checking that hosted layout goes through the shared shell instead of a cloud-only toolbar lane

## Stack

- Depends on #3225 (`quod/notebook-shared-editable-view`)
- Merge order: #3223, #3224, #3225, then this PR

## Validation

- `pnpm exec vp test run src/components/notebook-shell/__tests__/NotebookDocumentShell.test.tsx`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/viewer-shared-cell-surface.test.ts`
- `pnpm --workspace-root exec tsc -p apps/notebook/tsconfig.json --noEmit`
- `pnpm --dir apps/notebook-cloud typecheck`
- `cargo xtask lint --fix`

PR #3225, opened 2026-05-31
refactor(notebook): share editable view frame

rgbkrk:
## Summary

- add `NotebookEditableView` as the shared shell frame for host-provided editable cell rendering
- move cloud's markdown/code/fallback cell-type routing into the shared notebook-shell layer
- leave cloud-specific renderer callbacks, live-sync writes, presence projection, and hosted output policy in the cloud adapter

## Stack order

1. #3215
2. #3220
3. #3221
4. #3222
5. #3223
6. #3224
7. This PR

Merge bottom-up. This PR is intentionally stacked on #3224 and should be retargeted by GitHub as lower PRs merge.

## Validation

- `pnpm exec vp test run src/components/notebook-shell/__tests__/NotebookEditableView.test.tsx src/components/notebook-shell/__tests__/NotebookCellList.test.tsx`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/viewer-shared-cell-surface.test.ts`
- `pnpm --workspace-root exec tsc -p apps/notebook/tsconfig.json --noEmit`
- `pnpm --dir apps/notebook-cloud typecheck`
- `cargo xtask lint --fix`

PR #3224, opened 2026-05-31
refactor(notebook): share cell presence dots

rgbkrk:
## Summary

- extract the cell presence dot renderer into shared `src/components/cell`
- keep desktop cursor-registry subscription logic in the desktop host wrapper while delegating rendering to the shared dots
- replace cloud's duplicated markdown/code presence dot renderers with one cloud adapter over the shared dots

## Stack order

1. #3215
2. #3220
3. #3221
4. #3222
5. #3223
6. This PR

Merge bottom-up. This PR is intentionally stacked on #3223 and should be retargeted by GitHub as lower PRs merge.

## Validation

- `pnpm exec vp test run src/components/cell/__tests__/CellPresenceDots.test.tsx`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/viewer-shared-cell-surface.test.ts`
- `pnpm --workspace-root exec tsc -p apps/notebook/tsconfig.json --noEmit`
- `pnpm --dir apps/notebook-cloud typecheck`
- `cargo xtask lint --fix`

PR #3223, opened 2026-05-31
refactor(notebook): share read-only cell list frame

rgbkrk:
## Summary

- loosen `NotebookCellList` to a minimal shared cell-list item contract
- route `ReadOnlyNotebook` through `NotebookCellList` instead of maintaining its own duplicated section/error-boundary frame
- move read-only cell list data contracts into `notebook-shell` so shared shell projections no longer import the read-only cell wrapper
- preserve existing read-only cell props, reset keys, empty state, and custom error rendering

## Stack order

1. #3215
2. #3220
3. #3221
4. #3222
5. This PR

Merge bottom-up. This PR is intentionally stacked on #3222 and should be retargeted to `main` by GitHub as lower PRs merge.

## Validation

- `pnpm exec vp test run src/components/cell/__tests__/ReadOnlyNotebook.test.tsx src/components/notebook-shell/__tests__/NotebookCellList.test.tsx src/components/notebook-shell/__tests__/NotebookReadOnlyView.test.tsx src/components/notebook-shell/__tests__/view-model.test.ts`
- `pnpm --workspace-root exec tsc -p apps/notebook/tsconfig.json --noEmit`
- `pnpm --dir apps/notebook-cloud typecheck`
- `cargo xtask lint --fix`

PR #3227, opened 2026-05-31
test(notebook-cloud): smoke shared frame layout

rgbkrk:
## Summary

- extend the hosted notebook-cloud smoke with a shared frame layout check
- assert the shared header, body, rail, stage, and notebook scroll container exist
- verify scrolling notebook cells does not move the host header or left rail

## Stack

- Depends on #3226 (`quod/notebook-shell-host-frame`)
- Merge order: #3223, #3224, #3225, #3226, then this PR

## Validation

- `pnpm --dir apps/notebook-cloud smoke:hosted`
- `cargo xtask lint --fix`


rgbkrk:
Closing as superseded by #3232, which takes the reset path: notebook-cloud now renders through the shared desktop NotebookView surface instead of continuing this incremental cloud/shared-component stack. Follow-up work should build from main with the convergence review lens.

PR #3292, opened 2026-06-01
feat(notebook): show shared toolbar identity

rgbkrk:
## Summary

- show desktop notebook actor identity through the shared `NotebookToolbarIdentity` surface
- make `NotebookToolbar` consume the full `NotebookShellCapabilities` projection instead of a command-only subset
- cover the desktop toolbar identity path with a focused component test

## Verification

- `pnpm test:run apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx`
- `pnpm --dir apps/notebook exec tsc -b`
- `cargo xtask lint --fix`

## Notes

This keeps the work in the shared notebook toolbar path: desktop now renders the same identity primitive cloud uses, while host-specific actor facts still come from each host capability adapter.

### DIFF
--- a/apps/notebook/src/components/NotebookToolbar.tsx
+++ b/apps/notebook/src/components/NotebookToolbar.tsx
@@ -3,11 +3,12 @@ import { useCallback, useEffect, useState, type ReactElement, type ReactNode } f
 import { HoverCard, HoverCardContent, HoverCardTrigger } from "@/components/ui/hover-card";
 import {
   NotebookCommandToolbar,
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing this after rebasing against current main. The remaining diff would reintroduce desktop toolbar identity UI that Kyle intentionally removed until desktop can connect to cloud/remote identity in a more complete way. We should revisit from the shared projection/adapter contract side rather than landing this visual surface now.

PR #3316, opened 2026-06-01
feat(notebook-cloud): grant editors full cell editing in hosted rooms

rgbkrk:
Editors in a hosted room can now add, delete, reorder, and edit cells of any type, not just markdown source. Creating cells with a collaborator is the baseline for an editable notebook, so the editor write surface is no longer markdown-only. Notebook-level metadata (kernelspec, trust, environment, path, project) stays owner-authored.

## Design decision

The boundary that matters is server-side, not the capability flag. The room host runs a semantic diff validator (`validate_editor_notebook_changes` in `runtimed-wasm`, reached from `receive_notebook_sync` whenever `can_write_all_notebook_changes` is false): it clone-previews the incoming `NotebookDoc` sync, diffs the changed cells/fields, and rejects any change that touches notebook-level metadata. Owners skip the validator. The cloud capability (`canEditCells`/`canEditStructure` for editor scope) only mirrors what the room already enforces. Flipping the capability alone would have let the client show add-cell while the room rejected the frames, reintroducing the focus/desync class of bug. Both moved together, server-first.

This also converges the two hosts: desktop already granted editors full cell editing. Cloud now derives the same rule instead of an owner-only special case.

Execution stays a separate axis. There is no kernel provider in the hosted prototype, so run/restart/interrupt remain hidden; that gate is kernel availability, not an ACL scope, and generalizing it into a host-neutral runtime-availability capability is a follow-up.

## Behavioral coverage

| Scope | Cell source (md/code/raw) | Add/delete/move | Notebook metadata | Execute |
|-------|---------------------------|-----------------|-------------------|---------|
| viewer | no | no | no | no |
| editor | yes | yes | no (owner-only) | no (no kernel) |
| owner | yes | yes | yes | no (no kernel) |

## Notes

- `validate_markdown_source_only_changes` is renamed to `validate_editor_notebook_changes` and now returns `Result<(), String>` so it is unit-testable on native targets (the caller maps to `JsError`).
- Decision 7 of `docs/architecture/hosted-room-authorization.md` is rewritten to record the new editor surface, with "lock it down later" captured as a future owner capability rather than the baseline.
- Follow-up, not in this PR: a host-neutral runtime-availability capability so run controls hide when no kernel provider is attached (desktop benefits too), and the larger `useAutomergeNotebook` controller extraction tracked in `frontend-sync-bridge.md`.

Predecessor: #3314.

## Test plan

- [x] `cargo test -p runtimed-wasm` (17 passing, incl. 3 new editor-policy tests)
- [x] `apps/notebook-cloud` suite (430 passing; room-materializer integration tests prove editor add/edit-code accepted and editor metadata rejected against rebuilt wasm)
- [ ] Manually click "add cell" as an editor in the local cloud viewer and confirm round-trip

### DIFF
--- a/apps/notebook-cloud/test/room-materializer.test.ts
+++ b/apps/notebook-cloud/test/room-materializer.test.ts
@@ -185,7 +185,29 @@ describe("RoomHostHandle", () => {
           false,
           encodeTypedFrame(FrameType.AUTOMERGE_SYNC, message),
         ),
-      /cannot change notebook metadata/,
+      /editor scope may only edit cells/,
+    );
+  });
+
+  it("rejects editor-scoped runtime_state_doc_id repointing", async () => {
+    const host = await createEmptyRoomHost("demo", "system/schema:notebook-cloud-room");
+    const editor = NotebookHandle.create_bootstrap("user:dev:bob/desktop:editor");
+
+    syncHostWithClient(host, "peer-editor", "user:dev:bob", true, false, editor);
+    editor.set_runtime_state_doc_id("forged-runtime-doc");
+    const message = editor.flush_local_changes();
+    assert.ok(message);
+
+    assert.throws(
+      () =>
+        host.receive_peer_frame(
+          "peer-editor",
+          "user:dev:bob",
+          "editor",
+          false,
+          encodeTypedFrame(FrameType.AUTOMERGE_SYNC, message),
+        ),
+      /editor scope may only edit cells/,
     );
   });
 
--- a/apps/notebook-cloud/test/room-materializer.test.ts
+++ b/apps/notebook-cloud/test/room-materializer.test.ts
@@ -130,10 +130,11 @@ describe("RoomHostHandle", () => {
     );
   });
 
-  it("rejects editor-scoped source edits to code cells", async () => {
+  it("allows editor-scoped source edits to code cells", async () => {
     const host = await createEmptyRoomHost("demo", "system/schema:notebook-cloud-room");
     const owner = NotebookHandle.create_bootstrap("user:dev:alice/desktop:owner");
     const editor = NotebookHandle.create_bootstrap("user:dev:bob/desktop:editor");
+    const viewer = NotebookHandle.create_bootstrap("user:dev:carol/desktop:viewer");
 
     syncHostWithClient(host, "peer-owner", "user:dev:alice", true, true, owner);
     owner.add_cell(0, "code-cell", "code");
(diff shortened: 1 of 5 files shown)


rgbkrk:
@pullfrog deep-dive review requested. This PR widens the hosted-room **editor** write surface from markdown-only to full cell editing, enforced server-side in `validate_editor_notebook_changes` (`crates/runtimed-wasm/src/lib.rs`, renamed from `validate_markdown_source_only_changes`). It now rejects only notebook-level metadata changes (`changeset.metadata_changed`) for non-owners; structural + any cell-source edits are allowed. Owners skip the validator via `can_write_all_notebook_changes`. The cloud capability (`apps/notebook-cloud/viewer/shell-capabilities.ts`) was flipped to mirror this.

Please look hard at the authorization boundary, with full codebase context (not just this diff):

1. **Metadata carve-out completeness.** `metadata_changed` is the only editor reject. Are trust / environment / path / project all under NotebookDoc top-level metadata (so covered), or could any of them live in a cell field or a place an editor can now write? Trace where each is stored.
2. **Runtime-owned fields.** Can an editor now author NotebookDoc cell `outputs` / `execution_count`? Is that a fabricated-output or trust-bypass risk given RuntimeStateDoc is the live execution authority, or is it harmless persisted-nbformat state?
3. **Rejection paths intact.** Confirm only the `can_write_all_notebook_changes == false` (editor) branch loosened, and that viewer (`!can_write`) and runtime_peer NotebookDoc rejection are unaffected.
4. **Canonical schema seed.** The validator folds seed-change hashes into `heads_before`. Any way an editor can smuggle a metadata edit past that folding?

Note: stacked PR #3317 sits on top of this branch; please scope to this branch's diff. Thanks.

rgbkrk:
@pullfrog updated head (`c970ddfa`). The editor write surface is now restricted server-side to the `cells` subtree via an allowlist in `validate_editor_notebook_changes` (`crates/runtimed-wasm/src/lib.rs`): every diff patch must begin with the `cells` map key, and any other root write is rejected, including `metadata`, `schema_version`, `notebook_id`, and `runtime_state_doc_id`. This closes the earlier gap where only the `metadata` flag was checked.

Please re-review this head with that in mind:
1. Is the `patch.path.first() == Map("cells")` allowlist a complete and correct boundary? Any cell op whose patch path does NOT begin with `cells` (so it would be wrongly rejected), or any non-cell write whose path DOES begin with `cells` (wrongly allowed)?
2. The genesis seed is folded into `heads_before` so a fresh peer's bootstrap is not read as a write. Any bypass there?
3. Editors can still write cell-level `execution_count` / `execution_id` (they live under `cells/{id}`). Given live execution state is RuntimeStateDoc (separately gated), is treating those as benign cell edits correct, or should they be carved out too?

PR #3321, opened 2026-06-02
fix(notebook): stop frontend first-cell seeding

rgbkrk:
## Summary

- Remove NotebookView's implicit first-cell auto-seeding effect.
- Keep empty frontend projections as render-only states; hosts/daemons own initial document seeding.
- Preserve explicit empty-notebook add buttons for users with cell mutation capabilities.

## Why

Hosted cloud joins can briefly have an empty React cell store while the live Automerge handle already contains cells. The previous shared NotebookView effect interpreted that UI projection as an empty notebook and called `onAddCell("code")`. The cloud controller can accept that mutation once the CRDT handle has cells, producing an extra cell during room join.

The frontend should not infer document genesis from projected emptiness. New-notebook seeding belongs to the authoritative host/daemon path, while the frontend renders loading, empty, or synced states without authoring cells implicitly.

## Verification

- `pnpm --workspace-root exec vp test run apps/notebook/src/components/__tests__/notebook-view-capabilities-source.test.ts`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/viewer-shared-cell-surface.test.ts`
- `pnpm --dir apps/notebook exec tsc -b`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/test/viewer-shared-cell-surface.test.ts
+++ b/apps/notebook-cloud/test/viewer-shared-cell-surface.test.ts
@@ -15,6 +15,7 @@ test("cloud notebook body renders through the desktop NotebookView surface", () 
   assert.doesNotMatch(sourceText, /<CloudLiveNotebook/);
   assert.doesNotMatch(sourceText, /NotebookReadOnlyView/);
   assert.doesNotMatch(sourceText, /<NotebookReadOnlyView/);
+  assert.doesNotMatch(sourceText, /autoSeedFirstCell/);
 });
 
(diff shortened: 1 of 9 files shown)


rgbkrk:
I'm not sure this is the right approach. Backend should seed it (from the daemon) in the new case. I think I'd rather we just open empty on the frontend and get new cells.

rgbkrk:
Closing this draft because the frontend/test adaptation is not the desired architecture. We want to revisit the new-notebook initialization policy from the Rust/daemon side instead of normalizing empty synced notebooks in the frontend.

PR #3327, opened 2026-06-02
feat(elements): expose notebook host shell scenarios

rgbkrk:
## Summary

This branch keeps the shared NotebookView host-convergence work moving through Elements instead of adding more cloud-only UI.

- adds source-fact and host-boundary metadata to `ElementsNotebookScenario`
- renders those source facts and authority boundaries in the Notebook shell capabilities catalog
- adds a docs-local `ElementsNotebookEnvironment` provider for shell pages that need capabilities, rail state, document facts, outputs, runtime/package projections, projected host notices, and inert host callbacks
- projects host notice policy from scenario facts and renders it through the shared `NotebookNotice` primitive
- routes the cloud notebook shell catalog example through that fixture host instead of owning scenario/rail state locally

## Why

The current ADR/PRD direction says desktop, cloud, and Elements should share the notebook app surface while hosts own transport, auth, ACLs, filesystem permissions, runtime authority, notice actions, diagnostics, and storage/blob resolution. This makes Elements a stronger pressure test for that boundary: catalog pages can now show which facts come from the host adapter and which shared components consume them.

## Verification

- `cargo xtask lint --fix`
- local visual smoke with Elements dev server for:
  - `http://localhost:5176/docs/notebook-shell-capabilities`
  - `http://localhost:5176/docs/cloud-notebook-shell`

### DIFF
--- a/apps/elements/components/cloud-notebook-shell-example.tsx
+++ b/apps/elements/components/cloud-notebook-shell-example.tsx
@@ -41,8 +41,11 @@ import {
   type NotebookShellCapabilities,
 } from "@/components/notebook";
 import { EnvironmentSummary } from "@/components/environment";
-import type { NotebookRailPanelId } from "@/components/notebook-rail";
 import { cn } from "@/lib/utils";
+import {
+  ElementsNotebookEnvironment,
+  useElementsNotebookEnvironment,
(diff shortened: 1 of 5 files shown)


rgbkrk:
@pullfrog review please

PR #3339, opened 2026-06-03
docs(adr): propose CommsDoc split for widget comm state

rgbkrk:
## What this is

A proposed ADR (0002, in the new `docs/adr/` home from #3336) for extracting widget **comm state** out of RuntimeStateDoc into its own document, CommsDoc. Draft, because the decision is up for review before any implementation. No production code here.

This is the first decoupled piece of the host-convergence work (the handoff's "thread 1"). It builds directly on ADR 0001 (#3335) and shares its discipline: reason about the document model from invariants, not incidental co-location.

## The problem

RuntimeStateDoc is daemon-authoritative except for one carve-out: editors must write `comms/{comm_id}/state/{key}` so a slider pushes its value back to the kernel. That carve-out is enforced by a before/after policy-snapshot diff (`crates/runtimed-wasm/src/lib.rs:663-697`), not a document boundary. It is exactly the field-subtree authorization pattern the #3316 review proved dangerous, and it is the one thing keeping RuntimeStateDoc from being cleanly read-only to clients.

## What the ADR proposes

Split comm **topology** (which comms exist, target, model, owning cell - stays runtime-owned in RuntimeStateDoc) from comm **state** (mutable trait values - moves to CommsDoc, writable by anyone in the room plus the runtime). "Anyone writes CommsDoc" is safe because state written for a comm_id with no topology is orphaned and ignored. Editors stop writing RuntimeStateDoc entirely, and the policy-snapshot carve-out is deleted.

Co-location was silently providing three invariants the split has to rebuild deliberately, so the ADR gates the change on all three landing in the same change set:

1. **Membership gate** anchored to RuntimeStateDoc topology (re-create Guard A as an explicit runtime-side drop of orphan state).
2. **Kernel-echo filter** re-homed structurally onto CommsDoc's receive path, with a CI guard that a kernel-authored change is never forwarded back to the kernel.
3. **GC authority**: the daemon gets CommsDoc write authority so `remove_comm`/`clear_comms` reclaim topology and state atomically.

## Design decisions worth a reviewer's attention

- **Substrate: CRDT spine, presence fast-lane deferred.** CommsDoc is a CRDT doc, chosen for continuous free durability, free late-join via the existing sync protocol, and the cheapest bidirectional kernel path (re-homed `diff_comm_state`). The ephemeral-presence alternative (Option D) is the right idea in the wrong slot - its peer-to-peer-convergence payoff is moot because a room already has a single authority, while it loses the durability and late-join this needs. It is retained as an *optional* in-flight fast-lane that commits to the CRDT on settle, built only if measured churn hurts.
- **Representation gates the seed shape.** Comm state must be a native per-key Automerge map, not a JSON-blob string per comm (a blob rewrite is one fat op per keystroke and bloats history). The code is ambiguous on this today; it must be nailed before the CommsDoc seed is frozen.
- **Dependency on ADR 0001.** CommsDoc adds a `comms_doc_id` ROOT pointer. ADR 0001's `is_pristine` keys on a ROOT identity-put allowlist and says "if creation writes a new ROOT scaffolding key, add it." `comms_doc_id` must be added, or a fresh notebook reads as non-pristine and never gets its starter cell.
- **Why not the full five-document split now (Option C).** The all-or-nothing endpoint is logically forced (the deep-dive confirms it), but shipping it as one program couples this cheap, additive, reversible change to an irreversible `cells`/`metadata` data-move out of a frozen-genesis document - the source of the premortem's most-dangerous failures. The structural split is its own ADR, gated separately.

## Supporting material in this PR

- `docs/handoffs/2026-06-01-host-convergence-deep-dive.md` - the 50-agent investigation that confirmed the forced five-document derivation and recommended decoupling, with the full Phase A / Gate 1 / Phase B / Gate 2 / Phase C sequencing and the premortem synthesis (19 verified failure modes; the "most dangerous" cluster is the basis for splitting C out).
- `docs/handoffs/2026-06-01-notebook-host-convergence.md` - the seed handoff that framed the two threads.

## Review asks

- Does the CRDT-spine / presence-fast-lane substrate call land, or do you want comm state ephemeral from the start?
- Is gating the change on all three rebuilt invariants (membership, echo filter, GC) the right bar, or too strict for a first cut?
- Scope check: this ADR is CommsDoc only. The divergence-recovery frame (`ResyncRequired`/`Sever`) and the MetadataDoc/CellsDoc structural split are separate ADRs. Agree with that decomposition?

Implementation lands as a follow-up once the decision is signed off.


rgbkrk:
> Automated multi-agent review (4 code-tracing reviewers + synthesis) run locally against the comm code. Posted for sign-off consideration - not an approval. Every claim below was checked against `file:line` on `main`.

# Review: ADR 0002 (CommsDoc split)

**Verdict: Sign off with changes.** The decision is sound. Option B is the right call, the substrate reasoning holds, and gating CommsDoc on all three co-location invariants landing in one change set is the correct bar. But three things must change before this is implementation-ready: the safety claim is scoped too loosely (it conflates "ignored by the kernel" with "ignored, period"), invariant #2 is mischaracterized (it's three suppression layers, not one actor diff), and the representation open question is a phantom (the code already does per-key maps). None of these reverse the decision. All of them would mislead an implementer working from the ADR text alone. Fix the prose and the open questions, keep the structure.

Four reviewers (safety, invariant completeness, substrate, ADR-0001 dependency) independently landed at "sound-with-caveats." No reviewer found a reason to reject Option B or to unship the gating bar.

## Answers to the three review asks

### Ask 1: Is "anyone writes CommsDoc" genuinely safe?

Directionally yes, but the ADR's one-line safety claim is more reassuring than the diff warrants, and it needs scoping.

The claim "state written for a comm_id with no topology is orphaned and ignored" is true **only at the kernel-forward boundary.** The membership gate that backs it (`diff_comm_state`, `runtime_agent.rs:1495-1522`) only forwards state for comm_ids present in the daemon's authoritative topology snapshot. Orphan state never reaches the kernel. Good.

But "ignored" overstates it. Today's carve-out (`policy.rs:391-489`) does far more than drop orphan state. It rejects comm creation by editors (`policy.rs:464-472`), rejects comm deletion (`policy.rs:454-461`), freezes all topology/metadata (`validate_comm_metadata_unchanged`, `policy.rs:491-513`), and rejects any write outside `comms/*/state`. Deleting it means a malicious or buggy peer **can** mint arbitrary comm-state entries. Those are inert to the kernel, but they are **not** inert to durable persistence or cross-client rendering. Today the carve-out blocks them from ever landing. An open-write CommsDoc lands them durably and broadcasts them to every peer's widget layer.

So the honest statement is: orphan state is dropped at the forward boundary, but persisted and rendered until a GC pass reaps it. The rebuilt Guard A changes from **write-rejection** (orphan write never enters the doc) to **forward-suppression** (orphan state lands, never forwards). That is a real semantic shift the ADR does not name.

Net on safety: there is no per-comm owner concept in the code today (`CommDocEntry` at `doc.rs:276` has no owner field), so the ADR is not weakening per-comm authorization. None exists; the current carve-out already lets any editor write state for any comm with topology. The decision does not regress authorization. It does shift the entire safety burden onto a gate that today governs only kernel-forwarding, not storage or render. That needs to be stated, and the orphan-reclamation path needs to be a named requirement, not an "or" fallback.

### Ask 2: Are the three rebuilt invariants complete?

Each named invariant is real and correctly anchored. But "co-location was silently providing three invariants" undersells the surface, and invariant #2 is mischaracterized in a way that would mislead an implementer.

The three are correct as gates. The split removes all three simultaneously, so gating the change set on all three is the right bar, not too strict. The premortem's blind spot is well-founded: the echo filter is actor-prefix-based (`runtime_agent.rs:351`), so a single-principal dev session cannot exercise the cross-principal forward path. The dangerous failure mode is partial re-homings that pass desktop CI and fail only in multi-principal rooms.

Gaps the ADR understates:

- **Invariant #2 is three layers, not one.** The ADR describes per-change actor discrimination. The real mechanism: (a) per-**field** LWW-authorship, `receive_sync_and_foreign_comms` walks each comm's state map key-by-key and retains a key only if its current LWW winner's actor is foreign (`doc.rs:3202-3204`); the doc comment at `doc.rs:3169-3175` explains why the naive fork-and-apply approach is unsafe (frontend writes are causally after kernel writes during a slider drag). (b) a content-hash `EchoSuppressor` with 30s TTL (`runtime_agent/echo_suppression.rs`, consulted at `runtime_agent.rs:407, 440-443`). (c) a no-op scalar skip in `merge_comm_state_delta` (`doc.rs:2545-2559`). The CI guard the ADR asks for ("kernel-authored change never forwarded") covers only layer (a). Add coverage that a content-hash echo within TTL is also dropped post-split.
- **GC has two properties, not one.** "Who may delete" (verified: `remove_comm` on comm_close at `jupyter_kernel.rs:2236`, `clear_comms` on relaunch at `launch_kernel.rs:92`) is correct. But co-location also gave **delete atomicity within one Automerge transaction**. Post-split, two docs have independent sync heads, so `remove_comm` becomes a delete in RuntimeStateDoc plus a delete in CommsDoc that cannot share a transaction. A peer can observe topology-gone/state-present or state-gone/topology-present mid-GC. Invariant #1 makes the topology-absent direction inert; the reverse (topology-present/state-gone) surfaces a widget with identity but no values, the same hazard the ADR flags only for export. Specify delete ordering (state before topology) or have readers treat absent-state-with-present-topology as "not yet loaded."
- **The comm entry co-locates more than three invariants.** `comms/{comm_id}` also holds `capture_msg_id` (the Output-widget routing field, `doc.rs:71`, written at `jupyter_kernel.rs:2161`) and an inline `outputs` list. `validate_comm_metadata_unchanged` (`policy.rs:497-502`) treats `target_name`, `model_module`, `model_name`, `outputs`, `seq`, and `capture_msg_id` as daemon-owned topology, so "topology" is wider than the ADR's parenthetical "(target, model, owning cell)." Two daemon-local comm-keyed caches (`capture_cache`, `output_widget_replay_cache`, `jupyter_kernel.rs:2232-2233`) are cleared on the same `remove_comm`. Extra readers of `state.comms` exist at `daemon.rs:4603-4612` and `output_commit_measure.rs:287-297`. Widen the Problem framing from "three invariants" to "a cluster of co-location guarantees, three load-bearing," and confirm the extra readers and the policy-sense topology fields stay in RuntimeStateDoc.

### Ask 3: Is the substrate / representation choice right?

The substrate choice (CRDT spine, presence fast-lane deferred) is sound and matches the code. The representation open question is overstated to the point of being a phantom.

**Substrate: accept as written.** A room has a single authority (the daemon is the sole RuntimeStateDoc writer), so Option D's P2P-convergence payoff genuinely does not apply, exactly as the ADR argues. The two properties the room needs (continuous durability, free late-join) come from the existing sync protocol at zero cost. Deferring the ephemeral fast-lane to measured need is reasonable. One unstated hazard worth a note: `put_comm` uses the deprecated `automunge::put_json_at_key` (`doc.rs:2481-2482`), which the crate warns is dangerous in multi-peer scenarios (two peers calling `put_object` at the same key produce competing objects, `automunge lib.rs:87-92`). Per-property updates correctly use the safe `update_json_at_key`, but initial state-object creation does not. Under "anyone writes CommsDoc," concurrent creation of the same comm's state object could fork. Call this out.

**Representation: the code is not ambiguous. Fix the ADR, not the code.** Comm state is **already** a native per-key Automerge map. The write path (`put_comm` -> `put_json_at_key` -> `put_object(Map)`, `doc.rs:2482`, `automunge lib.rs:125-130`), the per-property setters (`set_comm_state_property` requires `Value::Object(ObjType::Map)`, `doc.rs:2508-2513`), the read path (`automunge::read_json_value`, `doc.rs:2635`), the kernel-forward diff (per-key on `state.as_object()`, `runtime_agent.rs:1502-1517`), and the policy gate (`policy.rs:474-486`) are all already per-key. No `serde_json::to_string` of comm state exists anywhere. The only thing claiming "Str" is a stale doc-comment at `doc.rs:68`. The ADR cites `doc.rs:63`; the actual stale line is `:68`.

This **strengthens** the ADR: the per-key prerequisite is already satisfied, lowering implementation risk. Downgrade open question 1 from "confirm and, if needed, migrate the representation" to "the representation is already a per-key native map; delete the stale `doc.rs:68` comment; treat per-key as a non-regression invariant." Also: the per-comm `state` map is created lazily (`scaffold_map("comms")` on first comm_open, `doc.rs:2476`), not part of frozen genesis, so "before the seed is frozen" overstates the coupling. Per-comm state shape is enforced by the write helpers and unit tests, not by the seed bytes. There is one load-bearing connection the ADR misses: per-field LWW-authorship (the echo filter's layer a, `doc.rs:3202`) only works if state is a native map, so the representation question is a prerequisite for invariant #2, not just for history bloat.

## Blockers and required changes (deduped)

These must be addressed in the ADR text before sign-off. None require a different decision.

1. **Scope the safety claim.** Rewrite "orphaned and ignored" to "dropped at the kernel-forward boundary (`diff_comm_state`), but persisted and rendered until reaped." State the orphan-reclamation mechanism explicitly and make the topology-absence GC pass a **named requirement**, not an "or" fallback to daemon delete authority. Add a CommsDoc-size / orphan-floor consideration to Consequences, since anyone can now mint unbounded orphan comm_ids. (Safety dimension, concern.)

2. **Name both carve-out enforcement sites.** Consequences lists only the cloud host (`crates/runtimed-wasm/src/lib.rs:663-697`). The carve-out is enforced at a second site: the daemon peer-sync path (`crates/runtimed/src/notebook_sync_server/peer_runtime_sync.rs:114-116`), which runs the identical `runtime_state_policy_snapshot` before/after diff. An implementer deleting only the wasm gate leaves the daemon gate dangling. Add `peer_runtime_sync.rs:114-116` to the removal list and to the conformance-test scope (handoff item 13). (Safety dimension, concern.)

3. **Rewrite invariant #2 to name all three suppression layers** and which substrate each moves to: per-field LWW-authorship (`doc.rs:3202`, must preserve the causality argument at `doc.rs:3169-3175`); the content-hash `EchoSuppressor` (`echo_suppression.rs`, stays daemon-local but its forward path now reads CommsDoc); the no-op scalar skip (`doc.rs:2545`). Require the CI guard to cover content-hash echoes, not just actor-authored ones. (Completeness dimension, concern.)

4. **Split invariant #3 into "delete authority" and "delete atomicity."** State that cross-document GC is non-atomic post-split, that correctness rests on invariant #1 making the topology-absent direction inert, and specify ordering for the topology-present/state-absent direction. (Completeness dimension, concern.)

5. **Correct the representation open question** (ask 3 above): it is already a per-key map; fix the `doc.rs:68` comment, drop the migration framing, cross-link to invariant #2. (Substrate + ADR-0001 dimensions, concern.)

6. **Promote "membership gate lands atomically with carve-out deletion" from prose to a hard acceptance gate** in the Decision, mirroring the handoff's Phase A item 6. Deleting the carve-out before the cross-doc membership gate lands opens a real window: orphan or abusive comm state reaching the kernel or persisting unfiltered. The ADR forbids that ordering, but only in prose. Add a regression test: an editor writes CommsDoc state for a comm_id with no RuntimeStateDoc topology; assert (a) never forwarded to the kernel, (b) reaped by the orphan-drop pass. Add a multi-principal integration test (editor + runtime_peer in one room) since single-principal testing is the identified blind spot. (Safety + completeness + ADR-0001 dimensions, concern.)

## Action item: the ADR-0001 `comms_doc_id` dependency

This is a real sequencing constraint, and the ADR states it almost right but omits that the predicate it depends on does not exist yet.

ADR 0001 (#3335) merged as docs-only, "Proposed. No code changes." There is no `is_pristine` method in the tree. The live seeding decision is still `is_uninitialized_notebook_doc` (`crates/runtimed/src/notebook_sync_server/load.rs:1063`, `cell_count()==0 && metadata.is_none()`), which has no ROOT-key allowlist and would **not** misfire on a `comms_doc_id` put. So as written today, adding `comms_doc_id` breaks nothing. The hazard materializes **only** once ADR 0001's predicate ships.

Once it ships, the dependency is exact. `new_inner` writes precisely `notebook_id` (`lib.rs:1007`) and `runtime_state_doc_id` (`lib.rs:1008-1012`) to ROOT. A third ROOT scaffold put (`comms_doc_id`) without an allowlist entry makes a fresh notebook read non-pristine and never get its starter cell. ADR 0001's allowlist is `notebook_id || runtime_state_doc_id` (`docs/adr/0001-notebook-seeding-invariant.md:144-145`) and the predicate fails closed on any unlisted ROOT op. The proposed `every_fresh_constructor_is_pristine` guard would catch the omission with correct directionality, but that test lives only in the implementation plan, not in the codebase.

Concrete actions:

- Reword the Consequences line: the dependency is a sequencing constraint on ADR 0001's **implementation**, not its current state. "When `is_pristine` lands (ADR 0001 is still Proposed/unimplemented), `comms_doc_id` must be added to its ROOT identity-put allowlist in the same change set that stamps it." Cite `load.rs:1063` as the current pre-predicate code so a reader knows the allowlist does not exist yet.
- Ensure `comms_doc_id` stamping goes through `new_inner` so the enumerated constructor guards cover it, not a side path.
- Stamp `comms_doc_id` via an `ensure_`-based idempotent path (fail-closed on absent-for-existing) like `ensure_runtime_state_doc_id` (`lib.rs:1435-1443`), not a bare put, and make the pointer deterministically derivable as `comms:{notebook_id}` (mirrors `default_runtime_state_doc_id`, `crates/notebook-doc/src/lib.rs:118-120`). Verified accurate.
- Add a constructor-guard acceptance test mirroring ADR 0001: a freshly created notebook that stamps `comms_doc_id` is still `is_pristine == true` and still gets exactly one starter cell. Cheap insurance against the silent never-seeds failure.

Open question 5 (CommsDoc needs its own `is_canonical_comms_seed_change` hash-pin) is correctly identified as a real gap, not duplicated machinery; there is no existing such helper in runtime-doc. Keep it.

rgbkrk:
Verified every claim against `file:line` on `main` before editing. All six blockers, the ADR-0001 action item, and the two extra hazards check out; none reverses the decision, all sharpen it. Pushed in 16dbfb33. Per blocker:

1. **Safety claim scoped.** Rewrote "orphaned and ignored" to name the real shift: Guard A moves from **write-rejection** to **forward-suppression**. Orphan state is dropped at the `diff_comm_state` kernel-forward boundary but lands durably and renders until reaped, so the orphan-reclamation pass is now a **named requirement**, not an "or" fallback. Added an orphan-floor / doc-size consequence since anyone can mint unbounded orphan comm_ids. Also recorded that there's no per-comm owner field today (`CommDocEntry`, `doc.rs:276`), so the split doesn't regress authorization, it relocates the safety burden onto a gate that currently governs only forwarding.

2. **Both enforcement sites named.** Added the daemon peer-sync gate (`peer_runtime_sync.rs:114-116`) alongside the wasm site to the removal list and the conformance-test scope. Confirmed both run the identical `runtime_state_policy_snapshot` before/after diff.

3. **Invariant #2 rewritten as three layers.** Per-field LWW-authorship (`doc.rs:3200-3205`, preserving the causality argument at `:3169-3175`), the content-hash `EchoSuppressor` (`echo_suppression.rs`), and the no-op scalar skip (`doc.rs:2545`). The CI guard now must cover content-hash echoes within TTL, not just actor-authored ones.

4. **Invariant #3 split into authority + atomicity.** Stated that cross-doc GC is non-atomic post-split (independent sync heads), that invariant #1 makes the topology-absent direction inert, and required delete ordering (state before topology) so present-topology/absent-state never strands a values-less widget.

5. **Representation corrected.** You're right that the code already does per-key maps - I'd misread it. Downgraded open question 1 from "confirm/migrate" to "already per-key, non-regression invariant," fixed the citation (the stale comment is `doc.rs:68`, not `:63`) and flagged it for deletion, and cross-linked it to invariant #2 since per-field LWW-authorship depends on the native map. Added the `put_comm` concurrency hazard: it creates the state object with the deprecated `put_json_at_key` (`doc.rs:2481-2482`), which can fork under concurrent same-key creation in an open-write doc - the initial-create path needs the safe API or daemon-only creation.

6. **Ordering promoted to a hard gate.** The carve-out deletion (both sites) and the membership gate now ship in one change set or not at all, stated as an acceptance gate rather than prose. Added the three required acceptance tests, including the multi-principal (editor + runtime_peer) integration test for the actor-prefix blind spot.

**ADR-0001 dependency reworded.** Corrected to a sequencing constraint on 0001's *unshipped* predicate: the live seeding code is `is_uninitialized_notebook_doc` (`load.rs:1063`) with no allowlist, so `comms_doc_id` breaks nothing today; the hazard is only live once `is_pristine` ships. Specified stamping through `new_inner` via an `ensure_`-based path, deterministic `comms:{notebook_id}`, with a constructor-guard test that a freshly stamped notebook stays `is_pristine` and still seeds.

**Also widened** the Problem framing from "three invariants" to "a cluster of co-location guarantees, three load-bearing," noted that policy-sense topology is wider than the parenthetical (`capture_msg_id`, `outputs`, `seq`), and added the extra `state.comms` readers (`daemon.rs:4603-4612`, `output_commit_measure.rs:287-297`) plus the daemon-local comm-keyed caches to the migration check.

Open question 5 (CommsDoc needs its own `is_canonical_comms_seed_change`) stays - confirmed no such helper exists.

rgbkrk:
> Automated verification of `16dbfb33` against the prior review (3 read-only verifiers + synthesis). Each blocker re-checked against `file:line` on `main`.

## Verdict: implementation-ready

All 6 blockers and the ADR-0001 item are resolved in `16dbfb33`. Spot-checked every load-bearing citation against `main`; they back the revised text.

- **B1 (safety scoping):** scoped to the kernel-forward boundary (`diff_comm_state`, `runtime_agent.rs:1495-1522`), Guard A's write-rejection→forward-suppression shift named, orphan-reclamation made a named requirement (not an "or"), orphan-floor/doc-size consequence added, no-owner-field recorded (`doc.rs:276`, verified).
- **B2 (both enforcement sites):** wasm host (`runtimed-wasm/src/lib.rs:663-697`) and daemon peer-sync (`peer_runtime_sync.rs:114-116`) both named for removal; daemon path provably runs the same `runtime_state_policy_snapshot` before/after diff; conformance scope extended to the daemon, and no existing test covers either site.
- **B3 (kernel-echo filter):** described as three layers (per-field LWW `doc.rs:3200-3205`, `EchoSuppressor` 30s TTL `echo_suppression.rs:4`, no-op scalar skip `doc.rs:2545-2559`); CI guard upgraded to require content-hash-within-TTL echoes, not just actor-authored.
- **B4 (GC):** split into delete-authority vs delete-atomicity, non-atomic cross-doc reality stated, topology-absent direction grounded as inert via the `before.get(comm_id)` gate, delete ordering (state before topology) required.
- **B5 (representation):** corrected to "already a per-key native Automerge map" (`put_comm`→`put_json_at_key`→`put_object(Map)`, `doc.rs:2481-2482`), stale `Str` doc-comment pinned to `doc.rs:68` and flagged for deletion, `put_comm` concurrency-fork hazard real (`put_json_at_key` is `#[deprecated]`, `automunge/lib.rs:90-91`).
- **B6 (atomic landing):** promoted to a hard acceptance gate (one change set or not at all), multi-principal editor+runtime_peer test named, tied to the actor-prefix blind spot at `runtime_agent.rs:351` (exact).
- **ADR-0001 item:** reworded as a sequencing constraint on the unshipped `is_pristine` predicate (live gate is `is_uninitialized_notebook_doc`, `load.rs:1063`, no allowlist, so `comms_doc_id` breaks nothing today); stamping via `new_inner` through an `ensure_`-idempotent path mirroring `ensure_runtime_state_doc_id` (`lib.rs:1435-1443`), deterministic `comms:{notebook_id}` (`default_runtime_state_doc_id`, `lib.rs:118-120`), constructor-guard test. Consistent with PR #3342's scalar-pointer-only `is_pristine` allowlist; open question 5 correctly retained (`is_canonical_comms_seed_change` doesn't exist).

### Non-material defects (no conclusion changes, fix on the next pass)

- `EchoSuppressor` cited as "consulted at `runtime_agent.rs:407,440-443`": 407 is the consult entry, but 440-443 is `record_outgoing` (the post-send record side). The actual `is_recent_echo` read is `runtime_agent.rs:1598`, uncited.
- `clear_comms` relaunch GC site cited as `launch_kernel.rs:92`; real path is `crates/runtimed/src/requests/launch_kernel.rs:92` (drops `requests/`). Line+fn correct.
- Substrate says the per-comm `state` map is "created lazily on first comm_open (`scaffold_map`, `doc.rs:2476`)"; `scaffold_map` lazily creates the top-level `comms` map. The per-comm `state` object is created by `put_json_at_key`/`put_object` at `doc.rs:2481-2482`.

No new factual errors introduced.

PR #3342, opened 2026-06-03
feat(notebook): causal is_pristine seeding gate (ADR 0001)

rgbkrk:
Implements [ADR 0001](docs/adr/0001-notebook-seeding-invariant.md): replace the cell-count/metadata proxy that decides whether a fresh notebook gets a starter cell with a causal `is_pristine` predicate derived from the document's own change history, shared by the daemon and the cloud wasm host.

## What changed

- **`NotebookDoc::is_pristine`** (`crates/notebook-doc/src/lib.rs`) - the new gate. A document is pristine iff its change graph contains nothing beyond the canonical schema seed and the two identity puts (`notebook_id`, `runtime_state_doc_id`). It matches on `op.action` (an inserting `Put`) and `op.pred.is_empty()`, not just `obj`/`key`, so a post-creation delete or overwrite of an identity key correctly fails the gate. A `metadata.is_some()` fast-path short-circuits the common already-seeded case; the history walk is the real backstop.
- **Daemon** (`crates/runtimed/src/daemon.rs`, `notebook_sync_server/load.rs`) - all three seeding sites switch to `doc.is_pristine()`; the old `is_uninitialized_notebook_doc` (`cell_count() == 0 && metadata.is_none()`) is deleted.
- **Cloud wasm host** (`crates/runtimed-wasm/src/lib.rs`) - `seed_initial_code_cell_if_empty` gates on `!self.doc.is_pristine()` instead of `cell_count() > 0`. Daemon and cloud now evaluate one shared predicate.
- **Tests** - unit coverage in `notebook-doc` (fresh / one-cell / emptied / metadata-stamped / overwritten-or-deleted-identity-key / every-fresh-constructor), and a runtimed reconnect test asserting an emptied notebook is not re-seeded.
- ADR 0001 marked Accepted.

## Why it's better

`cell_count() == 0` is a materialized-view proxy that isn't convergence-safe (a peer can observe zero cells transiently mid-sync); the old code was correct only because the daemon seeds under a write lock before the handoff. `is_pristine` derives the fact from immutable history: monotonic (an emptied notebook keeps its add+delete history, so it's never re-seeded), convergence-safe, and identical for both hosts with no metadata coupling.

## Verification

- **codex local review**: no correctness issues. Confirmed all call sites switched, the old guard fully removed, and the action/pred logic rejects post-creation identity edits.
- **Adversarial pass**: 25 probes (fresh constructors, save/load round-trips, schema migration v3/v4→v5, foreign-actor skeleton, sync-delivered cells/metadata, set-then-delete-to-empty) - could not produce a misclassification.
- **Gate**: `cargo xtask lint` (all four sections), `notebook-doc` tests, `runtimed-wasm` idempotency test, and `runtimed` sync tests all green.

One defensive note baked into the code from the adversarial pass: the allowlist must never grow to include a structural ROOT key (`cells`/`metadata`), or a foreign-authored skeleton would read pristine. Comment added at the allowlist.

Draft pending your review. Built and verified in an isolated worktree; rebased onto current `main`.

### DIFF
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -2450,6 +2450,14 @@ impl NotebookDoc {
                 // genesis: an inserting `Put` (empty `pred`) to ROOT's
                 // `notebook_id` / `runtime_state_doc_id`. A delete or an overwrite
                 // of those keys is post-creation history and must fail the gate.
+                //
+                // INVARIANT: this allowlist must stay limited to scalar identity
+                // pointers and must never grow to include `cells`, `metadata`, or
+                // any structural ROOT key. Those Maps are admitted only via the
+                // exact-hash genesis change (the `seed.contains` skip above), so a
+                // foreign actor that authors its own `cells`/`metadata` skeleton
+                // reads non-pristine. Allowlisting a structural key would let such a
+                // foreign skeleton read pristine and be wrongly re-seeded.
                 let is_creation_scaffold_put = matches!(op.action, LegacyOpType::Put(_))
                     && op.pred.is_empty()
                     && matches!(op.obj, LegacyObjectId::Root)
--- a/crates/notebook-doc/src/lib.rs
+++ b/crates/notebook-doc/src/lib.rs
@@ -2446,19 +2446,29 @@ impl NotebookDoc {
                 continue;
             }
             for op in change.decode().operations {
-                // Only the *initial creation* of an identity key is allowed beyond
-                // genesis: an inserting `Put` (empty `pred`) to ROOT's
-                // `notebook_id` / `runtime_state_doc_id`. A delete or an overwrite
-                // of those keys is post-creation history and must fail the gate.
+                // Scaffold allowed beyond genesis: the daemon writes these at room
+                // creation, before any seeding or user interaction.
                 //
-                // INVARIANT: this allowlist must stay limited to scalar identity
-                // pointers and must never grow to include `cells`, `metadata`, or
-                // any structural ROOT key. Those Maps are admitted only via the
-                // exact-hash genesis change (the `seed.contains` skip above), so a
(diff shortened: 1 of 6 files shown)


rgbkrk:
## Verification detail

**codex local review** (`codex exec review`, run in the worktree against `main...HEAD`):

> I found no concrete correctness issues in the diff. The code call sites were switched to `is_pristine`, the old helper has no remaining code references, and the predicate rejects cell/metadata history plus sequential identity overwrites/deletes while preserving fresh constructor/save-load behavior.

**Adversarial pass** - 25 probes, none produced a misclassification:
- Fresh constructors (`new`, `new_with_actor`, `new_with_encoding`, `bootstrap`, the daemon `load_or_create_with_actor` path) all read pristine; Utf16 + cross-encoding round-trips preserve it; multi-actor save/load ×5 keeps the genesis seed-hash match stable (actor-table re-indexing doesn't perturb the content-addressed change hash).
- The dangerous direction (non-fresh reading pristine) - all rejected: `save→load` preserves both `action` and `pred`, so add+delete, identity overwrite, and identity delete stay non-pristine across round-trips; schema migration v3/v4→v5 re-puts `schema_version` as an overwrite (non-empty `pred` to a non-allowlisted key) and reads non-pristine; a foreign-actor v5 skeleton reads non-pristine because its `cells`/`metadata` `put_object`s aren't the canonical seed change.
- Metadata fast-path never misfires (populated-only); reserved keys (`runtime`/`ephemeral`) bypass the fast-path but the history walk catches them (their op targets the `metadata` object, not a ROOT identity key).

**Gate**: `cargo xtask lint` (Rust fmt, JS/TS 852 files, ruff, ty), `notebook-doc` unit tests, `runtimed-wasm` idempotency, `runtimed` sync tests - all green.

Issue #1334, opened 2026-03-31
feat(runtimed): SSH remote runtimes


rgbkrk:
## Summary

Run notebook kernels on remote machines over SSH. The protocol the local side speaks to the remote side is the same Automerge sync wire we already use for in-process and subprocess peers; SSH is just a transport. The kernel's own ZMQ / IPC transport stays local on whichever box the kernel runs on. Nothing about kernel control needs to cross the SSH boundary except NotebookDoc deltas, RuntimeStateDoc deltas, and blob frames.

This is the SSH-transport implementation of the URI-scheme + daemons-as-peers primitive. The wire and ownership semantics are shared with the same-machine cross-daemon case; this issue is about the transport, discovery, and lifecycle pieces that are SSH-specific.

## Architecture

There are two viable shapes, both compatible with the existing peer model. They differ in what runs on the remote box.

### Shape A: Remote agent (lighter)

```
LOCAL MACHINE                              REMOTE MACHINE

┌──────────────────────┐                   ┌──────────────────────────┐
│  Desktop App / MCP   │                   │  runt agent              │
└─────────┬────────────┘                   │                          │
          │ Unix socket                    │  Connects back to daemon │
┌─────────▼────────────┐                   │  via SSH-tunneled socket │
│  runtimed (daemon)   │  SSH tunnel       │                          │
│                      │◄═════════════════►│  Watches RuntimeStateDoc │
│  NotebookRoom        │  (socket forward) │  for queued executions   │
│  - RuntimeStateDoc   │                   │                          │
│  - execution queue   │                   │  ┌───────────────────┐   │
│  - blob store (local)│                   │  │ Python/Deno       │   │
└──────────────────────┘                   │  │ kernel process    │   │
                                           │  └───────────────────┘   │
                                           └──────────────────────────┘
```

The agent subprocess work that shipped (#1333, #1431, #1433, #1449) gives us this shape almost for free: the agent is already a regular Automerge peer on the daemon's socket. SSH adds a forwarded socket and a way to spawn the agent on the remote side. Single daemon, single source of truth.

Trade-off: the kernel's lifetime is tied to the local daemon. Laptop closes, local daemon dies, remote kernel becomes orphan.

### Shape B: Remote daemon (heavier, generalizes further)

```
LOCAL MACHINE                              REMOTE MACHINE

┌──────────────────────┐                   ┌──────────────────────────┐
│  Desktop App / MCP   │                   │  runtimed (remote)       │
[...]

PR #3359, opened 2026-06-03
feat(daemon): add hidden SSH runtime-agent launcher

rgbkrk:
## Summary

Adds a hidden daemon/runtime-agent override that lets a local Desktop daemon spawn a runtime agent through SSH. This gives us a concrete Desktop <> SSH runtime target primitive without adding UI yet.

The new hidden path is:

- `runtimed --dev run --runtime-agent-exe <path>`
- `RUNTIMED_RUNTIME_AGENT_EXE=<path>` as an environment fallback
- `cargo xtask dev-daemon --runtime-agent-exe scripts/ssh-runtime-agent`

The checked-in SSH wrapper defaults to `lab-runner1` and `/usr/local/bin/runtimed-nightly`, starts a local token-gated proxy in front of the daemon socket, reverse-forwards that proxy over SSH, starts a user-owned Unix-socket bridge on the remote host, and runs the remote `runtimed-nightly runtime-agent` against the forwarded local daemon path.

The tunnel now keeps retry stderr inside the trapped local proxy directory, passes `GatewayPorts=no` for the reverse forward, sends the proxy token over stdin rather than argv/env, and keeps the runtime socket bridge user-owned with `0600` permissions.

## Alignment with workstation/runtime peer work

This PR is aligned with the newer workstation split by staying deliberately on the SSH/direct-access side:

- The local Desktop daemon remains notebook room authority.
- The remote process is still the local `RuntimeAgent` protocol reached through SSH, not a hosted `runtime_peer`.
- Registered workstations continue to use the mainline `cloud-runtime-agent` / `runtime_peer` path with room-mediated authorization.
- The SSH bridge can still be shown as a compute source in the future UI, but its trust boundary is Desktop ownership plus SSH keys, not hosted workstation registration.

After rebasing over #3491, #3492, and #3495, the boundary is clearer: SSH is Desktop-only direct access. It should not use hosted workstation registration, hosted attach jobs, or `runtime_peer` ACL flow. Future Desktop SSH work can reuse the shared Workstations projection shape, but its source data should come from local Desktop/daemon SSH target configuration. This PR stays as the hidden launch primitive and lab-runner1 smoke.

## Why hidden

This is intentionally a hidden/dev-only feature for now. The control-plane path works, but the Desktop product model still needs first-class configuration for:

- compute target state that distinguishes Local Desktop, registered workstations, and SSH/direct access
- local-to-remote working-directory mapping
- remote project/environment policy instead of local prewarmed or inline env paths
- blob replication for large/rich/binary outputs
- UI placement that keeps compute target selection separate from the active interpreter/packages shown in Environment

## Verification

- `bash -n scripts/ssh-runtime-agent`
- `node --check scripts/smoke-ssh-runtime-agent.mjs`
- `cargo test -p runtimed runtime_agent_exe`
- `cargo test -p xtask parse_dev_daemon_options`
- `cargo clippy --workspace --exclude notebook --exclude runtimed-wasm --exclude sift-wasm --exclude runtimed-py --all-targets -- -D warnings`
- `cargo xtask lint --fix`
- `pnpm --dir packages/runtimed-node smoke:ssh-runtime-agent -- --timeout-ms 180000`
- `uv run pr-review https://github.com/nteract/nteract/pull/3359 --max-turns 120 --out .context/reviews/pr-3359-rerun-10.json`
- `uv run pr-review https://github.com/nteract/nteract/pull/3359 --max-turns 120 --extra-prompt "<security audit prompt>" --out .context/reviews/pr-3359-security-rerun-7.json`

Live smoke output confirmed execution on `lab-runner1` without depending on the raw hostname value, which is redacted because it is present in the runtime-agent environment:

```text
hostname_matches_expected True
hostname_sha256 473c187f2840b8595e91ed5703842d17568213c9a3e6bb68d45890b279028b3b
Linux-6.17.0-1015-aws-x86_64-with-glibc2.39
```

### DIFF
--- a/crates/runtimed/src/main.rs
+++ b/crates/runtimed/src/main.rs
@@ -575,7 +575,14 @@ async fn main() -> anyhow::Result<()> {
 }
 
 fn resolve_runtime_agent_exe(cli_value: Option<PathBuf>) -> Option<PathBuf> {
-    cli_value.or_else(|| std::env::var_os(RUNTIME_AGENT_EXE_ENV).map(PathBuf::from))
+    let path = cli_value.or_else(|| std::env::var_os(RUNTIME_AGENT_EXE_ENV).map(PathBuf::from))?;
+    Some(std::fs::canonicalize(&path).unwrap_or_else(|e| {
+        eprintln!(
+            "Runtime agent executable not found: {}: {e}",
+            path.display()
+        );
+        std::process::exit(1);
+    }))
 }
 
 async fn run_daemon(config: DaemonConfig) -> anyhow::Result<()> {
@@ -667,6 +674,16 @@ mod tests {
             ),
             other => panic!("expected run command, got {other:?}"),
         }
+    }
+
+    #[test]
+    fn resolve_runtime_agent_exe_canonicalizes_existing_path() {
+        let current_exe = std::env::current_exe().unwrap();
+        let canonical = std::fs::canonicalize(&current_exe).unwrap();
+        assert_eq!(
+            resolve_runtime_agent_exe(Some(current_exe)),
+            Some(canonical)
+        );
     }
 }
 
--- a/crates/runtimed/src/main.rs
+++ b/crates/runtimed/src/main.rs
@@ -14,6 +14,8 @@ use runtimed::daemon::{Daemon, DaemonConfig};
 use runtimed::service::ServiceManager;
 use tracing::info;
 
+const RUNTIME_AGENT_EXE_ENV: &str = "RUNTIMED_RUNTIME_AGENT_EXE";
+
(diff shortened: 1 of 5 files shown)


rgbkrk:
> *This was generated by AI during triage.*

Closing this draft as superseded rather than merging it.

The hidden local-daemon SSH launcher predates the workstation architecture that has since landed. The current boundary puts a daemon and runtime agent beside the remote kernel, then represents accepted compute through the existing room-scoped `runtime_peer` model. A future Desktop SSH path should transport users to that remote daemon boundary instead of introducing a parallel direct-launch mechanism.

That updated design is tracked in #1334.

PR #3373, opened 2026-06-04
fix(notebook): move code cell presence into status line

rgbkrk:
## Summary

- Move code-cell presence activity into the left side of the current-line status bar.
- Keep the execution/runtime readout on the right side of the same line.
- Remove the now-unneeded presence separator and lock the new DOM order with a focused unit test.

## Verification

- `cargo xtask lint --fix`
- `pnpm test:run src/components/cell/__tests__/CodeCellCurrentLine.test.tsx apps/notebook/src/components/__tests__/code-cell-output-focus.test.tsx`
- `git diff --check`


rgbkrk:
Opted to keep it as is instead.

Issue #3381, opened 2026-06-04
feat(notebook-cloud): register remote workstation doc agents for hosted compute


rgbkrk:
## Summary

Track the provider-neutral remote compute path for nteract workstations. The goal is for a process running inside a provider environment to authenticate to `preview.runt.run`, register itself as an available workstation/doc agent, and attach to selected hosted notebook rooms as a scoped `runtime_peer` when the room selects it for execution.

In this model, **workstation** is the nteract product noun. JupyterHub can provide an nteract workstation. Outerbounds can provide nteract workstations. Provider adapters differ below the control plane, but hosted rooms see the same shape: online targets with capabilities, attach status, and runtime authority only after selection.

The UI also needs a broader compute-source taxonomy:

- Local Desktop: owner-only local compute from the desktop daemon, similar in shape to local agent compute surfaces.
- Workstations: API-registered doc agents, such as Outerbounds or JupyterHub, that can attach to hosted rooms through room-mediated runtime authority.
- SSH: direct key-based cross-compute access from the Desktop prototype; it is workstation-like in the UI but not API-registered.

The selected compute target is room state. The current Python, kernelspec, packages, and readiness are environment state for the current notebook after a target is attached.

This is related to, but not the same as, the SSH runtime-agent path in #1334 and #3359. The SSH path proves that execution can move to a remote host while the local daemon remains room authority. The workstation/doc-agent path inverts connection direction: the provider-side process dials home, and hosted room authority stays in the Cloudflare Durable Object.

It should also account for the shared Content rail prototype in #3380. That PR is not a runtime transport prototype, but it establishes the host-provided catalog surface where future workstation/project artifacts can appear without making the rail own runtime attachment or execution authority.

Docs:

- Architecture: `docs/adr/remote-workstation-doc-agents.md`
- Operator path: `docs/remote-workstation.md`

Elements prototype:

- `apps/elements/components/cloud-notebook-shell-example.tsx`
- `apps/elements/content/docs/compute-placement.mdx`

## Architecture shape

```mermaid
sequenceDiagram
    participant User as Browser/Desktop
    participant Room as preview.runt.run Room DO
    participant Control as preview.runt.run Workstation Control
    participant Agent as Workstation Doc Agent
    participant Adapter as Provider Runtime Adapter
    participant Runtime as Remote runtimed or lightweight agent
    participant Kernel as Provider Kernel

    Agent->>Control: register workstation + heartbeat
    User->>Room: select workstation target
[...]

PR #3387, opened 2026-06-04
docs(notebook-cloud): prototype remote workstation compute

rgbkrk:
## Summary

This draft explores remote compute as provider-neutral nteract workstations/doc agents for hosted notebook rooms, with Elements prototypes for the shell and compute-placement questions.

It adds an ADR for the dial-home workstation/doc-agent model, updates hosted topology references, and refreshes the Elements catalog so the design can be inspected at `/docs/cloud-notebook-shell` and `/docs/compute-placement`.

Issue: https://github.com/nteract/nteract/issues/3381

## What changed

- Added `docs/adr/remote-workstation-doc-agents.md` covering registered workstations, doc-agent control channels, `runtime_peer` room attachment, current-Python policy, and the Local Desktop / Workstations / SSH compute-source taxonomy.
- Added a cloud-shell Elements prototype showing an attached Outerbounds current-Python workstation alongside JupyterHub and SSH-style targets.
- Added `/docs/compute-placement`, which compares rail placement, environment grouping, and an explicit connect-compute flow for Local Desktop, registered Workstations, and SSH/direct access.
- Replaced the oversized Elements home preview with a compact catalog that includes the new compute placement page.

## Verification

- `cargo xtask lint --fix`
- `pnpm exec tsc --noEmit -p apps/elements/tsconfig.json`
- `git diff --check`
- Playwright render checks for `/docs/cloud-notebook-shell` and `/docs/compute-placement`

## Notes

This is intentionally a draft/design PR. The implementation does not add Worker routes, connector commands, or room dispatch yet; it is meant to make the product, security, and protocol split reviewable before the backend work starts.


rgbkrk:
## Design notes after grounding this against the hosted code

I traced these decisions through the actual `apps/notebook-cloud/` Worker and the Rust sync stack. The authority model here matches what is already shipped, which is the good news. The risk is concentrated in one place the ADR underplays: the dial-out `runtime_peer` client does not exist in any form yet.

### What the real code already backs

- **Decision 3** (attach at `wss://preview.runt.run/n/<id>/sync` as `runtime_peer`) is the real route. `src/index.ts` `routeRoomSync` authenticates, origin-checks, then authorizes scope before handing to the room DO. The endpoint shape in the ADR is exact.
- **Decision 4** ("runtime peers may update accepted executions but may not create them") is not just pointed at, it is enforced. `crates/runtime-doc/src/policy.rs` `validate_runtime_peer_runtime_delta` rejects root-key changes (room-host owned), allows `queued -> running -> done/error` plus append-only outputs, and blocks creating execution records. There are tests. The ADR could state this more strongly.
- **The `runtime_peer` boundary** (cannot be derived from editor/owner, needs an explicit ACL row) is in `src/authorization.ts` `aclRowsCoverScope`, with a test asserting owner/editor do not imply `runtime_peer`. Decision 3 and the Security section are consistent with it.
- **Decision 7 / UI prototype** maps directly to `viewer/shell-capabilities.ts`: `canExecute = runtimeAvailable && (editor||owner)`, and a `runtime_peer` keeps `access.level = viewer` while `runtime.canWriteRuntimeState = true`. Every prototype state listed is implementable against this today.
- **Open Q7 (blob path)**: both options already exist. `scopeAllowsFrame` permits `PUT_BLOB` for `runtime_peer` over the room socket, and the Worker exposes `PUT /api/n/:id/blobs/:hash`. This is a product/accounting choice, not a build.

### The gap to put on the critical path

There is no outbound `runtime_peer` WebSocket client anywhere in the Rust tree.

- `crates/runtimed/src/runtime_agent.rs` hosts the kernel and writes RuntimeStateDoc, but `connect_and_handshake` is hardcoded to a local Unix socket / named pipe via `Handshake::RuntimeAgent`.
- `crates/notebook-sync/src/connect.rs` is Unix-socket only.
- `runt-publish` reaches the cloud over HTTP POST, not the sync WebSocket.
- There is no `Handshake` variant for an authenticated remote `runtime_peer`, and no `wss://` dialer.

The one bit of good news: `spawn_peer_writer` in the sync server is already generic over `AsyncWrite`, so the sink side is not the blocker. Steps 4 and 7 of the implementation sequence quietly depend on a new outbound WS sync client plus a remote auth handshake. That deserves promotion from Decision 6's "may eventually become a native mode in runtimed" to an explicit v1 build item. It is the longest pole.

### The auth bridge is the fuzziest decision

Decision 3 offers "a normal validated bearer credential or a short-lived room attachment ticket minted by the hosted service." Today `src/identity.ts` validates exactly three things: a dev token, an Anaconda OIDC bearer, and an Anaconda API key. An Outerbounds workstation's own OIDC token (issuer `outerbounds-public...s3`, aud `merced.obp.outerbounds.com`) is none of them. So the attachment ticket is load-bearing and unspecified: what signs it, what audience the DO checks, how it maps to `user:anaconda:*`. For the prototype the dev-token path (`X-Notebook-Cloud-Dev-Token` + `X-Scope: runtime_peer`) plus a seeded `runtime_peer` ACL row is enough, and Security already blesses that for early dev. The ticket contract is the thing to pin before this leaves dev.

### Outerbounds transport reality (closes part of Open Q5)

I checked a live workstation through the bundled CLI in the VS Code extension. The box is web-access-only: `ssh_access.enabled=false`, named ports disabled, reachable through a k8s-exec credential (`generate-workstation-token`) and `prepare-for-ssh-access` (alias `<id>-0`). Two consequences:

- Dial-home is not just "preferred" for Outerbounds, it is required. The #3359 SSH path cannot reach this workstation inbound. Decision 1's inversion is the only option.
- Stable workspace identity is already available: `instance_id` (e.g. `ws-...`) and `workstation_name`. Enough to fill `provider_instance` / `hostname` in the registry schema without further extension reverse-engineering.

Asking @pullfrog for a focused pass on the seams above.

rgbkrk:
@pullfrog please review this design PR against the hosted implementation in `apps/notebook-cloud/` and the Rust sync stack in `crates/`. I do not need a general pass. I want judgment on these specific points, in priority order.

**1. Is the implementation sequence honest about the dial-out client? (highest priority)**
There is no outbound `runtime_peer` WebSocket client in Rust today: `crates/runtimed/src/runtime_agent.rs` is Unix-socket only via `Handshake::RuntimeAgent`, `crates/notebook-sync/src/connect.rs` is Unix-only, and `runt-publish` uses HTTP POST. Decision 6 frames the remote daemon as something that "may eventually become a native mode inside runtimed," yet steps 4 and 7 both require a working `wss://.../n/<id>/sync` `runtime_peer` connection. Should a minimal WS sync client plus a new authenticated `Handshake` variant be its own explicit early step? Is there a smaller seam than rebuilding the agent, given `spawn_peer_writer` is already generic over `AsyncWrite`?

**2. What is the minimal correct attachment-ticket contract? (highest priority)**
Decision 3 and Security depend on a "short-lived room attachment ticket," but `src/identity.ts` only validates dev-token / Anaconda OIDC / Anaconda API key. Specify the smallest correct ticket: issuer, audience, principal mapping to `user:anaconda:*`, TTL, and replay protection, such that a connector presents it on the `/n/:id/sync` upgrade and the DO verifies it without a network round-trip. Does this belong in `hosted-credential-transport.md` instead of here?

**3. Does Decision 4 describe existing behavior or net-new room-host code?**
Decision 4 lays out the ExecuteCell flow, but does `src/notebook-room.ts` currently turn a `NotebookRequest::ExecuteCell` into a queued execution entry in RuntimeStateDoc? `runtime-peer-and-blob-authority-audit.md` lists "define which request variants editors/owners can send and how they route to the active runtime peer" as unresolved. The ADR should say which decisions describe shipped behavior and which are still to build.

**4. Where does target state live, and does it collide with `policy.rs`?**
Decision 7 leaves open whether the active-workstation target lives in a room-host portion of RuntimeStateDoc or in DO side-state mirrored via `SessionControl`. A late-joining viewer must correctly see ready/disconnected. Which is correct, and if it goes in RuntimeStateDoc, does it conflict with `validate_runtime_peer_runtime_delta` treating root keys as room-host owned?

**5. What is the smallest correct disconnect/revocation state machine?**
Security says a disconnected workstation must fail queued work after a bounded timeout and revocation must close `runtime_peer` sockets. Where does that timer live (room DO?), and how do we avoid wedging the room if the doc agent dies mid-execution? Sketch the minimal `target = ready|busy|disconnected` machine with in-flight executions.

**6. Who writes the env projection under current-Python?**
Decision 5 wants package management disabled and "environment repair provider-owned," but env fields in RuntimeStateDoc are room-host owned per `policy.rs`. When the workstation's current Python differs from notebook metadata, who writes the `env`/`in_sync` projection the viewer renders, and does that fit a `runtime_peer`'s allowed delta?

PR #3406, opened 2026-06-05
fix(notebook): guard shell projection cache fields

rgbkrk:
## Summary

- replace shell projection cache-key field tripwires with exhaustive field-reader maps
- derive shell/access/auth/runtime/interaction/actor cache keys from those maps so new projection fields must get cache-key readers
- include the top-level shell capabilities projection in the same cache-key coverage path

## Verification

- `cargo xtask lint --fix`
- `pnpm test:run packages/runtimed/tests/notebook-shell-capabilities.test.ts packages/runtimed/tests/projection-cache.test.ts`


rgbkrk:
@pullfrog review

rgbkrk:
@pullfrog review

PR #3408, opened 2026-06-05
Proof (not for merge): runtime_peer kernel host - cloud-driven execution on a remote workstation

rgbkrk:
A working proof that a remote workstation can host a kernel and run cells driven from a cloud notebook room. **Not for merge** - it reimplements the daemon's kernel drive, which the production shape reuses instead. Opened as a permanent record of the diff and the result, then closed.

## What it proves

`runt-cloud-peer --host-kernel --scope runtime_peer` attaches to a hosted room as a `runtime_peer`, launches a local python kernel (`python -m ipykernel_launcher`, driven directly over `jupyter-protocol` + `jupyter-zmq-client`), picks up queued executions the room creates, runs them, and streams `running` + outputs + `done` back into the RuntimeStateDoc.

Verified end to end against live preview, including cross-machine: a cell submitted from one machine ran on a separate Linux workstation attached as the `runtime_peer`, and its output rendered in the cloud viewer (`platform.node()` came back as the remote host). The output path is confirmed too - a plain nbformat manifest via `append_output` persists across peer disconnect and renders in the viewer without the daemon's richer `OutputManifest` shape.

## Why it doesn't merge

`kernel_host.rs` reimplements what the daemon already does well: launch, IOPub to RuntimeStateDoc, lifecycle. The daemon's `JupyterKernel` is welded to daemon-only shared state so it can't be reused directly, but the right fix isn't a second kernel driver. The production shape keeps the daemon as the kernel manager (it owns the env pools, the launcher, supervision) and makes `runtime_agent` transport-agnostic, so a daemon-managed kernel syncs to a cloud room over the same wire. runt-cloud-peer's WS transport (already on main) becomes that cloud sink. This spike retires into that work.

## Gaps this surfaced

Running as a bare peer with no supervisor exposed the lifecycle question the production design must answer. Kernel-event detection (death, hang, idle, error) is transport-independent and survives the cloud swap untouched. But `kernel.lifecycle` is `runtime_peer`-only-writable, so when the runtime itself vanishes, no surviving room participant can correct the doc and the room has no watchdog - a dropped workstation leaves the room showing a phantom-live kernel with cells stuck `running`. The production design needs a cloud-room watchdog that reconciles in-flight executions on runtime departure, plus a narrow policy relaxation (or a `Disconnected` lifecycle the room can stamp). The spike also never polls child exit, so it writes no `Error` on kernel death.

Kept as a closed reference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7437,6 +7437,8 @@ dependencies = [
  "automerge",
  "clap",
  "futures-util",
+ "jupyter-protocol",
+ "jupyter-zmq-client",
  "notebook-doc",
  "notebook-wire",
  "runtime-doc",
--- a/crates/runt-cloud-peer/Cargo.toml
+++ b/crates/runt-cloud-peer/Cargo.toml
@@ -23,3 +23,8 @@ notebook-doc = { path = "../notebook-doc" }
 runtime-doc = { path = "../runtime-doc" }
 automerge = { workspace = true }
 uuid = { version = "1", features = ["v4"] }
+# Kernel hosting: drive a local Jupyter kernel directly over ZMQ (the daemon's
+# JupyterKernel is welded to daemon-only shared state, so we reuse the protocol
+# crates instead). See docs/adr/remote-workstation-doc-agents.md.
+jupyter-protocol = { workspace = true }
+jupyter-zmq-client = { workspace = true, features = ["tokio-runtime", "aws-lc-rs"] }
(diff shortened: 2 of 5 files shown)


rgbkrk:
Closing as a preserved reference. This proved the cloud-driven cross-machine execution path end to end; the production version reuses the daemon as the kernel manager and makes runtime_agent transport-agnostic rather than carrying this second kernel driver. The diff and result stay linkable here.

PR #3409, opened 2026-06-05
refactor(runtimed): extract FrameTransport and port runtime_agent (behavior-preserving)

quillaid:
Makes the daemon's `runtime_agent` transport-agnostic. Today it only speaks the local Unix socket (`Handshake::RuntimeAgent`); this moves the transport behind a trait so a daemon-managed kernel can later sync its RuntimeStateDoc and NotebookDoc to a cloud notebook room over WebSocket. The daemon stays the kernel manager (env pools, launcher, supervision); only the sync sink becomes pluggable. The Unix-socket path is moved behind the trait byte-for-byte, so there's no functional change.

## The seam

New `crates/notebook-protocol/src/connection/transport.rs`:

- `FrameSource::recv_frame()`: read half, cancel-safe, backed by `FramedReader`.
- `FrameSink::send_frame(ty, payload)`: write half, length-preamble typed framing.
- `FrameTransport::connect()`: owns the transport-specific dial and handshake, returns the source/sink pair.
- `UdsFrameTransport`: the daemon's current transport, behavior for behavior.

`runtime_agent.rs` is ported onto the trait. `connect_and_handshake` becomes `UdsFrameTransport::connect`, `reconnect_with_backoff` is generic over `FrameTransport`, and frame send/recv route through the halves. The `select!` loop, queue dispatch, `KernelConnection`, `RuntimeStateHandle`, `BlobStore`, `kernel_state`, and reconnect policy are untouched.

## Why split source and sink instead of one transport

The `select!` loop borrows the read half across the whole `select!` (one arm awaits `recv_frame`) while other arms call `send_frame`. A single `&mut self` transport makes those borrows conflict; the current code only compiles because the reader and writer are separate variables. Two halves preserve that exactly. The traits use generics rather than `dyn` so `notebook-protocol` stays `async-trait`-free and wasm-safe, matching the existing `KernelConnection` pattern.

## Verification

`cargo test -p runtimed` passes (944 tests, including the `tokio_mutex_lint` and `tokio_select_cancel_safe` checks). `cargo clippy` on `runtimed` and `notebook-protocol` is clean under `-D warnings`, `cargo build --workspace` is clean, and `cargo xtask lint` passes. Unix-socket behavior is unchanged.

Rationale is recorded in `docs/adr/remote-workstation-doc-agents.md` and `docs/handoffs/16-decisions-log.md`. This targets `quod/runtime-agent-transport-handoff` (the ADR and handoff docs) so the diff is exactly the refactor; the cloud-WS transport and lifecycle work stack on top.

### DIFF
--- a/crates/notebook-protocol/src/connection.rs
+++ b/crates/notebook-protocol/src/connection.rs
@@ -7,6 +7,7 @@ mod env;
 mod framing;
 mod handshake;
 mod pipe;
+mod transport;
 
 pub use env::{CreateNotebookEnvironmentMode, EnvSource, LaunchSpec, PackageManager};
 
@@ -23,6 +24,10 @@ pub use handshake::{
 
 #[cfg(windows)]
 pub use pipe::{connect_named_pipe_client, is_retryable_named_pipe_connect_error, ERROR_PIPE_BUSY};
+
+pub use transport::{
+    FrameSink, FrameSource, FrameTransport, FramedReaderSource, UdsFrameTransport, WriterFrameSink,
+};
 
 #[cfg(test)]
 mod tests {
--- a/crates/notebook-protocol/src/connection/transport.rs
+++ b/crates/notebook-protocol/src/connection/transport.rs
@@ -1,0 +1,262 @@
+//! Transport abstraction for the runtime-agent sync wire.
+//!
+//! The runtime agent's `select!` loop reads typed frames from a source and
+//! writes typed frames to a sink. Today that wire is the daemon's Unix-domain
+//! socket (a Windows named pipe on Windows) carrying length-preamble framing.
+//! A hosted runtime peer will instead carry the same typed frames over a cloud
+//! WebSocket. The loop, queue dispatch, kernel drive, and RuntimeStateDoc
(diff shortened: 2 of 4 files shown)


quillaid:
**Stacking update:** Phase 2 (cloud-WS `FrameTransport` library, crate `notebook-cloud-transport`) is built and stacked on this branch. Since this branch isn't on upstream yet, the Phase 2 PR is opened within the fork: quillaid/desktop#1 (base `quod/16-frame-transport`). A compile-time assertion in the new crate proves `CloudWsFrameTransport` is drop-in for the generic `FrameTransport` bound this PR introduces. Phase 3 (daemon spawn path + EOF-policy-by-transport split + cloud-room watchdog) will stack next. Decision-log entries 16–19 cover the Phase 2 scoping.

rgbkrk:
Superseded by #3411, which consolidates Phases 1 through 3a (the FrameTransport extraction here, plus the cloud-WS transport and the transport-aware clean-EOF policy) into one branch off `main`, authored under my handle and reviewed there. This branch was Phase 1 only, from the fork, targeting the handoff docs branch. Closing in favor of the consolidated PR.

PR #3411, opened 2026-06-05
feat(runtimed): transport-agnostic runtime_agent + cloud-WS FrameTransport

rgbkrk:
Makes the daemon's `runtime_agent` transport-agnostic and adds a cloud WebSocket transport, so a daemon-managed kernel can sync its RuntimeStateDoc and NotebookDoc to a cloud notebook room, not only over the local Unix socket. The daemon stays the kernel manager (env pools, launcher, supervision); only the sync sink becomes pluggable. This is the foundation for a workstation hosting compute for a hosted viewer.

## What's here

**Transport-agnostic agent.** A `FrameTransport` trait (split into `FrameSource`/`FrameSink` read and write halves) lands in `notebook-protocol`, with the existing Unix-socket path moved behind it byte-for-byte (`UdsFrameTransport`). `runtime_agent.rs` is ported onto the trait: `connect_and_handshake` becomes `UdsFrameTransport::connect`, `reconnect_with_backoff` is generic over the transport, and frame send/recv route through the halves. The `select!` loop, queue dispatch, `KernelConnection`, `RuntimeStateHandle`, `BlobStore`, and reconnect policy are untouched. No functional change to the desktop path.

**Cloud-WS transport.** A new `notebook-cloud-transport` crate implements `FrameTransport` over `tokio-tungstenite`: dial, header auth, the typed-frame wire, the `cloud_room_ready` ready-wait with pre-ready frame buffering, and `cloud_frame_rejected` surfaced as a connect error. The crate only moves bytes. The consumer-side receive switch (`receive_sync_message_with_changes`, so a cloud peer consumes the room's authoritative state instead of overwriting it) lands with the daemon spawn path in a follow-up, not here.

**Lifecycle safety.** A clean WebSocket close must not be treated like a Unix-socket EOF that tears down a healthy kernel. The clean-EOF policy is now transport-aware, with a reconnect floor between cycles so a flapping cloud link can't spin a reconnect storm.

## Why split source and sink

The agent's `select!` borrows the read half across the whole `select!` while other arms call send. A single `&mut self` transport makes those borrows conflict; two halves preserve the existing structure exactly. The traits use generics rather than `dyn` to keep `notebook-protocol` `async-trait`-free and wasm-safe, matching the existing `KernelConnection` pattern.

## Deferred

The daemon spawn path (allocating a cloud-attached runtime in env X for room Y with the right doc-actor identity) is not here. It's the most behaviorally entangled piece and its verification needs staging credentials and a live room, so it lands in a follow-up alongside the live cross-machine re-proof. `docs/handoffs/16-lifecycle-analysis.md` maps the remaining safety-net work: a cloud-room watchdog and the `kernel.lifecycle` policy deadlock when a runtime peer departs.

## Verification

Each phase was verified with `cargo test -p runtimed` (944 passing, including the `tokio_mutex_lint` and `tokio_select_cancel_safe` checks) plus `cargo clippy` under `-D warnings` and `cargo xtask lint`. Unix-socket behavior is unchanged. CI gates the rest here. Design rationale is recorded in `docs/adr/remote-workstation-doc-agents.md` and `docs/handoffs/16-decisions-log.md`.


rgbkrk:
@pullfrog please review this against the daemon agent in `crates/runtimed/src/runtime_agent.rs`, the new transport seam in `crates/notebook-protocol/src/connection/transport.rs`, and the cloud transport in `crates/notebook-cloud-transport/src/lib.rs`. I don't need a general pass. I want judgment on these specific points, in priority order.

1. **Behavior-preservation of the UDS path.** The refactor splits the transport into `FrameSource` (recv) + `FrameSink` (send) halves behind a generic `FrameTransport`, and ports `runtime_agent` onto it. Is the `select!` borrow structure preserved exactly? Does anything about the UDS dial, handshake, framing, or reconnect change behavior versus before the extraction? This is meant to be a zero-functional-change refactor.

2. **Consumer-side receive correctness.** The cloud transport consumes the room's authoritative state with `receive_sync_message_with_changes`, not the daemon-authoritative `receive_sync_message` (which strips incoming changes). Is the consumer semantics correct here, and is there any frame or state where a cloud peer should *not* accept incoming changes?

3. **The clean-EOF policy split.** `clean_eof_is_recoverable()` defaults `false` (UDS: clean close means "daemon gone, tear down") and is `true` for the cloud transport (clean close means "blip, reconnect"). Does the recoverable arm faithfully mirror the framing-error arm (keep kernel + queue + local doc, reconnect, resync)? Is there any case where a UDS clean EOF now behaves differently than before? Is the 1s reconnect floor (`CLEAN_EOF_RECONNECT_FLOOR`) sufficient against a flapping room, and does it leak into the UDS path?

4. **Cloud connect ready-wait.** Rejection handling (`cloud_frame_rejected` → `PermissionDenied`), pre-`cloud_room_ready` frame buffering, and the principal-mismatch-on-reconnect warning. Any way a rejection or a late control frame hangs the connect or drops frames?

Context and the deferred work (the daemon spawn path and the lifecycle safety net) are in `docs/handoffs/16-decisions-log.md` and `docs/handoffs/16-lifecycle-analysis.md`.

PR #3413, opened 2026-06-05
fix(cloud-transport,runtimed): bound cloud ready-wait + extend reconnect flap floor (Phase 3b warm-up)

quillaid:
Based on `main` (#3411 merged 2026-06-05). Closes the two connect-hardening gaps pullfrog flagged on #3411, as the Phase 3b warm-up. Pure Rust in `notebook-cloud-transport` + `runtime_agent`; no creds, no deploy.

## 1. Bounded cloud ready-wait timeout (`notebook-cloud-transport`)

`CloudWsFrameTransport::connect`'s `cloud_room_ready` wait blocked on `recv_frame()` with no deadline. A room that completes the WS upgrade but never sends `cloud_room_ready`/`cloud_frame_rejected` and never closes would hang `connect()` forever — and since `reconnect_with_backoff` calls `connect()` with no per-attempt timeout of its own, the whole recovery path wedges on one silent room.

- Ready loop extracted into `wait_for_ready<S: FrameSource>` (no deadline of its own → unit-testable against a mock source), wrapped in `tokio::time::timeout(READY_WAIT_TIMEOUT)` (30s).
- On elapse, returns a `TimedOut` io error, which `reconnect_with_backoff` already treats as a failed connect and retries.
- Also fixes a latent buffering bug: the old loop pushed pre-ready data frames into `source.pending` while reading via `source.recv_frame()` (which pops `pending` first). Now uses a separate deque, assigned into `pending` only after ready.

## 2. Flap floor on the framing-error reconnect arm (`runtimed`)

Phase 3a added the 1s reconnect floor only to the clean-EOF (`None`) arm. The framing-error (`Some(Err)`) arm + `reconnect_with_backoff` only sleep between *failed* connects, so a cloud room that accepts the connection then errors the stream every cycle spins reconnects at RTT rate (self-inflicted DoS on the room).

- New pure helper `reconnect_floor_delay(recoverable, since_last_reconnect, floor) -> Option<Duration>`, called by **both** reconnect arms.
- Timestamp renamed `last_recoverable_reconnect`; `CLEAN_EOF_RECONNECT_FLOOR` → `RECOVERABLE_RECONNECT_FLOOR`.
- The helper returns `None` immediately when `!recoverable`, so the **UDS framing-error path is byte-for-byte unchanged** (no sleep, no timestamp).

## STATUS

- **State:** Complete. Both gaps fixed + unit-tested.
- **Tests:** `cargo test -p notebook-cloud-transport` → 16 (was 12). `cargo test -p runtimed` → 881 lib + integration suites green (UDS path unchanged). 8 new unit tests total.
- **Lint:** `cargo xtask lint --fix` clean; clippy `-D warnings` clean on both crates.
- **Bump targets:** No new workspace crate → no `xtask` bump-targets change.
- **NEEDS US:** None for this PR — fully headless, no live room / deploy / creds required. (The broader Phase 3c live cross-machine re-proof and 3d worker deploy remain `NEEDS US`, tracked in the decision log; they are not part of this PR.)
- **Decision log:** entries #24, #25 in `docs/handoffs/16-decisions-log.md`.

### DIFF
--- a/crates/notebook-cloud-transport/Cargo.toml
+++ b/crates/notebook-cloud-transport/Cargo.toml
@@ -20,4 +20,4 @@ tokio-tungstenite = { version = "0.26", features = ["native-tls"] }
 tracing = "0.1"
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"] }
--- a/crates/notebook-cloud-transport/src/lib.rs
+++ b/crates/notebook-cloud-transport/src/lib.rs
@@ -41,6 +41,7 @@
 
 use std::collections::VecDeque;
 use std::sync::Arc;
+use std::time::Duration;
 
 use futures_util::stream::{SplitSink, SplitStream};
 use futures_util::{SinkExt, StreamExt};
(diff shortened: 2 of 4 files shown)


rgbkrk:

PR #3414, opened 2026-06-05
feat(cloud-transport): reconnect re-auth seam + reconnect_with_backoff tests (Phase 3b)

quillaid:
**Stacked on #3413** (Phase 3b warm-up). Until #3413 merges, this PR's diff includes #3413's two commits; it narrows to just the 3b commits below once #3413 lands. Review the last two commits, or wait for #3413 to merge.

Phase 3b of #16: WS reconnect/re-auth + full-resync on cloud reconnect (lifecycle-analysis req #2). The reconnect *mechanism* was already complete after Phase 3a (`reconnect_with_backoff` is generic over `FrameTransport`; both reconnect arms reset the sync state and fire the transport-agnostic resync kick `state_kick_tx`, so a full RuntimeStateDoc re-send happens on every recoverable reconnect, cloud included). So 3b is two concrete deliverables, not a rewrite:

## 1. `TokenRefresher` re-auth seam (`notebook-cloud-transport`)

`CloudAuth` holds a *static* token — fine for a short session, but a long-lived peer outlives its OIDC token's expiry, and a reconnect that re-presents an expired token is rejected at the upgrade, so `reconnect_with_backoff` burns all its attempts and gives up.

- `CloudWsFrameTransport::with_token_refresher(config, refresher)`: an `async` closure (`TokenRefresher = Arc<dyn Fn() -> Future<io::Result<String>>>`) the transport calls **before each connect** via `effective_auth()`.
- The fresh token is re-wrapped in the configured auth variant by `CloudAuth::with_token` (variant + any `Dev` user label preserved).
- The default `new()` keeps the static-token path **byte-for-byte**. Refresher error surfaces as a connect error (kind preserved) → retryable.

## 2. `reconnect_with_backoff` unit tests (`runtimed`)

A mock `FlakyTransport` whose `connect` fails the first N calls then succeeds. Three tests under `start_paused` time: recovers after transient failures; recovers on the final allowed attempt (9 fail → 10th succeeds); gives up after exactly `MAX_ATTEMPTS` (10) with the last error preserved.

## STATUS

- **State:** Complete. Both deliverables implemented + unit-tested.
- **Tests:** `notebook-cloud-transport` 20 (was 16, +4 refresher). `runtimed` 884 lib (was 881, +3 reconnect) + integration suites green; UDS path unchanged.
- **Lint:** `cargo xtask lint --fix` clean; clippy `-D warnings` clean on both crates.
- **Bump targets:** No new workspace crate.
[...]

PR #3417, opened 2026-06-05
feat(runtimed): cloud-transport spawn path + principal-derived doc actor (Phase 3c, code only)

quillaid:
**Stacked on #3414 → #3413.** Until those merge, this PR's diff includes their commits; it narrows to the single 3c commit once they land. Review the last commit.

Phase 3c (**CODE ONLY**) of #16: the daemon spawn path that runs the *real* `runtime_agent` over a cloud-WebSocket transport instead of the UDS socket.

## What this adds

- `run_cloud_runtime_agent(config, operator, blob_root)` — builds a `CloudWsFrameTransport` and runs the daemon's existing agent against a hosted room.
- A generic inner `run_runtime_agent_on_transport<T, F>` that **both** entry points funnel through. The kernel drive, queue, RuntimeStateDoc writes, lifecycle handling, and reconnect/resync are identical across transports.
- The only per-transport difference — the doc-actor label — is localised in a `resolve_actor: FnOnce(&T) -> Result<String>` closure run **after** connect (a cloud room's principal is only known from `cloud_room_ready`).
  - **UDS** passes `|_| Ok(runtime_agent_id)` — the daemon-supplied id verbatim, so the desktop/daemon path is **byte-for-byte unchanged**.
  - **Cloud** resolves `<principal>/<operator>:<nonce>` via `cloud_actor_label`. The principal satisfies the room's `validate_room_notebook_change_actors` (decision #6); the nonce avoids the `(actor, seq 1)` → `DuplicateSeqNumber` collision the spike documented; the resolver errors rather than authoring under a bogus actor when no principal is present.

## Correction to the planned 3c surface

The original plan had a premise (b): swap the agent's "stripping" receive for `receive_sync_message_with_changes`. **That was a misreading and is dropped.** The agent's receive path (`receive_sync_and_foreign_comms_recovering`) already calls **raw** `receive_sync_message` (applies all changes) and computes a foreign-comms fork purely for kernel-echo suppression — the doc comment states "The main doc still absorbs every applied change". It is *already* consumer-side; swapping to plain `receive_sync_message_with_changes` would **lose** the `rt:kernel:` echo filter (a regression). So 3c reduces to the doc-actor identity (premise a). Full reasoning in decision log #29.

## STATUS

- **State:** CODE complete + unit-tested behind the transport gate. UDS path byte-for-byte unchanged.
- **Tests:** `runtimed` 888 lib (was 884, +4: label format/uniqueness, no-principal error, UDS pass-through) + integration suites green. `xtask` 28 (bump-targets guard green; `notebook-cloud-transport` already registered, now also a `runtimed` dep).
- **NEEDS US (live re-proof — cannot run headlessly):** Nothing calls `run_cloud_runtime_agent` from a CLI yet, and it has never run against a real room. An interactive session with creds must: (1) deploy/obtain a preview room + grant the agent's principal the explicit `runtime_peer` ACL row (decision #9); (2) wire a CLI subcommand + `CloudAuth` source (trivial, omitted here as unverifiable headlessly); (3) spawn as `runtime_peer` and confirm a cloud-submitted cell runs on the daemon kernel and renders in the viewer, through the *real* agent (not the spike); (4) exercise the 3b token-refresher against an actually-expiring token. The path is additive + gated; it cannot affect the desktop/daemon path.
- **Decision log:** entries #29–#31 in `docs/handoffs/16-decisions-log.md`.

### DIFF
--- a/crates/notebook-cloud-transport/Cargo.toml
+++ b/crates/notebook-cloud-transport/Cargo.toml
@@ -20,4 +20,4 @@ tokio-tungstenite = { version = "0.26", features = ["native-tls"] }
 tracing = "0.1"
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"] }
--- a/crates/notebook-cloud-transport/src/lib.rs
+++ b/crates/notebook-cloud-transport/src/lib.rs
@@ -86,6 +86,40 @@ pub enum CloudAuth {
     Dev { token: String, user: String },
 }
 
+impl CloudAuth {
+    /// Return a copy of this auth with its credential replaced by `token`,
+    /// preserving the variant (and the `Dev` user label). Used by the
+    /// token-refresher path so a reconnect re-auths with a freshly-minted token
+    /// without changing the auth *kind* the room expects.
+    fn with_token(&self, token: String) -> Self {
+        match self {
+            CloudAuth::OidcBearer { .. } => CloudAuth::OidcBearer { token },
+            CloudAuth::AnacondaApiKey { .. } => CloudAuth::AnacondaApiKey { token },
+            CloudAuth::Dev { user, .. } => CloudAuth::Dev {
+                token,
+                user: user.clone(),
+            },
+        }
+    }
+}
+
+/// A closure the transport calls *before each connect* to obtain a fresh
+/// credential, replacing the static token in [`CloudWsConfig::auth`].
+///
+/// The cloud auth token is static for a single short session today, but a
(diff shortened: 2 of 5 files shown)


quillaid:
CI note: the only failing check is **Browser E2E (Playwright)**, a frontend browser job — the same nondeterministic flake seen on #3414. This PR is Rust-only (`runtimed` cloud spawn path + tests); the identical-base #3413 and the sibling #3418 both passed the same Playwright job, so this is not caused by the change. All Rust checks (Clippy, Linux Rust Tests, Lint & Format, runtimed-node) are green. I lack rerun permission on this fork PR; a maintainer rerun or the next push should clear it.

rgbkrk:

PR #3418, opened 2026-06-05
feat(runtime-doc): kernel.last_seen liveness heartbeat (Phase 3e model half)

quillaid:
**Stacked on #3417 → #3414 → #3413.** Until those merge, this PR's diff includes their commits; it narrows to the single 3e commit once they land. Review the last commit.

Phase 3e (model half) of #16: a `last_seen` liveness heartbeat on `KernelState`. This is the **fully-headless** half of lifecycle-analysis requirement #4(b) — the model field the Phase 3d room watchdog reads to detect a stale `runtime_peer`, without a new terminal `RuntimeLifecycle` variant.

## What this adds

- `KernelState.last_seen: Option<String>` — ISO-8601 timestamp the runtime peer was last observed present on the sync link.
- `RuntimeStateDoc::set_last_seen(Option<&str>)` + `read_state` wiring.

## Design

- **Not in the frozen genesis scaffold** → a fresh doc reads `None`, the `runtime_state_genesis_artifact_matches_scaffold` test stays green, no schema-version bump. Additive on the serde model (`#[serde(default)]`), so `..Default::default()` call sites in `runtimed-node`/`runtimed-py` are untouched.
- **Clears to CRDT `Null` → reads back `None`** (like `last_saved`), not the empty-string "scaffolded but unset" convention — a liveness field has no meaningful `""`.
- **Idempotent set** (no head churn) — a watchdog/peer stamps it frequently and must not trigger spurious sync rounds.
- **No policy change needed**: `last_seen` lives under `state.kernel`, so the existing kernel-ownership rules already make it `runtime_peer`-writable and editor-blocked (pinned by a test).

## Reordering + scope calls (decision log #32–#33)

3e is reordered **ahead of** 3d and **split**, after two findings:

1. **The policy relaxation is NOT on 3d's critical path.** `validate_runtime_state_sync_scope` (and its `policy.rs:403-405` deadlock) gates only *incoming peer sync frames* (`runtimed-wasm/src/lib.rs:715`). A DO-internal `alarm()` watchdog authoring *directly* on the room host's `state_doc` — the recommended 3d design — bypasses it. So the relaxation matters only for an Owner-peer-push design, which 3d should avoid.
2. **A `RuntimeLifecycle::Disconnected` variant has a ~28-Rust-file + TS blast radius** and its payoff is viewer-facing UX, verifiable only against a deployed viewer. The analysis sanctions `last_seen` as the alternative form of req #4(b); that's the low-risk headless piece.

## STATUS

- **State:** Model half complete + unit-tested.
- **Tests:** `runtime-doc` 215 (was 211, +5: default-None, roundtrip/clear, idempotent-no-head-churn, runtime_peer-writable/editor-blocked). Genesis-artifact match test green. Dependent crates (`runtimed` 888, `notebook-sync`, `runtimed-py`, `runtimed-node`) green.
- **NEEDS US (deferred to land with 3d under live verification):** (1) the `RuntimeLifecycle::Disconnected` variant — viewer UX validated against a live viewer (the watchdog can express "stale" via `last_seen` + existing `Error`/`Shutdown` in the meantime); (2) the `policy.rs:403-405` relaxation — only if 3d is built as Owner-peer-push rather than a DO-internal watchdog.
- **Decision log:** entries #32–#33 in `docs/handoffs/16-decisions-log.md`.

### DIFF
--- a/crates/notebook-cloud-transport/Cargo.toml
+++ b/crates/notebook-cloud-transport/Cargo.toml
@@ -20,4 +20,4 @@ tokio-tungstenite = { version = "0.26", features = ["native-tls"] }
 tracing = "0.1"
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"] }
--- a/crates/notebook-cloud-transport/src/lib.rs
+++ b/crates/notebook-cloud-transport/src/lib.rs
@@ -86,6 +86,40 @@ pub enum CloudAuth {
     Dev { token: String, user: String },
 }
 
+impl CloudAuth {
+    /// Return a copy of this auth with its credential replaced by `token`,
+    /// preserving the variant (and the `Dev` user label). Used by the
+    /// token-refresher path so a reconnect re-auths with a freshly-minted token
+    /// without changing the auth *kind* the room expects.
+    fn with_token(&self, token: String) -> Self {
+        match self {
+            CloudAuth::OidcBearer { .. } => CloudAuth::OidcBearer { token },
+            CloudAuth::AnacondaApiKey { .. } => CloudAuth::AnacondaApiKey { token },
+            CloudAuth::Dev { user, .. } => CloudAuth::Dev {
+                token,
+                user: user.clone(),
+            },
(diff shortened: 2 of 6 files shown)


rgbkrk:

PR #3419, opened 2026-06-05
fix(runtimed): writer-error reconnects instead of killing the kernel on a recoverable transport (Phase 3f req #6)

quillaid:
**Stacked on #3418 → #3417 → #3414 → #3413.** Until those merge, this PR's diff includes their commits; it narrows to the single 3f commit once they land. Review the last commit.

Phase 3f **req #6** of #16: a writer send error must not tear down a healthy kernel on a recoverable (cloud) transport.

## The gap

After Phase 3a fixed the *read*-side EOF/framing arms, the two **outbound (writer)** `select!` arms were the last place a transient cloud blip still destroys a healthy kernel:
- the `state_changed_rx` arm sending a `RuntimeStateSync` frame, and
- the `async_response_rx` arm sending a reply envelope

both `break` the loop (→ `kernel.shutdown()`) on a single `send_frame` error.

## The fix

New writer-only helper `reconnect_after_writer_error` — applies the shared flap floor, reconnects with backoff, stamps `last_recoverable_reconnect`. Both writer arms branch on `transport.clean_eof_is_recoverable()`:
- **`false` (UDS)** → `break` exactly as before — **byte-for-byte unchanged**.
- **`true` (cloud)** → drop source, reconnect, reset `coordinator_sync_state`, kick `state_kick_tx`, `continue`.

**No explicit egress buffer.** The failed `RuntimeStateSync` delta is durable in the local doc and re-sent by the fresh sync state + resync kick on reconnect — **Automerge's own resync is the replay**, the same mechanism the framing-error arm has always used. The one-shot async reply envelope is lost across reconnect, but its *effect* (env progress/state from `SyncEnvironment`) lives in the RuntimeStateDoc and does resync; only the reply is dropped (documented at the site). A dedicated buffer was considered and rejected as redundant + a new source of ordering bugs.

## STATUS

- **State:** req #6 complete + unit-tested. Read arms left byte-for-byte unchanged.
- **Tests:** `runtimed` 891 lib (was 888, +3: recover+stamp, give-up propagation, the UDS discriminant the loop branches on) + integration suites green.
- **NEEDS US (3f req #5):** the inbound request channel (interrupt/restart `RuntimeAgentRequest`s) needs the Phase 3d DurableObject hosted REQUEST dispatch to deliver triggers from the room to the agent. The kernel-side *effect* already survives; only the trigger path is missing, and it can't be built/verified without the worker + a live room. Build alongside 3d.
- **Decision log:** entries #34–#35 in `docs/handoffs/16-decisions-log.md`.

### DIFF
--- a/crates/notebook-cloud-transport/Cargo.toml
+++ b/crates/notebook-cloud-transport/Cargo.toml
@@ -20,4 +20,4 @@ tokio-tungstenite = { version = "0.26", features = ["native-tls"] }
 tracing = "0.1"
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "test-util"] }
--- a/crates/notebook-cloud-transport/src/lib.rs
+++ b/crates/notebook-cloud-transport/src/lib.rs
@@ -86,6 +86,40 @@ pub enum CloudAuth {
     Dev { token: String, user: String },
 }
 
+impl CloudAuth {
+    /// Return a copy of this auth with its credential replaced by `token`,
+    /// preserving the variant (and the `Dev` user label). Used by the
+    /// token-refresher path so a reconnect re-auths with a freshly-minted token
+    /// without changing the auth *kind* the room expects.
+    fn with_token(&self, token: String) -> Self {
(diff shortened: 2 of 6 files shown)


rgbkrk:

PR #3423, opened 2026-06-05
fix(runtimed): type captured environment disk state

quillaid:
## Summary
- introduce typed captured environment disk states: Missing, Partial, Usable
- route source overrides, launch-config captured env selection, and eviction preservation through typed disk state instead of unified_env_on_disk().is_some()
- keep Missing on the historical fallback path while Partial and Usable enter captured/prewarmed handling
- mark CEL-1 done in the ADR cleanup punchlist

## Tests
- cargo xtask wasm
- cargo test -p runtimed captured_env -- --nocapture
- cargo test -p runtimed launch_kernel -- --nocapture
- cargo xtask lint --fix

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/metadata.rs
+++ b/crates/runtimed/src/notebook_sync_server/metadata.rs
@@ -1647,25 +1647,62 @@ pub(crate) fn captured_env_for_runtime(
     }
 }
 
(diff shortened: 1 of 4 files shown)


rgbkrk (approved):

rgbkrk:
@pullfrog anything to be wary of here?

PR #3426, opened 2026-06-06
feat(runtimed): cloud runtime lifecycle (reconnect/re-auth, spawn path, peer-gone watchdog)

rgbkrk:
Hardens the cloud runtime path built in #3411 (transport-agnostic `runtime_agent` + cloud-WS transport) into a survivable lifecycle. Where #3411 made the daemon's agent able to sync to a cloud room, this makes that path reconnect, re-auth, spawn correctly, and recover when the workstation or the room drops. The desktop/UDS path is byte-for-byte unchanged throughout: every behavioral change is gated on the transport being the recoverable cloud one.

## What's here

- **Connect hardening.** The `cloud_room_ready` wait is now bounded by a timeout (a room that upgrades the WS but never sends ready/rejected no longer hangs the connect, and therefore the whole reconnect loop). The 1s reconnect flap floor added in 3a now also covers the framing-error arm, so a room that accepts then immediately errors the stream can't spin reconnects at RTT rate. (Both were latent gaps flagged on #3411.)
- **Reconnect re-auth.** An optional `TokenRefresher` seam lets the cloud transport mint a fresh token on reconnect, for sessions that outlive the original token. The static-token path is unchanged. `reconnect_with_backoff` is now covered by a mock-`FrameTransport` test (fail N times, then succeed).
- **Spawn path (code only).** `run_cloud_runtime_agent` builds a `CloudWsFrameTransport`-backed agent: it authors the doc actor as `<principal>/<operator>` derived from `cloud_room_ready` (so the room's actor validation accepts its changes) and uses consumer-side `receive_sync_message_with_changes`. All gated on a transport-kind discriminant; the UDS spawn path is untouched.
- **Liveness heartbeat.** A `kernel.last_seen` timestamp on `RuntimeStateDoc`, additive, so a viewer can distinguish a recently-live kernel from a stale one.
- **Writer-error recovery (req 6).** A writer error on a recoverable transport now reconnects instead of tearing down a healthy kernel. The dropped delta is durable in the local doc and Automerge's own resync re-sends it, the same replay the framing-error arm already relies on; no separate egress buffer.
- **Room watchdog (code only).** A DurableObject `runtime_peer`-gone watchdog: when the last runtime peer leaves, an `alarm()` armed `RUNTIME_PEER_GONE_GRACE_MS` (30s) out reconciles orphaned state via `RoomHostHandle::reconcile_runtime_peer_gone`, marking in-flight executions failed and flipping a still-live kernel to `Error` (already-terminal states are left alone, so a clean shutdown racing the disconnect isn't relabeled). A runtime peer rejoining inside the grace disarms it. This is the safety net the daemon structurally cannot provide, since the daemon dying is the trigger. It authors directly on the room-host `state_doc`, which doesn't pass through the runtime-state scope validation, so it needs no policy relaxation.

## Verification

Each piece is unit-tested: `cargo test -p runtimed` (891 lib + integration), `-p notebook-cloud-transport`, `-p runtime-doc`, `-p runtimed-wasm` (+4 reconciliation), and the notebook-cloud node tests (+3 watchdog). Clippy `-D warnings` and `cargo xtask lint` clean. The desktop/UDS path is unchanged and its tests still pass. CI is green aside from a Browser E2E flake unrelated to these Rust/TS-cloud changes.

## Deferred (needs a deployed worker + a live room)

Out of scope for this code-review merge, and documented per phase in `docs/handoffs/16-decisions-log.md`:

- Live cross-machine re-proof of the spawn path and the reconnect re-auth.
- Deploying the watchdog worker and the live peer-drop re-proof, plus tuning the 30s grace against real reconnect latencies.
- The `RuntimeLifecycle::Disconnected` variant (large blast radius, viewer-facing, only verifiable against a deployed viewer); `last_seen` is the lighter form that lands here.
- The inbound interrupt/restart request channel (req 5), which needs the watchdog worker's hosted REQUEST dispatch.

Design rationale: `docs/handoffs/16-decisions-log.md` and `docs/handoffs/16-lifecycle-analysis.md`. Builds on #3411.

### DIFF
--- a/apps/notebook-cloud/src/cloudflare-types.ts
+++ b/apps/notebook-cloud/src/cloudflare-types.ts
@@ -62,6 +62,12 @@ export interface DurableObjectStorage {
     limit?: number;
     reverse?: boolean;
   }): Promise<Map<string, T>>;
+  // Alarm API (Cloudflare-provided). Optional here so the existing fake
+  // storage in tests need not implement it; the runtime_peer-gone watchdog
+  // feature-detects these before arming.
+  setAlarm?(scheduledTime: number | Date): Promise<void>;
+  getAlarm?(): Promise<number | null>;
+  deleteAlarm?(): Promise<void>;
 }
 
 export interface D1Database {
--- a/apps/notebook-cloud/src/notebook-room.ts
+++ b/apps/notebook-cloud/src/notebook-room.ts
@@ -52,6 +52,18 @@ interface PeerAttachment {
 
 const MAX_STORED_FRAMES = 500;
 
+/// Grace window after the last `runtime_peer` leaves before the room reconciles
+/// its in-flight state (lifecycle-analysis reqs #3/#7). Tolerates a transient
+/// disconnect/reconnect (idle eviction, network blip) without terminalizing a
+/// kernel that is about to come back: if a `runtime_peer` rejoins inside the
+/// window the alarm is disarmed.
+const RUNTIME_PEER_GONE_GRACE_MS = 30_000;
+
+/// Storage key holding the notebook id whose `runtime_peer` departure armed the
+/// reconciliation alarm. Persisted so a DO that hibernates between the alarm
+/// being set and firing still knows which room to reconcile.
+const RUNTIME_PEER_WATCH_KEY = "runtime_peer_gone_watch";
+
 export class NotebookRoom {
   private readonly peers = new Map<string, Peer>();
   private readonly pendingRemovals = new Map<string, { notebookId: string; peer: Peer }>();
@@ -99,6 +111,11 @@ export class NotebookRoom {
 
     this.acceptPeerSocket(notebookId, peer);
     this.peers.set(peer.id, peer);
+    // A runtime_peer (re)joining cancels any pending reconciliation alarm: the
+    // kernel host is back, so its in-flight state must not be terminalized.
+    if (identity.scope === "runtime_peer") {
+      this.refreshRuntimePeerWatch(notebookId);
+    }
     cloudLog("info", "room.connection.accepted", {
       notebook_id: notebookId,
       peer_id: peer.id,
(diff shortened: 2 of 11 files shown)


rgbkrk:
@pullfrog please review this against the daemon agent in `crates/runtimed/src/runtime_agent.rs`, the cloud transport in `crates/notebook-cloud-transport/src/lib.rs`, the room host in `crates/runtimed-wasm/src/lib.rs`, and the DurableObject in `apps/notebook-cloud/src/notebook-room.ts`. Most of this is "code only" (the spawn path and the watchdog were never live-verified, that's deferred), so this review is the main verification they get. I don't need a general pass; judgment on these specific points, in priority order:

1. **Is the desktop/UDS path truly untouched?** Every behavioral change is supposed to be gated on the recoverable cloud transport. Verify the UDS path is byte-for-byte unchanged across: the spawn path (`run_cloud_runtime_agent` vs the existing UDS spawn), the writer-error arm, the clean-EOF policy, and the reconnect flap floor (`last_clean_reconnect` should stay `None` on UDS).

2. **Spawn path doc-actor + receive (Phase 3c).** The cloud agent authors the doc actor as `<principal>/<operator>` derived from `cloud_room_ready`, and uses consumer-side `receive_sync_message_with_changes`. Confirm the actor derivation is correct (the room's `validate_room_notebook_change_actors` must accept it) and that consumer-receive is applied only on the cloud path, not the UDS path.

3. **Writer-error recovery relies on Automerge resync as the replay (req #6).** No dedicated egress buffer. Confirm a `RuntimeStateSync` delta written while the writer errors is actually re-sent on reconnect (durable in the local doc + the resync kick), and that no terminal state is silently lost beyond the documented one-shot reply envelope.

4. **Watchdog correctness (Phase 3d).** `reconcile_runtime_peer_gone` authors directly on the room-host `state_doc`, claimed to bypass `validate_runtime_state_sync_scope` so no policy relaxation is needed, verify that's actually true and safe. Confirm already-terminal `Error`/`Shutdown`/`NotStarted` states are left untouched (clean-shutdown-race), the 30s `alarm()` grace + rejoin-disarm logic is correct, and the alarm feature-detect leaves alarm-less storage backends unaffected.

5. **`kernel.last_seen` (Phase 3e) is additive** and doesn't break existing `RuntimeStateDoc` serialization or policy.

Deferred live/deploy work is documented in `docs/handoffs/16-decisions-log.md` and is intentionally out of scope here. Builds on #3411.

PR #3428, opened 2026-06-06
feat(runtimed): cloud-runtime-agent CLI subcommand (workstation endpoint, Phase B1)

quillaid:
## STATUS — Phase B1: cloud-runtime-agent CLI (headless, complete)

Stacked on #3427 (Phase A scope). Closes the missing-caller gap for
`run_cloud_runtime_agent` (merged in #3426, which had no non-test caller —
decision 31).

### What's here
- A hidden `runtimed cloud-runtime-agent` subcommand mirroring the internal
  `runtime-agent` subcommand. It maps non-secret routing flags (`--cloud-url`,
  `--notebook-id`, `--scope`, `--operator`, `--auth-kind`) plus an
  environment-supplied credential to a `CloudWsConfig` + `CloudAuth` and drives
  `run_cloud_runtime_agent`.
- `runtimed::workstation::cloud_agent_cli` — the pure arg/env → config mapping,
  unit-tested.
- Cargo.lock fix: adds the `notebook-cloud-transport` entry under `runtimed`'s
  deps. main's lockfile was missing it (pre-existing drift from the #3426 merge —
  the Cargo.toml dep is present); a `--locked` build on main would fail without
  this.

### Security
The credential is read from `RUNT_CLOUD_TOKEN` (and `RUNT_CLOUD_DEV_USER` for
`dev` auth), **never argv**, per the ADR security constraint — no API keys in a
long-running process's command line / `ps` output. Flags carry only non-secret
routing and the auth *kind*.

### Tests
- 6 new lib tests: each auth-kind → expected `CloudAuth` variant; missing /
  blank / whitespace token rejected with an error naming the env var; token
  trimmed; `dev` requires a user label.
- `cargo test -p runtimed --lib` green; clippy `-D warnings` clean;
  `cargo xtask lint --fix` clean. Verified `--help` parses.

### Deferred — needs us
The actual dial against a real preview room (staging creds + deployed worker +
the `runtime_peer` ACL row, decision 9) is **not attempted headlessly**. Exact
run steps in `docs/handoffs/16-workstation-endpoint.md` ("Deferred — needs us").

PR #3429, opened 2026-06-06
feat(runtimed): list_environments workstation projection (Phase B2)

quillaid:
## STATUS — Phase B2: list_environments (headless, complete)

Stacked on #3428 (Phase B1 CLI). Implements the first endpoint operation: a
workstation *lists the environments it has*.

### What's here
- `WorkstationEnvironment` + `environments_from_pool_state(&PoolState)` — a pure
  projection over the daemon's existing pool state (`PoolDoc::read_state`),
  **not** a new enumerator (decision 39). One environment per env kind that
  offers something, carrying available/warming counts, the `environment_policy`
  (prewarmed pools are `ManagedProject`), and the pool's health note passed
  through so a failing pool is visible rather than silently offered.
- `list_environments(&Daemon)` — thin async wrapper reading the live pool doc.

The shape is kept stable so it can later back both the hosted
workstation-target API and the Content-rail catalog (ADR decision 8).

### Tests
4 new lib tests (stable-order projection; empty offers nothing; unhealthy pool
surfaced even at zero availability; target-only is offered). `cargo test -p
runtimed --lib` green; clippy `-D warnings` clean; `cargo xtask lint --fix` clean.

### Next in the stack
- B3: launch-on-attach in the agent loop (gated, so the endpoint can *start* a
  runtime headlessly while req #5 stays deferred — decision 38).
- B4: `allocate_runtime_for_room` glue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7589,6 +7589,7 @@ dependencies = [
  "loro_fractional_index",
  "nbformat",
  "nix 0.31.2",
+ "notebook-cloud-transport",
  "notebook-doc",
  "notebook-protocol",
  "notebook-sync",
--- a/crates/runtimed/src/lib.rs
+++ b/crates/runtimed/src/lib.rs
@@ -68,6 +68,7 @@ pub mod user_error;
 pub(crate) mod uv_project;
 #[doc(hidden)]
 pub mod warm_env;
+pub mod workstation;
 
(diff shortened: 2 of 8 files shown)


rgbkrk:
Superseded by #3434, which consolidates the full workstation-endpoint stack into one branch off main (squashed, re-authored, #N-scrubbed). Closing in favor of the consolidated PR.

PR #3430, opened 2026-06-06
feat(runtimed): launch-on-attach for the cloud runtime agent (Phase B3)

quillaid:
## STATUS — Phase B3: launch-on-attach (headless, complete)

Stacked on #3429 (B2 list_environments). This is the keystone that lets the
workstation endpoint **start** a runtime, not just attach.

### The problem
The agent's steady state waits for an inbound `RuntimeAgentRequest::LaunchKernel`
before starting a kernel. Over the daemon UDS that comes from the daemon; over
the **cloud** transport it must come from the room — and that inbound channel is
**req #5, Deferred** (needs the 3d worker + a live room). So a cloud agent spawned
today *attaches but is never told to launch*. The ADR says the endpoint
*allocates **and starts*** a runtime in env X (decision 38).

### What's here
- `workstation::launch_on_attach::build_current_python_launch` — builds an
  initial `LaunchKernel` for the ADR's first-class `current_python` policy
  (Decision 5): an explicit interpreter, **no** daemon env pool, package mutation
  disabled, labelled as current Python. Pure + unit-tested.
- `run_cloud_runtime_agent` / `run_runtime_agent_on_transport` gain an optional
  `initial_launch`, applied once after bootstrap through the **same**
  `handle_runtime_agent_request` path an inbound RPC uses (identical kernel drive,
  queue release, command-receiver install — only the trigger differs).

### Safety
The apply is gated on `clean_eof_is_recoverable()` — the recoverable/cloud
discriminant (same gate as decisions 22/35). The UDS/daemon path passes `None`
and would ignore a `Some` regardless, so it is **byte-for-byte unchanged**. The
bare cloud CLI entry passes `None` (attach-only); the launch trigger is wired by
`allocate_runtime_for_room` (B4).

### Tests
4 new lib tests (3 builder: explicit interpreter, no pool/inline deps leak,
env-source label routes to uv family but not Prewarmed; 1 gate discriminant).
`runtimed --lib` 909 green; integration 51 green; `tokio_mutex_lint` +
`tokio_select_cancel_safe` green; clippy `-D warnings` clean; `lint --fix` clean.

### Deferred — needs us
The live kernel launch + attach against a real preview room (staging creds +
deployed worker + `runtime_peer` ACL row). Steps in
`docs/handoffs/16-workstation-endpoint.md`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7589,6 +7589,7 @@ dependencies = [
  "loro_fractional_index",
  "nbformat",
  "nix 0.31.2",
+ "notebook-cloud-transport",
  "notebook-doc",
  "notebook-protocol",
  "notebook-sync",
--- a/crates/runtimed/src/lib.rs
+++ b/crates/runtimed/src/lib.rs
@@ -68,6 +68,7 @@ pub mod user_error;
 pub(crate) mod uv_project;
 #[doc(hidden)]
 pub mod warm_env;
+pub mod workstation;
 
 pub fn trusted_packages_db_path() -> std::path::PathBuf {
     runt_workspace::daemon_base_dir().join("trusted-packages.sqlite")
--- a/crates/runtimed/src/main.rs
+++ b/crates/runtimed/src/main.rs
@@ -469,7 +469,9 @@ async fn main() -> anyhow::Result<()> {
                         eprintln!("[cloud-runtime-agent] Config error: {}", e);
                         e
                     })?;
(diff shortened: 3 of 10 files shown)


rgbkrk:
Superseded by #3434, which consolidates the full workstation-endpoint stack into one branch off main (squashed, re-authored, #N-scrubbed). Closing in favor of the consolidated PR.

PR #3454, opened 2026-06-06
feat(cloud): enable owner execution with runtime peers

quillaid:
## Summary
- surface attached runtime peers to the cloud viewer via room session-control presence
- enable owner-scoped browser execution by sending hosted `execute_cell` requests through `NotebookClient`
- keep runtime availability visible separately from execution authority, with editors/viewers still unable to submit execution intent
- document the interim owner-only policy and the hosted request-ack gap in ADRs

## Stacking
This is stacked on #3452 (`qu/lab2/runtime-peer-execution-pickup`). The new commit is `ca7f3d8d feat(cloud): allow owner execution through runtime peers`; GitHub may show the #3452 commits until that PR lands.

## Verification
- `pnpm --filter notebook-cloud exec node --import tsx --test test/live-sync.test.ts test/viewer-presence.test.ts test/viewer-shell-capabilities.test.ts test/notebook-room.test.ts`
- `pnpm test:run packages/runtimed/tests/notebook-shell-capabilities.test.ts`
- `pnpm --filter notebook-cloud typecheck`
- `cargo xtask lint --fix`

## Reviewer Notes
@pullfrog please pay particular attention to the hosted request path: browser `execute_cell` currently completes on `cloud_frame_accepted` / `cloud_frame_rejected` session controls because the Durable Object room does not yet emit correlated `Response` frames for hosted requests. The ADR calls this out as intentional interim behavior; durable execution completion still comes through `RuntimeStateDoc`.

Also note the authorization shape: controls are owner-only while a runtime peer is attached. This avoids showing execute controls to editor connections that the current materializer would reject, and leaves room for a future explicit execute scope.

PR #3455, opened 2026-06-06
test(cloud): add hosted runtime peer smoke

quillaid:
## Summary
- adds `smoke:hosted:runtime-peer`, an API-key backed preview smoke that creates a private room, grants `runtime_peer`, launches `runtimed cloud-runtime-agent`, runs `runt-cloud-peer` as owner, and asserts `queued -> done` plus kernel execution logs
- keeps API keys out of argv for the owner peer via `--token-file`, validates the chosen Python can import `ipykernel`, exits as soon as terminal state is observed, and writes smoke logs under `.context/smokes/`
- supports detached keep-alive runtime peers for manual browser inspection, and replaces the long manual shell runbook with the package script

## Stack / merge note
- this smoke is intended to become the repeatable acceptance check for the runtime-peer execution stack in #3454
- current `origin/main` binaries do not include #3454, so they can launch a kernel and observe `queued/running` but not the final `done`; the live verification below points the script at #3454-built binaries

## Verification
- `cargo xtask artifacts ensure sift,renderer`
- `node --check apps/notebook-cloud/scripts/hosted-runtime-peer-smoke.mjs`
- `pnpm --filter notebook-cloud typecheck`
- `cargo build --release -p runtimed -p runt-cloud-peer`
- `cargo xtask lint --fix`
- live preview smoke using #3454 binaries:
  `NTERACT_CLOUD_URL=https://preview.runt.run NOTEBOOK_CLOUD_RUNTIMED_BIN=/home/ubuntu/codex/nteract-3454-smoke/target/release/runtimed NOTEBOOK_CLOUD_RUNT_CLOUD_PEER_BIN=/home/ubuntu/codex/nteract-3454-smoke/target/release/runt-cloud-peer pnpm --dir apps/notebook-cloud smoke:hosted:runtime-peer`
  passed with room `01KTFDJHYJRKFVAREG48NTBPNG`, viewer https://preview.runt.run/n/01KTFDJHYJRKFVAREG48NTBPNG/lab2-dualpeer, total 10.3s, owner execution 1.5s
- live keep-alive diagram smoke passed with detached runtime peer PID 202426, room `01KTFDPPPY5PNFGSVQ1CXMWN2X`, viewer https://preview.runt.run/n/01KTFDPPPY5PNFGSVQ1CXMWN2X/lab2-runtime-map-2

PR #3456, opened 2026-06-06
docs(adr): record generated runtime artifact policy

quillaid:
## Summary
- adds a Generated Runtime Artifacts ADR covering LFS-tracked renderer bundles, gitignored WASM/plugin outputs, and the `xtask artifacts` ownership boundary
- records why Rust build scripts should verify embedded generated assets but not run Node/Vite/wasm-pack generation
- documents the fresh-worktree direct runtime build sequence: `cargo xtask artifacts ensure sift,renderer` before `cargo build -p runtimed`

## Verification
- `cargo xtask lint --fix`

## Context
This was split out after a fresh #3454 worktree reproduced the direct-build failure: `runtimed` needs gitignored `sift.js` / `sift.css` and `sift_wasm_bg.wasm` before `cargo build -p runtimed` can succeed. The ADR makes the existing `xtask artifacts` policy durable instead of leaving it in scattered comments.

### DIFF
--- a/docs/adr/README.md
+++ b/docs/adr/README.md
@@ -60,6 +60,7 @@ These entries define the center of gravity for the system:
 | [Cell Execution Pipeline and Control-Plane Separation](execution-pipeline.md) | Draft | Execution state, output ordering, and lifecycle priority. |
 | [Execution Liveness](execution-liveness.md) | Design memo | Divergence-detection framing; not a recovery decision yet. |
 | [Tokio Mutex Discipline](tokio-mutex-discipline.md) | Draft | Async lock and cancel-safety invariants. |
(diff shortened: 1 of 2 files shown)


quillaid:
`pr-review` is currently blocked by the same Bedrock credential issue seen on #3455 (`infra_uncertain`: refreshed AWS credentials are still expired).

@pullfrog please review this ADR PR, especially whether the boundary is right:
- `xtask artifacts` owns generated WASM/renderer/MCP-widget preparation
- Rust build scripts verify embedded generated assets but do not invoke Node/Vite/wasm-pack
- stable renderer bundles stay LFS-tracked, volatile Sift/runtime WASM outputs stay gitignored and fingerprint-verified

PR #3461, opened 2026-06-06
perf(cloud): project live runtime updates incrementally

quillaid:
## Summary

- Stops cloud live `runtimeState$` updates from triggering full notebook view materialization.
- Wires cloud live sync into the same narrow runtime stores desktop uses: runtime state, execution view changes, output id changes, and pool state.
- Lets `applyOutputChangeset` accept a host-provided blob resolver so cloud can resolve manifest-backed output updates through its authenticated HTTP resolver without depending on the desktop blob-port store.

## Why

The cloud viewer was still using full `materializeCloudNotebookView(...)` passes for live runtime changes. For output-heavy notebooks, that means runtime/output updates can re-read and re-resolve the full cell list. This PR keeps initial/full cell materialization for notebook cell changes, but runtime-state churn now projects through stable per-runtime/per-execution/per-output stores.

## Verification

- `pnpm test:run apps/notebook/src/lib/__tests__/project-runtime-stores.test.ts`
- `pnpm --filter notebook-cloud typecheck`
- `pnpm --dir apps/notebook exec tsc -b`
- `cargo xtask lint --fix`

PR #3462, opened 2026-06-06
docs(adr): define live projection policy

quillaid:
## Summary

- Adds a draft ADR for live notebook projection policy.
- Records full materialization as a bounded fallback for bootstrap, snapshots, structural changes, and unclassified changesets.
- Records runtime/execution/output/pool/presence/broadcast updates as stable narrow-store projections, with host-provided blob resolver support as the shared boundary.

## Why

This documents the performance invariant behind the cloud live projection work: output-heavy runtime churn should not cause full cell-list re-read/re-resolution. It also gives future desktop/cloud/Elements projection changes a reviewable policy instead of leaving the rule implicit in adapters.

## Verification

- `cargo xtask lint --fix`


quillaid:
Local verification passed:

- `cargo xtask lint --fix`

`uv run pr-review doctor` is still blocked by expired Bedrock credentials on this machine, so I could not run the structured reviewer.

@pullfrog please review the ADR scope, especially whether the full-materialization fallback boundaries are stated tightly enough and whether the runtime/output hot-path rule matches the direction in #3461.

quillaid:
CI is green for #3462 at `98254d12`: source detection completed and all non-doc jobs were skipped as expected for this ADR-only PR. Local verification was `cargo xtask lint --fix`.

PR #3463, opened 2026-06-06
refactor(runtimed): share snapshot widget comm projection

quillaid:
## Summary

- Moves the pure snapshot widget comm projection helper from notebook-cloud into `packages/runtimed`.
- Exports `snapshotWidgetCommsFromRuntimeAndCommsState`, `resolveSnapshotWidgetComms`, `normalizeSnapshotWidgetComms`, and `widgetCommStoreState` from `runtimed`.
- Leaves `apps/notebook-cloud/src/widget-comms.ts` as a compatibility re-export while cloud viewer/snapshot/test call sites import from `runtimed` directly.

## Why

Live comm projection already lives in `SyncEngine`, but snapshot comm projection was still cloud-local. This removes that app-level duplicate and gives future non-cloud hosts a pure runtime package helper with host-provided blob resolution.

## Verification

- `pnpm test:run packages/runtimed/tests/snapshot-widget-comms.test.ts`
- `pnpm --dir apps/notebook-cloud test test/widget-comms.test.ts test/widget-runtime.test.ts test/snapshot-render.test.ts` (script runs the full notebook-cloud test suite)
- `pnpm --filter notebook-cloud typecheck`
- `cargo xtask lint --fix`

PR #3464, opened 2026-06-06
refactor(runtimed): reuse projection cache helpers

quillaid:
## Summary

- Reuses `projection-cache.ts` from `notebook-actor-projection.ts` instead of keeping local LRU/cache-key helper copies.
- Leaves actor projection behavior otherwise unchanged.

## Why

This closes a low-priority projection audit cleanup: actor projections and other runtimed projections should share the same bounded-cache utility instead of drifting on key and eviction behavior.

## Verification

- `pnpm test:run packages/runtimed/tests/notebook-actor-projection.test.ts packages/runtimed/tests/projection-cache.test.ts`
- `cargo xtask lint --fix`

Note: I also accidentally ran `pnpm test:run packages/runtimed/tests/snapshot-widget-comms.test.ts` while on this `origin/main`-based branch; that file only exists on #3463, so the runner correctly reported no matching files. It is not a failure for this branch.

### DIFF
--- a/packages/runtimed/src/notebook-actor-projection.ts
+++ b/packages/runtimed/src/notebook-actor-projection.ts
@@ -4,6 +4,7 @@ import type {
   NotebookShellAuthCapabilities,
   NotebookShellRuntimeCapabilities,
 } from "./notebook-shell-capabilities";
+import { getBoundedCacheValue, setBoundedCacheValue, stableCacheKey } from "./projection-cache";
 
 export type ParsedNotebookActorKind = "agent" | "runtime" | "system";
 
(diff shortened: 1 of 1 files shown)


quillaid:
Local verification passed:

- `pnpm test:run packages/runtimed/tests/notebook-actor-projection.test.ts packages/runtimed/tests/projection-cache.test.ts`
- `cargo xtask lint --fix`

`uv run pr-review doctor` has been failing with expired Bedrock credentials on this host, so I am using the same review fallback.

@pullfrog please review this small cleanup, especially whether switching actor projection cache keys to the shared `stableCacheKey` helper is acceptable. The shared helper preserves `undefined` explicitly, while the deleted local helper used raw `JSON.stringify`.

PR #3478, opened 2026-06-07
feat(mcp): prefer resources for cell reads

rgbkrk:
## Summary

- Hide legacy `get_cell` and `get_all_cells` from the advertised MCP tool list while keeping them callable for cached clients.
- Surface notebook/cell resource URIs and resource links from session, cell mutation, and execution responses.
- Point MCP server instructions at `nteract://notebooks/{notebook_id}/cells` and `nteract://notebooks/{notebook_id}/cells/{cell_id}` as the read path.

## Verification

- `cargo test -p runt-mcp`
- `cargo xtask lint --fix`

## Notes

- Stable notebook IDs across daemon restarts are left as a follow-up design problem; this PR makes the current read surface resource-addressed without changing identity persistence.


rgbkrk:
@pullfrog review please

PR #3507, opened 2026-06-09
refactor(notebook-cloud): decompose the cloud viewer and rebuild the workspace dashboard

rgbkrk:
The hosted cloud viewer entry, `apps/notebook-cloud/viewer/index.tsx`, had grown to 3,136 lines: routing, three route views, the notebook viewer, auth hooks, presence, workstation state, capability construction, and a pile of helpers in one file. This decomposes it into focused modules, makes the cloud host adapter an explicit seam, rebuilds the workspace dashboard sidebar as a real surface, and lifts the dashboard projection into shared code so the Elements catalog renders the real thing instead of a copy.

`index.tsx` goes from 3,136 lines to a 120-line entry. The decomposition is behavior-preserving; the dashboard sidebar and the shared-projection move are deliberate changes.

## What moved

`index.tsx` now holds only module side-effects, the route switch in `App`, `CloudNotebookProviders`, and the mount. Everything else is a focused module under `viewer/`:

| Module | Holds |
|---|---|
| `cloud-viewer-types.ts` | shared view types |
| `cloud-viewer-config.ts` | config/bootstrap loaders, route path helpers |
| `use-cloud-auth.ts` | `useCloudPrototypeAuth` / `AppSessionStatus` / `AppSessionBridge` |
| `use-cloud-shell-capabilities.ts` | the cloud host adapter (see below) |
| `use-cloud-workstations.ts` | `useCloudWorkstationManager` |
| `home-view.tsx`, `oidc-callback-view.tsx`, `notebook-list-view.tsx` | the three route views |
| `cloud-notebook-dashboard-view.tsx` | dashboard presentation |
| `cloud-presence-status.tsx`, `cloud-notebook-title.tsx`, `cloud-auth-controls.tsx` | leaf chrome |
| `notebook-viewer.tsx` | `NotebookViewer` (~1,000 lines, down from the inline 1,030) |

Workstation state, which was three `useState` slots plus a refresh callback, an attach/default mutation pair, a polling effect, and an attach-completion effect inline in `NotebookViewer`, is now single-owned in `useCloudWorkstationManager`.

## The cloud host adapter is explicit

`NotebookViewer` built `NotebookShellCapabilities` inline. That construction (the edit-access projection, the `canAcceptCellMutations` gate, the shell-capabilities memo) moves into `useCloudShellCapabilities`: the named adapter that turns live hosted-room facts (OIDC auth, room access scope, connection and runtime-peer presence, the workstation-attachment snapshot) into the host-neutral capability object the shared shell consumes. `NotebookViewer` renders `NotebookDocumentShell` / `NotebookDocumentToolbar` / `NotebookDocumentRail` from that output. Desktop and Elements build the same capability object from their own adapters; this is the cloud edge of a shared contract, now visible as one hook instead of buried in the component.

## Dashboard rebuild

The `/n` dashboard aside was two blocks of static prose ("Workstations" and "Public previews"). It is now driven by the notebook list the page already loads:

- **Published previews**: the published notebooks (`latest_revision_id`), linked, with an empty state.
- **Access**: an owned / editable / view-only role breakdown.

`projectCloudNotebookDashboard` returns a `sidebar` projection (unit-tested). The aside becomes a real right rail (it was a full-width band pinned above the list via `order: -1`) and stacks below the list on narrow widths. On `main` the grid was single-column at every width, so this both adds the desktop right rail and puts the notebook list first when stacked.

The hosted viewer's teal/green accent (the shared notebook tokens are deliberately monochrome) moves into named `--cloud-accent` tokens, with the raw hex replaced throughout `index.css`, so the accent is themeable in one place. The off-scale `font-weight: 720` display weights on the two largest dashboard headings normalize to 700.

## Shared projection, real Elements fixture

The dashboard projection (`projectCloudNotebookDashboard` and its types) lived in `apps/notebook-cloud`, and the Elements catalog re-implemented its own `projectDashboard` beside it. That is the second-app-model smell the host-shell ADR warns about. The pure projection moves to `src/components/notebook/workspace/notebook-dashboard.ts`, so cloud and Elements import the same code, and the Elements `cloud-dashboard` fixture now drives ordering, metrics, and the published/access sidebar from the shared projection rather than a copy. The catalog can no longer drift from production behavior; only per-notebook display flavor stays local.

The dashboard *view* stays in cloud. It is styled with the hosted viewer's bespoke CSS while Elements renders with Tailwind, so a shared view component is a larger move (shared stylesheet or a Tailwind rewrite) than this PR takes on. The projection is the part that was genuinely duplicated, and that is now shared.

## Design decisions

**Source guardrails read a corpus, not one file.** `viewer-render-props.test.ts` and `viewer-shared-cell-surface.test.ts` grep the viewer's raw source for shared-shell invariants (cloud routes chrome through the shared shell, presents one host notice, does not reintroduce report-mode cells, etc.). They pinned `index.tsx` directly. The first commit routes them through a `viewer-source-corpus.ts` helper so each invariant is asserted across the decomposed entry modules regardless of which one owns it. That commit moves no product code and keeps all 36 assertions green, which is the proof the later moves are faithful. The corpus is an explicit allowlist of the entry surface; siblings with their own concern (`notices.tsx`, `sharing-controls.tsx`) are read directly so the absence-assertions stay meaningful.

**Two suspected "gaps" are intentional host-specific decisions, left alone.** Cloud passes `identityControls={null}` and renders presence as a bespoke avatar stack (`data-slot="cloud-presence-stack"`) rather than the shared `NotebookPresenceStatus`. Both are pinned by guardrail tests, and hosted smoke selectors depend on the presence slot.

## Judgment calls, easy to overrule

- **Stopped the `NotebookViewer` hook extraction after the workstation manager and the capability adapter.** It is ~1,000 lines; the rest is cohesive shell orchestration and JSX. The access-request and cell/runtime-action hooks are the obvious next extractions.
- **Dashboard sidebar shows Published + Access, not a workstation-default summary.** `/n` loads the list bootstrap and has no workstation endpoint, so a live workstation panel there would be fabricated. The Elements fixture's old mocked workstation aside is replaced with the same published/access panels for the same reason.
- **The off-scale `font-weight: 720` normalized to 700** on the two largest dashboard headings. Deliberate, called out here because it is a visual change bundled with the rebuild.

## Verification

- Cloud `tsc --noEmit`: clean. Elements `tsc --noEmit`: clean (the one `collections/server` error is the pre-existing fumadocs codegen module, unrelated).
- Full cloud TS suite (`test/*.test.ts`): 541 passing, 0 failing, including the corpus guardrails and the dashboard projection test.
- `vp check apps/notebook-cloud`: formatted, 0 lint errors.
- A multi-agent adversarial review of the branch raised 15 findings, 6 confirmed (all low-severity), all addressed.

## Not verified

The cloud viewer was not run in a browser (it needs a hosted room with auth and data), and a Vite bundle build of the viewer was not run. The dashboard changes are covered by the projection unit test, the source guardrails, and the Elements fixture, not a live visual check.

## Follow-ups

- Extract `useCloudAccessRequest` and the cell/runtime action handlers from `notebook-viewer.tsx`.
- Share the dashboard *view*, not just the projection, once its styling is reconciled between the cloud stylesheet and Elements' Tailwind.


rgbkrk:
Superseded. origin/main #3509 independently landed the same viewer decomposition and a stronger filter-based dashboard. The only non-redundant pieces from this exploration - lifting the dashboard projection to src/components/notebook/workspace and pointing the Elements catalog at the real projection - are now in #3512, built on #3509's code. Closing this; the branch stays for reference.

PR #3509, opened 2026-06-09
feat(cloud): polish notebook home and workstation panel

quillaid:
## Summary

- Makes hosted `/n` a search-first notebook home with a compact Continue affordance, work/cleanup filter groups, quieter row facts, and cleanup queues for generated and untitled rooms.
- Temporarily redirects `/` to `/n` so signed-in users land on their notebooks immediately.
- Removes visible filter counts from the dashboard filter row; sections still carry count context where it helps, and filter buttons keep count-aware accessible labels.
- Preserves OIDC `given_name` for the `by <first name>` subtitle without rendering email addresses.
- Keeps personal rows quiet by omitting repeated `owner` facts while preserving editor/viewer/runtime and published facts where they change user expectations.
- Compacts the shared workstation panel around active target facts and structured registered-workstation facts.
- Fixes stale workstation attachment recovery so a recovered online/default registered workstation row remains visible and attachable.
- Lets the hosted workstation agent run with either API-key auth (`anaconda-key`, default) or OIDC bearer auth (`NOTEBOOK_CLOUD_WORKSTATION_AUTH_KIND=oidc`) and passes the matching auth kind to `runtimed cloud-runtime-agent` without putting credentials in argv.
- Extracts cloud viewer config/types, auth hooks, shell-capabilities, workstation manager, hosted route views, auth controls, dashboard view, notebook title, and presence chrome out of `viewer/index.tsx`.
- Waits for dev auth or the app-session cookie before browser app APIs fetch hosted data, avoiding fresh-context anonymous request races and console 401s.

## Preview

Deployed to preview.runt.run at head `6db3932e`.

- Main Worker version: `84ee6a61-5d58-41d6-8e68-ab5aaec5544d`
- Renderer assets Worker version: `54391f75-06f7-49db-88ab-8c4f85b814b9`

## Validation

- `pnpm --dir apps/notebook-cloud exec node --test test/hosted-workstation-agent.test.mjs`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/notebook-dashboard.test.ts`
- `node --check apps/notebook-cloud/scripts/hosted-workstation-agent.mjs && node --check apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs`
- `pnpm --dir apps/notebook-cloud exec tsc --noEmit --pretty false`
- `cargo xtask lint --fix`
- `pnpm --dir apps/notebook-cloud run deploy`
- `pnpm --dir apps/notebook-cloud run deploy:check`
[...]

PR #3508, opened 2026-06-09
docs(elements): polish cloud notebook surfaces

rgbkrk:
## Summary

Polishes the cloud notebook surfaces in Elements and removes the throwaway dashboard design-lab app from the final branch diff.

This keeps the durable work focused on:

- A search-first `/docs/cloud-dashboard` fixture for hosted notebook home, with pinned precedence, recent activity, cleanup filters, workstation context, and safe sharing metadata.
- Current cloud shell and compute-placement fixtures that remove outdated sync, workstation, presence, package, and fake-notebook chrome.
- Production-backed rail, workstation, environment, toolbar, and remote-cursor surfaces so the catalog examples do not teach agents placeholder UI.

## Product Notes

- `/n` should read as a notebook switcher first, not a metrics dashboard or project-management surface.
- Workstation readiness stays host-owned context until a room is opened; execution controls stay in the notebook shell.
- Sharing facts are safe host metadata. Private notebooks should not leak preview content; public previews require explicit published revision data.
- The Elements catalog should keep rendering shared notebook components and deterministic fixtures rather than synthetic placeholder controls.

## Validation

- `pnpm --dir apps/elements build`
- `cargo xtask lint --fix`
- In-app browser verification against `http://localhost:5176/docs/cloud-dashboard` confirmed the dashboard fixture renders, final copy is present, and old design-lab/ranking copy is gone.

PR #3512, opened 2026-06-09
refactor: share the cloud dashboard projection and fixture it in Elements

rgbkrk:
Two small convergence pieces on top of #3509 (which already decomposed the cloud viewer and reworked the notebook home). #3509 left the dashboard projection in `apps/notebook-cloud/viewer` and the Elements catalog kept re-implementing its own copy of it. This shares the projection and points the catalog at the real one.

## Changes

1. **Lift the projection to shared.** `projectCloudNotebookDashboard`, `projectCloudNotebookDashboardView`, and their types move from `apps/notebook-cloud/viewer/notebook-dashboard.ts` to `src/components/notebook/workspace/notebook-dashboard.ts`. The module is pure (no imports), so it carries cleanly. Cloud importers resolve it through `@/components/notebook/workspace`. Pure code move, no behavior change.

2. **Elements renders the real projection.** The `cloud-dashboard` catalog fixture re-implemented its own `projectDashboard`. It now drives ordering, the filter groups and their counts, and the sections from the shared `projectCloudNotebookDashboard` / `projectCloudNotebookDashboardView`. The fixture shows the real filter-based home with interactive filter chips instead of a parallel mock; only per-notebook summary copy stays local. This removes the second-projection divergence the host-shell ADR warns about, and #3509 never touched `apps/elements`.

## Verification

- Cloud `tsc --noEmit`: clean. Elements `tsc --noEmit`: clean (the lone `collections/server` error is the pre-existing fumadocs codegen module, present on `main` too).
- Full cloud TS suite: 546 passing. The 3 failures (`html.test.ts`, `worker-routes.test.ts`, root path `302 != 200`) are **pre-existing on `origin/main`** — the Worker now redirects `/` to `/n` and those route tests still expect a 200 shell. Confirmed by running the suite on a clean `origin/main` checkout (same 3 red). Not touched here; worth a separate fix.
- `vp check`: formatted, 0 lint errors.

## Context

This supersedes draft #3507, an exploratory branch that did the full decomposition independently before #3509 landed. #3509 covers the decomposition and a stronger filter-based dashboard, so the only non-redundant pieces from that exploration are the two above. #3507 will be closed.


rgbkrk:
Superseded by #3508. The Elements cloud dashboard moved to a richer bespoke design (search-first, pinned rooms, project grouping, compute states) whose fixtures use fields the shared projection does not model (project, compute, pinned, share vs notebook_id, title, scope, latest_revision_id). So the convergence here is incompatible: the shared projection cannot drive #3508's design without dropping its data, and the projection lift loses its cross-app consumer. Closing rather than force a regressive rebase. Branch kept for reference.

PR #3517, opened 2026-06-09
fix(notebook-cloud): sync hosted widget state across viewers

quillaid:
## Summary
- project hosted widget comm state into the browser WidgetStore from the shared SyncEngine `commChanges$` stream
- let cloud live handles expose RuntimeStateDoc and CommsDoc snapshots so late subscribers can replay durable comm projection state
- fall back to raw CommsDoc JSON for plain widget state only when no ContentRefs are present; ContentRef-backed state still defers instead of being mis-projected
- converge desktop and hosted cloud on a shared `applyWidgetCommChangesToStore` / `applyWidgetCommBroadcastToStore` bridge so both surfaces consume document truth with the same WidgetStore semantics
- harden the hosted two-window widget smoke so it proves IntSlider sync in both directions and captures first-party diagnostics on failure

## Reviewer focus
- Please scrutinize the raw JSON fallback in `SyncEngine.projectComms()`. It is deliberately constrained to no-ContentRef state, but the likely next step is a real cloud resolver for blob/manifest-backed widget state. The important boundary is that plain JSON widgets should project from CommsDoc truth, while manifest/blob-backed state should be resolved properly or deferred, never plucked half-correctly.
- Please also check the shared widget store bridge: desktop and cloud now use the same store transition semantics for `commChanges$` and custom comm broadcasts.

## Validation
- `cargo xtask lint --fix`
- `node --import tsx --test test/room-materializer.test.ts`
- `pnpm test:run -- src/components/widgets/__tests__/comm-changes-store-bridge.test.ts src/components/widgets/__tests__/cloud-widget-runtime.test.tsx packages/runtimed/tests/sync-engine.test.ts src/components/isolated/__tests__/comm-bridge-manager.test.ts`
- `pnpm --filter notebook-ui test:e2e:browser -- --project=chromium apps/notebook/e2e/execute-cell-output.spec.ts`
- `pnpm --filter notebook-ui test:e2e:browser -- --project=chromium apps/notebook/e2e/widget-cross-window-sync.spec.ts`
- deployed preview.runt.run before the rebase: assets worker `ec818943-b1b8-4a66-9ced-0cea4320b6b7`, main worker `6c9a35f0-0061-4664-bff4-ac5da7d02768`
- hosted smoke passed before the rebase against `https://preview.runt.run/n/01KTPTP7PY4YS9YN4A1DNAGBQN/widget-cross-window-smoke`: primary drag synced to peer, then peer drag synced back to primary

## Follow-up
- cloud should grow a host-provided resolver for live comm states that contain ContentRefs, probably using the existing blob/manifest resolution path that feeds iframe buffer paths. This PR intentionally keeps those states deferred rather than mis-projecting them.

### DIFF
--- a/apps/notebook-cloud/package.json
+++ b/apps/notebook-cloud/package.json
@@ -26,6 +26,7 @@
     "smoke:hosted:runtime-peer": "node scripts/hosted-runtime-peer-smoke.mjs",
     "smoke:hosted:workstation-agent": "node scripts/hosted-workstation-agent.mjs",
     "smoke:hosted:workstation-toolbar": "node scripts/hosted-workstation-toolbar-smoke.mjs",
+    "smoke:hosted:widget-cross-window": "node scripts/hosted-widget-cross-window-smoke.mjs",
     "smoke:hosted:browser-execute": "node scripts/hosted-browser-execute-smoke.mjs",
     "smoke:hosted:runtime-browser-execute": "node scripts/hosted-runtime-browser-execute-smoke.mjs",
     "smoke:hosted:collab": "node --import tsx scripts/hosted-collab-smoke.mjs",
--- a/apps/notebook-cloud/scripts/hosted-widget-cross-window-smoke.mjs
+++ b/apps/notebook-cloud/scripts/hosted-widget-cross-window-smoke.mjs
@@ -83,11 +83,38 @@ async function main() {
       );
     }
 
-    const browserRun = await runBrowserSmoke({
-      browser,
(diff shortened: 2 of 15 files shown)


rgbkrk:
@pullfrog review this with an eye towards suggestions. It seems we have architectural funk if we're plucking out json separate from the automerge document. I also wouldn't mind if you looked at this in terms of architectural consistency.

rgbkrk:
@pullfrog try again lil buddy

PR #3520, opened 2026-06-09
fix(notebook-cloud): trust app sessions over stale oidc state

quillaid:
## Summary
- hide the notebook header sign-in CTA while a first-party app-session cookie is present or still loading
- keep /n from reporting an expired local OIDC renewal token as an expired browser session when the app-session cookie is usable
- move the header sign-in decision into the auth model layer and cover the cookie-over-localStorage behavior

## Why
The live room and first-party APIs use the app-session cookie. A stale localStorage OIDC token is only renewal material at that point, so it should not make the notebook chrome say “Sign in again” while widgets/live sync still work.

## Validation
- pnpm --dir apps/notebook-cloud exec node --import tsx --test test/viewer-render-props.test.ts test/viewer-notices.test.ts test/collaborator-auth.test.ts
- pnpm --dir apps/notebook-cloud run typecheck
- cargo xtask lint --fix

### DIFF
--- a/apps/notebook-cloud/test/collaborator-auth.test.ts
+++ b/apps/notebook-cloud/test/collaborator-auth.test.ts
@@ -13,6 +13,7 @@ import {
   clearCloudPrototypeDevAuth,
   cloudSyncAuthFromAppSessionCookie,
   cloudSyncAuthFromPrototypeAuthState,
+  shouldShowCloudHeaderSignIn,
   withCloudPrototypeAuthHeaders,
   isCloudPrototypeAuthStorageKey,
   prepareCloudOidcViewerLogin,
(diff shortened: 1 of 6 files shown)


quillaid:
Preview deployed and live-smoked.

Live smoke setup: established a valid app-session cookie from the local preview token, deliberately planted an expired `nteract:notebook-cloud:oidc-token` in localStorage, then checked `/n/topic-viz/topic-viz` and `/n`.

Observed:
- notebook `Sign in again` buttons: 0
- notebook body contains `Sign in again`: false
- `/n` contains `Session expired`: false
- `/n` contains `Sign in again`: false

Deploy versions:
- assets worker: `4f62d398-1a43-46ab-89ca-f539cd6f7905`
- main worker: `e24a5da3-2b36-476f-9495-00943f9ccdd4`

PR #3521, opened 2026-06-09
fix(notebook): recover Tauri frame channel after reload

rgbkrk:
## Summary

- Rebind the Tauri notebook frame channel before each frontend bootstrap so HMR or manual webview reloads can receive daemon frames in the new JS world.
- Make the Rust relay read the current same-generation frame channel for every frame instead of holding the first channel forever.
- Detect an already-running dev daemon by probing the worktree socket instead of relying on `daemon.json`, which is being retired.

## Why

Vite HMR could leave the desktop notebook stuck after reload: the new frontend registered a replacement Tauri channel, but the Rust relay kept sending sync frames to the old channel captured before reload. The new `SyncEngine` could announce readiness, then never receive the session-control frame that makes the notebook interactive.

## Verification

- `cargo xtask lint --fix`
- `pnpm test:run packages/notebook-host/tests/relay-bootstrap.test.ts packages/notebook-host/tests/tauri-host.test.ts`
- `cargo check -p notebook`
- `cargo test -p xtask`
- `cargo test -p notebook sync_ready`

### DIFF
--- a/apps/notebook/e2e/run-browser-e2e.mjs
+++ b/apps/notebook/e2e/run-browser-e2e.mjs
@@ -2,7 +2,6 @@ import crypto from "node:crypto";
 import fs from "node:fs";
 import http from "node:http";
 import https from "node:https";
-import os from "node:os";
 import path from "node:path";
 import { spawn, spawnSync } from "node:child_process";
 import { fileURLToPath } from "node:url";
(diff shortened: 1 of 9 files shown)


rgbkrk:
@pullfrog review and verify claims

PR #3531, opened 2026-06-10
fix(cloud): stop runtime peer rejection storms

quillaid:
## Summary
- classify steady-state cloud `cloud_frame_rejected` controls as terminal stream errors for cloud runtime peers
- keep ordinary cloud WebSocket closes/resets recoverable, but stop reconnecting when the room rejects our sync frames
- make the hosted workstation agent parse non-JSON 429/1027 responses safely and cool down on 429/503 responses

## Why
Preview hit Cloudflare Error 1027 after the `NotebookRoom` Durable Object consumed roughly 837k hibernation requests. Two old runtime peers for `with anil` and `Renting Cognition` produced most of the traffic after room-host RuntimeStateDoc actor-label rejections, then kept reconnecting/resyncing.

## Validation
- `cargo xtask lint --fix`
- `cargo test -p notebook-protocol connection::transport`
- `cargo test -p notebook-cloud-transport`
- `cargo test -p runtimed reconnect_with_backoff -- --nocapture`
- `pnpm --dir apps/notebook-cloud exec node --test test/hosted-workstation-agent.test.mjs`

## Deployment
Not deployed yet. `preview.runt.run` is currently over the Cloudflare free-plan Durable Object request cap, so live smokes would add noise until the cap resets or the plan is upgraded.

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
@@ -2,6 +2,7 @@ import os from "node:os";
 import path from "node:path";
 
 export const DEFAULT_WORKSTATION_AUTH_KIND = "anaconda-key";
+export const DEFAULT_WORKSTATION_RETRY_AFTER_MS = 60_000;
 export const WORKSTATION_AUTH_KINDS = new Set([DEFAULT_WORKSTATION_AUTH_KIND, "oidc"]);
 
 export function buildWorkstationRegistrationPayload({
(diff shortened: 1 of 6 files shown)


rgbkrk (approved):
Correct diagnosis and clean fix. The two sides address different parts of the same storm: Rust stops the runtime peer from reconnecting after a permanent room rejection; JS stops the workstation agent from hammering a rate-limited endpoint. Both are guarded by tests. Approving with a few notes.

**Rust - `cloud_frame_rejection_error` in the hot path**

The function runs on every `SessionControl` frame. In steady state that's `Some(Ok(frame))` — the rejection check only fires on `SessionControl` frames, which are infrequent, so the `serde_json::Value` allocation is acceptable. One small cosmetic thing: `value.to_string()` on a `serde_json::Value::Number` produces the bare number (`5`), but on a `Value::String` it produces the JSON-quoted form (`"sync_frame"`). Harmless for diagnostics, but a `control["frame_type"].as_u64().unwrap_or_default()` or similar would give a cleaner message. Not a blocker.

**Rust - `terminal_error` + `break` + `Err` propagation**

Clean. The inner reconnect loop breaks, the outer function returns `Err`, and the two call sites in `main.rs` and `allocate.rs` don't restart the agent on error — so the reconnect storm stops at the process level. Verified there's no outer retry wrapper around `run_cloud_runtime_agent`.

**JS - cooldown double-check in `main()`**

The pattern is correct:
```js
await runAgentStep("heartbeat", ...);
if (cooldownUntil > Date.now()) {
  continue; // heartbeat itself 429'd, skip poll
}
await runAgentStep("poll_attach_jobs", ...);
```
It reads a bit like belt-and-suspenders since the top-of-loop check also catches this on the next iteration, but it's the right call — you don't want to poll immediately after a heartbeat 429 even if the sleep would be short.

**`retryAfterMs` floor at 1 second**

Good guard. A header of `Retry-After: 0` would otherwise spin at poll frequency, which is exactly the failure mode this PR is fixing.

**`parseHttpResponseBody` and the old `responseJson`**

The old `.json().catch(() => response.text())` was a streaming trap: once `response.json()` fails and consumes the body, `response.text()` on the same response object returns an empty string in some runtimes. The new implementation reads text once and parses — this is the right fix, especially for Cloudflare HTML error pages.

One thing worth confirming before undraft: does the mjs test cover the case where `retryAfterMs` receives an HTTP-date `Retry-After` header? The test only exercises the numeric seconds form and the missing-header fallback. The `Date.parse(header)` branch is reachable but not tested. Not a merge blocker, but worth a follow-up test or a comment noting the branch.

PR #3534, opened 2026-06-10
fix(notebook): harden sync-bridge output projection (FSB-1, FSB-2)

quillaid:
Two punchlist rows from `cleanup-punchlist.md`, both in the frontend sync bridge. Follow-up to #3523/#3524; sibling of #3535.

## FSB-1 — output projection failures surfaced, not swallowed

`applyOutputChangeset` used to log failed manifest resolutions at `warn` and move on — a partial output projection silently showed a stale manifest with no retry, banner, or error count.

Now:
- **Bounded retry** (250ms, 1s; per-attempt at `debug` per logging rules, final failure at `warn`).
- **Per-output write generations** guard the unserialized projection path: each tick stamps a generation before its first await, so a suspended older tick's retry can never resurrect a removed output or downgrade a newer manifest (Codex P1; both interleavings pinned by regression tests). Chosen over serializing the path, which would head-of-line-block all outputs behind one failing output's backoff.
- **Subscribable failure surface** (`getOutputProjectionFailures` / `subscribeOutputProjectionFailures` / `useOutputProjectionFailures`) with a production reader: `NotebookView` renders an "N outputs failed to load" banner — cross-host, since desktop and cloud render the same `NotebookView` (Codex P3). Entries clear on later success, output removal, or projection reset.

## FSB-2 — CI guard for the stable DOM order invariant

The invariant (cells render in sorted-ID DOM order; visual position is CSS `order` in a flex column; regression = `insertBefore` destroying output iframes) lived in three uncoordinated sites in `NotebookView.tsx` with no guard. A source-shape test now pins all three sites and additionally rejects mapping the cell list over `cellIds` directly. jsdom can't observe iframe teardown, so the guard is at source level — same pattern as the existing `notebook-view-capabilities-source.test.ts`.

Punchlist updated: FSB-1/FSB-2 → Done.

## Verification
- Desktop suite 628 passed (incl. 5 interleaving/failure-surface tests + 3 DOM-order guards). Typechecks clean both apps. `cargo xtask lint` clean.


quillaid:
Addressed both Codex findings in dd1b7bc5:

**P1 (interleaved retry races):** went with the per-output generation guard rather than serializing the projection path — serialization would head-of-line-block every subsequent output behind one failing output's backoff (~1.25s), which is the projection-latency regression the LNP-1 line of work exists to avoid. Each `applyOutputChangeset` tick stamps a generation per output id before its first await; writes/failure records land only while current, and removals bump the generation so suspended resolutions become no-ops. Both interleavings (resurrect-after-remove, downgrade-after-replace) are pinned by regression tests.

**P3 (no production reader):** `NotebookView` now renders an "N outputs failed to load (may be stale)" banner from the new `useOutputProjectionFailures()` hook — cross-host since both apps render the same `NotebookView` — and the surface is exported via `notebook-surface.ts`. Punchlist row updated to match what actually shipped.

PR #3542, opened 2026-06-10
docs(cloud): add EC2 fallback runbook

quillaid:
## Summary

- add `pnpm --dir apps/notebook-cloud run dev:ec2`, a wrapper for running the existing notebook-cloud Worker locally on one EC2 instance with persisted Miniflare state
- default the EC2 fallback to same-origin renderer/runtime assets, `srcdoc` output frames, and prototype dev-token browser auth so it does not depend on preview-specific domains or OIDC redirect registration
- document the EC2 fallback runbook, reverse proxy shape, systemd unit, workstation runner setup, smoke checks, and limitations

## Why

`preview.runt.run` can be blocked by Cloudflare plan limits even when we still need a hosted notebook room for demos. This keeps the fallback path on the existing Worker/Durable Object/D1/R2-compatible model instead of inventing a second room host.

## Validation

- `pnpm --dir apps/notebook-cloud exec node --test test/ec2-local-worker.test.mjs`
- `node --check apps/notebook-cloud/scripts/ec2-local-worker.mjs`
- `pnpm --dir apps/notebook-cloud test`
- `cargo xtask lint --fix`

## Not yet done

I have not brought up a public EC2 hostname with nginx/Caddy and a running workstation against it in this PR. The wrapper/runbook are the first mergeable slice.

### DIFF
--- a/apps/notebook-cloud/README.md
+++ b/apps/notebook-cloud/README.md
@@ -591,6 +591,12 @@ pnpm --workspace-root exec wrangler tail nteract-notebook-cloud \
   --format pretty
 ```
 
+If `preview.runt.run` is unavailable or rate-limited, use
+[`docs/notebook-cloud-ec2-fallback.md`](../../docs/notebook-cloud-ec2-fallback.md)
+to run this same Worker bundle on one EC2 instance through
+`wrangler dev --local --persist-to`. That fallback is intended for demos and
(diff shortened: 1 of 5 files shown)


quillaid:
Closing per direction: this was a Wrangler/Miniflare fallback slice, but we want a deeper proper AWS architecture pass instead.

PR #3548, opened 2026-06-10
docs(adr): propose AWS rust room host

quillaid:
## Summary
- add a draft ADR for a native Rust hosted room service on AWS
- define Postgres/RDS as the catalog and authorization store, with sqlx as the default Rust client
- define S3 as the durable snapshot/blob byte store while keeping the typed-frame WebSocket protocol stable
- register the ADR in the architecture index

## Validation
- cargo xtask lint --fix

### DIFF
--- a/docs/adr/README.md
+++ b/docs/adr/README.md
@@ -97,6 +97,7 @@ These entries define the center of gravity for the system:
 | Entry | Status | Notes |
 |-------|--------|-------|
 | [Deployment Topology](deployment-topology.md) | Draft | Hosted room topology, compute attachment, and durable state split. |
+| [AWS Rust Room Host](aws-rust-room-host.md) | Draft | Proposed native Rust room host on AWS with Postgres/RDS and S3. |
 | [Hosted Room Authorization and Cloud Room Host](hosted-room-authorization.md) | Draft | D1 ACLs, scope derivation, and DO room host. |
 | [Runtime Principal Promotion](runtime-principal-promotion.md) | Draft | Local runtime principals, hosted promotion, and execution authority split. |
 | [Remote Workstation Doc Agents](remote-workstation-doc-agents.md) | Draft | Provider-neutral workstation/doc-agent registration and runtime-peer attachment model. |
--- a/docs/adr/aws-rust-room-host.md
+++ b/docs/adr/aws-rust-room-host.md
@@ -339,7 +339,68 @@ The stable split:
 - revision/checkpoint metadata in Postgres;
 - room coordination in the Rust room actor, not S3.
 
-## Decision 7: One Big Room Host First, Leases Later
+## Decision 7: Cloud Notebooks Use Native Document Set Identity
+
+Hosted notebooks are not persisted as `.ipynb` files with a cloud sync wrapper.
+The cloud-native source of truth is a durable document set:
+
+- `NotebookDoc` for cells, metadata, order, and notebook-level authoring state;
+- `RuntimeStateDoc` for execution records, queue state, kernel/session facts,
+  output manifests, workstation attachment state, and other runtime-visible
+  notebook state;
+- `CommsDoc` for widget/comm state needed to render interactive outputs;
+- content-addressed blobs for output bytes and renderer assets referenced by
+  those documents.
+
+`.ipynb`, HTML, and other file formats are import/export projections from that
+document set. They are not the hosted persistence boundary.
+
+Keep identifiers boring and single-purpose:
+
+- `notebook_id` is the stable opaque room/catalog id. It drives routes, ACLs,
+  room placement, leases, workstations, and live sync.
+- document ids identify the individual Automerge documents whose immutable
+  published snapshots live under `docs/{docId}`.
+- `revision_id` identifies an immutable published snapshot bundle for one
+  notebook, including the exact `NotebookDoc`, `RuntimeStateDoc`, `CommsDoc`,
+  and blob references needed to render it.
+- `checkpoint_id` identifies an operational recovery point for a live room.
+  Checkpoints may include runtime state, but they are not the default public
+  sharing surface.
+- titles and slugs are mutable display/navigation metadata. They are never
+  authorization, placement, publication, or checkpoint authority.
+
+The canonical hosted routes should be identity-first with optional decorative
+slugs:
+
+```text
+/n/:notebook_id/:slug?                 live room
+/n/:notebook_id/r/:revision_id/:slug?  immutable published revision
+```
+
+The router trusts only `notebook_id`, the reserved control segment `r`, and
+`revision_id`. The slug exists so copied links remain recognizable to humans;
+it may be stale, missing, or regenerated after a title change. By default,
+shareable links include the slug. If title sensitivity becomes a product
+requirement, add an explicit "copy link without title" affordance later rather
+than making all links opaque now.
+
+Reserve control segments after `notebook_id` for route modes. At minimum, `r`
+is reserved for revisions. If checkpoint links become a user-facing or admin
+feature, use another reserved control segment rather than overloading vanity
+slug parsing. Vanity slugs should be single path segments and generated to
+avoid reserved control words.
+
+Persist `RuntimeStateDoc` and `CommsDoc` in both live checkpoints and published
+revisions so outputs, execution history, widgets, and workstation attachment
+context do not disappear when a room reloads or when a notebook is viewed
+without an active runtime peer. Persistence should compact these documents
+before storage: keep durable execution/output/widget facts, but reconcile
+process-local claims such as "kernel is running on host X" when the room loads.
+If the runtime peer does not reconnect and reclaim ownership, mark process
+state stale while preserving the user-visible execution and output history.
+
+## Decision 8: One Big Room Host First, Leases Later
 
 The first AWS service should be one vertically scaled room-host process behind
 an ALB. One active process owns all live room actors.
@@ -406,7 +467,7 @@ architectural requirements for the native service. A long-running room host can
 keep active rooms in memory and should rely on document sync plus checkpoints
 for recovery rather than porting hibernation behavior mechanically.
 
-## Decision 8: Room Checkpointing Is Debounced But Durable
+## Decision 9: Room Checkpointing Is Debounced But Durable
 
 On room load, port the current Cloudflare materializer's checkpoint precedence
 logic rather than replacing it with a simpler "checkpoint else snapshot" rule:
@@ -463,7 +524,7 @@ becomes an in-process Tokio timer in the single-host service. Because that timer
 dies with the process, room load must also reconcile stale runtime-peer state so
 phantom running executions and kernels do not survive a host crash.
 
-## Decision 9: Authority Boundaries Stay The Same
+## Decision 10: Authority Boundaries Stay The Same
 
 Moving from Durable Objects to Rust/AWS must not weaken hosted authorization:
 
@@ -493,7 +554,7 @@ transport. `KernelIdle`, `ExecutionDone`, `CellError`, and `KernelDied` cannot
 be backpressured by stdout floods, display churn, manifest commits, or blob
 writes.
 
-## Decision 10: Hosted Origins Stay Split
+## Decision 11: Hosted Origins Stay Split
 
 The AWS deployment keeps the three-origin security model from
 `hosted-output-origin-isolation.md`:
@@ -599,11 +660,12 @@ without AWS.
 6. Implement `HostedObjectStore` and `HostedCheckpointStore`; port checkpoint
    precedence logic with shared test vectors for deterministic actor label,
    starter cell id, and keep/discard decisions. Include holder/generation
-   fields in checkpoint pointer metadata from the start.
+   fields in checkpoint pointer metadata from the start. Persist compacted
+   `RuntimeStateDoc` and `CommsDoc` with checkpoints.
 7. Implement Postgres migrations and a `sqlx` `HostedCatalogStore` /
    `HostedAuthz` for catalog, ACL, sharing, workstations, and attach jobs.
    Preserve never-zero-owner ACL protection, partial unique active attach jobs,
-   and transactional revision publication.
+   document-set ids, and transactional revision publication.
 8. Implement the app-session/OIDC credential layer and key-management contract
    needed by the Rust service.
 9. Add integration tests covering:
--- a/docs/adr/aws-rust-room-host.md
+++ b/docs/adr/aws-rust-room-host.md
@@ -63,13 +63,18 @@ The shared protocol surface already exists:
   `/n/:id/sync`, so the browser does not require Durable Object semantics.
 
 The native daemon already has pieces the hosted service should reuse or
-extract:
+extract, but the native `NotebookRoom` itself is not the hosted-room shape. It
+also owns local file bindings, file watchers, autosave, trust state, local
+runtime-agent handles, and destructive kernel teardown. Those are daemon/local
+concerns, not hosted-room concerns.
 
 - `crates/runtimed/src/notebook_sync_server/peer_loop.rs` has the biased
-  frame/readiness/runtime/presence loop.
+  frame/readiness/runtime/presence loop. It is generic over `AsyncRead` and
+  `AsyncWrite`; the remaining daemon couplings should be carved out as
+  explicit parameters or hosted adapters.
 - `peer_writer.rs` classifies reliable frame lanes and gates request authority.
-- The native room model owns `NotebookDoc`, `RuntimeStateDoc`, `CommsDoc`,
-  presence, request workers, blob upload workers, and runtime lifecycle.
+- The shared document and wire crates (`notebook-doc`, `runtime-doc`,
+  `notebook-wire`, `notebook-protocol`) are daemon-free.
 
 The cloud Worker has hosted product pieces that should be ported through proper
 interfaces:
@@ -81,6 +86,13 @@ interfaces:
 - `apps/notebook-cloud/src/room-materializer.ts` has the checkpoint and
   published-snapshot recovery policy, currently backed by Durable Object
   storage, D1, R2, and `runtimed-wasm`.
+
+`crates/runtimed-wasm/src/lib.rs` `RoomHostHandle` is the closest reference
+design for the hosted room's shape: per-peer Automerge sync states, scope
+validation, execution-queue handling, workstation attachment publication, and
+`reconcile_runtime_peer_gone`. The native hosted room should mirror that
+structure in Rust; it should not treat the WASM boundary as the implementation
+to embed or mechanically extract.
 
 ## Decision 1: The Hosted Room Host Is Native Rust
 
@@ -121,10 +133,16 @@ Outputs:
 - catalog/revision writes to Postgres;
 - metrics/logs.
 
+One Tokio task owns the three room documents and all per-peer sync state for a
+given notebook. Peer tasks communicate with that owner through channels. Do not
+share `NotebookDoc`, `RuntimeStateDoc`, `CommsDoc`, or per-peer
+`automerge::sync::State` through `Arc<RwLock<_>>` across peer tasks; that erodes
+the single-writer guarantee that Durable Objects previously gave implicitly.
+
 The service should reuse native `runtimed` room-loop mechanics where possible.
-It should not embed `runtimed-wasm` as its core room engine. WASM remains a
-browser/client boundary and a compatibility/testing asset; the server is native
-Rust.
+It should model the hosted document owner after `RoomHostHandle`, but should
+not embed `runtimed-wasm` as its core room engine. WASM remains a browser/client
+boundary and a compatibility/testing asset; the server is native Rust.
 
 ## Decision 2: The WebSocket Protocol Stays Stable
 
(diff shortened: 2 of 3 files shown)


quillaid:
## Architectural review

Read against the related ADRs and the actual code on `main`. The direction is right: protocol stability (Decision 2), one-host-first with explicit leases later (Decision 6), the Postgres-metadata / S3-bytes split, and rejecting Wrangler-on-EC2 are all the correct calls. The findings below are mostly about (a) places where the ADR's framing doesn't match what the code actually offers, (b) consistency with sibling ADRs, and (c) what to carve out **before** any Terraform exists.

### 1. The reuse story is slightly miscast — and the real one is better

The ADR says "extract the room-host document logic out of `runtimed-wasm` into native Rust." In practice `runtimed-wasm`'s `RoomHostHandle` (`crates/runtimed-wasm/src/lib.rs:366`) is already the right *shape* for the hosted room — per-peer Automerge sync states, scope validation, execution-queue handling, `reconcile_runtime_peer_gone` — but it's a reference design to mirror, not a codebase to extract from. Meanwhile the native `NotebookRoom` (`crates/runtimed/src/notebook_sync_server/room.rs`) is the one tangled with daemon concerns: `NotebookFileBinding` + file watcher + autosave debouncer, local `.automerge` persistence, `trust_state`, `runtime_agent_handle` (local kernel ownership), and `kernel_teardown_destructive`.

The genuinely reusable pieces are:

- `run_sync_loop_v2` in `peer_loop.rs` is generic over `AsyncRead`/`AsyncWrite` and ~95% transport-agnostic. Its daemon couplings are narrow: `daemon.pool_doc_changed`, `daemon.idle_peer_timeout()`, and the request-worker dispatch.
- `FrameSource`/`FrameSink`/`FrameTransport` already exist (`crates/notebook-protocol/src/connection/transport.rs:52-75`) and `CloudWsFrameTransport` already proves the one-WS-message-per-frame mapping from the client side. The Axum server adapter is the mirror image — on the order of 50 lines, not a new subsystem.
- The document-model crates (`notebook-doc`, `runtime-doc`, `notebook-wire`, `notebook-protocol`) are shared and daemon-free.

Suggested edit: rephrase implementation-plan step 2 to "build the hosted `RoomActor` after `RoomHostHandle`'s structure, reusing `peer_loop`/`peer_writer` mechanics and the shared document crates" rather than "extract out of `runtimed-wasm`."

### 2. The `Hosted*` traits are new construction, not extraction

`runtimed` has no persistence trait layer — `RoomPersistence` is a struct with atomic flags writing directly to the local cache dir, and the blob store is daemon-local. Step 1 of the plan ("define native hosted room traits") is the right move, but the ADR should be honest that this is fresh abstraction design. The good news is each trait has an obvious local test double (filesystem/minio for `HostedObjectStore`, dockerized Postgres for `HostedCatalogStore`), which is what makes the pre-Terraform sequencing below possible.

### 3. S3 layout: align with `hosted-notebook-artifacts.md`'s target namespace, don't invent a third

`hosted-notebook-artifacts.md` Decision 2 records the current compat layout (`n/{id}/snapshots/…`, `n/{id}/blobs/{sha256}`) **and** a stated long-term target (`docs/{docId}/snapshots/{headsHash}.am`, `blobs/{sha256}`). This ADR's global `blobs/{sha256}` matches the target — good. But `notebooks/{notebook_id}/snapshots/{heads_hash}.am` introduces a third namespace alongside `n/{id}` and `docs/{docId}`. A fresh S3 bucket is exactly the moment to land the target layout: use `docs/{docId}` for all three document types (NotebookDoc included) and reserve a `notebooks/{id}/checkpoints/` prefix only for room-host checkpoints, which are an operational artifact, not a publish artifact.

Two follow-ons:

- Going from per-notebook `n/{id}/blobs/` to a global content-addressed pool changes blob lifecycle: today notebook deletion can reap by prefix; a shared pool needs the `notebook_blobs` table to act as the ref/authz layer and an explicit GC stance (even if "no GC yet" — say so).
- Revision rows record exact `snapshot_key`/`runtime_snapshot_key` values, so mixed-layout reads must stay supported, same as the existing "legacy nested runtime snapshot keys remain readable" rule.

### 4. Checkpoint semantics: port the precedence logic exactly, and make the pointer write atomic

`room-materializer.ts:174-296` is more subtle than "load checkpoint else snapshot": the keep/discard decision on `published_revision_id` mismatch hinges on `hasUnpublishedCheckpointChanges` (heads-array comparison) plus a legacy-version timestamp fallback, and the deterministic identities are load-bearing in two places — the room actor label `system:notebook-cloud-room/room:{fnv1a64(notebook_id)}` **and** the starter cell ID `cell-room-{stableRoomKey}` (`room-materializer.ts:511-527`). Decision 7 covers the labels; it should name the precedence algorithm and starter-cell determinism as ported-verbatim behavior, ideally with a shared test vector.

One thing the move makes *better*: the DO checkpoint is split across four storage keys with no transaction, so a crash mid-checkpoint leaves partial state that load-time validation has to reject. The proposed S3 layout already supports doing this right — write the immutable `{checkpoint_id}/*.am` objects first, then swap `checkpoints/latest.json` last as the single atomic commit point. Worth stating explicitly in Decision 7 so it doesn't get implemented as four independent PUTs.

### 5. Single-writer discipline needs one sentence of commitment

Durable Objects gave per-room single-threaded execution for free; that guarantee is why the bootstrap-actor/duplicate-seq class of bugs stayed contained. Decision 1's `RoomActor` sketch is right, but the ADR should commit to the discipline: one tokio task owns the three docs and all per-peer sync states, peers talk to it via mpsc — not `Arc<RwLock<NotebookDoc>>` shared across peer tasks (which is the daemon's current pattern and is exactly the shape that erodes into implicit multi-writer later, undermining Decision 6's reasoning).

### 6. Request authority: reject, don't ignore, daemon-only requests

The daemon request dispatcher (`crates/runtimed/src/requests/mod.rs:156-200`) handles `LaunchKernel`/`Interrupt`/`ShutdownKernel`/`SaveNotebook`, all of which are local-daemon concerns (kernel process control, `.ipynb` writes). The hosted-safe set is `ExecuteCell`/`RunAllCells`/`SendComm`, which only touch `RuntimeStateDoc`/`CommsDoc`. Decision 8 covers ACL-level rejection of unauthorized `REQUEST` frames; it should also state that out-of-scope request *types* get explicit error responses rather than silent drops, since current desktop clients can emit them.

### 7. The auth/session dependency is bigger than one tracked follow-up

`hosted-credential-transport.md` defines the whole chain this service inherits: OIDC bearer → app-session cookie → WebSocket, subprotocol fallback for browser bearer tokens, origin checks as part of browser WS auth, and `Authorization: Bearer` for native peers. Two pieces are currently Cloudflare-shaped in ways that affect infra design:

- App-session cookie encryption/signing is platform-provided today; the Rust service needs its own session key material, which determines the Secrets Manager + IAM shape in Decision 4's Terraform sketch. That's a pre-Terraform design decision, not a post-deploy follow-up.
- The Worker authenticates and stamps trusted identity headers before the DO sees the connection. In a single binary that boundary collapses (fine), but the ADR should say where authn happens relative to the WS upgrade and the room actor, so the "trusted headers" pattern doesn't get accidentally re-created across an internal boundary that no longer exists.

Recommend promoting this from Tracked Follow-Ups into a short Decision, and adding `hosted-credential-transport.md` and `hosted-output-origin-isolation.md` to References.

### 8. AWS-specific gaps in the current draft

- **ALB idle timeout**: default is 60s; a quiet viewer-only room will get its WebSocket severed. State the requirement (raised idle timeout and/or protocol-level keepalive) rather than just linking the attribute doc.
- **WS message size**: `MAX_FRAME_SIZE` is 100 MiB with 64 MiB caps on `AutomergeSync`/`RuntimeStateSync` (`crates/notebook-wire/src/lib.rs:47-128`). ALB doesn't care post-upgrade, but axum/tungstenite default max-message-size will, and it must be configured to match `notebook-wire`'s limits — one constant, sourced from the crate, not re-declared.
- **Three-origin model**: `hosted-output-origin-isolation.md` requires app origin / renderer-asset origin / output-document origin separation. The Terraform sketch has one ALB and no answer for the other two origins (CloudFront + S3, presumably) or for viewer/static asset serving generally. Decision 8 says sandboxed outputs stay on a separate origin, but the infra list doesn't provision one.
- **Runtime-peer grace timer**: the DO `setAlarm` 30s grace window becomes an in-process tokio timer — fine for one host, but it dies with the process. Note that load-time `reconcile_runtime_peer_gone` is the recovery backstop so phantom executions don't survive a crash.
- **Dropped DO behaviors worth recording as intentional**: hibernation and the 500-frame replay buffer don't apply to a long-running native process; saying so preempts someone porting them.

### 9. Carve-out order before the first Terraform file

This is the sequencing I'd propose — every step verifiable with `cargo test` + docker-compose (Postgres + minio), zero AWS:

1. **Axum WS frame adapter** — implement `FrameSource`/`FrameSink` over an Axum WebSocket (`[type byte][payload]`, sizes from `notebook-wire`). Smallest step, unlocks everything, immediately testable against `CloudWsFrameTransport` as the client.
2. **Decouple `peer_loop` from `Daemon`** — pass `idle_peer_timeout: Duration` and the pool-doc handle (or drop `PoolStateSync` for hosted) as parameters; delete the `daemon` parameter from the loop signature.
3. **Hosted request policy** — route the request worker through a trait so hosted scopes to `ExecuteCell`/`RunAllCells`/`SendComm` with explicit rejections, and handlers take `Arc<Room>` rather than the daemon.
4. **`HostedRoom` actor** — modeled on `RoomHostHandle`, owning docs + per-peer sync states in one task, no file binding / trust state / kernel handle.
5. **`HostedObjectStore` + `HostedCheckpointStore`** — fresh traits with filesystem and minio impls; port the checkpoint precedence logic with shared test vectors against the TS implementation (deterministic actor label, starter cell ID, keep-reason decisions).
6. **`HostedCatalogStore` + `HostedAuthz` on sqlx** — real migrations porting the D1 schema with the improvements this ADR lists, plus the invariants the current code enforces in app code: never-zero-owner ACL, partial-unique active attach jobs, transactional revision-publish (the D1 `recordRevision` batch can leave `latest_revision_id` drift — Postgres should close that with a transaction).
7. **Session/credential layer decision** — cookie signing + OIDC callback shape, since it dictates the Secrets Manager/IAM Terraform.

Step 8 in the existing plan (Terraform) then arrives with a binary that already passes the Decision-7/8 integration suite locally. The plan already orders Terraform second-to-last; the edit I'd make is stating the gate explicitly: *no Terraform until the integration tests in step 6 pass against Postgres + minio.*

### Bookkeeping

- This PR should also touch `docs/adr/deployment-topology.md` with a one-line forward pointer at Decision 1 ("partially superseded by aws-rust-room-host.md for the next hosted target"); right now the supersession is only discoverable from the new file. "Supersedes in part" is also a new convention for this index — fine, but the pointer should be bidirectional.
- `hosted-room-authorization.md` Decision 2 specifies the ACL contract in D1 terms (schema, `updated_at` helper rules, orphan-room protection). This ADR's Decision 3 should either restate those invariants as Postgres-portable or note that the authorization ADR needs a follow-up amendment — otherwise the ACL source of truth is split across two drafts.

quillaid:
Addressed the architectural review in 7337d7ce.

Changes made:
- reframed reuse around RoomHostHandle as a reference design, peer_loop/peer_writer as carve-out targets, and Hosted* traits as new service abstractions
- added explicit single-writer RoomActor discipline and hosted request-type rejection
- added auth-before-room-admission as a decision, including session key material / Secrets Manager implications
- aligned the S3 publish layout with the docs/{docId} target namespace and noted global blob GC/authz implications
- clarified checkpoint precedence, deterministic starter-cell identity, atomic latest.json pointer swaps, and runtime-peer-gone recovery
- added ALB idle timeout / keepalive, server message cap, and three-origin deployment requirements
- added the bidirectional pointer from deployment-topology.md

Validation: cargo xtask lint --fix

quillaid:
## Follow-up: consistent-hashing readiness without overdesign

7337d7ce addresses the review well — the `RoomHostHandle` reframing, single-writer actor discipline, auth-before-admission decision, and `docs/{docId}` alignment all read correctly now.

One more dimension worth a short subsection in Decision 7: being *ready* for consistent hashing at the notebook-room layer when scale-out arrives. The good news is that readiness here is almost entirely about preserving seams the design already has, not adding machinery. The framing that keeps this from being overdesigned:

**Leases are authority; hashing is only a placement hint.** When ring views differ between hosts (they always do, transiently), the hash answers "who should *try* to acquire the lease," and the lease answers "who owns the room." If the design ever treats the hash as authority, we've rebuilt sticky-session split-brain — the exact bug class Decision 7 lists. Consistent hashing later is an optimization (lease-acquisition locality, less thundering herd), never a correctness mechanism. Stating that one sentence now prevents the wrong mental model from forming.

Given that framing, the readiness list is four items, all of which the single-host service should do anyway:

1. **The unit of placement is `notebook_id`, full stop.** The room's three docs (`NotebookDoc`, `RuntimeStateDoc`, `CommsDoc`) co-locate under it and are never placed independently — worth saying explicitly now that the publish layout moved to per-doc `docs/{docId}` namespaces, which could tempt per-doc placement later. Nothing else (vanity names, doc IDs, revision IDs) routes. Notebook IDs are ULIDs (time-prefixed), so any future placement must hash, never range-partition — `stableRoomKey`'s FNV-1a is already a fine shape.

2. **Make `RoomRegistry` resolution the single admission gate.** Decision 1 already sketches `RoomRegistry: notebook_id -> RoomActorHandle`. Define its resolve as the one place a peer is admitted to a room, with a signature that can return "not local" even though v1 always succeeds locally: `resolve(notebook_id) -> Owned(handle) | OwnedElsewhere(holder)`. The WS upgrade path goes authn → authz → resolve → attach. That seam — trivially `Owned` in v1 — is the entire integration point for hashing and leases later; nothing else in the room actor needs to know placement exists.

3. **Implement room eviction on day one.** Flush checkpoint, drop per-peer sync states, close peers. Single-host needs this anyway for idle-room memory management, and it is exactly the handoff primitive room movement needs later ("evict and don't reacquire"). If eviction is only built when leases arrive, it gets built under pressure.

4. **Clients treat WS close as "reconnect and re-resolve."** The ADR already says room movement resets per-peer sync state and document truth comes from checkpoints. The client contract should be equally dumb: on close, reconnect through the front door with backoff; routing happens server-side at admission. No redirect protocol, no handoff frames in v1 — a clean close is sufficient, because resolve-at-admission does the rest.

One genuinely cheap pre-provision beyond the seams: **include the holder identity/generation in `checkpoints/latest.json`** when Decision 8's atomic pointer swap is implemented. The lease sketch already carries `generation bigint` — that's a fencing token, and S3 conditional writes (`If-Match` on the pointer's ETag) plus a recorded generation are what stop a zombie host from clobbering the checkpoint pointer after losing its lease. Retrofitting a field into checkpoint metadata later is painless; realizing a zombie clobbered `latest.json` in production is not. One field, one sentence.

And the explicit not-now list, to keep Decision 7's discipline: no ring membership service, no virtual-node configuration, no hash-algorithm selection, no cross-host forwarding implementation, no routing tier, and no `room_leases` table — all unchanged from the current draft. When multi-host does arrive, the likely v2 is ALB → any host → resolve → lease lookup → serve or hand back a reconnect, with consistent hashing arriving last, purely to make the first lease attempt usually land on the right host.

quillaid:
## Notebook ID format: drop the time prefix now

Since we're at the stage where this is still cheap, it's worth deciding the notebook ID format in this ADR rather than inheriting it. Today's IDs come from a hand-rolled ULID (`apps/notebook-cloud/src/index.ts:1613`): 48-bit millisecond timestamp + 80 bits random, 26-char Crockford base32.

To be clear about what does *not* motivate this: consistent hashing doesn't care. Placement should hash the ID regardless of format (and `stableRoomKey` already does), so the time prefix costs nothing there. The real arguments are elsewhere:

- **Every shared notebook URL leaks its creation timestamp.** Notebook URLs are the shareable artifact — invites, public snapshots, links pasted in chat. A ULID's first 10 characters decode to creation time at millisecond precision, forever. That's a small but permanent metadata leak we don't need.
- **Order-meaningfulness is an attractive nuisance.** A time-prefixed ID invites someone, someday, to sort by it, range-partition by it, or infer recency from it. A uniformly random ID makes "IDs are opaque" self-enforcing instead of a rule in a doc.
- **Slightly better collision/enumeration margin**: 80 random bits → 130, for free (the create path keeps its uniqueness check either way).

The counterarguments are weak at our scale: time-ordered PKs give append-only B-tree growth in Postgres, but the PK is `TEXT` regardless and we're not at the volume where index locality matters; creation-time sortability already lives in `created_at`.

The switch itself is forward-only and schema-free, because every consumer already treats IDs as opaque URL-safe strings: keep the 26-char Crockford base32 shape (so any path validation and URL aesthetics are untouched), fill all 130 bits from `crypto.getRandomValues`, and existing ULIDs remain valid as opaque strings. No migration, no dual-format handling beyond "old IDs keep working," which is already implied.

Suggested ADR addition (Decision 3 or a one-paragraph decision of its own): *Notebook IDs are opaque, URL-safe, uniformly random 26-char Crockford base32 strings. Nothing may infer time, order, or placement from ID structure; creation time lives in `created_at`, placement comes from hashing the full ID.* That also lets the scale-out readiness language drop its "ULIDs are time-prefixed, so never range-partition" caveat — there'd be nothing to range-partition on.

If we'd rather keep time-prefixed IDs for ops ergonomics (eyeballing recency in logs/S3 listings), that's a defensible trade — but it should then be recorded as an accepted trade-off against URL timestamp leakage, not an accident of the prototype's generator.

quillaid:
Addressed the consistent-hashing readiness follow-up in 128d59fc.\n\nChanges made:\n- added a Decision 7 subsection that keeps leases as authority and hashing as a future placement hint only\n- made notebook_id the only future placement unit, with NotebookDoc/RuntimeStateDoc/CommsDoc explicitly co-located\n- documented RoomRegistry resolution as the single post-auth admission seam and room eviction as a day-one primitive\n- kept redirect/handoff/ring/forwarding/room_leases machinery out of v1\n- added holder/generation metadata to latest.json checkpoint guidance for future lease fencing\n\nAlso rebased onto current origin/main. Validation: cargo xtask lint --fix.

quillaid:
## Review status after 128d59fc

The placement-seams subsection in Decision 7 resolves the consistent-hashing follow-up faithfully: leases-as-authority framing, `notebook_id` as the only placement unit with the three docs explicitly co-located, `RoomRegistry` resolve as the single post-auth admission gate with the `Owned | OwnedElsewhere` shape, day-one eviction, dumb-client reconnect, and the explicit not-now list. The Decision 8 fencing guidance also landed well — making holder/generation nullable or sentinel-valued in v1 is the right refinement so the single-host service doesn't carry fake lease semantics.

One item from the earlier comments remains open: the **notebook ID format decision** (timestamps in the comment thread overlapped — it was posted two minutes before 128d59fc's summary). The current draft still records "Notebook ids are ULID-like and time-prefixed, so any future placement must hash; it must not range-partition by id." That's accurate for today's generator, but per the ID-format comment above, this ADR is the natural place to either (a) switch new IDs to uniformly random 26-char Crockford base32 and replace that sentence with "IDs are opaque and uniformly random; nothing infers time, order, or placement from ID structure," or (b) keep time-prefixed IDs and record the URL creation-timestamp leakage as an accepted trade-off. Either resolution is fine; it just shouldn't stay an accident of the prototype.

No other open findings from my side. Once the ID decision is recorded one way or the other, this reads ready for @rgbkrk's final pass.

quillaid:
Added the hosted notebook identity/persistence update in 76a146bc.

This now records:
- hosted notebooks persist as a cloud-native document set, not as `.ipynb` files
- `notebook_id`, document ids, `revision_id`, `checkpoint_id`, and slugs each have separate authority
- live and published revision routes are identity-first: `/n/:notebook_id/:slug?` and `/n/:notebook_id/r/:revision_id/:slug?`
- slugs are intentionally shareable/decorative by default, not authority
- `RuntimeStateDoc` and `CommsDoc` should be compacted and persisted for checkpoints and published revisions, with process-local runtime claims reconciled on load

Rebased onto current origin/main. Validation: `cargo xtask lint --fix`.

quillaid:
## Decision 7 is good — and it makes the ID-format question smaller, not moot

The identity decision in 76a146bc is a genuinely strong addition: the document-set-not-`.ipynb` persistence boundary, identity-first routes with decorative slugs, reserved control segments, and compacting `RuntimeStateDoc`/`CommsDoc` while reconciling process-local claims on load are all the right calls and go beyond what the review asked for.

But it leaves the ID-format question from the earlier comment open in a way that's now internally inconsistent rather than just unrecorded:

- Decision 7 declares `notebook_id` "the stable **opaque** room/catalog id."
- Decision 8's placement seams still say "Notebook ids are ULID-like and time-prefixed, so any future placement must hash; it must not range-partition by id."

Those two sentences disagree about what "opaque" means. Decision 7 means opaque *in role*; Decision 8 concedes the structure is readable — the first 10 characters decode to a creation timestamp in every link, forever. Opaque-in-role but transparent-in-structure is exactly the state the earlier comment argued against leaving by accident.

Worth noting that Decision 7 already does careful privacy reasoning for the *mutable* part of the URL: "If title sensitivity becomes a product requirement, add an explicit 'copy link without title' affordance later." That affordance works for slugs because a slug can be stripped from a link. The creation timestamp in the ID cannot — it's in the one path segment the router trusts. The ADR currently handles the removable leak thoughtfully and is silent on the permanent one.

The fix is one sentence plus a generator swap, and Decision 7 is the natural home: *"`notebook_id` is uniformly random (26-char Crockford base32, 130 bits); nothing may infer time, order, or placement from ID structure — creation time lives in `created_at`."* Then Decision 8's range-partition caveat deletes itself, because there'd be nothing to range-partition on. Existing ULIDs stay valid as opaque strings; no migration.

Alternatively, if time-readable IDs are wanted for ops ergonomics (eyeballing recency in logs and S3 listings), keep them — but then Decision 7 should say so: *"notebook_id is ULID-shaped; the embedded creation timestamp in shared URLs is an accepted trade-off for operational legibility."* Either sentence closes the question. The only outcome worth avoiding is the current one, where the ADR asserts opacity in one decision and disclaims it in the next.

quillaid:
Addressed the final ID-format inconsistency in fd9dc2ca.

Changes made:
- new hosted notebook ids are uniformly random 26-character Crockford base32 strings with 130 bits of randomness
- `created_at` owns creation time; ids do not encode time, order, tenancy, or placement
- existing prototype ULID-shaped ids remain valid as opaque legacy strings with no special routing or ordering behavior
- Decision 8 now says future placement hashes the full id and must not infer structure
- the implementation plan calls out the random hosted notebook id generator

Self-review:
- searched for stale ULID/time-prefix wording; only the intentional legacy compatibility note remains
- checked decision numbering after the inserted identity decision
- ran `git diff --check`
- validation: `cargo xtask lint --fix` passed

Issue #3421, opened 2026-06-05
Local-first notebook state: browser-persisted Automerge, offline edits, unified connection/identity (desktop & cloud)


rgbkrk:
## Problem

On a thin or flapping network the cloud viewer degrades badly: a dropped WebSocket or a failed asset fetch leaves a blank "live room needs attention" viewer, and there is no local persistence, so disconnecting can strand in-flight edits. Desktop has the *same shape* of concern (the runtime daemon being up or down), so this should be modeled once, not patched as a cloud-only special case, especially now that a desktop client can attach to a remote notebook.

## Direction: connection-aware, local-first notebook state

The same model on desktop and cloud:

1. **Local WASM does the work; only Automerge crosses the wire.** The room/doc logic (`runtimed-wasm`) runs locally in the client. The only thing syncing over the network is Automerge sync messages (small deltas), not server round-trips or repeated asset fetches. On a thin link, only a few WS bytes need to make it across.

2. **Browser-persisted Automerge (the automerge-repo pattern).** Persist the notebook doc to browser storage (IndexedDB), the way automerge-repo's storage adapter does, so a user can disconnect, keep editing locally, and have those edits apply and sync on reconnect. Today a disconnect can lose edits.

3. **One connection-state model, surfaced once.** Model connection/runtime health (daemon up/down on desktop; room/WS up/down on cloud) as projected room state, and render it in a single place: a self-identity + connectivity slot in the upper-right. The slot is **conditional**, absent on a purely local desktop session, present when there is a remote/collaborative context (cloud viewer, or a desktop attached to a remote notebook). Earlier designs were pulled for not matching the app's aesthetic; redo to fit.

4. **Graceful degradation, not hard failure.** Retry asset/import/sync failures with backoff and show a coherent reconnecting/degraded state, instead of a blank viewer or a cascade of cross-origin iframe errors when a renderer asset fails to load.

## Notes

- This parallels the runtime_peer disconnect-detection and state-propagation work in #16: the same "detect a disconnect, project the state up" shape, on the client side.
- An interim cache fix already shipped (renderer assets now revalidate instead of being pinned `immutable`); content-hashing the runtimed WASM (reusing the `?v=<hash>` pattern `sift_wasm` already uses) is the durable follow-up.
- Related: #3381 (cloud compute), #3387 (remote-workstation ADR).

This can be split into focused issues (browser persistence, connection-state model + slot, resilience/retry) as each is scoped; filing it as one to capture the direction.

## Progress

[...]

PR #3580, opened 2026-06-11
fix(runtimed-wasm): preserve handle actor across doc loads and recovery rebuilds

quillaid:
## Summary

Long-tail follow-up to #3579 (slider edits rejected with `actor label "<random hex>" is invalid`). The hotfix made `NotebookHandle.set_actor()` cover the CommsDoc; this closes the two sibling holes with the same root shape — a doc replaced via bare `automerge::AutoCommit::load(bytes)` gets a fresh **random** actor, silently discarding the authenticated `<principal>/<operator>` actor the handle had:

1. **Load order** — `NotebookHandle::load_state_doc` / `load_comms_doc` now capture the current actor before the load and restore it after. Today's callers happen to load before `set_actor`, so #3579 masks this; any future caller that sets the actor first would have silently regressed.
2. **Error recovery (the live failure loop)** — the `rebuild_doc` / `rebuild_state_doc` / `rebuild_comms_doc` / `rebuild_pool_doc` helpers fire when a sync message fails to apply. They rebuilt via bare save→load, so one transient sync error left the client authoring under a random actor, and every subsequent widget edit got rejected by the room: error → rebuild → invalid actor → more rejections, self-reinforcing until page reload. They now delegate to the crate-level `rebuild_from_save` methods (`NotebookDoc` / `RuntimeStateDoc` / `CommsDoc`), which already preserve the actor — and for `rebuild_doc`, also refuse cell-losing rebuilds. `PoolDoc` has no crate helper, so its rebuild preserves the actor manually.

Five new unit tests pin the invariant: `set_actor` covers all docs on both handle types, both `load_*_doc` setters preserve the actor, and the recovery rebuild preserves the actor.

## Note for the IndexedDB persistence workstream (lab1)

Deliberately **not** included here: persisted seeds containing changes authored under a bad actor are *permanently* unacceptable to the room — `shouldDiscardPersistedSeedOnRejection` in `apps/notebook-cloud/viewer/cloud-viewer-session.ts` should treat actor-label rejections (`actor label .* is invalid`, `actor principal .* is not authorized`) as an unconditional poison-pill discard, and persistence may want to record the authoring actor with the seed. This PR fixes the *sources* of bad actors in runtimed-wasm; the persistence-side defense belongs with the in-flight IndexedDB work to avoid colliding with it.

Also dormant but worth knowing: `RoomHostHandle::load_comms_doc` / `load_snapshot` still give the room host itself a random comms actor. Nothing validates the room host's own actor today, so no change here — but if the room host ever authors comms changes that peers validate, it needs the same treatment.

## Validation

- `cargo test -p runtimed-wasm` — 39/39 (34 existing + 5 new actor tests)
- `cargo test -p runtime-doc` — 225/225 (regression guard on the delegation targets)
- `cargo xtask lint --fix` — clean, no resulting diff

### DIFF
--- a/crates/runtimed-wasm/src/lib.rs
+++ b/crates/runtimed-wasm/src/lib.rs
@@ -2205,9 +2205,15 @@ impl NotebookHandle {
     /// (outputs, executions) alongside the notebook doc. Replacing the state
     /// doc also resets RuntimeStateDoc sync state so later room-host sync starts
     /// from the loaded snapshot, not the previous empty/bootstrap doc.
+    ///
+    /// Preserves the current actor so a load never silently discards an
+    /// authenticated actor set via `set_actor` (room hosts reject
+    /// non-`<principal>/<operator>` actors).
     pub fn load_state_doc(&mut self, bytes: &[u8]) -> Result<(), JsError> {
-        let doc = automerge::AutoCommit::load(bytes)
+        let actor = self.state_doc.doc().get_actor().clone();
+        let mut doc = automerge::AutoCommit::load(bytes)
             .map_err(|e| JsError::new(&format!("load_state_doc failed: {}", e)))?;
+        doc.set_actor(actor);
         self.state_doc = RuntimeStateDoc::from_doc(doc);
         self.state_sync_state = sync::State::new();
         self.prev_output_by_id.clear();
@@ -2216,9 +2222,15 @@ impl NotebookHandle {
     }
 
     /// Load a CommsDoc from saved bytes.
+    ///
+    /// Preserves the current actor so a load never silently discards an
+    /// authenticated actor set via `set_actor` (room hosts reject
+    /// non-`<principal>/<operator>` actors).
     pub fn load_comms_doc(&mut self, bytes: &[u8]) -> Result<(), JsError> {
-        let doc = automerge::AutoCommit::load(bytes)
+        let actor = self.comms_doc.doc().get_actor().clone();
+        let mut doc = automerge::AutoCommit::load(bytes)
             .map_err(|e| JsError::new(&format!("load_comms_doc failed: {}", e)))?;
+        doc.set_actor(actor);
         self.comms_doc = CommsDoc::from_doc(doc);
         self.comms_sync_state = sync::State::new();
         Ok(())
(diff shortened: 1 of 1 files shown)


quillaid:
**Coordination note for the persistence-side follow-up** (from the #3421 stack workstream — for whichever agent picks this up):

The "Note for the IndexedDB persistence workstream" above is accepted as a carve-off. Scope it as ONE self-contained PR against main:

1. **Unconditional poison-pill on actor rejections.** In `apps/notebook-cloud/viewer/cloud-viewer-session.ts`, the recoverable-rejection handling (`CloudRecoverableRejectionTracker` flow + `shouldDiscardPersistedSeedOnRejection`) currently strike-counts ALL recoverable `AUTOMERGE_SYNC` rejections: first strike → in-place `resetAndResync`, delivery-gated repeat → teardown + seed discard. For rejection reasons matching `actor label .* is invalid` or `actor principal .* is not authorized`, skip the resync-first dance entirely: a seed carrying bad-actor changes can never become acceptable, so escalate immediately to the existing discard path (dispose-flush → clear → skip-seed-once ordering already exists — reuse it, don't reimplement). Plain transport divergence keeps the current strike behavior.
2. **Record the authoring actor in the envelope meta.** `packages/runtimed/src/persistence/notebook-doc-persistence.ts` — add optional `actorLabel` to `NotebookDocPersistenceMeta` (additive: no `schemaVersion` bump; decoder must tolerate its absence in existing records). The cloud session passes the current connection's actor label. This gives future seed-time validation something to check against.
3. **Tests:** actor-rejection → immediate discard with the dispose-flush-then-clear ordering preserved (no strike absorbed first); non-actor rejection keeps strike behavior; meta roundtrip with and without `actorLabel`.

**Do-not-touch (in-flight on the #3421 stack, will conflict):** `isTransportReconnectError` / the sustained-reconnecting notice routing, `NotebookConnectionIdentity`, `CloudConnectionStatusBridge`, anything in `src/components/isolated/` or `runtimed-wasm-client.ts` (PR 4 in review), and the `armPersistence` body beyond threading the actor label. Notice/UI treatment of the discard is deliberately out of scope here.

Context docs in-repo: `docs/adr/local-first-notebook-state.md` (PR 1/PR 2 sections explain the poison-pill rails and why discard ordering matters).

PR #3605, opened 2026-06-12
feat(notebook-cloud): resync heal loop with convergence verification and quiet exhaustion

quillaid:
The final live slice of #3585 — the two operational patterns from the automerge-repo/subduction deep-read, applied to our own shapeless spot: **resync verification**.

## Resync heal loop

`resetAndResync()` was fire-and-forget — nothing verified that a recovery actually converged. Now every fire-and-forget sync exchange arms a per-doc verification deadline in the upstream `SyncScheduler` shape: the **initial bootstrap** (the compact adversarial review's confirmed major: the replayed first handshake trips the `applyRoomReady` peer-id dedupe, so roomReady arming never covers the cold load — the most common connection event was the one left unverified, exactly the scenario where instant paint shows editable content under an honest *online* dot while a wedged room host answers liveness pings without converging), the per-roomReady re-establish, and the rejection tracker's strike-1 in-place recovery.

The confirmation primitive is `notebook_doc_caught_up()` polled on `notebookSyncApplied$` (#3588's pair) — *"confirmation is just another sync"*: a re-verification that returns true iff the previous exchange landed. Not converged by deadline → re-kick on a bounded ladder (2s→60s, ×2, ±25% jitter, 10 attempts) → exhaustion is ONE quiet signal: "Sync is stalled. / Your edits are kept locally." (info tone, yields to the sustained-reconnecting line, clears on late convergence) + a `[sync-heal]` log.

Interplay, all test-pinned: heal kicks never count as rejection-tracker strikes; `shouldKick` gates on the link being online (held deadlines don't consume attempts while reconnecting); roomReady resets the ladder but a standing exhaustion clears only on real convergence (same stall, one signal); exhaustion never touches the tab-bridge quarantine. Per-doc keyed — RuntimeStateDoc/Comms adoption recorded open in the ADR.

## Persistence self-disable: one second chance

The controller's 3-failure self-dispose was terminal-silent. It now signals `onSelfDisabled` once per controller pair, and a `PersistenceRearmGate` grants exactly one re-arm on the next `online` transition or heal recovery — a transient IDB hiccup gets a second chance; a genuinely dead backend stays disabled after one retry.

## Review

Compact adversarial pass (1 finder + verification): 1 confirmed major (the initial-bootstrap arming gap above — fixed with a wiring pin and ADR acknowledgment), 1 refuted (the cross-reset starvation claim — the notice survives ladder resets once fired, and flapping *accelerates* healing). Also clears the four pre-existing lint warnings from #3595's tests.

## Verification

vp suite **2244 passed, 0 failed** · notebook-cloud **946 passed, 0 failed** (sync-heal 16, notices 21) · no Rust changes · `cargo xtask lint` **0 warnings** · ADR updated.

With this, #3585 reduces to its watch-list (fragments successor + subduction protocol items).

### DIFF
--- a/apps/notebook-cloud/test/instant-paint.test.ts
+++ b/apps/notebook-cloud/test/instant-paint.test.ts
@@ -691,7 +691,7 @@ describe("empty live room displacement", () => {
   it("kicks one full materialization when the doc first catches up to the room", () => {
     assert.match(
       sessionSource,
-      /notebookSyncApplied\$\.subscribe\(\(\) => \{\s*if \(caughtUpMaterializeKicked\) return;\s*if \(!cloudNotebookHandleCaughtUp\(liveRuntime\.handle\)\) return;\s*caughtUpMaterializeKicked = true;\s*materializeLiveCellsSafely\(liveRuntime\);/,
+      /notebookSyncApplied\$\.subscribe\(\(\) => \{\s*const caughtUp = cloudNotebookHandleCaughtUp\(liveRuntime\.handle\);[\s\S]{0,400}?if \(caughtUpMaterializeKicked\) return;\s*if \(!caughtUp\) return;\s*caughtUpMaterializeKicked = true;\s*materializeLiveCellsSafely\(liveRuntime\);/,
     );
   });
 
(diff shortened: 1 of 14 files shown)


rgbkrk:
@pullfrog is this a smart move for us?

PR #3607, opened 2026-06-12
docs(adr): define cloud-connected local MCP clients

quillaid:
## Summary

Adds a draft ADR for connecting the existing local `runt mcp` server directly to hosted notebook rooms such as `https://preview.runt.run/n/<id>`, instead of requiring one MCP registration per notebook or relying on a remote MCP facade.

The ADR records:

- `connect_notebook(target=...)` as the preferred addressable entrypoint while preserving `path` and `notebook_id` compatibility.
- Hosted notebook URLs and remote aliases as room locators, not MCP resource URIs.
- A separate machine-local remote registry for hosted origins and credential references, rather than putting API keys into synced settings.
- Principal/operator handling: principal comes from the credential, operator comes from local config, and live Automerge actors still need nonce/persistence protection.
- The separation between document access and workstation/runtime attachment.
- Rejoin and parked-session implications for remote room targets.

## Why

This is the next likely foundation for agent collaboration against cloud notebooks: a local stdio MCP process should be able to sync directly with the hosted Automerge document set using local credentials and operator attribution.

## Review focus

- Does the settings/config split match the existing synced settings model?
- Is the target-resolution model compatible with current `connect_notebook` semantics?
- Are the principal/operator and actor uniqueness constraints stated clearly enough?
- Are the non-goals strict enough to avoid drifting into remote MCP or workstation authority changes too early?

## Validation

- `cargo xtask lint --fix`

### DIFF
--- a/docs/adr/cloud-connected-local-mcp.md
+++ b/docs/adr/cloud-connected-local-mcp.md
@@ -1,0 +1,339 @@
+# Cloud-Connected Local MCP Clients
+
+**Status:** Draft, 2026-06-12.
+
+## Context
+
+nteract's MCP server is currently local-first: one stdio MCP process talks to a
+local `runtimed` daemon over a Unix socket, holds one active
+`NotebookSession`, and exposes tools such as `connect_notebook`,
+`create_notebook`, `insert_cell`, and `execute_cell`. That model works well for
+local notebooks because the agent edits the same Automerge document set the
+desktop app sees.
+
+Hosted notebooks at `preview.runt.run` now use the same document split:
+`NotebookDoc`, `RuntimeStateDoc`, and `CommsDoc` travel over typed-frame v4
+WebSockets. The browser viewer and workstation/runtime peers already prove that
+native clients can speak the hosted room protocol with explicit credentials,
+connection scope, principal, and operator attribution.
+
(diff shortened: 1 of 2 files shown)


docs/adr/cloud-connected-local-mcp.md
```diff
@@ -0,0 +1,294 @@
...
+`path`, or `notebook_id` may be provided.
+
+Resolution order:
+
+| Input | Connector |
+| --- | --- |
+| `path` or path-like `target` | Local daemon `connect_open(path)` |
+| `notebook_id` or local UUID target | Local daemon `connect(uuid)` |
+| `https://<host>/n/<id>` or `https://<host>/n/<id>/<slug>` | Hosted room on the matching remote |
+| `<remote-alias>:<notebook-id>` | Hosted room using the configured remote alias |
```
rgbkrk: To keep initial work simple and to prevent having to maintain extra, let's not introduce remote aliases just yet. We can reserve the space but not support it.

rgbkrk:
@pullfrog look over this draft plan of how we'll connect desktop to remote instances

rgbkrk:
I feel like we should skip the new tool endpoints for listing and instead allow an additional optional parameter for the domain (which can include desktop/local as an option).

PR #3609, opened 2026-06-12
feat(notebook-cloud): workstation pairing codes mint scoped workstation credentials

quillaid:
## Summary

Workstation registration previously required an externally issued bearer (Anaconda API key / OIDC token) or a dev token plus a hand-seeded ACL row. This PR adds the provider-neutral pairing flow from #3381's registration UX track:

1. A signed-in owner mints a **pairing code** — `POST /api/workstations/pairing-codes` — short-lived (10 min), single use, 12 chars from an unambiguous alphabet (~59 bits), stored hashed.
2. The remote agent redeems it — `POST /api/workstations/pairing-codes/redeem` (the code is the entire credential for this call) — and receives a long-lived **workstation credential** (`nwc_` prefix, 256-bit random, stored hashed, revocable via `revoked_at`).
3. The agent registers/heartbeats with `Authorization: Bearer nwc_...`; the browser polls `GET /api/workstations/pairing-codes/:id` and watches `pending → redeemed → registered`.

## Least privilege

The credential authenticates **only the workstation surface**: registration/heartbeat, attach-job polling/patching, and `runtime_peer` room dials. The gate is default-deny in the central request-auth wrapper (`authenticateRequestOrResponse`), so every route that doesn't explicitly opt in rejects the prefix before provider dispatch — a leaked agent token cannot create notebooks, mutate ACLs, or read account state. Scope requests are clamped to `viewer` / `runtime_peer`.

Because the credential maps to the minting owner's canonical principal, the existing attach path (owner-principal `runtime_peer` ACL row) and the Rust cloud transport's `Authorization`-header WS dial work unchanged.

## Contract record

`docs/adr/hosted-credential-transport.md` Decision 9, including why codes are 59-bit (first registration auto-selects the default workstation target — a guessed code must not be able to plant an attacker workstation), why redeem responses don't oracle live codes, and where ticket-based hardening (Decision 4) still applies.

## Tests

8 new route tests in `worker-routes.test.ts` (mint auth, full happy path, replay, expiry, code normalization, least-privilege denials, scope clamp, attach-job polling with the credential) plus FakeD1 emulation for the new statements. 120 pass / 0 fail.

Part of #3381 (step 6 — registration/attachment credentials). UI follow-up (panel pairing card) and `runt workstation` CLI are stacked next.

### DIFF
--- a/.claude/.gitignore
+++ b/.claude/.gitignore
@@ -1,0 +1,1 @@
+worktrees/
--- a/apps/notebook-cloud/src/identity.ts
+++ b/apps/notebook-cloud/src/identity.ts
@@ -43,7 +43,13 @@ export interface IdentityEnvironment {
 }
 
 export interface AuthenticatedConnectionMetadata {
-  provider: "anonymous" | "app-session" | "dev" | "oidc" | "anaconda-api-key";
+  provider:
+    | "anonymous"
+    | "app-session"
+    | "dev"
+    | "oidc"
+    | "anaconda-api-key"
+    | "workstation-credential";
   transport:
     | "anonymous"
     | "app-session-cookie"
@@ -52,11 +58,14 @@ export interface AuthenticatedConnectionMetadata {
     | "dev-token-subprotocol"
     | "api-key-bearer"
     | "oidc-bearer"
-    | "oidc-subprotocol";
+    | "oidc-subprotocol"
+    | "workstation-credential-header";
   principalNamespace: string;
   displayName?: string;
   email?: string;
   emailVerified?: boolean;
+  workstationCredentialId?: string;
+  workstationPairingCodeId?: string;
 }
 
 const ANONYMOUS_PRINCIPAL_PREFIX = "anonymous:";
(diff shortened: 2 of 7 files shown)


quillaid:
Adversarial review (3 lenses: auth surface, correctness/races, API contract) confirmed three findings; all fixed in the latest push:

1. **Revocation was documented but unimplemented** — Decision 9 leaned on `revoked_at`, which nothing wrote. Now real: `GET /api/workstations/credentials` + `POST /api/workstations/credentials/:id/revoke`, owner-auth only (deliberately closed to workstation credentials, so a stolen agent token cannot enumerate or revoke its siblings). Revoked bearers lose the surface immediately; covered by tests.
2. **Non-atomic redeem could burn a code** — consume (UPDATE…RETURNING) and mint (INSERT) were two D1 round-trips; a transient failure between them left the code dead with no credential and the panel stuck at "registering…". Now one `DB.batch` transaction, with the INSERT joined on a per-request `redeemed_by_credential_id` marker instead of the redeem timestamp — which also closes a same-millisecond concurrent-redeem race the naive timestamp join would have introduced.
3. **Hook state machine had zero coverage** (stacked UI PR) — added `use-cloud-workstations.test.tsx` rendering the real hook with the client mocked: mint success/failure, poll transitions with terminal shutdown, abort on unmount, client-driven expiry flip, display-name resolution.

Worker suite: 123 pass / 0 fail. Stack: panel UI in quillaid#6, `runt workstation` CLI in quillaid#7 — both based on this branch, to be promoted here as it merges.

PR #3612, opened 2026-06-12
feat(install): macOS support in the release installer

quillaid:
## Summary

The release installer now handles macOS (Apple silicon + Intel) alongside Linux x64, so the `sh.nteract.io` one-liner — and the `--headless` workstation path — work on a Mac.

- **Desktop**: downloads `nteract-<channel>-darwin-<arch>.app.tar.gz` (the signed bundle already carries the `runt`/`runtimed`/`nteract-mcp` sidecars in `Contents/MacOS`), installs it to `~/Applications` (where the channel's app-launch candidates look), and links the commands from the bundle into `~/.local/bin`.
- **Headless** (`--headless`): keeps the bundle under the prefix instead — a workstation box never grows an Applications entry — and links only the commands. The completion hint now leads with the `runt workstation connect` pairing flow (#3611).
- The daemon-service step is unchanged: `runt daemon doctor --fix` already speaks launchd.
- The script keeps its historical `install-linux-release` name because releases and the `sh.nteract.io` bootstrap fetch that asset name; the header comment records this.
- Docs: README + `docs/remote-workstation.md` now lead with the live `sh.nteract.io` entrypoint (the "once the domain is live" note was stale — it serves today) and the Linux-or-macOS framing.

## Tests

`scripts/test-install-release` runs the real installer against `--from-dir` fixtures with a sandboxed `HOME`; Darwin paths run on any host via a PATH-shimmed `uname`. Covers: linux desktop, darwin desktop, darwin reinstall (atomic bundle replace), darwin headless, unsupported-platform rejection. All green.

## Sequencing note

`sh.nteract.io` fetches this script from the **latest release's** assets, so macOS support flows through the one-liner once the next release carries it. A companion install-sh PR extends the bootstrap's platform gate and adds a self-healing fallback to `main`'s script for Darwin until then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,20 +8,24 @@ Built on [jupyter-zmq-client](https://crates.io/crates/jupyter-zmq-client) and [
 
 Download the latest release from [GitHub Releases](https://github.com/nteract/nteract/releases).
 
-Linux desktop users should use the AppImage from GitHub Releases, or run
-`scripts/install-linux-release` to install the AppImage plus CLI/daemon and
-the per-user systemd service in one step. DEB/RPM/APT installs are not
-currently supported because `runtimed` is a per-user daemon managed by the app
-and CLI, not by system package-manager scripts.
+Linux x64 and macOS (Apple silicon or Intel) can install with one command —
+the AppImage or signed .app bundle plus CLI/daemon and the per-user service
+(systemd or launchd) in one step:
 
-For remote Linux workstations (Outerbounds, JupyterHub) that offer compute to
+```bash
+curl --proto '=https' --tlsv1.2 -sSf https://sh.nteract.io | bash
+```
+
+macOS users can equally drag the DMG from GitHub Releases. DEB/RPM/APT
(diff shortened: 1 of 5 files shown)


quillaid:
Back to **~/Applications** as the default — per-user by design, matching the daemon service and bin links; `--applications-dir` stays as the override. (The /Applications detour is reverted; squash will fold it away.)

Also traced the service question through the code: the script's existing `runt daemon doctor --fix` is the complete macOS story. Doctor resolves the daemon via the **sibling of the running runt** — so from `<bundle>/Contents/MacOS/runt` it finds the bundled `runtimed` whether the bundle is in ~/Applications or under the headless prefix. `ServiceManager::install` sees an in-bundle source and points the launchd plist **directly at it** (no copy); the legacy `~/Library/LaunchAgents` plist is always written as primary registration with SMAppService as best-effort Login-Items sugar; doctor then starts the agent (unless `--no-start`) and the script's `runt daemon status` confirms. Upgrades ride the existing version-mismatch path with the atomic new-inode swap. Six fixture scenarios green.

PR #3616, opened 2026-06-12
perf(runtime): batch initial widget comm writes

quillaid:
## Summary

This makes the shared runtime widget `comm_open` path faster by using Automerge's batched object creation for fresh JSON trees.

- adds `automunge::put_json_at_key_batched` for new/replacement map keys
- uses it for initial RuntimeStateDoc comm topology entries
- uses it for initial CommsDoc comm state, while keeping existing comm updates on `update_json_at_key` so mutable widget object identity is preserved
- adds a reproducible runtimed measurement harness for widget comm-open write paths

## Local measurement

Command:

```bash
cargo test -p runtimed print_comm_open_measurement_paths -- --ignored --nocapture
```

On lab2, the synthetic 128-widget/1KB-state comm-open path improved from the earlier baseline of about `2512ms` total to about `290ms` total after the batched JSON writes.

Current branch timing:

| Count | RuntimeStateDoc | CommsDoc | Total |
|---:|---:|---:|---:|
| 32 | 5.8ms | 16.3ms | 27.6ms |
| 128 | 54.1ms | 215.4ms | 289.6ms |

## Validation

- `cargo test -p automunge put_json_at_key_batched -- --nocapture`
- `cargo test -p runtime-doc comm -- --nocapture`
- `cargo test -p runtimed output_commit_measure -- --nocapture`
- `cargo test -p automunge`
- `cargo test -p runtime-doc`
- `cargo test -p runtimed`
- `cargo xtask lint --fix`
- `cargo build -p runtimed -p runt`

## Notes

The batched helper is deliberately documented as safe only for fresh or replaceable object trees. Shared/mutable existing JSON objects should continue to use `update_json_at_key`.

### DIFF
--- a/crates/automunge/src/lib.rs
+++ b/crates/automunge/src/lib.rs
@@ -8,7 +8,10 @@
 //! can cover the workspace Automerge API we need. Until then, this is the
 //! single source of truth for JSON/Automerge conversion.
 
-use automerge::{transaction::Transactable, AutoCommit, AutomergeError, ObjId, ObjType, ReadDoc};
+use automerge::{
+    hydrate, transaction::Transactable, AutoCommit, AutomergeError, ObjId, ObjType, ReadDoc,
+    ScalarValue,
+};
 
 /// Read a string scalar at `prop`, distinguishing "absent" from "empty".
 ///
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog verify that this isn't a regression from 927a44ee705b3daf7b4a7d7759b9a5c8d7aa528d. Make suggestions if you see any for improving the performance here.

quillaid:
Checked Pullfrog's hydrate Map allocation note against the pinned Automerge fork (`992effb`). `hydrate::Map` exposes `From<HashMap<&str, Value>>` and `From<HashMap<String, Value>>`; the inner map is private and there is no Vec/from-iterator constructor we can use from `automunge`, so the transient HashMap is unavoidable without changing the fork API. I agree the cross-comm transaction batching point is a follow-up; this PR intentionally lands the intra-record `batch_create_object` win first.

PR #3619, opened 2026-06-12
docs(memos): introduce shared memo space

quillaid:
## Summary

- introduce `docs/memos/` as shared thinking space for product, design, engineering, research, and AI collaborators
- promote the Markdown Plan document research note into `docs/memos/markdown-plan-documents.md`
- point the ADR register at memos for exploratory framing before durable decisions graduate into ADRs

## Verification

- `cargo xtask lint --fix`

PR #3620, opened 2026-06-12
docs: organize documents by kind

quillaid:
## Summary

- add a root docs index and first-class homes for PRDs, plans, audits, measurements, and runbooks
- move self-declared non-ADR documents out of `docs/adr/` into their matching homes
- keep `docs/adr/` focused on durable technical decisions and update cross-links after the moves

## Verification

- docs local Markdown link check
- `cargo xtask lint --fix`

Follow-up to #3619.


rgbkrk:
@pullfrog review the locations of these documents being accurate for placement and that the content has not changed to be inaccurate relative to where the repo is at and going.

quillaid:
Addressed PullFrog's stale neighbor-reference note in f22bb7f4.

- updated moved audit/PRD prose refs to explicit `../adr/...` paths
- included the lower-priority inline refs PullFrog called out
- reran docs Markdown link check and `cargo xtask lint --fix`

PR #3621, opened 2026-06-12
docs: update agent documentation taxonomy guidance

quillaid:
## Summary

- add the new documentation taxonomy to the root `AGENTS.md` agent map
- update the frontend-dev skill to route exploration through `.context/` and `docs/memos/` before graduating durable decisions/requirements/plans
- confirm no stale moved-doc `docs/adr/...` paths or old "graduate to ADRs" guidance remain in tracked agent/skill/rule docs

## Verification

- focused `git grep` checks for stale guidance and moved-doc paths
- `cargo xtask lint --fix`


quillaid:
Updated after Quill's note: the guidance now says agents should start working notes in `.context/`, avoid persistent docs for routine implementation/test notes, and only promote to `docs/` when the artifact should outlive the immediate change.

quillaid:
Adjusted after the frontend-dev concern: removed the general docs taxonomy bullets from `.agents/skills/frontend-dev/SKILL.md` and added a focused `.agents/skills/architecture/SKILL.md` instead. Root `AGENTS.md` still keeps the minimal always-visible rule: start in `.context/`, promote into docs only when the artifact should be durable for product, design, engineering, research, or AI collaborators.

PR #3624, opened 2026-06-12
fix(workstations): drop inactive attach jobs

quillaid:
## Summary
- classify workstation attach-job PATCH failures using structured HTTP status/body data
- treat `409 workstation attach job is no longer active` as a terminal local job state
- drop and terminate local runtime peers for replaced/cancelled jobs so the agent stops heartbeating stale work
- keep retryable/server failures on the existing error/cooldown path

## Validation
- `cargo test -p runtimed workstation::agent_loop`
- `cargo xtask lint --fix`

## Context
This came out of the hosted workstation toolbar smoke while validating #3623. The smoke intentionally creates replacement attach jobs for restart and restart-and-run-all. When the room replaces an earlier job, the workstation agent currently keeps the cancelled job in its active map and repeatedly logs 409s from heartbeat/exit patches. This keeps the replacement path noisy and can leave old runtime peers around longer than needed.

### DIFF
--- a/crates/runtimed/src/workstation/agent_loop.rs
+++ b/crates/runtimed/src/workstation/agent_loop.rs
@@ -412,6 +412,8 @@ pub fn parse_response_body(text: &str) -> Value {
 pub struct AgentHttpError {
     pub message: String,
     pub retry_after_ms: u64,
+    status: Option<u16>,
+    body: Value,
 }
 
 impl AgentHttpError {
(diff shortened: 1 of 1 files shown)


rgbkrk:
@pullfrog review please

quillaid:
Addressed Pullfrog's inline note in 23f1d727: replaced-job cleanup now awaits a bounded child reap after signalling, so the workstation agent does not leave a killed runtime peer as a zombie.

PR #3625, opened 2026-06-12
fix(cloud): stop persisting transient room frames

quillaid:
## Summary
- remove Durable Object per-frame history writes from the notebook room hot path
- keep authenticated presence live-only: broadcast + ack, no storage row
- keep materialized document recovery on the RoomMaterializer checkpoint path instead of a separate frame log
- update hosted-room authorization docs to spell out the document checkpoint boundary

## Validation
- pnpm --dir apps/notebook-cloud test test/notebook-room.test.ts (984 tests passed; package script runs the full cloud test suite)
- cargo xtask lint --fix

## Notes
This removes the old `frame:<sequence>` / `frame_sequence` metadata writes. Those rows were not read for room replay or recovery; they only amplified Durable Object storage writes for accepted live frames.


rgbkrk:
@pullfrog go deep on this one, we're cleaning up our use of Durable Objects to not do unnecessary writes

PR #3628, opened 2026-06-13
fix(workstations): stabilize hosted runtime peers

quillaid:
## Summary
- add a per-workstation advisory singleton lock for the hosted workstation agent so one local workstation identity cannot be served by multiple runner processes
- replace older runtime peers in a room when a newer peer connects with the same workstation id, including restored hibernated peers, while leaving other workstation ids/kernels alone
- shut down replaced/inactive runtime-peer children gracefully so `cloud-runtime-agent` reaches its normal kernel cleanup path before the workstation agent falls back to a hard kill
- add timing/session-cookie support to the hosted workstation toolbar smoke harness for latency/debug runs

## Validation
- `cargo xtask lint --fix`
- `cargo test -p runtimed workstation::agent_loop::tests --lib`
- `cargo test -p runtimed runtime_agent --lib`
- `pnpm --dir apps/notebook-cloud test -- notebook-room` (992/992 cloud tests passed)
- rebuilt local `runt`/`runtimed`, started one `ws-lab2` workstation runner, and verified a second runner exits with `another workstation agent is already running for ws-lab2`

## Notes
- Pullfrog's hibernation-restore edge is addressed: restored duplicate runtime peers collapse to the newest peer for the same notebook/workstation key before room-host sync/publish.
- Live smoke found that hard-killing inactive `cloud-runtime-agent` jobs could orphan launched kernel children; this branch now sends SIGTERM first and runtime agents exit through their normal `k.shutdown().await` cleanup path.

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-toolbar-smoke.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-toolbar-smoke.mjs
@@ -18,6 +18,8 @@ import {
   writeSmokeJsonReport,
 } from "./smoke-paths.mjs";
 
+const smokePhaseTimings = [];
+
 if (isMainModule()) {
   main().catch((error) => {
     console.error(error instanceof Error ? error.message : String(error));
@@ -57,6 +59,10 @@ async function main() {
   const source =
     process.env.NOTEBOOK_CLOUD_WORKSTATION_TOOLBAR_SMOKE_CODE ??
     `print(${JSON.stringify(runMarker)})`;
+  const runOutputProbes = parseBoolean(
+    process.env.NOTEBOOK_CLOUD_WORKSTATION_TOOLBAR_SMOKE_OUTPUT_PROBES,
+    false,
+  );
   const title =
     process.env.NOTEBOOK_CLOUD_WORKSTATION_TOOLBAR_SMOKE_TITLE ??
     `Toolbar attach smoke ${new Date().toISOString()}`;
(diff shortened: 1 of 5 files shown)


rgbkrk:
@pullfrog Focus on:

  - Whether the duplicate-peer room policy is correct: newer runtime peer replaces older only for the same notebook + workstation ID, while still allowing multiple kernels across different notebooks/workstations.

  - Whether closing the older peer with suppressRuntimePeerWatch: true is safe, or if it could mask a real runtime loss.

  - Whether the workstation-agent advisory lock path is the right singleton scope for dev/nightly/installed setups.

  - Whether the smoke harness auth-cookie/timing additions are appropriate in this PR, or should be split out.

  - Any missed edge cases around hibernated Durable Object peers and restored runtime peers with the same workstation ID.

Note: live testing found a separate follow-up, not solved here. Force-killing inactive cloud-runtime-agent children can orphan nteract_kernel_launcher processes; I intentionally did not fold that into this PR.

quillaid:
Followed up on the review findings in the latest commits.

- Hibernation restore now collapses duplicate restored runtime peers by notebook/workstation key before room-host sync/publish, with coverage in `notebook-room.test.ts`.
- The earlier orphaned-kernel follow-up is now addressed too: workstation cleanup sends SIGTERM first, and `cloud-runtime-agent` exits through the normal runtime-agent cleanup path before hard-kill fallback.
- I retitled/reframed the PR as hosted runtime-peer stabilization so the smoke harness timing/auth-cookie additions fit the validation/debug scope instead of reading as duplicate-peer-only scope creep.

Keeping this draft until the new CI run is green.

PR #3650, opened 2026-06-13
fix(notebook-cloud): hide offline chrome while access loads

quillaid:
## Summary
- hide the cloud connection identity and presence slots while catalog access is still loading
- keep the existing hidden behavior for access-blocked pages and visible behavior for normal readable notebook pages
- reuse the computed live-room disabled status for both the session hook and header projection

## Verification
- pnpm --dir apps/notebook-cloud exec node --import tsx --test test/notebook-view-loading.test.ts test/cloud-notebook-catalog-access.test.ts
- pnpm --dir apps/notebook-cloud exec tsc -p tsconfig.json --noEmit
- pnpm --dir apps/notebook-cloud build:viewer
- local Wrangler app-session smoke with delayed /api/n?limit=100 catalog response: pending state showed no visible Offline text
- cargo xtask lint --fix


rgbkrk:
@pullfrog review

PR #3655, opened 2026-06-13
perf(notebook-cloud): preload notebook route from dashboard

quillaid:
## Summary
- Preload the notebook route asset manifest from `/n` when the Worker already bootstraps a signed-in notebook list with rows.
- Keep signed-out and empty dashboard shells light, and keep dashboard HTML free of notebook runtime config.
- Update workstation attachment route expectations for the existing `runtime_session_id` field so the worker route suite is green.

## Verification
- `node --import tsx --test test/html.test.ts test/worker-routes.test.ts`
- `cargo xtask lint --fix`


rgbkrk:
@pullfrog review

PR #3656, opened 2026-06-13
perf(notebook-cloud): allow cached blob fetches

quillaid:
## Summary
- Stop applying `cache: "no-store"` to ordinary cloud `blobResolver.fetch` calls when authenticated binary display URLs are enabled.
- Keep protected binary display URL fetches on their explicit no-store path.
- Add coverage that text/html blob resolution can use the browser HTTP cache while binary display fetches remain protected.

## Live preview evidence
On deployed preview build `38c4dd5e`, opening the workstation smoke notebook showed repeated parent-side HTML blob requests with `cache-control: no-cache` for the same immutable blob hashes. Those responses were Cloudflare cache hits, but the browser still had to make network requests because the resolver forced no-store for all blob fetches.

## Verification
- `node --import tsx --test test/cloud-session-auth-stability.test.ts test/blob-resolver.test.ts`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/src/blob-resolver.ts
+++ b/apps/notebook-cloud/src/blob-resolver.ts
@@ -28,7 +28,6 @@ export function createNotebookCloudBlobResolver(input: {
     url,
     ...(authenticatedBinaryDisplayUrls
       ? {
-          requestInit: { cache: "no-store" },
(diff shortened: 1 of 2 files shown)


rgbkrk:
@pullfrog review

PR #3660, opened 2026-06-13
fix(notebook-cloud): clean up workstation exit messages

quillaid:
## Summary
- format workstation runtime-peer exit messages without empty `signal=` details in both workstation-agent implementations
- use the formatter when attach jobs fail after the spawned peer exits
- add regression coverage for code-only, signal-only, combined, and unknown exits

## Context
Live preview showed a failed workstation start as `Runtime peer exited with code=1, signal=`. The running preview agent is the Rust `runtimed workstation-agent`, which had the same empty-field formatter as the older JS hosted agent. This keeps the same status flow but avoids surfacing blank signal metadata in the notebook Workstations panel.

The runtime-peer log for that failed attach showed `runtime peer session unknown does not match selected runtime session ...`; that was from a stale local workstation-agent process whose executable had already been replaced. After restarting this machine's workstation agent on the current binary, the same owner notebook attached successfully, launched the current-Python kernel, and ran `2 + 2` to `4`.

## Verification
- `cargo test -p runtimed runtime_peer_exit_messages_omit_empty_details`
- `cargo test -p runtimed workstation::agent_loop::tests`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/hosted-workstation-agent.test.mjs`
- `cargo xtask lint --fix`
- Live preview: restarted the local `runtimed workstation-agent`, attached `ws-lab1-quill` to notebook `01KV1BKBQ83XVZ13DZ0T28EWXH`, and executed `2 + 2` via `Control+Enter` with output `4`.

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
@@ -209,6 +209,22 @@ export function retryCooldownMs({
   return Math.min(maxDelay, Math.ceil(baseDelay + jitter));
 }
 
+export function runtimePeerExitMessage(code, signal) {
+  const exitCode = Number.isInteger(code) ? code : null;
+  const exitSignal = typeof signal === "string" && signal.length > 0 ? signal : null;
+
+  if (exitCode != null && exitSignal != null) {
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

PR #3663, opened 2026-06-13
fix(notebook-cloud): close public viewers on revoke

quillaid:
## Summary
- close existing anonymous live-room viewers when an owner removes the public-link ACL row
- add an internal room control path for access revocation that only disconnects anonymous viewer peers
- keep signed-in viewers, owners, editors, and runtime peers connected when public-link access changes

## Live finding
While auditing preview, I enabled public link access on a disposable owned notebook, opened it in a signed-out browser context, then disabled public link access again. The owner page still showed the anonymous tab as present, which means existing anonymous WebSocket sessions can outlive public-link revocation until they disconnect or reload.

## Verification
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/notebook-room.test.ts --test-name-pattern "disconnects anonymous viewers"`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/worker-routes.test.ts --test-name-pattern "closes anonymous live viewers"`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/src/index.ts
+++ b/apps/notebook-cloud/src/index.ts
@@ -3209,6 +3209,9 @@ async function routeNotebookAcl(request: Request, env: Env, notebookId: string):
       counter: "acl_revocations",
       counter_delta: 1,
     });
+    if (revoked && isPublicViewerAclInput(aclInput)) {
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

PR #3666, opened 2026-06-13
docs(notebook-cloud): clarify access projection store boundary

quillaid:
## Summary
- records cloud access/share projection as the next shared-store follow-up
- clarifies that OIDC, D1 ACL rows, public links, invites, access requests, and URL normalization stay in the cloud host boundary
- explicitly keeps this policy out of `runtimed-wasm`, which should only consume already-negotiated room/document facts

## Verification
- `cargo xtask lint --fix`


rgbkrk:
@pullfrog review

rgbkrk (approved):

rgbkrk:
Excellent shared store for Facts.

PR #3671, opened 2026-06-13
fix(notebook-cloud): resolve app-session invites before listing

quillaid:
## Summary
- resolve pending email invites from the stored verified OIDC profile when a browser is authenticated only by the first-party app-session cookie
- run that resolution before `/api/n` list responses, `/n` notebook-list bootstrap, first-party notebook API authorization, and app-session WebSocket authorization
- keep the app-session cookie identity-only; the verified email stays server-side in the existing profile table

## Why
If a user is already signed in and another user shares a notebook with their email later, `/n` can be backed only by the app-session cookie. That cookie intentionally does not contain email, so fresh pending invites were not accepted before listing or opening direct links unless another provider-backed auth request happened first.

## Verification
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test --test-name-pattern "post-login pending invites" test/worker-routes.test.ts`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/app-session.test.ts test/sharing-storage.test.ts`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/worker-routes.test.ts`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/src/index.ts
+++ b/apps/notebook-cloud/src/index.ts
@@ -81,6 +81,7 @@ import { collectBlobRefs } from "./blob-refs.ts";
 import { cloudLog, durationMs } from "./observability.ts";
 import {
   createPendingNotebookInvite,
+  getPrincipalProfile,
   getPrincipalProfiles,
   listNotebookInvites,
   resolveNotebookInvitesForLogin,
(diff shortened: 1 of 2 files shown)


rgbkrk:
@pullfrog review

PR #3689, opened 2026-06-14
feat(notebook-cloud): wake workstations through SSE control stream

quillaid:
## Summary
- Add a workstation event-stream Durable Object and `/api/workstations/:id/events` route authenticated by existing workstation credentials.
- Notify the stream when an owner creates an attach job, while keeping D1 attach jobs as the source of truth.
- Update the Rust workstation agent to use SSE as the fast wakeup path with 60s fallback polling, and stretch heartbeat defaults/stale UI window to match.

## Request-cost shape
Before this slice, one idle workstation did roughly 43,200 attach-job polls/day plus 4,320 registration heartbeats/day. After this slice, idle cost is one long-lived SSE stream plus roughly one heartbeat and one recovery poll per minute, before real runtime sync work.

This is a large reduction, not the final hundreds-of-idle-workstations model. The next reduction should split live stream presence from active job status heartbeats so idle connected workstations do not need minute heartbeats.

## Validation
- `cargo xtask lint --fix`
- `cargo test -p runtimed --lib`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/worker-routes.test.ts`
- `pnpm --dir apps/notebook-cloud exec wrangler deploy --dry-run --config wrangler.toml`

No preview deploy in this pass; validation stayed local to avoid spending Worker requests while the account is near the daily limit.

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-agent.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-agent.mjs
@@ -42,12 +42,12 @@ const workingDirectory = path.resolve(process.env.NOTEBOOK_CLOUD_WORKSTATION_CWD
 const pollIntervalMs = parsePositiveInteger(
   process.env.NOTEBOOK_CLOUD_WORKSTATION_POLL_MS,
   "NOTEBOOK_CLOUD_WORKSTATION_POLL_MS",
-  2_000,
+  60_000,
 );
 const heartbeatIntervalMs = parsePositiveInteger(
   process.env.NOTEBOOK_CLOUD_WORKSTATION_HEARTBEAT_MS,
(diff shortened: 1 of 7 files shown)


rgbkrk:
@pullfrog review please

quillaid:
Addressed Pullfrog's two comments in 1ae26e01:

- Added a 60s idle timeout around workstation SSE reads so half-open streams reconnect instead of waiting forever in response.chunk().
- Made the agent loop treat a closed event-listener channel as disabled SSE and fall back to the timer path, avoiding a hot-spin.

Validation rerun after the review fixes:
- cargo xtask lint --fix
- cargo test -p runtimed --lib
- pnpm --dir apps/notebook-cloud exec node --import tsx --test test/worker-routes.test.ts

rgbkrk:
@pullfrog we're ready for another round of reviews

PR #3690, opened 2026-06-14
perf(notebook-cloud): use workstation stream presence for online status

quillaid:
## Summary

Follow-up to #3689. This keeps the SSE control stream as the live workstation presence signal so idle connected workstations do not need to keep writing D1 heartbeats just to remain visible as online.

- Overlay WORKSTATION_EVENTS /status onto user-facing workstation list and attach-request decisions.
- Allow attach requests for stale last_seen_at rows when the workstation event stream is connected.
- Stop the Rust workstation agent, and the legacy JS smoke agent, from re-registering while idle; startup registration and active-job status heartbeats remain intact.
- Avoid adding DO /status lookups to the agent attach-job polling path, so idle fallback remains one poll/minute rather than poll plus presence check.

## Request-budget effect

After #3689, an idle workstation still had:

- one long-lived SSE stream;
- one fallback attach-job poll per minute;
- one registration heartbeat/write per minute.

This PR removes the idle registration heartbeat/write. Online status for connected workstations comes from the stream DO, and last_seen_at remains the fallback when the stream binding/status path is unavailable.

## Validation

- cargo xtask lint --fix
- cargo test -p runtimed workstation::agent_loop --lib
- pnpm --dir apps/notebook-cloud exec node --import tsx --test test/worker-routes.test.ts
- pnpm --dir apps/notebook-cloud exec wrangler deploy --dry-run --config wrangler.toml

No preview deploy from this branch while the Cloudflare request budget is hot.

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-agent.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-agent.mjs
@@ -142,7 +142,9 @@ async function runAgentStep(step, fn) {
 
 async function heartbeatIfNeeded(pythonPath) {
   const now = Date.now();
-  if (now - lastHeartbeatAt < heartbeatIntervalMs) {
+  const shouldRegister =
+    lastHeartbeatAt === 0 || (activeJobs.size > 0 && now - lastHeartbeatAt >= heartbeatIntervalMs);
+  if (!shouldRegister) {
     return false;
   }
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

rgbkrk:
@pullfrog review again please

PR #3692, opened 2026-06-14
fix(notebook): preserve runtime session targets

quillaid:
## Summary

- keep `runtimeSessionId` on stabilized notebook shell runtime targets
- add a regression test for replacing a workstation runtime session under the same target id

## Why

The cache key already included `runtimeSessionId`, but the frozen target object dropped it. Downstream workstation projections key on this fact, so replacing a stale runtime session could lose the session identity after stabilization.

## Verification

- `pnpm test:run -- packages/runtimed/tests/notebook-shell-capabilities.test.ts packages/runtimed/tests/notebook-workstation-panel.test.ts packages/runtimed/tests/notebook-workstation-launch.test.ts` (runner executed full Vitest suite: 218 files passed, 1 skipped)
- `cargo xtask lint --fix`

### DIFF
--- a/packages/runtimed/src/notebook-shell-capabilities.ts
+++ b/packages/runtimed/src/notebook-shell-capabilities.ts
@@ -764,6 +764,7 @@ function stableNotebookShellRuntimeTarget(
 
   const stableTarget = Object.freeze({
     id: target.id ?? null,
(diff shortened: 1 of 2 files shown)


rgbkrk:
@pullfrog review

PR #3695, opened 2026-06-15
feat(notebook-doc): add comms doc identity pointer

quillaid:
## Summary
- add deterministic `NotebookDoc.comms_doc_id` (`comms:{notebook_id}`) with constructor stamping, load/migration repair, pristine allowlist coverage, WASM accessors, and room bootstrap assertions
- align runtime-state write capability so only `runtime_peer` can author `RuntimeStateDoc` frames, with generated TypeScript and cloud authorization updated
- update ADRs/plans/cloud docs to reflect the now-required CommsDoc pointer and runtime-agent-only RuntimeStateDoc writes

## Verification
- `cargo test -p notebook-doc -p nteract-identity -p notebook-protocol`
- `cargo test -p runtimed test_room_load_or_create_new`
- `cargo test -p runtimed test_new_fresh_creates_empty_doc`
- `cargo test -p runtimed-wasm editor_changes_reject`
- `pnpm --dir apps/notebook-cloud typecheck`
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/authorization.test.ts test/snapshot-render.test.ts`
- `cargo xtask lint --fix`
- `git diff --check`

## Notes
- The checked-in notebook fixture LFS pointers changed because regenerated fixture NotebookDoc bytes now include the new `comms_doc_id` root scalar.
- Hosted CommsDoc snapshot routes still use runtime-state-keyed object paths; the ADR now calls out first-class `comms_doc_id` keying as follow-up hosted persistence cleanup.

PR #3696, opened 2026-06-15
docs(adr): reconcile cleanup punchlist drift

quillaid:
## Summary

- Align ADR/PRD/runbook text with current runtime-state, CommsDoc, hosted auth, invite, artifact, and workstation implementations.
- Document the workstation SSE wakeup path and why missed attach-job recovery uses durable queue polling instead of SSE `Last-Event-ID` replay.
- Remove or reframe landed stale follow-ups, including 3D-9 and EP-11, while leaving current-code follow-ups explicitly deferred.

## Disposition

| Item | Disposition | Evidence |
|------|-------------|----------|
| Workstation control channel documented as polling/WebSocket-first | updated | `runtimed workstation-agent` opens `/api/workstations/:id/events` as an SSE wakeup stream, falls back to attach-job polling, and avoids idle heartbeat request churn. Docs updated in the workstation ADR, operator doc, and cloud README. |
| Workstation SSE missed-event recovery via `Last-Event-ID` | reframed | The SSE stream is a wakeup/presence channel without event ids; attach jobs are durable D1 rows claimed through `/attach-jobs`, so reconnect recovery polls the queue. |
| 3D-9 `comms_doc_id` tracked follow-up after #3695 | removed | `NotebookDoc.comms_doc_id` is now stamped/migrated; the landed 3D-9 bullet was removed from the document-split tracked follow-ups. |
| EP-11 empty `flush_then_signal` priority path | updated | The stream committer test now proves empty flushes ride the priority queue; the execution ADR was updated and stale `QueueCommand` references were replaced. |
| Hosted WebSocket transport status | updated | The typed-frame ADR now names browser `CloudWebSocketTransport` and runtime-peer `CloudWsFrameTransport`. |
| Browser hosted auth defaults | updated | The direct-OIDC runbook now matches the app-session cookie default and leaves `nteract-bearer.*` as explicit fallback. |
| Hosted sharing route names | updated | The PRD now uses `/api/n/:id/invites` and records public viewer access as an ACL row, not a `/public-viewer` route. |
| Captured environment disk-state wording | updated | Docs and comments now refer to typed `CapturedEnvDiskState` / `captured_env_disk_state(...)` where they describe current reopen behavior. |
| ADR register status/index drift | updated | The ADR register defines `Proposed` and `In progress`, indexes local-first and shared-store ADRs, and normalizes the captured-env status. |
| Generated runtime artifact ownership | updated | The generated-artifacts ADR now matches `.gitattributes` and `.gitignore`: third-party renderer bundles are LFS-tracked; owned renderer/WASM outputs are regenerated. |
| CommsDoc snapshot routes keyed by runtime-state id | deferred | Current cloud code still keys CommsDoc snapshots with `runtimeStateDocId`; ADR 0002 keeps first-class `comms_doc_id` snapshot keying as remaining hosted persistence cleanup. |

## Verification

- `cargo xtask lint --fix`


rgbkrk:
@pullfrog review

PR #3699, opened 2026-06-15
fix(cloud): reject hosted restart request no-op

quillaid:
## Summary

Hardens hosted room REQUEST handling so raw `restart_kernel` frames fail honestly instead of falling through to materialization as accepted no-ops.

- adds `restart_kernel` to the unsupported response-bearing runtime request set;
- expands the notebook-room regression test to cover all unsupported response-bearing actions;
- updates the remote-workstation ADR list so it names `restart_kernel` explicitly.

This keeps direct restart routing behind the lifecycle design called out in the ADR. Interrupt remains a fire-and-forget runtime-peer command and is not changed here.

## Verification

- `node --import tsx --test test/notebook-room.test.ts` from `apps/notebook-cloud` (61 tests passed)
- `cargo xtask lint --fix`
- `git diff --check`


rgbkrk:
@pullfrog review

PR #3702, opened 2026-06-15
test(cloud): exhaust protocol helper coverage

quillaid:
## Summary

- derive cloud frame names from the generated `FrameType` keys instead of maintaining a switch
- make the cloud client-writable frame policy an exhaustive `Record<FrameTypeValue, boolean>` table
- update protocol helper tests to iterate every generated frame type, including name lookup, writable policy, and typed-frame split acceptance

## Verification

- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/protocol.test.ts`
- `git diff --check`
- `cargo xtask lint --fix`

## Notes

This preserves the existing writable policy. `BROADCAST` and `SESSION_CONTROL` remain server-originated from cloud clients' perspective; `POOL_STATE_SYNC` remains writable because the current cloud policy already allowed it.

### DIFF
--- a/apps/notebook-cloud/src/protocol.ts
+++ b/apps/notebook-cloud/src/protocol.ts
@@ -123,48 +123,42 @@ export function splitTypedFrame(message: ArrayBuffer | ArrayBufferView): TypedFr
   };
 }
 
+const KNOWN_FRAME_TYPE_ENTRIES = Object.entries(FrameType) as Array<
+  readonly [keyof typeof FrameType, FrameTypeValue]
+>;
+const KNOWN_FRAME_TYPES = KNOWN_FRAME_TYPE_ENTRIES.map(([, value]) => value);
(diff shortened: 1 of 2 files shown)


quillaid:
Side-review follow-up pushed in `097c0902`.

Disposition:
- Confirmed/fixed: the exhaustive protocol helper test was deriving expected names and coverage from `FrameType`, so it would miss some generated constant drift. The test now pins expected frame keys, wire bytes, display names, and client-writable bits explicitly, and separately checks the generated `FrameType` object has exactly those keys and numeric wire-order values.

Verification:
- `pnpm --dir apps/notebook-cloud exec node --import tsx --test test/protocol.test.ts`
- `git diff --check`
- `cargo xtask lint --fix`

CI is queued on the pushed follow-up commit.

rgbkrk:
@pullfrog review

PR #3703, opened 2026-06-15
refactor(kernel-env): centralize base package lists

quillaid:
## Summary
- expose owned UV/Conda base-package helpers from `kernel-env` and reuse them in install/prewarm paths
- derive daemon UV/Conda prewarm bases from the `kernel-env` constants while preserving the existing `nbformat`/`pyarrow` override behavior and UV package order
- refresh the kernel-env agent guidance table for the Conda base package set

No daemon package set changes are intended. The daemon still re-adds display packages after user defaults so `nbformat`/`pyarrow` overrides keep the prior behavior.

One latent helper behavior does intentionally change with the centralization: the currently uncalled `create_prewarmed_environment_in` helpers now draw from the full `kernel-env` base package constants instead of their previous narrower inline lists. Pullfrog verified this has no runtime impact today because the daemon prewarms through `spawn_warm_env` and the derived `*_prewarmed_packages(...)` path.

## Verification
- `cargo test -p kernel-env strip_base --lib`
- `cargo test -p kernel-env managed_specs --lib`
- `cargo test -p runtimed prewarmed_packages --lib`
- `cargo test -p runtimed effective_user_deps_from_launched --lib`
- `git diff --check`
- `cargo xtask lint --fix`

### DIFF
--- a/crates/kernel-env/AGENTS.md
+++ b/crates/kernel-env/AGENTS.md
@@ -100,7 +100,7 @@ Pool warmer and capture step strip a base set so captured metadata records only 
 | Constant | Value |
 |----------|-------|
 | `kernel_env::uv::UV_BASE_PACKAGES` | `[ipykernel, ipywidgets, anywidget, nbformat, pyarrow>=14, uv]` |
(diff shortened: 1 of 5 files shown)


rgbkrk:
@pullfrog review

PR #3715, opened 2026-06-17
fix(release): pin nightly installer notes to release asset

rgbkrk:
## Summary

- Change generated release notes to point Linux/macOS shell installs at the installer asset pinned to the current release tag.
- Keep the `https://sh.nteract.io` shortcut only on stable release notes so nightly users do not accidentally install latest stable.
- Add installer fixture coverage for the release-stamped nightly installer path.

## Verification

- `cargo xtask lint --fix`
- `bash -n scripts/install-linux-release && bash -n scripts/test-install-release`
- `bash scripts/test-install-release`
- `actionlint .github/workflows/release-common.yml`
- `git diff --check`

## Notes

- This remains draft pending GitHub CI.


rgbkrk:
@pullfrog review

PR #3719, opened 2026-06-18
refactor(notebook): parameterize document controls

quillaid:
## Summary

- lets `NotebookEditModeButton` accept document-specific labels, aria/data-slot values, and a disabled edit segment
- lets `NotebookOutlinePanel` accept document-specific aria and empty-state copy
- keeps notebook defaults unchanged and adds focused tests for the new knobs

## Why

The markdown prototype needed these as shared document affordances, but the useful part is smaller than the `/m` product surface. This PR makes the existing notebook controls reusable for future document surfaces without adding new routes or document models.

## Verification

- `pnpm test:run src/components/notebook/__tests__/NotebookEditModeButton.test.tsx src/components/notebook-rail/__tests__/NotebookRail.test.tsx`
- `cargo xtask lint --fix`

### DIFF
--- a/src/components/notebook-rail/NotebookRail.tsx
+++ b/src/components/notebook-rail/NotebookRail.tsx
@@ -121,6 +121,8 @@ export interface NotebookOutlinePanelProps {
   activeItemId?: string | null;
   selectedItemId?: string | null;
   selectedCellId?: string | null;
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review please

PR #3713, opened 2026-06-17
feat(comments-doc): add selected source comments

quillaid:
## Stack

Depends on #3712 (`codex/comments-doc-desktop-preview`). This branch is sixteen commits on top of `fffe5cad`.

## Summary

- add a CodeMirror selected-source comment affordance for code cells, raw cells, and Markdown edit mode
- add rendered Markdown prose comment affordances that map host-rendered text back to verified source-range anchors without forcing users into source edit mode
- add a keyboard-only rendered paragraph path via a focusable paragraph comment button, while retaining `Mod-Alt-m` for existing rendered selections
- keep rendered paragraph comment affordances inline with the prose so the button stays physically tied to the text on wide and narrow layouts
- require rendered Markdown comment anchors to come from a projection whose `source` exactly matches the current preview source and whose source quote exactly matches the rendered selection/run text, so stale projections and syntax-changing selections fail closed
- create bounded `source_range` anchors with zero-based UTF-16 line/column coordinates, exact quote, and prefix/suffix context
- route source comments through the existing local CommentsDoc write, flush, and runtime authority accept flow
- validate source anchors at the CommentsDoc model boundary, in MCP input handling, and again in daemon authority before writing trusted authority fields
- require MCP-created `source_range` anchors to select at least one source character and carry a non-empty exact quote
- show source draft/thread excerpts in the comments rail, preserve indentation, and focus the composer when a source draft is created
- add a keyboard path for selected source comments via `Mod-Alt-m`, including rendered Markdown preview selections
- reject whitespace-only/oversized selections and stop submit if the selected source changed before create
- align frontend source quote limits with the Rust byte limit so multibyte selections are rejected consistently
- consume the live editor selection once it becomes a source comment draft so returning to the cell does not resurrect a stale inline comment affordance
- label mixed thread lists per anchor type and add a `Show cell` action so source/cell threads can jump back to the notebook canvas, including narrow rail takeover
- require daemon-advertised `comments_doc_id` and `comments_authority_actor_label` for successful local daemon notebook bootstraps; browser/desktop ready payloads now fail early if required comments identity is missing
- refresh docs/guidance so local CommentsDoc/UI/MCP state, relay frame support, and hosted Cloud deferrals are aligned, and correct the ADR to distinguish landed rail/source UI from follow-up body markers/popovers
[...]

PR #3712, opened 2026-06-17
feat(comments-doc): add desktop comments preview

quillaid:
## Summary

Stacked on #3710. Adds a local Desktop preview path for CommentsDoc so users can try comments without Cloud.

- passes daemon-provided CommentsDoc identity into the Desktop/WASM bootstrap
- bridges `COMMENTS_DOC_SYNC` through Desktop, runtimed-wasm, SyncEngine, and the Vite browser relay
- adds an optional shared comments rail panel and wires Desktop to create, reply, resolve, and reopen comment threads
- follows the MCP authority pattern: local pending content mutation -> comments heads -> `flushSync()` -> daemon accept request; status transitions go through daemon-authoritative resolve/reopen requests with observed comment heads
- gates write UI to local/owner/editor comment scopes and keeps runtime peers read-only
- adds a browser E2E covering the local comments rail plus an MCP peer observing accepted create/reply and resolve/reopen state
- polishes the first product surface after playtesting: clearer `Add comment` action, hidden routine accepted debug badges, readable local actor labels, contextual accessible labels, live status/error roles, all projected anchors visible, composer text clears immediately on submit, failed submissions restore drafts, status actions wait for pending replies, and message-count badges are accessible

## Local Desktop Tryout

For a source checkout on this branch after #3710 is available locally:

```bash
cargo xtask dev-daemon
cargo xtask vite
cargo xtask notebook --attach
```

Open the Comments rail in the Desktop window. You can add a document comment, reply, resolve the thread, and reopen it. A local MCP peer connected to the same notebook should see the accepted comment state through `list_comments`.

For a bundled debug app instead of hot reload:

```bash
cargo xtask build
cargo xtask run
```

## Current State

- Latest head: `a739d5c1`.
- Merged the updated #3710 branch into this branch with a normal merge commit so review state remains trackable.
- GitHub could not use the fork-only #3710 branch as an upstream PR base, so this draft targets `main` until #3710 merges.
- GitHub CI for `a739d5c1` is green and the PR merge state is CLEAN. Prior head `fffe5cad` was also green/CLEAN.

## Latest Slice

- Added `a739d5c1`: docs-only ADR cleanup aligning published attribution wording with `authority_created_by_actor_label`.
- Added `fffe5cad`: fixed the TypeScript nullability failure from #3712 Build shared UI artifacts.
- The failing build reported nullable `handle` uses in `apps/notebook/src/App.tsx` after comments create/reply guards.
[...]

PR #3710, opened 2026-06-16
feat(comments-doc): wire local comments sync and MCP tools

quillaid:
## Summary

Draft stacked follow-up to #3709. Until #3709 merges, this PR includes the core CommentsDoc commits plus the local sync/MCP follow-up work.

This branch adds the local CommentsDoc sync frame, daemon sidecar storage, MCP create/list/reply/resolve/reopen tools, daemon-authoritative comment finalization/status requests, scoped comments resources, and a real multi-client MCP smoke harness.

## Current State

- Latest head: `39a4d5f2`.
- PR remains draft/stacked on #3709 until the review/stack state is cleared.
- GitHub CI for `39a4d5f2` is green and the PR merge state is CLEAN. Prior head `1ad1bde7` was also green/CLEAN.

## Latest Slice

- Added `39a4d5f2`: docs-only ADR cleanup aligning published attribution wording with `authority_created_by_actor_label`.
- Added `1ad1bde7`: hardened the runtime-peer CommentsDoc rejection test after Linux Rust Tests failed on a private Automerge sync-state assertion.
- The failed assertion compared `sync::State::their_heads` after a rejected runtime-peer payload. That is a brittle internal negotiation detail and failed on CI while the local full Rust shard and repeated direct reproductions passed.
- The test now verifies behavior through an observer CommentsDoc that receives the daemon's rejection reply and confirms the forged runtime-peer thread is not replayed to clients.
- Existing document-level assertions remain: runtime peer writes do not mutate `CommentsDoc`, do not persist a sidecar, and later daemon-authored broadcasts still advance to daemon heads.

## Prior Important Slice

- Added `921eb32e`: hardened sync denial coverage after local read-only review.
- Runtime-agent channel coverage now keeps the client connection open briefly after a forged `CommentsDocSync` frame and fails if any live comments negotiation reply appears before shutdown.
- V2 sync coverage now proves fixtures use an actual combined V2 sync chunk by asserting the single sync chunk does not decode as an individual Automerge change.
- Added negative V2 cases for a client using the reserved `runtimed:comments` actor and for client-authored `authority_*` fields.
- V2 fixtures now force separate AutoCommit commits before sync generation so the combined-chunk tests exercise the path that broke the old direct `Change::from_bytes` authorization approach.

## Verification

Latest GitHub CI after `39a4d5f2`:

[...]

PR #3709, opened 2026-06-16
feat(comments-doc): add durable comments document core

quillaid:
## Summary
- Add standalone `comments-doc` workspace crate for durable per-notebook comment state.
- Implement required `comments_doc_id` identity, frozen genesis asset, Automerge Text message bodies, fractional ordering, projection snapshots, sync helpers, and `CommentsDocHandle`.
- Harden trust projection around authority-authored `authority_*` snapshots, `Unverified` states, sync identity rollback, malformed ordering data, and subscriber notifications.
- Update the comments ADR to use the dedicated crate boundary, document the implemented `authority_*` schema/trust model, and call out principal-to-actor authentication as a future sync-ingress requirement.
- Include `crates/comments-doc/Cargo.toml` in the release bump target list so Linux Rust Tests cover the new versioned crate cleanly.

## Current State
- Latest head: `87951db7`.
- PR remains draft.
- GitHub CI for `87951db7` is green and the PR merge state is CLEAN. Prior head `aca85e5d` was also green/CLEAN.

## Verification
Latest local verification after `87951db7`:
- `git diff --check`
- `rg -n -P '(^|[^_])created_by_(actor_label|authority)' docs/adr/notebook-comments-document.md` (no stale non-authority matches)

Previous local verification after `aca85e5d`:
- `cargo test -p comments-doc` (32 passed, 1 ignored)
- `git diff --check`

Previous verification retained:
- `cargo check -p comments-doc --all-targets`
- `cargo clippy -p comments-doc --all-targets -- -D warnings`
- `cargo metadata --no-deps`
- `cargo fmt --check`
- `cargo test -p xtask rust_workspace_crate_versions_are_bump_targets`
- `cargo xtask lint --fix`

## Notes
- This PR intentionally stops at the pure CommentsDoc core. The next slice adds `COMMENTS_DOC_SYNC` and the first MCP/local sync mutation surface.
- Agent user testing is tracked as a goal lane: once a real comment creation surface exists, run multi-agent exercises adding/replying/resolving comments on the same notebook and report friction around addressing, context, permissions, and review flow.
- The crate can project trust from admitted Automerge actor ids, but it cannot authenticate remote peers; sync ingress must bind principal identity to authority actor ids before accepting authority-authored fields.
[...]

PR #3726, opened 2026-06-18
feat(comments-doc): add durable comments document core

rgbkrk:
Add the `comments-doc` crate, a durable per-document Automerge model for comment threads, decoupled from runtime state, output transport, and execution. This is internal substrate: nothing wires it into the deployed desktop or cloud app yet. The sync frame + MCP tools, desktop preview, and source-selection UI land in follow-ups.

This is a fresh extraction off current `main` of the core from the draft stack (#3709 → #3710 → #3712 → #3713), which was authored against an older main by an agent that's since gone offline. Just the crate, its ADR, and workspace registration. No daemon, wire, or UI changes.

## Trust comes from the sync ingress, not the document

Comments are plain authored Automerge changes. A client writes a comment into its local replica and it renders immediately from local heads, no request round trip.

The trust boundary is the sync ingress (daemon/cloud, a follow-up), which binds each connection to its own Automerge actor id before admitting changes. So the actor recorded against an admitted change is its real author. That makes stored attribution pointless: in a CRDT any editor-scope peer can write any field, so a `created_by` in the document is a forgeable claim, while the change author is not.

So **attribution is read, not stored**:

- A thread's or message's author is the actor that created the object (`actor_label_of`).
- A thread's resolver is the actor that wrote the winning `status` value (`field_writer_label`).

The document keeps only content and shared state: `status` (`open`/`resolved`), `resolved_at`, fractional positions, `Text` bodies, timestamps. No `mutation_state`, no `authority_*` keyspace, no accept/reject finalization. An authority-stamped field returns only for a concept attribution cannot express, such as a future moderation gate, never for author identity or resolve state.

Threads and replies order by fractional index, never list position, so concurrent inserts converge with an id tie-break.

## Schema reach

Anchors cover notebook, cell, cell-range, source-range, and output scopes (a tagged enum, additive). The thread/message/projection machinery is document-agnostic; only `CommentAnchor` and `NotebookCommentRef` carry notebook shape, which is the seam for commenting on non-notebook documents (a markdown file gets a cell-free anchor variant later).

## Verification

- [x] `cargo test -p comments-doc` — 23 passed, 1 ignored; includes a sync round-trip test proving attribution (author + resolver) survives CRDT merge between two peers
- [x] `cargo clippy -p comments-doc` clean
- [x] `cargo fmt` applied


rgbkrk:
@pullfrog review

PR #3757, opened 2026-06-19
fix(runtimed): bound pool warmup timeouts

rgbkrk:
## Summary

- split required Python runtime validation from optional pool warmup work
- add a pool readiness marker so validated environments can be leased even when full `.pyc` warmup does not finish
- replace timeout-wrapped `Command::output()` with explicit child ownership so timed-out warmup children are killed and drained
- cap optional Python compile/import warmup at 30s and log it as a warning instead of reporting a pool failure

## Verification

- `cargo xtask lint --fix`
- `cargo test -p runtimed`

## Notes

The current approach avoids detaching `compileall`: a detached compiler can race with pool leasing, eviction, and cleanup. Instead, the env is validated, marked ready, and allowed into the pool after a short best-effort warmup window. Full `.warmed` remains the marker for environments that completed the optional warmup.

### DIFF
--- a/crates/runtimed/src/daemon.rs
+++ b/crates/runtimed/src/daemon.rs
@@ -389,6 +389,7 @@ struct Pool {
 
 const MIN_WARM_BASES: usize = 2;
 const POOL_PACKAGE_HASH_FILE: &str = ".runt-pool-packages.sha256";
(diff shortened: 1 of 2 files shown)


rgbkrk:
@pullfrog review please

PR #3758, opened 2026-06-19
feat(comments): shared comment-highlight surface across editor and rendered markdown

rgbkrk:
The comment highlight is now one shared visual surface across planes: the CodeMirror editor and rendered markdown prose use the same definition, so they look identical and tune in one place. This also adds comment highlights to rendered markdown. Completes Phase 2 of rendered-markdown commenting (after #3751, #3755).

## Shared surface

The tint and open/resolved treatment move from the CodeMirror `baseTheme` into a shared stylesheet (`src/styles/comment-highlight.css`: `.comment-highlight` / `.comment-highlight-resolved`, driven by `--cm-comment-color`), imported by both the notebook and cloud-viewer entrypoints. The editor decoration co-classes onto it and no color rules remain in the theme, so tuning the stylesheet updates both planes. The editor highlight is visually unchanged.

## Rendered highlights

`ProjectedMarkdownView` gains a `commentHighlights` prop and tints the overlapping run spans (via `markdownRunsForSourceRange`, harvested here). The hosts pass per-cell threads through `NotebookView` to `MarkdownCell`, which resolves each anchor against the rendered source and builds the highlights. The rendered view has no editor, so it cannot use the CodeMirror highlight bridge.

## Elements

A fixture renders the highlight on prose (open + resolved) so the shared surface is tunable in the design system.

## Verification

`vp check` clean; full `vp test` green (2419), including new tests for `markdownRunsForSourceRange` overlap and the `ProjectedMarkdownView` highlight rendering, with the existing comment-highlight tests still passing; `apps/notebook-cloud` typecheck clean; `apps/elements` build clean (28/28 pages).

### DIFF
--- a/apps/elements/app/global.css
+++ b/apps/elements/app/global.css
@@ -5,6 +5,7 @@
 @import "../../../src/styles/ansi.css";
 @import "../../../src/styles/notebook-base.css";
 @import "../../../src/styles/cream-theme.css";
+@import "../../../src/styles/comment-highlight.css";
 
 @source "../**/*.{ts,tsx,mdx}";
 @source "../../../src/components/cell/**/*.{ts,tsx}";
--- a/apps/elements/app/page.tsx
+++ b/apps/elements/app/page.tsx
@@ -9,6 +9,7 @@ import {
   IdCard,
   LayoutDashboard,
   ListPlus,
+  MessageSquareText,
   MousePointer2,
   PackageCheck,
   Palette,
@@ -75,6 +76,12 @@ const catalogGroups = [
         description: "Target-specific notebook actions for menus and toolbars.",
         href: "/docs/context-controls",
         icon: MousePointer2,
+      },
+      {
+        title: "Comment surfaces",
+        description: "Anchored discussion, attribution, and rendered markdown highlights.",
+        href: "/docs/comment-surfaces",
+        icon: MessageSquareText,
       },
       {
         title: "Notebook toolbar",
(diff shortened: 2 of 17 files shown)


rgbkrk:
@pullfrog review

Extra focus areas for this PR:

- **Shared-CSS repoint (the load-bearing change):** the comment-highlight tint/treatment moved out of the CodeMirror `EditorView.baseTheme` into `src/styles/comment-highlight.css` (`.comment-highlight` / `.comment-highlight-resolved`), and the editor decoration now co-classes onto it. Please confirm the editor highlight is visually unchanged (the `color-mix` formulas match the removed baseTheme rules exactly) and that nothing inside `.cm-editor` overrides the shared class via specificity. The goal is a single tune-once definition shared by the editor and rendered-markdown planes.
- **Data path:** hosts (`App.tsx`, `notebook-viewer.tsx`) pass per-cell threads through `NotebookView` (`commentThreadsByCell`) to `MarkdownCell`, which resolves each anchor against the rendered source into `commentHighlights`. The prop is optional at every hop; worth checking no regression for other `NotebookView` callers and that resolution only runs in matched preview mode.
- **`ProjectedMarkdownView` overlap:** `commentHighlightForRun` wraps matching run spans (shortest match, half-open overlap) without disturbing the Phase-1 `data-source-run` attributes or the active-run span.

Context: the rendered view has no CodeMirror editor, so it deliberately cannot reuse the editor's highlight bridge; the prop-based data path is the alternative.

rgbkrk:
Pushed a follow-up: resolved comment highlights now drop the in-code underline (they keep only the faint tint). The underline is the open-comment "needs attention" cue, so it should not persist after resolve. Applies to both the editor and rendered planes via the shared `comment-highlight.css`.

PR #3761, opened 2026-06-19
feat(comments): show the local author's macOS full name

rgbkrk:
Desktop comment authors showed the humanized OS account name (e.g. "Kylekelley") instead of a real name, because the local actor label derives from `$USER`.

`get_username()` now returns the macOS full name via `NSFullUserName()`, falling back to `$USER`/`$USERNAME` when it is empty or off-macOS. That value already flows through `__NTERACT_USERNAME__` into the local peer label.

`resolveCommentAuthor` feeds that name through a `peers` entry keyed by the local principal, mirroring the cloud presence model. Only the local author is relabeled; synced peers and agents resolve their own friendly labels. A first pass applied the name as an `identityLabel` for every author, which would have relabeled remote and agent comments with this machine's user name; the peer-scoped approach avoids that and drops the unused `identityLabel` option.

Verified: `vp check`, the actor-display unit test, notebook-app and cloud typecheck, and `cargo check -p notebook` for the Rust change.

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -757,6 +757,7 @@ function AppContent() {
     (actorLabel: string): CommentAuthor => {
       const display = resolveActorDisplay({
         actorLabel,
+        identityLabel: peerLabel,
         peers: [],
         source: connectionScope ? "cloud" : "local",
       });
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -4,6 +4,8 @@ import {
   deriveRuntimeKind,
   NotebookClient,
   resolveActorDisplay,
+  splitNotebookActorPrincipalOperator,
+  type ActorDisplayPeer,
   type CommentAnchor,
   type CommentThreadSnapshot,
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

Local desktop comment authorship now shows the macOS full name. Worth focusing on:

- **Attribution scoping** (`resolveCommentAuthor`, App.tsx): the full name is fed through a `peers` entry keyed by the local principal, not as a blanket `identityLabel`. Confirm a synced peer's comment and an agent-authored comment still resolve their own principal labels (not this machine's user name), and that an agent acting on behalf of the local user shows the real name via the peer match. The local-principal key comes from `splitNotebookActorPrincipalOperator(localActor)`.
- **objc memory safety** (`crates/notebook/src/lib.rs` `macos_full_user_name`): `NSFullUserName()` returns an autoreleased `NSString`; confirm it is not released, the UTF8 contents are copied into an owned `String` before return, and nil/null are guarded.
- **Fallback**: empty full name or non-macOS falls back to `$USER`/`$USERNAME`.

PR #3762, opened 2026-06-19
fix(reconnect): survive a daemon restart during initial materialize

rgbkrk:
A daemon restart while the notebook was connected froze the open document: the rendered cells vanished even though the daemon still held them, and only a reload recovered.

The cause is a use-after-free of the WASM notebook handle across a reconnect. A restart cycles the relay generation (gen1 to gen2); the gen2 bootstrap frees the gen1 handle while an in-flight materialize is still awaiting blob resolution. After the await, the materialize calls back into the freed handle (`set_blob_port`, `get_cells_json`) and wasm-bindgen panics on `assert_not_null`, so the initial materialize fails and the cells disappear.

`materializeCells` and `materializeChangeset` now re-check live handle identity after each await and bail before any WASM call. The `initialSyncComplete` continuation does the same before it reuses the captured handle for `get_cell_ids` / output seeding / projection; otherwise the same panic could fire one frame downstream. The guard mirrors the existing `cellChanges` continuation, which already re-reads the live handle. When the handle was replaced, the gen2 bootstrap drives its own materialize cycle, so bailing loses nothing.

Verified: `vp check`, notebook-app typecheck, and the sync-store-bridge test suite, including a new case that proves the continuation bails when the handle is replaced mid-await.

### DIFF
--- a/apps/notebook/src/hooks/useAutomergeNotebook.ts
+++ b/apps/notebook/src/hooks/useAutomergeNotebook.ts
@@ -178,33 +178,39 @@ export function useNotebook() {
   // ── Core helpers ───────────────────────────────────────────────────
 
   /** Full materialization: WASM doc → resolve manifests → write to store. */
-  const materializeCells = useCallback(async (handle: NotebookHandle) => {
-    const start = performance.now();
-    // Resolve blob port BEFORE reading cells — WASM needs it to
-    // convert binary ContentRefs to Url variants in get_cells_json().
-    let blobResolver = getBlobResolver();
-    if (blobResolver === null) {
-      blobResolver = await refreshBlobResolver();
-    }
-    const blobPort = blobResolver?.port ?? null;
-    if (blobPort !== null) {
-      handle.set_blob_port(blobPort);
-    }
-    const json = handle.get_cells_json();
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

A use-after-free of the WASM notebook handle across a daemon-restart reconnect (gen1 to gen2). Worth focusing on:

- **Completeness of the guard**: confirm there is no remaining path where a handle captured before an `await` is called into after the await without re-checking live identity. The fix covers `materializeCells` and `materializeChangeset` (apps/notebook/src/hooks/useAutomergeNotebook.ts, apps/notebook/src/lib/frame-pipeline.ts) and the `initialSyncComplete` continuation (apps/notebook/src/lib/notebook-sync-store-bridge.ts). The sibling `cellChanges` continuation was already correct.
- **No stuck-loading**: when a continuation bails because the handle was replaced, the gen2 bootstrap drives its own materialize cycle. Confirm bailing cannot strand `setIsLoading(true)` with no follow-up.
- **Guard correctness**: the check compares the captured handle against the live one (`getHandle()` / `handleHost.current`), matching the existing `NotebookHandleHost.bootstrap` guard.

PR #3763, opened 2026-06-19
feat(comments): dot selection affordance shared across editor and rendered markdown

rgbkrk:
The "comment on selection" affordance was a 24px icon button that sat heavily over the selected text, and it was built twice: a CodeMirror tooltip button in the editor and a separate React button in rendered markdown.

This replaces both with a small resting dot that grows in place into a comment button on hover or keyboard focus. The visual is defined once in `src/styles/comment-affordance.css` and reused across both planes: a shared `CommentSelectionAffordance` component (`src/components/comments/`) for the rendered-markdown plane, and the same classes applied to the CodeMirror tooltip button for the editor. It's cataloged in Elements under Comment surfaces so the look stays tunable in one place.

Design notes:

- The dot stays quiet while you select and copy text, which is the common case. It only becomes a button when you reach for it.
- Growing the same element (rather than revealing a separate floating panel) keeps the target under the cursor. There's no gap to cross and nothing to flicker out as you move toward it, which is the hover-intent problem a "safe triangle" would otherwise solve.
- A 22px hit area wraps the 9px dot, so it's easy to land on without enlarging the resting footprint.
- The shared CSS strips the CodeMirror tooltip chrome so only the dot shows in the editor.

Both planes feed the same `onCreateSourceComment(anchor, rect)` path as before; only the trigger's look and shared definition changed.

Verified: `vp check`, notebook / cloud / Elements typecheck, and the full `vp test` suite (2445 passed). Worth an interactive look in Elements (Comment surfaces) and the running app for the hover and focus growth on both planes.

### DIFF
--- a/apps/elements/app/global.css
+++ b/apps/elements/app/global.css
@@ -6,6 +6,7 @@
 @import "../../../src/styles/notebook-base.css";
 @import "../../../src/styles/cream-theme.css";
 @import "../../../src/styles/comment-highlight.css";
+@import "../../../src/styles/comment-affordance.css";
 
 @source "../**/*.{ts,tsx,mdx}";
 @source "../../../src/components/cell/**/*.{ts,tsx}";
--- a/apps/elements/components/comment-selection-affordance-example.tsx
+++ b/apps/elements/components/comment-selection-affordance-example.tsx
@@ -1,0 +1,21 @@
+"use client";
+
+import { CommentSelectionAffordance } from "@/components/comments/CommentSelectionAffordance";
+
+export function CommentSelectionAffordanceExample() {
+  return (
+    <div className="not-prose my-6">
+      <div className="mx-auto max-w-[760px] border border-border bg-background px-6 py-5 text-foreground shadow-sm max-sm:px-4">
+        <p className="mb-4 text-sm text-muted-foreground">
+          Resting, the affordance is a quiet dot that stays out of the way while you select and copy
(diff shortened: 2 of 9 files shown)


rgbkrk:
@pullfrog review

Redesign of the comment-on-selection affordance into a shared dot, across the CodeMirror editor and rendered-markdown planes. Worth focusing on:

- **Shared-visual correctness**: the look lives in `src/styles/comment-affordance.css`, imported by both the notebook app (`src/index.css`) and Elements (`apps/elements/app/global.css`), mirroring the comment-highlight pattern. Confirm the CodeMirror tooltip chrome is actually stripped by `.cm-tooltip:has(.comment-affordance)` so the editor shows only the dot (no leftover tooltip box).
- **Behavior parity**: both planes still call the same `onCreateSourceComment(anchor, rect)`. In `MarkdownCell` the new component forwards the click event to `handleRenderedSourceCommentClick`, which uses `event.currentTarget` for its fallback rect. Confirm that still resolves to the button. The `source-comment-extension` keyboard path (`Mod-Alt-m`) and selection gating are unchanged.
- **Accessibility**: the dot grows on `:hover` and `:focus-visible`; the button keeps `aria-label`/`title` and the decorative dot/icon are `aria-hidden`. Confirm keyboard focus reveals and activates it.
- **Elements wiring**: `apps/elements/tsconfig.json` gained a `@/components/comments/*` path (Elements uses granular per-directory mappings, no catch-all). Confirm the catalog page builds.

PR #3765, opened 2026-06-19
feat(comments): comment on cell outputs with demote-to-notebook on detach

rgbkrk:
Comment on a cell's outputs.

A small "comment on outputs" button in the output gutter creates a comment anchored to the cell and its current execution. Anchoring uses the stable IDs we already have - `cell_id` + `execution_id` (+ an optional `output_id`) - rather than text ranges, so the anchor survives scrolling and re-rendering and degrades predictably when the execution changes. The `CommentAnchor::Output` variant already existed end to end (Rust, TS, validation, panel routing); this wires up the create path and detach handling.

The button is gated exactly like the existing source and rendered-markdown affordances (hidden when comments are off or read-only) and works on both the desktop app and the cloud viewer.

## Detach handling

When a cell re-runs, the execution an output comment was anchored to is gone. Rather than stranding the comment on a dead execution, it auto-demotes to a notebook-level comment (the existing `demote_thread_anchor_to_notebook`, newly exposed to the frontend through a wasm binding).

The important part is *when* it demotes. Demotion treats absence of data as "not loaded yet", never as detachment: it fires only on positive evidence - the cell is loaded and a current execution has landed, and the anchored execution differs from the current one (or the anchored output is gone from the loaded execution). This matters because on a reconnect or notebook switch the comments projection is preserved while the cell and execution stores momentarily reset to empty; treating that window as detachment would permanently demote a valid comment and sync the wrong anchor to every peer. The decision is a pure, tested function, and the once-per-thread guard only triggers on a genuine detach.

Validation before creation mirrors the source-range liveness check: you can't anchor an output comment to a stale execution.

## Verification

`cargo check -p runtimed-wasm`, `cargo test -p comments-doc` (incl. the Output demote test), full `vp test` (2450 passing, with load-window cases that assert no demote), and notebook / cloud / Elements typecheck. The new wasm binding is reflected in the generated and checked-in declarations.

Manual UX of the gutter button hasn't been exercised in a browser yet; worth a look when checking out.

### DIFF
--- a/apps/notebook-cloud/src/runtimed-wasm.generated.d.ts
+++ b/apps/notebook-cloud/src/runtimed-wasm.generated.d.ts
@@ -65,6 +65,7 @@ declare module "../../notebook/src/wasm/runtimed-wasm/runtimed_wasm.js" {
       afterThreadId: string | null | undefined,
       createdAt: string,
     ): unknown;
+    demote_comment_thread_to_notebook(threadId: string): void;
     flush_comms_doc_sync(): Uint8Array | undefined;
     flush_comments_doc_sync(): Uint8Array | undefined;
     flush_local_changes(): Uint8Array | undefined;
--- a/apps/notebook-cloud/viewer/live-sync.ts
+++ b/apps/notebook-cloud/viewer/live-sync.ts
@@ -1741,6 +1741,7 @@ export function syncableCloudHandle(handle: NotebookHandle): SyncableHandle {
     get_comments_doc_heads_hex: commentsSync.get_comments_doc_heads_hex,
     init_comments_sync_target: commentsSync.init_comments_sync_target,
     create_comment_thread: commentsSync.create_comment_thread,
+    demote_comment_thread_to_notebook: commentsSync.demote_comment_thread_to_notebook,
     reply_comment_thread: commentsSync.reply_comment_thread,
     resolve_comment_thread: commentsSync.resolve_comment_thread,
     reopen_comment_thread: commentsSync.reopen_comment_thread,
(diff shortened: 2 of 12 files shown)


rgbkrk:
@pullfrog review

Comment on cell outputs (stable-ID anchoring) with auto-demote-to-notebook on detach. A prior review flagged the demote-on-detach as a blocker (it demoted on absence-of-data during the load window, corrupting valid comments); that's now fixed. Focus on:

- **Demote-on-detach safety** (`src/components/notebook/output-comment-demotion.ts` `shouldDemoteOutputCommentAnchor`): it must demote ONLY on positive evidence - cell loaded AND a current execution landed (`currentExecutionId !== null`), then execution mismatch or missing output. Confirm a reconnect/notebook-switch window (cell store empty, execution null) cannot demote, and that the once-per-thread `attemptedThreadIds` guard therefore only fires on genuine detachment. Convergence: concurrent peers demote to the same notebook anchor (idempotent).
- **Validate-before-create** mirrors `sourceRangeAnchorMatchesCurrentCell` (rejects a stale execution).
- **Button gating parity**: the output-gutter button matches `onCreateSourceComment` gating (hidden when comments off / read-only / no outputs) on BOTH desktop and cloud; no regression to source/markdown comments.
- **wasm binding**: `demote_comment_thread_to_notebook` soundly wraps the Rust `demote_thread_anchor_to_notebook`, and the checked-in cloud `.d.ts` + shared handle interface expose it.

rgbkrk:
@pullfrog review

Addressed both findings (commit 181b5a2):

- **Guard-before-success**: `demoteThreadToNotebook` now returns whether the demote committed; the hook marks a thread attempted only on success, so a failed demote retries on the next execution/output bump instead of being permanently silenced. Both handlers return the boolean and no longer write the user-facing comments-error state (it's a background auto-repair).
- **Cross-execution output_id**: `shouldDemoteOutputCommentAnchor` now only evaluates execution-pinned anchors; an anchor with no `execution_id` never demotes on a missing `output_id` (different run). Added a test for the unpinned case.

Full `vp test` green (2451); notebook + cloud typecheck clean.

PR #3767, opened 2026-06-19
fix(notebook-cloud): recover room hosts from checkpoints

quillaid:
## Summary

- recover hosted room materializers from the durable room checkpoint when the live `RoomHostHandle` hits the recoverable WASM/Automerge aliasing boundary and no published snapshot exists
- keep malformed presence frames out of user-facing `cloud_frame_rejected` notices while still counting them toward the rejected-frame streak
- refresh current-main source guardrail tests that were stale after the shared title/comments work

## Why

Preview showed repeated websocket reconnects with `cloud sync socket closed (1011): room sync failed`. `wrangler tail` showed the room host repeatedly failing notebook/runtime/comms sync with:

```text
recursive use of an object detected which would lead to unsafe aliasing in rust
```

Recovery already fell back to a published snapshot, but live beta rooms without a published revision logged `latest_published_snapshot_unavailable` and stayed poisoned. Reloading the durable checkpoint creates a fresh room host from saved Automerge bytes, which preserves unpublished room state and clears the bad in-memory host object.

The visible `unsupported presence payload` banner was a separate symptom from an ephemeral presence frame. Bad cursor/focus/heartbeat payloads should not put the notebook in an unable-to-load state, so those rejects now avoid a `cloud_frame_rejected` control frame.

## Validation

- `node --import tsx --test test/room-materializer.test.ts test/notebook-room.test.ts` from `apps/notebook-cloud` (105/105)
- `pnpm --dir apps/notebook-cloud run test` (1128/1128)
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/src/notebook-room.ts
+++ b/apps/notebook-cloud/src/notebook-room.ts
@@ -75,6 +75,7 @@ interface SelectedRuntimePeerSession {
 
 interface RejectFrameOptions {
   countsTowardStreak?: boolean;
+  sendControl?: boolean;
 }
 
 interface PeerCloseOptions {
@@ -861,6 +862,7 @@ export class NotebookRoom {
           peer,
           frame.type,
           `unsupported presence payload: ${String(error)}`,
+          { sendControl: false },
         );
         return;
       }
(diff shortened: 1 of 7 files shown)


rgbkrk:
@pullfrog review

PR #3768, opened 2026-06-20
feat(markdown): keep edit mode on blur, clean rendered copy, comment menu in source

rgbkrk:
Markdown cell editing and copy fixes from the preview review.

- **Stay in edit mode on blur.** A markdown cell no longer auto-flips back to rendered when you click away or focus another cell. The editing/rendered choice is local per-viewer state (not synced, not persisted); rendering now happens only on an explicit toggle (the render button, Escape, Ctrl+Enter, or Tab navigation). `readOnly` still forces rendered.
- **Clean copy of rendered markdown.** Selecting rendered markdown and copying now writes `text/plain` as the raw markdown source for the selection (reconstructed from the run-span source mappings) plus a cleaned `text/html`, on both the right-click menu and keyboard Cmd/Ctrl+C. The heading `#` anchor is `select-none` and list markers are stripped, so neither leaks into the copy.
- **Comment menu in the markdown source editor.** The CodeMirror editor shown when editing a markdown cell now has the right-click "Add comment" item (it was only on code cells) and refreshes comment highlights on entering edit mode. The selection affordance and Mod-Alt-m already worked there.

## Verification

`vp check`, full `vp test` (2456), notebook + cloud typecheck. Built by codex, reviewed by Claude (verdict: clean).

### DIFF
--- a/apps/notebook/src/components/MarkdownCell.tsx
+++ b/apps/notebook/src/components/MarkdownCell.tsx
@@ -2,6 +2,7 @@ import type { EditorView, KeyBinding } from "@codemirror/view";
 import { Check, Pencil } from "lucide-react";
 import {
   memo,
+  type ClipboardEvent,
   type MouseEvent,
   type PointerEvent,
   type ReactNode,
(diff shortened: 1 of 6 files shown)


rgbkrk:
@pullfrog review

Markdown cell edit/copy from the preview review: (1) stop auto-rendering on blur/focus-loss (local state only; explicit toggles still render; readOnly still forces rendered), (2) clean rendered-markdown copy - text/plain = raw markdown via the source-span mapping + cleaned text/html, stripping the heading "#" anchor (select-none) and list markers, on both right-click and Cmd/Ctrl+C, (3) the markdown source editor gets the right-click "Add comment" menu + highlight refresh. Worth checking: the onCopy preventDefault only fires when it resolves a source anchor (no swallowing normal copies), removing the blur path leaves no dangling editor registration, and no regression to code cells or cloud.

PR #3769, opened 2026-06-20
feat(outputs): copy image from a context menu on raster outputs

rgbkrk:
Adds a context menu with "Copy image" to inline raster image outputs, from the preview review.

Raster images (png/jpeg/gif/webp/bmp) already render inline as a real `` in the host DOM (they are main-DOM-safe; svg stays isolated), so this is a UI-only change with no pipeline or isolation changes:

- `ImageOutput`'s image gets a `NotebookContextMenu` (surface `output`) with a "Copy image" action.
- `copy-image.ts` converts the image source (data URL or fetched blob) to a Blob and writes it via `navigator.clipboard.write([ClipboardItem])`, inside the user-gesture context-menu callback. Non-PNG rasters are re-encoded to PNG via a canvas first, since `ClipboardItem` image writes are effectively PNG-only in browsers. Failures are caught and logged, never thrown.

## Known limitation

For outputs that render inside the sandboxed iframe (isolated renderer), the frame lacks `clipboard-write` and a same-origin context, so the copy no-ops there (caught and logged). Raster images take the main-DOM path in the common case, where copy works. The iframe isolation model is intentionally left untouched.

## Verification

`vp check`, `vp test` (copy-image + image-output suites), notebook + cloud typecheck. Built by codex, reviewed by Claude (verdict: minor; the two findings - PNG-only writes and the sandbox-path no-op - are addressed/documented here).

### DIFF
--- a/src/components/outputs/__tests__/copy-image.test.ts
+++ b/src/components/outputs/__tests__/copy-image.test.ts
@@ -1,0 +1,86 @@
+import { afterEach, describe, expect, it, vi } from "vite-plus/test";
+import { logger } from "@/lib/logger";
+import { copyRasterImageToClipboard } from "../copy-image";
+
+vi.mock("@/lib/logger", () => ({
+  logger: { warn: vi.fn() },
+}));
+
+const originalClipboard = Object.getOwnPropertyDescriptor(navigator, "clipboard");
+const originalClipboardItem = Object.getOwnPropertyDescriptor(globalThis, "ClipboardItem");
+
+function readBlobBytes(blob: Blob): Promise<number[]> {
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

Copy image from a context menu on inline raster outputs. Raster images already render inline as  (not iframed); this adds a NotebookContextMenu + copy-image util. Addresses the prior review's two minors: non-PNG rasters are now re-encoded to PNG via canvas before ClipboardItem write (PNG-only browser constraint), and the sandbox-path no-op is documented (clipboard-write is blocked in the isolated iframe; raster takes the main-DOM path in the common case). Worth checking: the canvas PNG normalization (bitmap.close, null-guards), the copy runs in the user-gesture context-menu callback, the menu is scoped to the image only, and no change to the iframe isolation model.

PR #3770, opened 2026-06-20
feat(comments): redesign the selection affordance as a dot, fix markdown anchoring

rgbkrk:
Two comment-affordance fixes from the preview review, plus the dot redesign.

## Markdown selections now anchor to the actual text
`sourceRangeAnchorFromRenderedMarkdownSelection`/`...Runs` rejected an anchor whenever `exact_quote !== selectedText`. But `exact_quote` is the markdown *source* slice (e.g. `**bold**`) while `selectedText` is the rendered DOM text (`bold`), so they differ for every styled run, and the affordance silently disappeared on styled or cross-run rendered selections. Removed that guard; the remaining `sourceRangeAnchorFromOffsets` guards (empty/zero-length/out-of-bounds/whitespace/oversize) still reject degenerate anchors. The CodeMirror source-view path was never affected. Tests updated to assert styled/cross-run selections now resolve.

## The dot affordance, redesigned (no icon)
The comment-plus icon is gone. The affordance is now an author-colored dot that breathes at rest, **peeks open once when it appears** (so a fresh selection can tell what it is), and springs into a "Comment" speech bubble with a tail on hover or keyboard focus, using the same spring as the running motion. The bubble shape is the cue; no icon.

- Fixes the import bug from #3763: the app entry `apps/notebook/src/index.css` never loaded `comment-affordance.css`, so the dot rendered as a full-size icon. Now imported.
- The CodeMirror tooltip anchors at `selection.head` (the live end of the selection) instead of `selection.to`, so it follows your pointer.
- Shared by the editor and rendered-markdown planes via one stylesheet; showcased in Elements (Comment surfaces) with a `--comment-affordance-color` knob.

Faithful preview (exact shipped CSS): https://img.runt.run/2026/06/20/7602654e12fd.html

## Verification
`vp check`, full `vp test` (2452), notebook + cloud + Elements typecheck. Anchor fix + tests built by codex and reviewed clean; the dot redesign is hand-built.

### DIFF
--- a/apps/elements/components/comment-selection-affordance-example.tsx
+++ b/apps/elements/components/comment-selection-affordance-example.tsx
@@ -1,20 +1,47 @@
 "use client";
 
+import type { CSSProperties } from "react";
 import { CommentSelectionAffordance } from "@/components/comments/CommentSelectionAffordance";
+
+function authorColor(color: string): CSSProperties {
+  return { "--comment-affordance-color": color } as CSSProperties;
+}
+
+const SAMPLES: Array<{ text: string; color: string }> = [
+  { text: "selected prose", color: "#6366f1" },
+  { text: "another author", color: "#16a34a" },
+  { text: "a third voice", color: "#ea580c" },
+];
 
 export function CommentSelectionAffordanceExample() {
   return (
     <div className="not-prose my-6">
       <div className="mx-auto max-w-[760px] border border-border bg-background px-6 py-5 text-foreground shadow-sm max-sm:px-4">
-        <p className="mb-4 text-sm text-muted-foreground">
-          Resting, the affordance is a quiet dot that stays out of the way while you select and copy
-          text. Hover it, or focus it with the keyboard, and it grows in place into a comment
-          button. The editor and rendered-markdown planes share this one surface.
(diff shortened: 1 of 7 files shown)


rgbkrk:
@pullfrog review

Two parts. (1) Markdown anchoring: removed the `exact_quote !== selectedText` guard that rejected every styled/cross-run rendered selection (exact_quote is source markup, selectedText is rendered text); the sourceRangeAnchorFromOffsets guards still reject degenerate anchors; tests updated; CM source path unaffected. (2) Dot redesign (hand-built UI): the comment-plus icon is removed; the affordance is an author-colored dot that breathes, peeks open once on appear, and springs into a "Comment" bubble on hover/focus. Worth checking: the peek lifecycle cleans up (React clears the timeout on unmount; the CM tooltip create() clears its peek timer in destroy()), the #3763 CSS-import bug is fixed (apps/notebook/src/index.css now imports comment-affordance.css at the right relative depth), selection.head positioning doesn't change the built anchor, prefers-reduced-motion is honored, and there's no regression to code-cell or markdown commenting. Elements showcases it with a --comment-affordance-color knob.

PR #3771, opened 2026-06-20
fix(comments): grow the comment bubble past the hit-area on iOS Safari

rgbkrk:
Follow-up to #3770: the comment bubble clipped to a single glyph ("C"/"R") on iOS Safari.

The affordance button had a fixed `width: 26px` for the hit area. On WebKit/iOS Safari a flex child cannot exceed a fixed-width flex parent, so the open bubble was capped at the button width even though the dot's own `max-width` allowed more (Chromium let the child overflow, which is why desktop looked right). Make the button content-sized with `padding` for the hit area and `flex: none` on the dot, so the bubble grows to the full word on every engine.

Mobile preview (exact shipped CSS): https://img.runt.run/2026/06/20/3633437f8d0d.html

## Verification
`vp check` clean. CSS-only change to `src/styles/comment-affordance.css`.

### DIFF
--- a/src/styles/comment-affordance.css
+++ b/src/styles/comment-affordance.css
@@ -19,10 +19,10 @@
   display: inline-flex;
   align-items: center;
   justify-content: center;
-  /* Generous transparent hit area around the resting dot. */
-  width: 26px;
-  height: 26px;
-  padding: 0;
+  /* Content-sized with padding for the hit area. A fixed width would cap the
+     dot's growth: on WebKit/iOS Safari a flex child cannot exceed a fixed-width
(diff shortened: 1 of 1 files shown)


rgbkrk:
@pullfrog review

Follow-up to #3770. CSS-only: the affordance hit-area button was a fixed 26px width, and WebKit/iOS Safari caps a flex child to its fixed-width flex parent, so the open comment bubble clipped to one glyph on mobile. Now the button is content-sized with padding for the hit area and the dot is flex:none, so the bubble grows to the full label on every engine. Worth confirming: the resting hit area is still ~26px, the dot still centers, and the markdown/CM positioning isn't visibly shifted by the padding change.

PR #3772, opened 2026-06-20
fix(outputs): keep copy image inside the user gesture (Safari)

rgbkrk:
Follow-up to #3769 (pullfrog's user-gesture concern).

`copyRasterImageToClipboard` awaited `fetch`/canvas encode before `navigator.clipboard.write`, so on Safari the transient user activation was already gone by the time `write` ran - the copy silently rejected for fetched (daemon blob-store) or non-PNG images. Matplotlib PNG data URLs happened to survive because their await is a single microtask.

Now `write()` is called synchronously inside the gesture with a `ClipboardItem` whose value is a `Promise<Blob>`; the fetch/canvas resolution is deferred inside the ClipboardItem, which keeps the activation. Failures are still caught and logged.

Tests updated to assert the ClipboardItem receives a promise and `write` fires synchronously.

## Verification
`vp check`, `copy-image` + `image-output` test suites green. (tsc in the fresh worktree only reports un-generated WASM bindings; the change is isolated to `copy-image.ts`.)

### DIFF
--- a/src/components/outputs/__tests__/copy-image.test.ts
+++ b/src/components/outputs/__tests__/copy-image.test.ts
@@ -40,8 +40,8 @@ describe("copyRasterImageToClipboard", () => {
   it("writes a base64 data URL image through ClipboardItem", async () => {
     const write = vi.fn().mockResolvedValue(undefined);
     const ClipboardItemMock = vi.fn(function ClipboardItem(
-      this: { items: Record<string, Blob> },
-      items: Record<string, Blob>,
+      this: { items: Record<string, Promise<Blob> | Blob> },
+      items: Record<string, Promise<Blob> | Blob>,
     ) {
       this.items = items;
     });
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review

Follow-up to your #3769 user-gesture concern. copyRasterImageToClipboard now calls navigator.clipboard.write synchronously inside the gesture with a ClipboardItem whose value is a Promise<Blob>, instead of awaiting fetch/canvas before the write (which dropped the Safari transient activation). Worth confirming: the write is genuinely synchronous at the call site (onSelect -> copyRasterImageToClipboard, no await before it), the ClipboardItem promise-value pattern is the right Safari approach, and failures (rejected blob promise or rejected write) are still caught + logged.

rgbkrk:
@pullfrog review

Addressed the stale mock: image-output.test.tsx no longer uses mockResolvedValue(undefined) for copyRasterImageToClipboard (it returns void now) - plain vi.fn(). The Safari gesture fix is unchanged.

rgbkrk:
@pullfrog review

Addressed the test-fragility nit: the failure test now uses vi.waitFor to poll for the logged warning instead of counting microtask ticks.

PR #3773, opened 2026-06-20
fix(comments): author-color the selection affordance and clear resolved highlights

rgbkrk:
Polish for the comment-on-selection affordance and resolved-thread highlights, following the dot redesign in #3770 and the iOS clip fix in #3771. The affordance was rendering as a near-black bubble inside a boxed CodeMirror tooltip, a native "Comment on selection" tooltip duplicated its own label, and resolved threads kept their inline highlight.

## Fixes

**Author color.** The affordance tinted itself with `--primary`, a near-black neutral in the default theme. It now reads `--comment-author-color`, set on the document element from `colorForActorIdentity(localActor)`, the same color as the author's cursor and edits. `--primary` stays as the fallback when no author color is in scope.

**Tooltip chrome.** CodeMirror merges its `cm-tooltip` classes onto the affordance button itself, so the element is both `.comment-affordance` and `.cm-tooltip` and the tooltip border/background/shadow painted a box around the dot. The old `:has(.comment-affordance)` selector never matched, because the classes sit on the same element rather than a descendant. The chrome is now stripped on `.cm-tooltip.comment-affordance`; the theme sets it with editor-scoped specificity, so `!important` wins the cascade. The hit-area padding stays.

**Native tooltip.** Dropped the `title` attribute on both affordance planes (CodeMirror editor and rendered markdown). The bubble already labels itself "Comment", so the native title only duplicated it on hover. `aria-label` keeps the control accessible.

**Resolved highlights.** Resolved threads now leave no inline highlight in either the editor (`comment-highlights.ts`) or rendered markdown (`MarkdownCell.tsx`). They stay reachable under "Show resolved" in the rail.

## Behavioral coverage

| State | Before | After |
|-------|--------|-------|
| Affordance at rest | Near-black dot | Author-colored dot |
| Affordance open | "Comment" pill inside a dark tooltip box | Clean "Comment" pill, no box |
| Hover | Native title duplicates the label | Bubble label only |
| Thread resolved | Inline highlight stays | Highlight cleared, thread under "Show resolved" |

Verified live in the Vite dev app: author-colored dot springs to a "Comment" pill with no box and no native tooltip, and resolving a thread removes its inline highlight (confirmed zero highlight decorations remain in the editor).

The green left bar some builds show on a commented cell is the focused-cell ribbon (`colors.ribbon.focused`), driven by focus, not comments. Out of scope here.

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -1,5 +1,6 @@
 import { type ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
+  colorForActorIdentity,
   deriveEnvManager,
   deriveRuntimeKind,
   NotebookClient,
@@ -794,6 +795,19 @@ function AppContent() {
     const [localPrincipal] = splitNotebookActorPrincipalOperator(localActor);
     return [{ participantKey: localPrincipal, label }];
   }, [localActor, peerLabel]);
+
+  // Tint the comment-on-selection affordance with the local author's canonical
+  // color (the same color as their cursor and edits), exposed as a CSS var the
+  // shared affordance styles read. Without it the dot falls back to --primary,
+  // a near-black neutral.
+  useEffect(() => {
+    if (!localActor) return;
+    const color = colorForActorIdentity(localActor);
+    document.documentElement.style.setProperty("--comment-author-color", color);
+    return () => {
+      document.documentElement.style.removeProperty("--comment-author-color");
+    };
+  }, [localActor]);
 
   const resolveCommentAuthor = useCallback(
     (actorLabel: string): CommentAuthor => {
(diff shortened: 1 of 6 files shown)


rgbkrk:
@pullfrog review

Polish on the comment affordance after #3770/#3771. Four fixes:

- Author-colors the selection dot. It tinted with `--primary` (near-black in the default theme); now reads `--comment-author-color` set from `colorForActorIdentity(localActor)`, with `--primary` as fallback.
- Strips the CodeMirror tooltip box. The old `:has(.comment-affordance)` selector never matched because CM merges its `cm-tooltip` classes onto the affordance button itself, so the chrome is stripped on `.cm-tooltip.comment-affordance` (needs `!important` to beat editor-scoped theme specificity).
- Drops the duplicate native `title` tooltip on both planes; `aria-label` keeps it accessible.
- Resolved threads leave no inline highlight in editor or rendered markdown; still reachable under "Show resolved".

Verified live in the dev app and `vp check` / `vp test run` are green (2461 passed, 0 failures).

PR #3776, opened 2026-06-20
feat(comments): quiet the selection affordance and stage its open with WAAPI

rgbkrk:
The comment-on-selection affordance auto-folded into a "Comment" pill while you dragged a selection, sprang open on its own, and jumped into position before expanding. Most text selection is for editing code, not commenting, so this makes it stay out of the way and open deliberately when you reach for it. Follows #3773.

## Quiet at rest

A small (8px), flat, author-colored dot. No auto-peek, no breathe, no shadow. It stays a dot through the whole selection and opens only on hover or keyboard focus, on both the editor and rendered-markdown planes.

## Staged open via the Web Animations API

Hover and focus drive a staged morph: the dot grows to a circle, spreads to a pill, then the label fades in once there is room. WAAPI over a CSS keyframe for two reasons:

- It reverses cleanly from wherever the open got to. A CSS keyframe snaps back, and a transition-reverse cannot replay the intermediate "circle" stop.
- It measures the label's natural width and animates to that, so the pill fits its content instead of a guessed cap that clips long labels.

Events are the input, the `Animation` handle is the state: enter/focus play it forward, leave/blur play it back. Sizes and easing live in CSS custom properties the helper reads, so there is one source of truth.

## No reposition jump

The tooltip reserves the open pill's height, so CodeMirror measures and positions it once. The dot is centered in that height and grows from its center, instead of the box growing taller and re-anchoring above the line in an unanimated upward jump.

## Placement follows the cursor

In the editor the dot sits at the selection head; a leftward selection flips it to the left of the head and grows leftward, keeping it off the selected text. Rendered markdown has no cursor, so it stays to the right of the full selection.

## Scope

Removes the old peek capability (prop, state, timer, and CSS) rather than leaving it as dead code. The author-color tint and resolved-highlight behavior from #3773 are unchanged.

| State | Before | After |
|-------|--------|-------|
| Dragging a selection | "Comment" pill folded out | Quiet flat dot |
| Resting dot | 14px, breathing, shadowed | 8px, flat, still |
| Open | All dimensions tween at once, snaps back | Staged grow → spread → reveal, reverses cleanly |
| Tooltip on grow | Jumps up to re-anchor | Reserved height, grows from center |
| Leftward selection (editor) | Dot on the right, over the text | Dot flips left, off the text |

Verified live in the dev app across both planes; `vp check` and `vp test run` green (2461 passed, 0 failures).

### DIFF
--- a/apps/elements/components/comment-selection-affordance-example.tsx
+++ b/apps/elements/components/comment-selection-affordance-example.tsx
@@ -18,10 +18,11 @@ export function CommentSelectionAffordanceExample() {
     <div className="not-prose my-6">
       <div className="mx-auto max-w-[760px] border border-border bg-background px-6 py-5 text-foreground shadow-sm max-sm:px-4">
         <p className="mb-5 text-sm text-muted-foreground">
-          A small author-colored dot that breathes at rest, peeks open once when it appears so you
-          can tell what it is, then springs into a "Comment" speech bubble on hover or keyboard
-          focus. No icon: the bubble shape is the cue. The editor and rendered-markdown planes share
-          this one surface, tunable here via <code>--comment-affordance-color</code>.
+          A small, flat, author-colored dot that stays out of the way while you select. It opens
+          into a "Comment" pill only on hover or keyboard focus, a staged grow then spread then
+          reveal driven by the Web Animations API. No icon: the pill shape and label are the cue.
+          The editor and rendered-markdown planes share this one surface, tunable here via{" "}
+          <code>--comment-affordance-color</code>.
         </p>
         <div className="flex flex-wrap items-center gap-x-8 gap-y-4 text-sm">
           {SAMPLES.map((sample) => (
--- a/apps/notebook/src/lib/source-comment-extension.ts
+++ b/apps/notebook/src/lib/source-comment-extension.ts
@@ -1,5 +1,6 @@
 import { StateEffect, StateField, type Extension } from "@codemirror/state";
 import { EditorView, keymap, showTooltip, type Tooltip } from "@codemirror/view";
+import { wireCommentAffordanceMotion } from "@/components/comments/comment-affordance-motion";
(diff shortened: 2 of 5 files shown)


rgbkrk:
@pullfrog review

Follow-up to #3773, reworking the selection affordance's motion and placement.

- Quiet flat 8px dot at rest. No auto-peek, no breathe, no shadow. Opens only on hover/focus, both planes.
- Staged open via the Web Animations API (`src/components/comments/comment-affordance-motion.ts`): grow → spread → reveal. Chosen over a CSS keyframe for clean `reverse()` and to measure the label's natural width. Sizes/easing read from CSS custom properties.
- Tooltip reserves the open pill's height so CodeMirror positions it once; the dot grows from its center instead of re-anchoring with a jump.
- Editor placement follows the cursor: leftward selections flip the dot to the left of the head. Rendered markdown stays right of the full selection.
- Removed the old peek capability (prop, state, timer, CSS) rather than leaving dead code.

A couple of spots worth a careful look: the WAAPI helper's lifecycle (listeners + `Animation` handles disposed in the CM tooltip `destroy()` and the React effect cleanup), and the jsdom/SSR guard (`typeof button.animate !== "function"`). `vp check` and `vp test run` are green (2461 passed).

rgbkrk:
@pullfrog review

Addressed both rough edges from your last pass (commit 51ddf0f):

- **pointercancel gap:** the motion helper now also closes on `pointercancel`, so an interrupted touch (system takes over the pointer for a scroll/gesture, no `pointerleave`) no longer leaves the pill open. Added to the disposer too.
- **Stale Elements prose:** `comment-selection-affordance-example.tsx` no longer describes the removed breathe/peek/speech-bubble behavior; it now matches the quiet flat dot and the staged WAAPI open.

`vp check` and `vp test run` green (2461 passed).

PR #3789, opened 2026-06-21
fix(notebook): preserve state across open and save-as

### LINKED ISSUE #3787: Funky behavior on notebook open/save
Loving nteract as always, but wanted to report some funkiness that's been plaguing me the last couple of days. Running the latest stable build (2.5.1). While opening an existing notebook, it will overwrite the notebook in the current window with no warning to save before replacing the window's contents. And if you save a notebook that you're actively working on, it kills the state (not the kernel) and you have to reissue all commands. Weirdest yet is when I saved a notebook as "bst5" and it changed it to "bst2" and overwrote the other notebook. Runt diagnostics attached, let me know if you need anything else. 

[runt-diagnostics-2026-06-20-122150.tar.gz](https://github.com/user-attachments/files/29166291/runt-diagnostics-2026-06-20-122150.tar.gz)

quillaid:
## Summary

Fixes #3787.

This addresses two desktop state-loss paths found in the attached diagnostics:

- macOS file-open events no longer retarget an already-open pathless notebook window after startup. They still focus an existing matching file window, and deferred Finder startup opens may reuse the initial placeholder window, but ordinary later opens create a separate window so untitled work is not overwritten.
- Desktop Save As no longer shuts down and relaunches the kernel. The daemon already owns the UUID-stable Save As transition by rebinding the path, writing `RuntimeStateDoc.path`, refreshing project context, and publishing workstation attachment facts. Restarting in the Tauri command discarded live kernel/session state for a file-system rename.

## Validation

- `cargo test -p notebook --lib`
- `cargo test -p runtimed test_save_as_rebind_replaces_file_lifecycle_and_runtime_path`
- `cargo xtask lint --fix`
- `git diff --check`

### DIFF
--- a/crates/notebook/src/lib.rs
+++ b/crates/notebook/src/lib.rs
@@ -15,7 +15,6 @@ pub mod typosquat;
 extern crate runtimed_client as runtimed;
 pub use runtimed::runtime::Runtime;
 
-use notebook_protocol::connection::LaunchSpec;
 use notebook_protocol::protocol::{NotebookRequest, NotebookResponse, SaveErrorKind};
 use notebook_sync::RelayHandle;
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
@pullfrog review this for me

rgbkrk:
Slightly different UX direction to migrate the kernel on save as but maybe that's better for users.

PR #3793, opened 2026-06-21
refactor(frontend): converge icon sizing from h-N w-N to size-N

rgbkrk:
## Overview

Converges icon sizing across the shared `src/components/**` and `apps/notebook/src/**` trees from the older `h-N w-N` form to the Tailwind v4 `size-N` shorthand. 113 equal `h-N w-N` pairs converted across 22 files, so the frontend reads as one convention instead of the ~50/50 split it had drifted into.

## Design decision

The audit found 113 icons using `h-N w-N` while newer code (NotebookCommentsPanel, TrustDialog, the Elements catalog, and the shadcn-generated `button`/`avatar` primitives) uses `size-N`. Tailwind v4 ships `size-N` as the shorthand for equal width + height. The split was purely historical - older files kept the long form, newer files adopted the shorthand. One mechanical pass collapses it.

Rectangle patterns (`h-N w-M` where N != M, e.g. `h-9 w-48` for a combobox, `h-32 w-2` for a vertical progress bar) are left as-is: `size-N` only represents square dimensions. Seven such patterns exist and are correct as written.

Scope excludes `src/components/ui/**` (upstream shadcn primitives, a few still use `h-N w-N` by design) and test files.

## Test plan

- [x] `cargo xtask lint --fix` (Rust fmt, `vp check --fix`, raw-byte, ruff, ty) - all pass
- [x] `vp check` (format + lint + type, the CI gate) - clean, 1055 files
- [x] Notebook component vitest suite - 257 tests pass across 33 files (NotebookNotice, PoolErrorBanner, NotebookContextMenu, NotebookDocumentRail, EnvBuildDecisionDialog, etc.)
- [ ] Visual QA: spot-check the command toolbar, dependency panels, and global find bar in light + dark to confirm icon sizes unchanged (the conversion is width+height equivalent, so no visual change is expected)

### DIFF
--- a/apps/notebook/src/components/CodeCell.tsx
+++ b/apps/notebook/src/components/CodeCell.tsx
@@ -906,7 +906,7 @@ export const CodeCell = memo(function CodeCell({
                   title="Comment on outputs"
                   aria-label="Comment on outputs"
                 >
-                  <CommentMarkIcon className="h-3.5 w-3.5" aria-hidden="true" />
+                  <CommentMarkIcon className="size-3.5" aria-hidden="true" />
                 </button>
               ) : null}
               {onToggleOutputsHidden ? (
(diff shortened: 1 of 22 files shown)


rgbkrk:
@pullfrog verify for me

PR #3798, opened 2026-06-21
[codex] Document Codex plugin cache reset

rgbkrk:
## Summary

- document the Codex plugin cache reset path in the root README
- add the same reset guidance to the generated `agent-plugins` README template
- teach both bundled REPL skills how to route users through marketplace refresh, plugin reinstall, `codex mcp list`, and a fresh Codex session when tools do not appear

## Why

Codex can have a healthy `nteract/agent-plugins` marketplace checkout while the installed plugin cache or current session tool surface remains stale. The docs already covered restart-first troubleshooting, but not the practical local repair sequence.

## Validation

- `bash -n scripts/assemble-plugin-dist.sh`
- `python3 scripts/ci/check-docs-guidance.py`
- temp run of `scripts/assemble-plugin-dist.sh` with fake sidecars, confirming the generated README includes the new troubleshooting copy

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,17 @@ Restart Codex, then open the plugin directory, choose the nteract marketplace, a
 
 The distribution repository also includes a `nightly` plugin entry for pre-release builds.
 
+If Codex still does not show notebook tools after restarting, refresh the marketplace and reinstall the selected plugin:
+
+```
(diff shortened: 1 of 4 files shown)


rgbkrk:
@pullfrog review

PR #3807, opened 2026-06-22
fix(notebook-cloud): require live runtime peer for compute actions

quillaid:
## Summary

- treat a ready RuntimeStateDoc workstation attachment as executable only when the cloud room currently observes a live runtime peer
- keep stale attachments visible as previous compute needing attention instead of surfacing run/restart/interrupt controls
- hide stale kernel status labels when the runtime peer is gone so the toolbar does not report idle for an unreachable kernel

## Validation

- cargo xtask lint --fix
- pnpm test:run -- apps/notebook-cloud/test/viewer-shell-capabilities.test.ts packages/runtimed/tests/notebook-shell-capabilities.test.ts


rgbkrk:
@pullfrog review

PR #3808, opened 2026-06-22
fix(notebook): preserve title editor caret order

rgbkrk:
## Summary

- Keep the notebook/cloud title `contentEditable` DOM-owned while editing so React does not replace the text node after each keystroke.
- Track the draft title in a ref and seed the editable DOM once when editing begins.
- Add a regression test that types `Outbound` into the title and asserts the visible/submitted order.

## Verification

- `pnpm test:run src/components/notebook/__tests__/DocumentTitle.test.tsx`
- `pnpm test:run`
- `cargo xtask lint --fix`
- `pnpm --dir apps/notebook-cloud dev:browser --rebuild --next /n/title-caret-smoke/Title%20Caret%20Smoke`
- Browser against local wrangler `http://127.0.0.1:45221`: created a local owner notebook, renamed the cloud title to `Outbound`, and verified the visible title, document title, route slug, and `/api/n/:id` catalog title all stayed `Outbound`.

### DIFF
--- a/src/components/notebook/DocumentTitle.tsx
+++ b/src/components/notebook/DocumentTitle.tsx
@@ -66,13 +66,13 @@ export function DocumentTitle({
   title,
 }: DocumentTitleProps) {
   const [editing, setEditing] = useState(false);
-  const [draftTitle, setDraftTitle] = useState(renameTitle);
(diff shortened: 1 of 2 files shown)


rgbkrk:
@pullfrog review

PR #3809, opened 2026-06-23
fix(mcp-app): inline small raster fallback outputs

rgbkrk:
## Summary

- warn MCP App hosts when daemon output frames are unavailable because `frameDomains` does not grant the daemon blob origin
- fall back to data URI rendering for small static raster image blobs when the app has to use `srcdoc`
- keep large raster images, HTML, widgets, and other interactive or binary outputs on the existing non-inline paths

## Verification

- `cargo xtask lint --fix`
- `pnpm test:run apps/mcp-app/src src/components/isolated/__tests__`
- `pnpm test:run src/isolated-renderer/__tests__`
- `git diff --check`

## Notes

This remains a draft while CI runs. The host-side fix is still to include the daemon blob/output-frame origin in MCP Apps `hostCapabilities.sandbox.csp.frameDomains`.

### DIFF
--- a/apps/mcp-app/src/components/__tests__/shared-cell-outputs.test.tsx
+++ b/apps/mcp-app/src/components/__tests__/shared-cell-outputs.test.tsx
@@ -100,6 +100,38 @@ describe("SharedCellOutputs", () => {
     );
   });
 
+  it("waits for host CSP before enabling the raster image inline fallback", () => {
+    render(
+      <SharedCellOutputs
+        cell={{
+          cell_id: "cell-1",
+          cell_type: "code",
+          source: "display('hi')",
+          execution_count: 1,
+          status: "done",
+          outputs: [
+            {
+              output_id: "output-1",
+              output_type: "display_data",
+              data: {
+                "image/png": "http://localhost:47830/blob/image-hash",
+              },
+            },
+          ],
+        }}
+        blobBaseUrl="http://localhost:47830/"
+        hostCapabilities={null}
+      />,
+    );
+
+    expect(mcpAppOutputFrameSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        outputDocumentUrl: null,
+        inlineRasterBlobImages: false,
+      }),
+    );
+  });
+
   it("warns the host when daemon output frames are blocked by frameDomains", async () => {
     const sendLog = vi.fn();
     setHostLogSink({ sendLog });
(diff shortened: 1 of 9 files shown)


rgbkrk:
@pullfrog review please

rgbkrk:
Spec cross-check and Pullfrog follow-up:

- Verified against local `modelcontextprotocol/ext-apps` at `fa127449` and stable spec `specification/2026-01-26/apps.mdx`.
- The nteract server remains on path: resource UI metadata declares the daemon blob origin in `resourceDomains`, `connectDomains`, and `frameDomains`.
- The fallback remains app-local: structured tool output still carries daemon blob URLs; the MCP app fetches those URLs and only then converts small static raster image blobs to `data:` URIs when no daemon output-frame URL is available.
- This aligns with the spec model where `frameDomains` maps to `frame-src`, omitted or ungranted frame domains mean nested frames may be unavailable, and `data:` is allowed for images/media under restrictive host defaults.
- Pullfrog's dropped inline edge case was confirmed and fixed in `ab4d90a3f`: if `Content-Length` is absent, the app now reads the response and applies the existing byte cap after `arrayBuffer()` instead of falling back to an unloadable blob URL.

Local verification after the follow-up:

- `cargo xtask lint --fix`
- `pnpm test:run apps/mcp-app/src src/components/isolated/__tests__`
- `pnpm test:run src/isolated-renderer/__tests__`
- `git diff --check`

rgbkrk:
Follow-up after Pullfrog + MCP Apps spec audit:

- Pushed `303b0f148` to address the audit edge cases.
- `inlineRasterBlobImages` now keys off the same explicit `frameDomains` block predicate as the warning, so unknown/missing host CSP no longer enables the data-URI fallback early.
- `McpAppOutputFrame` now replays outputs when the iframe is recreated for an `outputDocumentUrl` / blob resolver mode transition.
- Server structured content now preserves raster output as Blob Store URLs only. Inline static raster bytes are omitted from MCP tool `structuredContent`, and omitted inline rasters no longer suppress a richer inline `text/html` fallback.

Verified locally:

```bash
cargo test -p runt-mcp structured_ -- --nocapture
pnpm test:run apps/mcp-app/src src/components/isolated/__tests__
pnpm test:run src/isolated-renderer/__tests__
cargo xtask lint --fix
git diff --check
```

Remaining broader audit item, intentionally not folded into this patch: the daemon-served output frame has a broader internal CSP than the MCP resource `_meta.ui.csp` declaration exposes. This PR still improves degradation when a host does not grant `frameDomains`, but the transitive policy for a granted daemon frame should get a separate design pass against the Apps spec and real host behavior.

Issue #3377, opened 2026-06-04
Support Jupyter-style TeX environments in projected markdown


rgbkrk:
## Summary

The Rust/WASM projected markdown path has landed. This issue is now the compatibility follow-up for Jupyter-style bare TeX display environments: standalone TeX environment blocks should project to host-rendered display math, not ordinary paragraph text.

Jupyter Notebook's typesetting examples show markdown source using bare TeX environments such as:

```tex
\begin{align}
a^2 + b^2 &= c^2 \\
\sin^2(\theta) + \cos^2(\theta) &= 1
\end{align}
```

and:

```tex
\begin{equation*}
\left( \sum_{k=1}^n a_k b_k \right)^2 \leq \left( \sum_{k=1}^n a_k^2 \right) \left( \sum_{k=1}^n b_k^2 \right)
\end{equation*}
```

The docs also note that explicit `$$` delimiters are accepted for legacy notebooks, but are not required when using TeX environments.

Reference: https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Typesetting%20Equations.html

## Why this matters

Projected markdown is now the fast host-rendered path for markdown cells and markdown outputs. Without bare TeX environment support, imported or existing Jupyter notebooks can render documented math examples as literal text even though the isolated markdown renderer/MathJax-compatible path would treat them as display math.

## Scope

- Keep `$...$` inline math and `$$...$$` display math working.
- Detect standalone bare display TeX environment blocks, including at least `align`, `align*`, `equation`, `equation*`, `array`, and `eqnarray`.
- Preserve source spans so cursor mapping, rich rendered selection, and future markdown editing hooks remain viable.
- Add focused engine, WASM serializer, and frontend projection tests matching the documented Jupyter examples.
- Keep mixed prose such as `Before \begin{align}...\end{align} after` as normal markdown text; this issue is about standalone display blocks.

## Out of Scope

- Reopening the projected markdown spike. That work has already merged.
- Full MathJax parity for every possible TeX environment or macro.
- Broad fallback-policy changes for ambiguous or unsupported math.

PR #3835, opened 2026-06-27
fix(markdown): widen bare TeX coverage to match JupyterLab

### LINKED ISSUE #3834: Markdown: expand bare TeX coverage to match JupyterLab (cases, matrices, \(...\) inline delimiters)
## Summary

Follow-up to #3377, which landed bare TeX display environments in projected markdown for a curated set (`align`, `align*`, `equation`, `equation*`, `array`, `eqnarray`). Writing the 2.6 changelog demo notebook surfaced three gaps against what JupyterLab renders. In each case the source shows up as literal paragraph text instead of math.

## What's missing today

Confirmed in `v2.6.0-stable.202606251403`:

1. **Inline `\(...\)` and display `\[...\]` delimiters** are not recognized. Only `$...$` and `$$...$$` render.
2. **`\begin{cases}`** renders as raw text.
3. **Matrix environments** (`pmatrix`, and the rest of the `matrix` / `bmatrix` / `vmatrix` family) render as raw text.

`$...$`, `$$...$$`, and the #3377 environment set all work today, including `align` auto-numbering.

## What JupyterLab does

Jupyter's markdown is MathJax-backed (tex2jax with `processEnvironments`). It renders:

- inline: `$...$` **and** `\(...\)`
- display: `$$...$$` **and** `\[...\]`
- any standalone `\begin{ENV}...\end{ENV}` as display math, no `$$` required. The typesetting examples render bare `align`, `equation*`, `eqnarray`, and `vmatrix`.

Reference: https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Typesetting%20Equations.html

The mechanism is the important part. MathJax does not keep an environment whitelist. `processEnvironments` wraps any bare `\begin{X}...\end{X}` in math mode and hands it to the TeX renderer. So `cases`, the matrix family, `gather`, `multline`, `aligned`, `split`, and `alignat` all work simply because they are valid once inside math mode.

## Repro

Each renders as literal text in 2.6 and as math in JupyterLab:

```tex
A complex number \(z = a + bi\) with conjugate \(\bar z = a - bi\).
```

```tex
\begin{cases}
x & \text{if } x \ge 0 \\
-x & \text{otherwise}
\end{cases}
```

```tex
\begin{pmatrix}
\cos\theta & -\sin\theta \\
\sin\theta & \cos\theta
\end{pmatrix}
```

## Suggested direction

#3377 deliberately scoped out "full MathJax parity." This is the case for widening that scope: the gaps hit ordinary notebook math (piecewise definitions, matrices, determinants), and imported Jupyter notebooks will render those as plain text.

Two concrete moves:

1. Add `\(...\)` and `\[...\]` to the recognized delimiters alongside `$` / `$$`.
2. Replace the environment whitelist with MathJax's `processEnvironments` behavior: detect any standalone `\begin{X}...\end{X}` block and route the whole block to the host math renderer, which already handles these environments once it is in math mode. That covers `cases`, `matrix` / `pmatrix` / `bmatrix` / `vmatrix`, `gather`, `multline`, `aligned`, `split`, and `alignat` without enumerating each.

Keep the #3377 constraints: mixed inline prose like `before \begin{align}...\end{align} after` stays normal text, and source spans stay intact for cursor mapping.

## Workaround for now

Authoring against 2.6: use `$...$` for inline, `$$...$$` for display, bare `\begin{align}` for aligned multi-line, and expand the rest by hand (a determinant becomes the shoelace cross-product, a rotation becomes a complex multiplication, `cases` becomes a markdown list).

## Environment

- nteract `v2.6.0-stable.202606251403`
- Surfaced while authoring the 2.6 changelog demo notebook (`Napoleon's Theorem - A MathNet Walkthrough.ipynb`).

rgbkrk:
Closes #3834. Scope narrowed from the original issue after prototyping — see [this comment](https://github.com/nteract/nteract/issues/3834#issuecomment-4819924794).

## What lands

Two clean wins for imported-Jupyter / ordinary-notebook math:

1. **`processEnvironments` parity.** Replaces the curated bare-TeX environment whitelist from #3377 with MathJax's behavior: any balanced `\begin{X}...\end{X}` paragraph routes to the host math renderer. Covers `cases`, `pmatrix`/`bmatrix`/`vmatrix`/`matrix`, `gather`, `multline`, `aligned`, `split`, `alignat` — without enumerating each. KaTeX already renders every one of these once the LaTeX reaches it, so this is pure routing on top of the existing `MathBlock` projection.
2. **`\[...\]` display blocks.** Standalone `\[...\]` paragraphs become display math. Recovered from the raw source span since markdown-rs treats `\[`/`\]` as backslash escapes and drops the leading backslash from the parsed text. The existing wasm `MathBlock` path handled this with zero glue changes.

The `cases` and matrix repros from the issue — the ones that hit ordinary notebook math and imported Jupyter notebooks — are fully covered by move (1).

## Bonus fix

Reading raw source for these blocks also fixes a latent bug: the old path routed environment text through `node.to_string()`, which collapsed LaTeX `\\` line-breaks to `\`. The raw slice preserves them faithfully. The `mixed-tex-proof` fixture test was updated to the correct two-backslash expectation.

## Intentionally out of scope

Inline `\(...\)`. markdown-rs collapses `\(` as a markdown escape into a single `Text` node with no clean node boundary (the leading backslash is stripped from the span), and recovering it means re-deriving markdown's escape-unescaping layer. An adversarial review (`codex`, o-series) of a prototype confirmed a cluster of edge cases — escaped literals (`\\(`), empty/`unterminated` delimiters, and prose-before-math losing its unescaping. Rather than ship an awkward escape-guessing layer, inline `\(...\)` stays as text; `$...$` remains the inline path.

The `\[...\]` case dodges all of this because it's whole-paragraph detection (no prose-splitting).

## Verification

- `cargo test -p nteract-markdown-engine` — 13 passed
- `cargo test -p nteract-markdown-wasm` — 12 passed
- `src/lib/__tests__/markdown-projection.test.ts` — 19 passed; `src/components/markdown/__tests__/ProjectedMarkdownView.test.tsx` — 35 passed
- `cargo xtask lint --fix` — all checks pass
- Rebuilt `runtimed-wasm` (genesis-verified)

The #3377 constraints are preserved: mixed inline prose like `before \begin{align}...\end{align} after` and `before \[ x \] after` stays normal paragraph text, and source spans stay intact for cursor mapping.

### DIFF
--- a/crates/nteract-markdown-engine/src/lib.rs
+++ b/crates/nteract-markdown-engine/src/lib.rs
@@ -358,6 +358,11 @@ fn project_node(node: &mdast::Node, context: &mut ProjectionContext<'_>) -> Proj
                 text = math_source.to_string();
                 copy_text = math_source.to_string();
                 children.clear();
+            } else if let Some(math_source) = bracket_display_tex(context.source, &span) {
+                kind = NodeKind::MathBlock;
+                text = math_source.to_string();
+                copy_text = math_source.to_string();
+                children.clear();
             } else {
                 kind = NodeKind::Paragraph;
             }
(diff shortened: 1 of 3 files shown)


rgbkrk:
@pullfrog review

PR #3849, opened 2026-06-29
fix(mcp): hint unapplied dependency edits

rgbkrk:
## Summary

- add a structured `apply.mode = "none"` response when dependency edits are recorded without an environment action
- include a concrete `suggested_next_call` so agents can immediately retry with `apply="sync"` for adds or `apply="restart"` for removals
- cover the add, remove, and no-hint cases with focused unit tests

## Verification

- `cargo test -p runt-mcp manage_dependencies -- --nocapture`
- `cargo xtask lint --fix`
- `git diff --check`

## Notes

This keeps `apply="none"` as the default behavior, but makes the successful no-install path explicit in the MCP response.

### DIFF
--- a/crates/runt-mcp/src/tools/deps.rs
+++ b/crates/runt-mcp/src/tools/deps.rs
@@ -602,6 +602,41 @@ async fn apply_dependency_changes(
     }
 }
 
+fn apply_none_hint(
+    add: &[String],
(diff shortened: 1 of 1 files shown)


rgbkrk:
@pullfrog review please

PR #3875, opened 2026-07-01
chore(ci): retire runt-mcp plugin snapshot

### LINKED ISSUE #3873: Retire crates/runt-mcp/assets/plugins/ (runtime-dead, pipeline-alive)
`crates/runt-mcp/assets/plugins/` holds a committed snapshot of the renderer plugin bundles (sift.css/js from May 11, pre token-unification; plotly, vega, leaflet, markdown; plus `sift_wasm.wasm`). Runtime serving no longer reads it: the daemon embeds plugins via `include_bytes!` from `apps/notebook/src/renderer-plugins/` (`crates/runtimed/src/embedded_plugins.rs`), and the drift test compares against that dir too.

But the directory is still wired into the build pipeline:

- `.github/workflows/build.yml` and `release-common.yml` include it in the `wasm-artifacts` upload, which ~10 downstream jobs restore before building.
- `build.yml`'s `wasm` path filter watches it.
- `cargo xtask wasm sift` copies `sift_wasm.wasm` into it.
- xtask's raw-control-byte scan excludes it as a known generated bundle.

So it's runtime-dead but pipeline-alive, and the committed CSS/JS content is stale (it will keep drifting since nothing regenerates the css/js copies). Cleaning it up means:

1. Confirm no CI job actually reads the restored path (vs it just riding along in the artifact).
2. Drop it from both workflow artifact lists and the path filter.
3. Remove the `wasm sift` copy step and the xtask scan exclusion (+ its test).
4. Delete the directory.
5. Validate with a full CI run and a release-common dry run.

Found while unifying sift tokens (#3872 noted it as "dead snapshot" - that was right about runtime, wrong about the pipeline).

rgbkrk:
## Summary

- remove `crates/runt-mcp/assets/plugins/` from build/release WASM artifact uploads and the CI path filter
- delete the stale committed renderer plugin snapshot under `crates/runt-mcp/assets/plugins/`
- remove the old xtask help text and raw-control-byte scan exemption for that dead snapshot path

Closes #3873.

## Verification

- `rg -n "crates/runt-mcp/assets/plugins|runt-mcp/assets/plugins|assets/plugins" .github crates apps packages scripts mcpb .gitignore .gitattributes Cargo.toml package.json -g '!target/**'`
- `./actionlint .github/workflows/build.yml .github/workflows/release-common.yml`
- `cargo test -p xtask`
- `git lfs pull --include='apps/notebook/src/renderer-plugins/**'`
- `cargo xtask artifacts ensure renderer`
- `cargo xtask artifacts verify renderer`
- `cargo xtask lint --fix`

## Notes

- `release-common.yml` is a reusable `workflow_call` workflow, so the local release-common check here is static validation via `actionlint`; the full CI/release caller coverage needs GitHub CI.


rgbkrk:
@pullfrog review

PR #3893, opened 2026-07-02
fix(notebook-cloud): bootstrap edit access on reload

quillaid:
## Summary
- seed notebook viewer HTML with app-session catalog access scope and title so logged-in reloads can start with the correct edit affordance
- switch notebook-viewer catalog access refresh from the broad `/api/n?limit=100` list fetch to the direct notebook catalog endpoint carrying `access.scope`
- keep edit mode pending while catalog grants edit access but the live room is still connected as viewer, avoiding the misleading "Request Sent" state

@rgbkrk @pullfrog would appreciate a review of the reload/auth bootstrap shape here.

## Tests
- `node --import tsx --test test/worker-routes.test.ts`
- `node --import tsx --test test/cloud-notebook-catalog-access.test.ts test/cloud-document-edit-readiness.test.ts test/viewer-shared-cell-surface.test.ts test/notebook-room.test.ts`
- `cargo xtask lint --fix`

### DIFF
--- a/apps/notebook-cloud/src/authorization.ts
+++ b/apps/notebook-cloud/src/authorization.ts
@@ -10,6 +10,7 @@ import {
   getNotebookRow,
   getPublicNotebookAclRows,
   type NotebookAclRow,
+  type NotebookRow,
 } from "./storage.ts";
 
 const CAP_READ = 1 << 0;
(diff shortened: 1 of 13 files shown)


quillaid:
Followed up on the Claude Fable notes in `23e49b62`:

- widened the `allowLiveScopeDowngrade` option comment to explicitly cover read-only effective-scope discovery routes, not just live-room tickets
- added `authorizeNotebookAccessWithNotebook()` so the viewer shell bootstrap reuses the notebook row already loaded during authorization instead of calling `getNotebookCatalog()` and loading revisions/blobs on the HTML path
- replaced the unchecked `authorized.scope as Exclude<..., "runtime_peer">` with a `browserCatalogAccessScope()` guard

On the stale-ref note: I did not add a scope-resolution-triggered reconnect. For `?mode=edit`, unresolved catalog facts still request `owner` through `cloudNotebookSyncScopeForCatalogAccess()`, and room auth downgrades that to the user's real scope. If the user starts in view mode, there is no need to upgrade the live room until they explicitly request edit; that path already goes through the edit-mode transition/retry flow. So I think the soft-stuck viewer connection is avoided without another reconnect effect.

Rechecked:
- `node --import tsx --test test/worker-routes.test.ts test/cloud-notebook-catalog-access.test.ts`
- `cargo xtask lint --fix`

PR #3902, opened 2026-07-03
feat(mcp): first-class comment tools + CommentsDoc replica in typed clients

quillaid:
## Summary

Implements the top item from `docs/plans/comments-rollout.md`: first-class MCP
comment tools so agents participate in notebook review through the same
authored-change model as humans.

- **Four mutation tools** — `create_comment` (cell or notebook anchor),
  `reply_comment`, `resolve_comment`, `reopen_comment` — registered in
  `runt-mcp`, following the existing cell-tool patterns.
- **Read surface** — `nteract://notebooks/{id}/comments` resource returning the
  comments projection (settles pending sync frames before reading so a read
  right after join does not race the daemon's initial CommentsDocSync).
- **CommentsDoc replica in typed clients** — `notebook-sync` now holds a
  comments replica in `SharedDocState` and syncs `0x0a` frames, mirroring the
  CommsDoc pattern. The daemon side needed no changes — ingress
  (`peer_comments_sync.rs`) and persistence (`comments_store.rs`) already
  existed; this fills the client half we documented as missing in #3899.
- **Contract test gap closed** — `CommentsDocSync` added to the
  frame-size coverage in `notebook-protocol`, making the typed-frame v4 ADR's
  coverage claim true.
- Docs updated in step: `notebook-sync/AGENTS.md`, automerge-sync skill stream
  table, and the rollout plan's MCP bullet retired (desktop polish and publish
  boundary remain open).

## Attribution invariant

The load-bearing constraint from `crates/comments-doc/AGENTS.md`: attribution
comes from admitted Automerge change actors, never stored fields.

- The comments replica authors under the connection's actor label — inherited
  from the notebook doc at construction and kept in step by
  `DocHandle::set_actor` (which now covers both docs, mirroring the WASM
  handle).
- The replica scaffolds from the shared CommentsDoc genesis
  (`schema_seed_doc`), so a mutation racing ahead of the first daemon sync
  frame cannot fork a conflicting ROOT `threads` map.
- Comment mutations notify the sync task (`changed_tx`) and flush through the
  same outbound path as other docs; tools confirm via `confirm_state_sync`
  (which flushes comments frames) rather than the NotebookDoc-heads-only
  `confirm_sync`.
- A client-side comments failure degrades alone (reset comments peer state,
  keep the connection), matching the daemon's explicit policy.

## Review provenance

Implemented via staged workflow (design → notebook-sync package → runt-mcp
package), then reviewed by two independent passes. The deep review pass caught
three must-fix bugs the local CRUD tests could not see — mutations never
flushed to the wire, the replica authoring under a random ActorId (admitted
anonymously on local daemons, rejected on hosted), and a genesis-fork race —
all fixed and re-verified, plus the schema/description mismatch on the anchor
parameter and connection-teardown-on-comments-error. Final state: 1,503 tests
pass across `comments-doc`, `notebook-sync`, `notebook-protocol`, `runt-mcp`,
`runtimed`; `CI=true cargo xtask lint --fix` clean.

## Follow-ups (tracked in rollout plan)

- Desktop product polish (rail/panel flows, stale anchors).
- Publish boundary policy (exclude private comments from public artifacts).
- Icon polish: comment tools currently reuse the EditCell icon.

### DIFF
--- a/.agents/skills/automerge-sync/SKILL.md
+++ b/.agents/skills/automerge-sync/SKILL.md
@@ -122,16 +122,18 @@ V1 is original; V2 allows compressed document encoding. Backward-compatible via 
 | Notebook | `0x00` AutomergeSync | `SharedDocState.doc` | Bidirectional |
 | RuntimeState | `0x05` RuntimeStateSync | `SharedDocState.state_doc` | Daemon-authoritative |
 | CommsDoc | `0x09` CommsDocSync | `SharedDocState.comms_doc` | Widget state, gated by RuntimeStateDoc topology |
-| CommentsDoc | `0x0a` CommentsDocSync | Frontend WASM replica (`runtimed-wasm`); daemon replica persisted by `comments_store.rs` | Notebook-room comments sidecar; ingress validates change actor labels against the connection principal |
+| CommentsDoc | `0x0a` CommentsDocSync | `SharedDocState.comments_doc` (typed clients + frontend WASM); daemon replica persisted by `comments_store.rs` | Notebook-room comments sidecar; ingress validates change actor labels against the connection principal |
 | PoolState | `0x06` PoolStateSync | PoolDoc | Frontend owns sync state; daemon carries `pool_peer_state` separately |
 
 For CommentsDoc, see `crates/comments-doc`, daemon persistence at
 `crates/runtimed/src/notebook_sync_server/comments_store.rs`, ingress at
-`peer_comments_sync.rs`. Optimistic client mutations apply via Automerge; the
-daemon validates change actor labels against the connection principal
-(clone-preview) and strips writes from scopes without comment authority. There
-is no daemon finalization step — attribution (`resolved_by_actor_label`,
-`resolved_at`) is projected from admitted change actors.
+`peer_comments_sync.rs`. Typed clients (`notebook-sync` crate) and frontend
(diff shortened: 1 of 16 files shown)


quillaid:
Local Codex review flagged a real P1: the typed client's CommentsDoc replica scaffolds from the empty sync seed with no cached `comments_doc_id`, and raw sync receive never populated it — so `read_projection()` failed its identity guard (`MissingCommentsDocId`) even after the daemon synced the identity into the document. The existing CRUD test masked it by swapping in an already-identified doc.

Fixed in 79c78b74:
- `CommentsDoc::adopt_synced_identity()` — adopts the document's synced `comments_doc_id` into the cached identity, validating format and rejecting conflicted values through the existing `ensure_raw_comments_doc_id_matches_value` guard.
- Client receive path (`SharedDocState::receive_comments_sync_message`) adopts after each apply; adoption failure logs and degrades alone rather than failing sync, matching the comments degrade-alone policy.
- Integration-style test `comments_projection_works_after_initial_daemon_sync`: starts from the real `SharedDocState::new()` empty seed, drives full bidirectional sync with a daemon-side identified doc, and asserts the projection is readable with the daemon's id — exactly the path the `nteract://notebooks/{id}/comments` resource and post-mutation reads exercise.

All targeted suites pass (394 tests across comments-doc, notebook-sync, notebook-protocol, runt-mcp); lint clean.

PR #3905, opened 2026-07-03
fix(workstations): abort stale refresh on sign-out, stack layout on narrow viewports

rgbkrk:
Two follow-up fixes to the workstations management surface from #3904.

**Stale refresh after sign-out.** The background refresh effect created an `AbortController` per interval tick but cleanup only cleared the timer, so an in-flight fetch keyed to the previous session could resolve after `canFetchCatalog` flipped false and overwrite the `signed_out` state with the old workstation list. One controller now covers the effect's lifetime and cleanup aborts it, matching the poll pattern already used in `notebook-viewer.tsx`. `loadWorkstations` already guards on `signal.aborted` before writing state.

**Narrow-viewport clipping.** `WorkstationsManagementPage` rendered a non-wrapping flex row with a fixed 392px sidebar inside `overflow-hidden`, so the live `/workstations` route lost the detail pane on narrow windows (the Elements fixture masked this by forcing a 1040px min-width). Below `lg` the panes now stack: the machine list caps at `max-h-72` and scrolls, the detail pane takes the rest. Desktop layout is unchanged.

### DIFF
--- a/apps/notebook-cloud/viewer/workstations-view.tsx
+++ b/apps/notebook-cloud/viewer/workstations-view.tsx
@@ -90,17 +90,23 @@ export function CloudWorkstationsView({ authConfig }: { authConfig: CloudViewerA
   }, [authState, canFetchCatalog, loadWorkstations, refreshIndex, waitingForAppSession]);
 
   // Background refresh keeps status spines honest while the page stays open.
+  // One controller covers every tick so cleanup aborts any in-flight refresh;
+  // otherwise a slow fetch from a signed-out session could land late and
+  // overwrite the signed_out state with stale registry data.
   useEffect(() => {
     if (!canFetchCatalog || viewState.kind !== "ready") {
       return;
     }
+    const controller = new AbortController();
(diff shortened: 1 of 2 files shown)


quillaid (commented):
Reviewed the follow-up patch. No blocking issues found.

The abort change now invalidates the interval refresh on cleanup, so sign-out/auth eligibility changes should no longer let an old `/api/workstations` response write over `signed_out`. The narrow layout change looks scoped: below `lg` the list stacks with a bounded scroll area and the detail pane keeps `min-h-0`, while desktop layout remains unchanged.

Verification:
- `pnpm --dir apps/notebook-cloud exec tsc -p tsconfig.json --noEmit --pretty false`

Small non-blocking note: the standalone page still uses `setInterval`, so very slow same-session refreshes can overlap; if that ever causes status regressions, the serialized timeout pattern in `use-cloud-workstations.ts` would be the tighter shape. Not a blocker for this stale sign-out fix.

PR #3908, opened 2026-07-05
docs(agents): route RxJS work through frontend guidance

rgbkrk:
## Summary

- route RxJS/shared-store work from root agent instructions to the `frontend-dev` skill and `frontend-sync-bridge` ADR
- refine the frontend-dev skill path examples for `ObservableStore`, `poll`, cloud store hooks, and browser signals
- augment the cloud viewer AGENTS file with an explicit read-before-edit trigger while keeping main's Decision 8 store doctrine
- add the shared-store projection memo to the docs front door for notebook shell/frontend convergence work

## Verification

- `ruby -e 'require "yaml"; YAML.load_file(".agents/skills/frontend-dev/SKILL.md"); puts "frontend-dev frontmatter ok"'`
- `git diff --check origin/main...HEAD`
- `cargo xtask lint --fix`


rgbkrk:
Rebase map for the doctrine that landed after this PR was opened (#3909, #3910, #3912 - the skill, cloud AGENTS, and review rubric all moved):

**Keep - unique value this PR still adds:**
- The root `AGENTS.md` routing section. Nothing at root routes RxJS/store work today; path-scoped rules only fire once files are touched. Suggest trimming it to the routing pointer (skill + `frontend-sync-bridge.md` Decision 8) and dropping the invariant-restatement paragraph - the invariants now live in the skill's "Module-Singleton Source Stores" section and Decision 8, and a second wording at root is drift surface loaded into every session.
- The skill trigger-list additions for `packages/runtimed/src/poll.ts`, `viewer/use-cloud-*-store.ts`, and `viewer/browser-signals.ts`. Main's list already has `apps/notebook-cloud/viewer/*store*.ts` (rebase around it); the three extra entries are real coverage.
- The broadened after-`await` line (auth/endpoint/activation generation) is compatible with main's wording; fine to keep.

**Drop - superseded on main:**
- The `apps/notebook-cloud/AGENTS.md` "Viewer reactive stores" section. Main now has "### Viewer async state" covering the same ground (domain hooks, activate/dispose, capture-at-issue/drop-at-apply with the activation epoch, do-not-converge placement) plus the comparator-manifest and typecheck-lane notes. Nothing in this PR's version is absent from main's; two wordings of one invariant is the drift the one-depth-home rule exists to prevent.

**Update - the audit doc:** `docs/audits/rxjs-agent-guidance-audit.md` describes the routing gaps as open. Most are now closed (#3909 encoded the doctrine + added `src/components/notebook/AGENTS.md` and the `notebook-state` rule; #3912 added the speculative-claim review lens). Landing the audit as a record is consistent with the docs taxonomy if each finding carries its resolution status; as-is it reads as open gaps that are not.

Context for the wording above: `docs/adr/frontend-sync-bridge.md` Decision 8 is the decision record; the frontend-dev skill holds the how-to depth; AGENTS.md files carry terse per-path invariants with cross-references, deliberately not copies.

rgbkrk (commented):
Rebase looks right - audit dropped, cloud AGENTS deduped to a pointer, trigger list merged around main's entries, and the docs front-door line adding the convergence memo is a real improvement. Two items before merge, one that matters:

1. **Terminology: "activation generation" should be "activation epoch"** (root `AGENTS.md` new section, last sentence; and the skill's broadened after-`await` line). The mechanism is named `activationEpoch` in `cloud-workstations-store.ts` / `cloud-access-request-store.ts`, and Decision 8 plus the skill's "Module-Singleton Source Stores" section both say epoch. A second name for the same mechanism in routing-level guidance is exactly the drift this PR is trying to prevent - an agent grepping "generation" finds nothing.

2. **Suggestion, not blocking:** the root `AGENTS.md` section's second paragraph restates invariants that live in the skill and Decision 8. The routing pointer (first paragraph) is the durable part; the restatement is a third wording that can drift. One sentence - "The invariants live in the skill's Module-Singleton Source Stores section and Decision 8; consult those, not this summary." - would outlast it. Fine to keep if you prefer the summary at root; item 1 applies either way.

With item 1 fixed this is good to land.

PR #3912, opened 2026-07-05
docs: store refactor measurements + speculative-claim review lens

rgbkrk:
Docs-only closeout for the store refactor arc (#3906, #3909, #3910, #3911): the before/after measurements memo and a review-rubric extension.

## Measurements memo

`docs/measurements/store-refactor-bundle-and-timers.md` - both sides built at their real commits (worktree at `4b4ed8420` vs main), no estimates. The honest story:

- Scope isolation is real: the workstations surface left every route's payload, and what moved shrank from a 70 kB gzip always-loaded chunk to 14 kB gzip route-local.
- The raw byte win on `/n` is small (-2 kB gzip) because vite's rechunking hoisted a 70 kB gzip lucide-icons chunk into the entry closure, and cold `/workstations` is +13 kB gzip vs the old monolith. The icons chunk is named as the next optimization lever, investigation before config.
- Timer census: 7 per-mount auth timer/listener installs, hand-duplicated in source across 4 views, collapse to one per-page driver set over 3 shared app-lifetime listeners; 3 hand-rolled poll loops become 3 `createPoll` sites; store test coverage goes 0 to 69, over surface where two of three poll loops previously had zero tests.
- Browser-side profiling (INP, render counts, reconnect behavior) is explicitly listed as not-measured-here with the candidate harness.

## Rubric extension

The review rubric's comment lens named history narration but not its mirror: speculative future-consumer claims. A month-old module header claiming the MCP surface consumes the TS store streams (it is Rust by design) survived every doctrine pass this week because no rule named speculation and cross-subsystem claims cannot be falsified from inside the module. The rubric now flags both, and carried-forward prose in a touched header is explicitly in scope.

### DIFF
--- a/.agents/reviewers/nteract-code-review-rubric.md
+++ b/.agents/reviewers/nteract-code-review-rubric.md
@@ -59,6 +59,13 @@ Prefer at most 12 findings.
   equivalent guards.
 - Tests: require focused tests at the changed boundary. Treat deleted tests as
   suspicious unless the behavior was removed and replacement coverage exists.
+- Comment and doc claims: module headers and doc comments state what is true
+  now. Flag past-tense history narration AND speculative future-consumer
+  claims - naming hosts, surfaces, or integrations that do not consume the
+  module. Claims about other subsystems deserve extra suspicion because the
+  diff cannot verify them (example: the MCP surface is Rust by design, so TS
+  store docs must never list it as a consumer). Carried-forward prose in a
+  touched header is in scope, not grandfathered.
 
 ## Finding Contract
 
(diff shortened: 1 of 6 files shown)


rgbkrk:
Cross-model review record for the merge:

- Codex adversarial review of the harness half returned BLOCKING with four P2s and two P3s - one theme: the harness could report success or clean metrics when the measured property failed or was never measured (unasserted reconnect stability, TBT defaulting to 0 without an attached observer, INP without a verified interaction, a visibility stimulus that never flipped `visibilityState`).
- All six fixed in 4f07f2b54: reconnect stability is now an assertion driving the process exit code, the visibility cycle overrides `Document.prototype.visibilityState` so `browser-signals.ts`'s real gate exercises both edges, TBT/INP report `null` plus an unavailable reason instead of fabricated values, the hook comment states the true invariant, and probe notebooks carry timestamp prefixes (the worker has no notebook DELETE route - verified against `src/index.ts`).
- Codex re-check: all six FIXED, new-hole hunt clean (override cannot leak into non-reconnect measurements; the exit-1 path is not swallowable), verdict NON-BLOCKING.
- Full gate green on the branch: cloud typecheck, 1273 node tests, viewer+runtimed vitest suites, canonical `build:viewer` unaffected by the profiling flag, harness syntax check.

Merging under the cross-model review + gates protocol.

PR #3914, opened 2026-07-06
feat(local-oidc): dev-only OIDC issuer for local wrangler

rgbkrk:
Adds a dev-only OIDC issuer so local wrangler can run the real sign-in, app-session, and token-renewal paths without a live IdP. This is the enabler for driving the cloud viewer's full auth lifecycle in a browser: before it, local wrangler only had loopback-trust auth, so sign-in, renewal, sign-out gating, and multi-user sharing were untestable locally.

## What lands

- `@nteract/local-oidc` (`packages/local-oidc`): a host-free, web-standard OIDC issuer. `jose` is its only dependency. `private: true` - it is never published. An ephemeral RS256 key is generated per boot, so a restart rotates the key and invalidates every prior token.
- The `/dev/oidc/*` mount (`apps/notebook-cloud/src/dev-oidc.ts`), served only when `NOTEBOOK_CLOUD_LOCAL_OIDC === "true"`. Production never sets the flag and exposes nothing new.
- `dev.mjs` wires the issuer vars, binds loopback only, and seeds a per-process random app-session secret.
- `profile-local.mjs` gains a renewal-path check: mint a token, exchange it for an app-session cookie, load a room, rotate the token, and assert the live-room WebSocket does not churn across the rotation.

## Security posture (this is auth code, dev-only)

The issuer mints real tokens, so it is built to be unexploitable by construction rather than by trusting the environment:

- **Loopback only.** The dev worker binds the derived loopback host and offers no `--ip` override; `--ip`/`--ip=` are stripped before args reach Wrangler. There is no way to bind it to a routable interface.
- **One-time bound authorization codes.** Codes are opaque 32-byte random values held in per-isolate memory with a 60s TTL, bound to the authorize request's `redirect_uri`, and consumed on lookup before any other check can fail, so replay and cross-redirect exchange both fail.
- **Real PKCE.** When an authorize request carries a `code_challenge`, the token exchange requires and verifies an S256 `code_verifier`; `plain` is rejected and discovery advertises S256 only. The no-PKCE flow stays for local harness scripts that do not run a browser.
- **Fixed mount.** `/dev/oidc` is gated on the exact path before the issuer sees a request, so a misconfigured `NOTEBOOK_CLOUD_OIDC_ISSUER` can never shadow `/api/auth/*`. Mount-relative paths are translated onto the issuer URL's path, which is the identity mapping under the default config.
- **Per-process secret.** The dev app-session secret is random per process; a restart invalidates dev cookies and forces a fresh sign-in.

## Design decision: why the issuer URL is a var, not the request origin

The token `iss` claim must equal what the worker's OIDC verifier and the viewer's auth config point at. Wrangler dev serves the worker under its configured route host, so the raw request origin is not the loopback origin the browser and verifier use. The mount reads `NOTEBOOK_CLOUD_OIDC_ISSUER` for the claim and falls back to request-origin + mount path only when unconfigured. Requests are still accepted only at the fixed `/dev/oidc` mount regardless of that value.

## Sign-in wiring

With the issuer mounted, OIDC becomes the primary sign-in across every entry point (header button, home view, notebook-view re-auth), so a real browser sign-in drives the local issuer's authorization_code + PKCE flow. The three entry points resolve through one `cloudSignInMethodForConfig` helper rather than duplicated priority logic. The loopback dev-token path stays reachable at `/local-auth` as a fallback and remains the only sign-in when the flag is off. Access and refresh tokens carry a `token_use` claim that the refresh grant, the issuer userinfo, and the resource-server verifier enforce, so the two roles cannot be swapped.

Predecessor context: the store refactor this measures lands in #3906, #3909, #3910; the profiling harness in #3912.


rgbkrk:
Flipping this to you rather than self-merging: it's the auth surface, and the goal doc says summon here regardless.

## How it was built and reviewed

Implemented by codex (GPT 5.5) under direction, then reworked by Fable to the quality bar. Reviewed by a three-family panel (Opus, Sonnet, codex) on distinct lenses (correctness/spec-fidelity, auth-abuse/edge-cases, adversarial security), with an adversarial refute pass on every finding before any of it was acted on.

## What the panel found, and what changed

- **`token_use` never enforced** (Sonnet + codex, independently). The issuer stamped a `token_use` claim but checked it nowhere, so a refresh token worked as a bearer access token and an access token could be POSTed to mint a fresh pair. Now stamped `access`/`refresh` and enforced at the refresh grant, the issuer userinfo, and the resource-server verifier (`identity.ts`). The verifier only rejects a token that explicitly declares itself a refresh token, so real-IdP access tokens (which omit `token_use` or set `access`) are unaffected. Negative tests both directions.
- **Sign-in never reached OIDC** (codex, confirmed). Every sign-in button chose the loopback dev-token path when both configs resolved, so the real authorization_code + PKCE flow was unreachable. Fixed, and the source-text convergence test surfaced a third copy of the localDev-first logic in `notebook-viewer.tsx` that the panel had not flagged. All three entry points now route through one `cloudSignInMethodForConfig` helper.
- **Renewal harness passed vacuously** (Opus). A run where no live-room socket opened (0 to 0 sockets reads as stable) or the app-session exchange failed still reported green. Now requires a live socket and a real app-session before the stability result counts.
- **Refuted:** codex's P3 auth-code map "leak" - pruning runs before insert, so the map self-bounds, and the retry loop it cited does not exist in the code.

The two P2s from the original hardening pass (forgeable codes, hardcoded secret) plus the `--ip` off-box bind were already closed in the earlier commits: opaque one-time PKCE-verified codes bound to `client_id`/`redirect_uri`, per-process random secret, loopback-only bind with no override.

## Re-review

codex re-reviewed the fixes (written by a different model than it) with focus on the production `identity.ts` change: no defects, real-IdP tokens confirmed unaffected, no false negatives in the harness gate, no new holes.

## Gate

1281 notebook-cloud tests, 27 issuer, 54 identity, three typechecks (notebook-cloud, local-oidc, runtimed), viewer build, lint - all green on the head commit.

The package is `private: true`, gated behind `NOTEBOOK_CLOUD_LOCAL_OIDC`, loopback-only. Not for public releases.

Issue #3947, opened 2026-07-07
Failed kernel launch renders as a healthy running kernel (stuck queued, no error)


rgbkrk:
## Summary

A failed kernel launch renders as a healthy running kernel. When `LaunchKernel` fails (notably launch-on-attach), the runtime peer stays connected, the notebook UI reports "Kernel running / running", and queued cells sit at "queued" forever with no surfaced error. This is the "everything queued, nothing running, no logs in the browser" symptom.

## Live repro (preview.runt.run, 2026-07-07)

Notebook `01KWX5Z6ZHMBS4SA7M53FTSRK0` (`yep`) on a fresh nightly workstation agent (`ws-lab1`, `runt 2.6.2-nightly.202607071009`). Started compute from the rail. The rail showed **lab1 Ready · Kernel running · Compute sessions 1 · Can run**, the toolbar showed **Python / running**, and all cells stayed at **queued**. Nothing executed. No error anywhere in the UI.

The failure was only visible on the compute host, in the runtime-peer log:

```
[jupyter-kernel] LaunchKernel: type=python source=uv:current_python
[jupyter-kernel] Starting Python kernel with current python at "/usr/bin/python3"
[kernel-stderr] ModuleNotFoundError: No module named 'ipykernel'
[jupyter-kernel] Kernel process exited before launch completed
[runtime-agent] LaunchKernel failed: ... kind=ProcessExited
[runtime-agent] Launch-on-attach failed: Failed to launch kernel
```

The operational cause (system `/usr/bin/python3` had no `ipykernel`) was fixed on the host with `apt install python3-ipykernel`; after a Restart the same notebook executed and returned `('lab1', 8)`. That is not the bug being filed. The bug is that the launch-on-attach failure never propagated to `RuntimeStateDoc`, so the UI could not show it.

## Root cause direction

`crates/runtimed/src/runtime_agent.rs` logs the launch-on-attach `LaunchKernel` failure at WARN but does not appear to write a failed/errored kernel state into `RuntimeStateDoc`. The runtime peer remains connected (room-link healthy, which #3943 now heartbeats), so the room and UI continue to treat the session as live. Queued execution requests written into `RuntimeStateDoc` have no live kernel to drain them, and there is no terminal signal to move the cells out of "queued".

This is the same class of gap as the control-plane liveness items (a real dead state that no surface reflects), one layer up from transport into kernel lifecycle. See `docs/memos/byoc-control-plane.md` failure-mode 7.

## Expected behavior

[...]

PR #3949, opened 2026-07-07
feat(cloud): workstation liveness lease in the registry DO

rgbkrk:
Consolidates workstation liveness into the owner-scoped registry Durable Object. Each heartbeat renews a lease with an expiry, and a single DO alarm sweeps a workstation offline the moment its lease lapses, then pushes the transition to any listeners. Today offline is only inferred lazily, at read time, from a stale D1 `last_seen_at`, so the rail and toolbar can keep showing a workstation online well after it went quiet. This makes the offline edge proactive and cheap, without standing up a separate ingress/egress service to register and track nodes.

## Design decision: the hibernation boundary is load-bearing

The registry DO (`OwnerComputeIndex`) holds only lease state and one alarm. It never holds a WebSocket. That is what lets it hibernate between a heartbeat write and the next sweep, so an idle fleet costs nothing while it waits. A DO that parks an open connection stays resident and bills wall-clock, which is the failure mode to avoid here.

The socket-holding DO (`WorkstationEvents`) already hibernates: it uses `setWebSocketAutoResponse` for ping/pong, so it answers keepalives without waking. On expiry the registry reaches it by `fetch` to deliver the `went_offline` push. It never grabs a socket of its own. Keeping the lease DO connection-less and the events DO socket-only is the boundary that keeps both hibernatable.

The alarm is single-earliest-wins: it arms for the soonest live lease expiry and rewrites only when that target moves, so N workstations share one alarm, not N.

## What ships (three commits)

1. **Lease store + alarm.** `OwnerComputeIndex` gains `/lease/upsert` and `/lease/list`; the heartbeat POST dual-writes a lease next to the D1 row (best-effort, never blocks registration). `alarm()` sweeps expired leases to offline and re-arms, or disarms when none are live.
2. **Read path consults the lease.** `GET /api/workstations` reads the lease so an alarm-swept lease reports offline proactively. It also checks the expiry timestamp directly, so a late alarm can't report a lapsed lease as live. A live event socket still overrides to online; a missing lease falls back to the old D1 staleness check.
3. **went_offline push + GC.** The sweep `fetch`es `WorkstationEvents /notify` with a new `went_offline` event so listeners learn a drop instead of inferring it from a missed poll. The same sweep garbage-collects leases that have stayed offline past a day, so DO storage doesn't accumulate dead records. D1 stays the registry of record, so a GC'd workstation still lists (offline) via the fallback.

## Behavioral coverage

| Situation | Reported status | Path |
|---|---|---|
| Heartbeat within the window | online | lease `online` + not expired |
| No heartbeat within the window | offline | alarm sweeps + pushes `went_offline` |
| Alarm late, lease already lapsed | offline | read checks `lease_expires_at` directly |
| Live event socket present | online | socket presence overrides the lease |
| No lease (pre-feature / dropped write) | D1 lazy staleness | backward-compatible fallback |
| Offline past the GC window | dropped from DO, still lists offline | alarm GC + D1 fallback |

## Cost

Liveness stays on DO-SQLite row writes. At ~200 nodes on a 20s heartbeat that's roughly 26M writes/month, under the 50M included row-writes, so ~$0 marginal. A Redis control plane would add a managed dependency plus separate ingress/egress and can't hold the persistent `SUBSCRIBE`/keyspace-notification connections a Worker needs anyway. The design rationale and the options considered are in the memo (#3946).

## Scope

Explicit user-initiated deregistration (a D1 cascade-delete plus a remove-workstation control) is a separate follow-up. It's registration lifecycle, not liveness, and the motivating never-connected workstation has no lease to sweep. The rail "Running" badge convergence (the room host reflecting `kernel.lifecycle` into attachment status) also stays out; that's the read-side remainder tracked on #3947.

Tests: DO lease upsert/list, alarm arm/disarm, sweep-to-offline, GC, and the `went_offline` push (fake `WorkstationEvents` namespace); the read-path derivation (live/swept/lapsed/fallback/socket-override); and `WorkstationEvents` accepting the new event. Full notebook-cloud node suite and the cloud viewer vitest suite are green.

Design doc: #3946. Part of the workstation reliability work alongside #3948 (which resolved the stuck-queue on launch failure); tracked with #3947.

### DIFF
--- a/apps/notebook-cloud/src/compute-session-index.ts
+++ b/apps/notebook-cloud/src/compute-session-index.ts
@@ -12,7 +12,10 @@ const MAX_NOTEBOOK_IDS_PER_LIST = 500;
 // alarm drops it so DO storage does not accumulate stale records; D1 remains the
 // registry of record, so the workstation still lists (offline) until it is
 // deregistered.
-const WORKSTATION_LEASE_GC_MS = 24 * 60 * 60_000;
+export const WORKSTATION_LEASE_GC_MS = 24 * 60 * 60_000;
+// Hard ceiling on a single went_offline delivery so a slow or dead events DO
+// cannot keep the background notify (and thus the registry DO) alive.
+const WORKSTATION_WENT_OFFLINE_NOTIFY_TIMEOUT_MS = 5_000;
 
 /**
  * A workstation liveness lease held in the owner-scoped registry DO. The
@@ -116,8 +119,11 @@ export class OwnerComputeIndex {
   }
 
   /**
-   * Single-earliest-wins: schedule the alarm for the soonest online lease
-   * expiry, or clear it when no lease is live. Cheap to call on every upsert -
+   * Single-earliest-wins: schedule the alarm for the soonest deadline that
+   * needs the sweep - an online lease's expiry, or an offline lease's GC
+   * deadline - or clear it when nothing is pending. Arming the GC deadline
+   * matters: without it, an all-offline fleet disarms the alarm and its dead
+   * lease rows would never be collected. Cheap to call on every upsert:
    * `setAlarm` is one row write, and we only rewrite when the target moves.
    */
   private async armLeaseAlarm(): Promise<void> {
@@ -127,8 +133,11 @@ export class OwnerComputeIndex {
     const leases = await this.readLeases();
     let earliest: number | null = null;
     for (const lease of leases.values()) {
-      if (lease.online && (earliest === null || lease.lease_expires_at < earliest)) {
-        earliest = lease.lease_expires_at;
+      const deadline = lease.online
+        ? lease.lease_expires_at
+        : lease.lease_expires_at + WORKSTATION_LEASE_GC_MS;
+      if (earliest === null || deadline < earliest) {
+        earliest = deadline;
       }
     }
     const current = (await this.state.storage.getAlarm?.()) ?? null;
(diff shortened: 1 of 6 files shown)


rgbkrk:
Cross-model review (Codex) surfaced three real issues, all fixed in `fix(cloud): address review findings on the workstation lease`:

- **Stale offline lease could override a fresher D1 row.** The heartbeat writes D1 first and the paired lease write is best-effort, so a lagged/failed lease write left the read reporting a live workstation as offline. The read now trusts the lease only when its `last_seen_at` is at least as new as the D1 row, else falls back to D1.
- **The alarm handler awaited `went_offline` delivery.** A slow or dead events DO kept the registry DO resident, breaking the hibernation invariant. Delivery now dispatches via `waitUntil` with a bounded `AbortSignal.timeout`, so the handler returns promptly.
- **GC deadline was never armed once the whole fleet went offline.** `armLeaseAlarm` only considered live leases, so an all-offline fleet disarmed the alarm and dead rows were never collected. It now arms for the earliest of live-lease expiry and offline-lease GC deadline.

The test harness now models Cloudflare consuming the pending alarm on delivery (`getAlarm` reads null inside the handler) and captures `waitUntil` dispatches so the push is observable.

Deployed to preview.runt.run at `455496518` for a smoke check: `/api/health` reports that build SHA, `/api/workstations` is up and auth-gated (401 unauthenticated), all three DO bindings resolve. Full notebook-cloud node suite (1353) and cloud viewer vitest (62) green; `cargo xtask lint --fix` clean.

PR #3951, opened 2026-07-07
feat(local-oidc): multiple dev users selectable by login_hint

rgbkrk:
Unblocks local multi-user QA. Driving the cloud viewer against local wrangler, the entire multi-user matrix (cross-user sign-in, request/grant edit access, public/private visibility, presence) was unreachable: the dev OIDC issuer mounted a single `dev@localhost` identity and the viewer never sent a `login_hint`, so there was no way to reach a second user. The `feat/local-oidc-issuer` goal doc anticipated exactly this gap.

## Change

- `dev-oidc.ts` mounts three dev identities: `dev@localhost` (default), `alice@localhost`, `bob@localhost`. The issuer already supported a `users` list selected by `login_hint`; it was only ever handed one user.
- `beginOidcLogin` forwards a `login_hint` query param to the authorize request, sourced from the current URL, **only for the local dev issuer** (detected by its fixed `/dev/oidc` mount in the issuer URL). Sign in as a non-default user from `/n?login_hint=alice@localhost`.

## Dev-only by construction

The extra identities exist only behind the `NOTEBOOK_CLOUD_LOCAL_OIDC` gate, and the hint is forwarded only when the configured issuer is the local `/dev/oidc` mount. Production sign-in is provably unchanged: a test asserts the prod issuer drops a `login_hint` present in the URL. `login_hint` is a standard OIDC parameter regardless, so the shape is conventional even where it is sent.

## Tests

- Viewer: `beginOidcLogin` forwards the URL `login_hint` for the dev issuer and drops it for the prod issuer; `buildOidcAuthorizationUrl` sets the param only when given one.
- Mount: authorize + token through `handleLocalOidcRequest` selects `alice@localhost` and `bob@localhost` by `login_hint`, and defaults to `dev@localhost` with none. Issuer-level `selectUser` is already covered in `packages/local-oidc`.

Part of the cloud-viewer QA arc (`.context/goal-cloud-viewer-qa.md`); this is the enabler for the multi-user rows in `.context/cloud-qa-findings.md`.

### DIFF
--- a/apps/notebook-cloud/src/dev-oidc.ts
+++ b/apps/notebook-cloud/src/dev-oidc.ts
@@ -29,11 +29,16 @@ export const NOTEBOOK_CLOUD_LOCAL_OIDC_MOUNT_PATH = "/dev/oidc";
 export const NOTEBOOK_CLOUD_LOCAL_OIDC_CLIENT_ID = "local-oidc-client";
 export const NOTEBOOK_CLOUD_LOCAL_OIDC_DEFAULT_TTL_SECONDS = 300;
 
-const DEV_USER = {
-  email: "dev@localhost",
-  givenName: "Local",
-  familyName: "Developer",
-} as const;
+// Multiple dev identities so local multi-user flows (sharing, presence,
+// cross-user gating) are drivable against one issuer. The first is the default
+// when the authorize request carries no login_hint; reach the others with
+// `?login_hint=<email>` on the sign-in URL (the viewer forwards it only for this
+// local issuer).
+const DEV_USERS = [
+  { email: "dev@localhost", givenName: "Local", familyName: "Developer" },
+  { email: "alice@localhost", givenName: "Alice", familyName: "Example" },
+  { email: "bob@localhost", givenName: "Bob", familyName: "Example" },
+] as const;
 
 // One issuer instance per resolved issuer URL, held for the worker's boot. Every
 // restart starts with an empty cache and a fresh signing key.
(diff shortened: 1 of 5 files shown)


rgbkrk:
Cross-model review (Codex) caught one real issue and it's fixed.

The first cut gated `login_hint` forwarding on `config.issuer.includes("/dev/oidc")` — a substring match. A future production issuer whose URL contained that path (e.g. `.../tenants/dev/oidc/...`) would have forwarded an attacker-supplied URL `login_hint` to the real IdP, breaking the "production sign-in unchanged" claim. Not active with today's Anaconda config (`/api/auth`), but the invariant would depend on config never drifting into that substring.

Fixed in `fix(local-oidc): gate login_hint on a server flag, not the issuer string`: the worker sets `localOidc` on the viewer OIDC config only inside `oidcAuthConfigForRequest` under the `NOTEBOOK_CLOUD_LOCAL_OIDC` gate, and `beginOidcLogin` forwards the hint only when that server-derived flag is set. Dev-ness is no longer inferred from the issuer string at all. A regression test asserts a production issuer whose URL contains `/dev/oidc` as a path substring still drops the hint.

Re-review confirmed clean: the substring vector is closed, the server-flag derivation is correct (a prod deploy without `NOTEBOOK_CLOUD_LOCAL_OIDC` never emits `localOidc`; normalization promotes only boolean `true`/string `"true"`), and no new issue. 29 notebook-cloud OIDC tests pass, typecheck + lint clean.

Auth-adjacent, so bringing it for review rather than self-merging. The change is dev-only by construction (extra identities and hint-forwarding both live behind the local-OIDC gate). Browser E2E of the alice flow is pending a clean re-verify — a wrangler restart rotated the dev session secret and issuer key, so a stale localStorage token drove a failing refresh loop that blocked fresh sign-in until storage was cleared; the mount integration test already proves `login_hint` selects alice end to end through the worker path.

PR #3960, opened 2026-07-07
feat(settings): add editor font and line number settings

### LINKED ISSUE #3950: Notebook Fonts Customization [Feature Request]
I'd like to customize the fonts the editor uses in the notebook itself.
I'd like to set different fonts for the code blocks and for the "comments"/however you'd call what is outside the static code blocks and their output.
The following devtools console snippet sets the font of the code block as I'd like, however it's not persistent.
```
const fontName = 'Hack';
const style = document.createElement('style');
style.innerHTML = `
    .cm-content, .cm-line, .monaco-editor, pre, code { 
        font-family: "${fontName}", monospace !important; 
    }
`;
document.head.appendChild(style);
```
---
**System Info**
- App: 2.6.2 (1f0d54c)
- OS: windows Microsoft Windows [Version 10.0.26200.8655] (x86_64)
- Released: 2026-06-29

rgbkrk:
## Summary

- Add synced `editor` settings for code font, Markdown font, and line numbers.
- Surface a Settings > Editor section with font-family inputs and a line-number toggle.
- Project editor settings into a narrow external store so CodeMirror reconfigures fonts and line numbers through compartments without remounting.
- Route host-DOM Markdown/code typography through the shared output font CSS variables.

Closes #3950

## Verification

- `cargo test -p runtimed-client --features ts-bindings settings_doc`
- `cargo test -p runtimed-settings-sync editor_settings`
- `cargo test -p notebook settings`
- `pnpm test:run src/hooks/__tests__/useSyncedSettings.test.tsx src/components/editor/__tests__/codemirror-editor.test.tsx src/components/editor/__tests__/editor-settings-store.test.ts`
- `pnpm --dir apps/notebook exec tsc -p tsconfig.json --pretty false`
- `cargo xtask lint --fix`
- `cargo xtask renderer-plugins`
- `pnpm --dir apps/notebook build`


rgbkrk:
Cross-model review (Claude over the Codex implementation). Verdict: solid - approve with one scoped follow-up and one nit. Verified rather than assumed:

**Checked and clean:**

- **Automerge shape**: the nested `editor/` map is written and read consistently everywhere - `put`/`get`/`put_bool`/`get_bool` all handle dotted keys (the string variants already did; this PR adds the bool paths), `from_json`, `apply_json_changes`, `SettingsDoc::get_all`, and `runtimed-settings-sync::get_all_from_doc` all agree on the nested layout. The doc-header comment documents it.
- **No remount**: line numbers and fonts reconfigure through dedicated CodeMirror compartments; the test proves the `EditorView` instance survives toggling. This is the right pattern.
- **Input hygiene**: `normalizeFontFamily` strips control chars, cuts at `;{}` (no style injection through the CSS var), and caps length. Good.
- **Store shape**: module store + `useSyncExternalStore`, equality-gated emissions, frozen default. Matches house RxJS/store doctrine even without RxJS.
- **CSS var baseline**: the bare `var(--output-mono-font)` reads in `notebook-base.css` and `markdownInlineCodeClassName` are safe - `notebook-tokens.css:56` defines the host baseline, so `removeProperty` on an empty setting falls back to the stylesheet default, not to an inherited proportional font. (`static-highlight`'s inline fallback stack is belt-and-suspenders; fine.) `--output-document-font` has no host baseline, but host prose already read it before this PR - unset-means-inherit is pre-existing behavior, and the empty default preserves it exactly.
- Bindings match ts-rs output; the schema test pins the new fields; settings.rs load fallbacks mirror the existing per-field pattern.

**Follow-up worth filing (scope boundary, not a bug in what's built):** the font settings do not reach isolated output frames. `host-context.ts:134` forwards `--output-mono-font` from the `OUTPUT_MONO_FONT` constant and `--output-document-font` from theme-palette constants (`:216,:232,:247`), so outputs - tracebacks, rich renders, rendered-markdown outputs in frames - keep default fonts while the editor above them changes. Visually that reads as "my font setting half-works." The natural extension is sourcing those two host-context values from the editor-settings store so frames follow (the forwarding machinery already exists). Decision on whether outputs should follow user fonts at all is a product call - either way it deserves an issue so it's chosen, not accidental.

**Nit:** `updateNotebookEditorSettings` is a pure alias of `setNotebookEditorSettings` - one of the two names should go before this pattern spreads.

Trivial observation, no action needed: markdown editors apply the prose font to `.cm-gutters` too, so line-number digits go proportional in markdown cells if gutters are ever shown there.

rgbkrk:
CI addendum to the review above - three failing checks, two with a clear cause:

- `MarkdownText.test.tsx > renders shared inline semantic text styling` and `ProjectedMarkdownView.test.tsx > renders inline code with the document code chip styling` both still pin the old `font-mono` class that `markdownInlineCodeClassName` now spells `font-[var(--output-mono-font)]`. Update the pinned strings.
- Browser E2E (Playwright) and Windows clippy also failed; logs were not yet fetchable while the run was in progress - re-check after the fix push. The E2E one is plausibly the same class-string change reaching a snapshot/selector.

The full JS suite locally (`pnpm test:run src/components/markdown`) should reproduce both quickly.

PR #3972, opened 2026-07-09
feat(cloud): runtime idle TTL and lazy launch-on-execute

rgbkrk:
Implements the runtime idle TTL + lazy launch-on-execute arc: visiting a notebook is never a compute-starting event, idle kernels shut down cleanly after 30 minutes without execution, the attachment stays sticky, and the next Run wakes compute on the same box. Follows the control-plane correctness PR (#3971). The decision record ships as `docs/adr/runtime-idle-ttl-lazy-launch.md` in this PR.

## Decisions

- **Execution-only idleness**: open viewer tabs never hold a kernel alive (WebSocket hibernation keeps tabs cheap). A future force-keep-alive pin per notebook is recorded as ADR future work, not built.
- **TTL = 30 min**, compile-time constant for v1.
- **Lazy launch**: kernels launch on first execution, never on room wake. Eager launch-on-attach survives only for user-initiated attaches (`trigger=user_attach`); resume jobs (`trigger=resume`) spawn with launch-on-execute.
- **Owner-only resume**: hosted execution is owner-only; resume jobs reuse the owner principal. No new ACL surface.
- Fail active attach jobs immediately on lease expiry (registry alarm as the host-independent driver).

## Commits

1. `fix(cloud): ignore stale cross-session workstation-attachment publishes` — the monotonic publish guard deferred from #3971; a raced publish can no longer re-point execution routing at a cancelled job.
2. `fix(cloud): fail active attach jobs on lease expiry` — the host-independent expiry driver deferred from #3971.
3. `fix(cloud): stop idle hosted runtimes and lazily resume execution` — the room-DO idle alarm (third task on the existing single-earliest alarm), execution-activity tracking, teardown via the close path, and the owner-only wake-on-execute resume path.
4. `feat(runtimed): treat runtime idle timeout close as graceful terminal` — a third close category (terminal AND graceful): the host shuts the kernel down cleanly, exits Ok, and the attach job patches `completed` instead of `failed`; the close-reason constant is pinned identical across TS and Rust by test.
5. `feat(runtimed): launch-on-execute trigger for workstation attach jobs` — `trigger` column (guarded ALTER), a one-shot launch template consumed on first execution after `runtime_state_doc_seen`; `user_attach` spawn argv byte-identical to before.
6. `feat(cloud): attached-idle workstation state in rail and Run affordance` — status `idle` renders as "Idle · starts on the next run", Run stays available as the wake affordance (owner connections only), no room-link "lost" chrome for a deliberately hibernated runtime, and stale idle-session peers get a 409 instead of resurrecting reaped compute.
7. `fix(cloud): idle-teardown ordering, close-match anchoring, and review polish` — reconcile frames deliver to peers before sockets close; the idle rewrite preserves `updated_at` and `runtime_session_id` so it can never outrank a later legitimate publish; the graceful-close matcher is anchored to the close-frame format; the lazy-launch template clears on any successful inbound launch RPC; a `user_attach` create that dedups onto an active resume job upgrades the trigger.
8. `docs(adr): runtime idle TTL and lazy launch-on-execute`.
9. `fix(cloud): dedupe active attach jobs before creating the owner unique index at runtime` — from the preview deploy incident (1101s on a DB with pre-#3971 duplicate active jobs); the runtime schema path now runs migration 0008's idempotent dedupe before the index swap.
10. `fix(cloud): peer-gone reconcile never demotes an idle attachment` — from the live preview smoke: a host-initiated socket close racing the idle teardown armed the peer-gone grace alarm, which stamped `error` over the fresh `idle` attachment ("Kernel failed to start" instead of the calm idle state). Guarded in both layers: the WASM peer-gone reconcile skips the attachment rewrite when the current status is `idle`, and the room's grace handler skips + the idle teardown clears any racing watch. Genuine lost-peer cleanup (abort executions, error a live kernel) is preserved.

## Review protocol

Same as #3971: each slice Codex-implemented, adversarially reviewed by an opposing Claude model before the next stacked, then a final gate re-verified the newest fix commit, swept 12 cross-slice invariants with file:line evidence, and ran the full gates (four Rust crates, cloud typecheck + 1392 tests, repo vitest, app build, lint).

The protocol caught two real bugs before they shipped: the s5 review empirically verified checkpoint-ordering and byte-identical argv; the s6 review caught a critical end-to-end break where the idle status flip would have made Run-as-wake reject the frame and demote the attachment to `error` — a notebook dead until manual re-attach.

## Known residuals (filed, not blocking)

- #3973 — when Run hits an `idle` attachment whose workstation is offline, the pre-existing no-peer relabel degrades `idle` to `error`, killing lazy resume until re-attach. Pre-existing shape on main; the fix is a small guard.
- #3974 — an unsuppressed socket-close during idle-teardown delivery can arm the peer-gone grace alarm. Hit live on preview; commit 10 makes the consequence harmless for idle rooms (the spurious alarm fires and skips), so what remains of #3974 is suppressing the pointless arm itself.
- Peer-loss language: "Kernel failed to start" is wrong copy for a running kernel whose peer went away (deploys surface this on every open notebook). Follow-up arc: map peer-gone to a calm Disconnected state with a Retry/Run affordance, keeping genuine launch failures loud.
- The `uses_fresh_port_retry` launch-failure path still skips the error banner (pre-existing #3947 shape), listed in the ADR's future work.

## Deploy note

The `trigger` column lands via the guarded ALTER TABLE pattern in the runtime schema path; no migration-ordering dependency.

The owner-level active-job unique index initially had one: a DB holding duplicate active attach-job groups (the pre-#3971 shape) failed the runtime `CREATE UNIQUE INDEX` with Cloudflare 1101s before migration 0008's cleanup could run — hit live on the preview deploy (see the deploy smoke comment below). Fixed on this branch: the runtime schema path now runs the same idempotent dedupe UPDATE as migration 0008 strictly before the index swap, verified by a test that seeds the duplicate shape and asserts init succeeds, the newest job survives, and the index enforces afterward. Fresh, migrated, and previously-duplicated databases all converge without manual migration ordering.

Binary rollout: ship workstation binaries (nightly, then stable) before the production Worker deploy. New binary + old Worker is fully compatible; old binary + new Worker degrades noisily on idle teardown (reconnect attempts fenced by 409s, jobs `failed` instead of `completed`).

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-agent-core.mjs
@@ -80,6 +80,10 @@ export function buildAttachJobSpawnPlan({
     displayName,
   ];
 
+  if (job.trigger === "resume") {
+    args.splice(args.length - 2, 0, "--launch-mode", "execute");
+  }
+
   if (typeof job.notebook_path === "string" && job.notebook_path.length > 0) {
     args.push("--notebook-path", job.notebook_path);
   }
--- a/apps/notebook-cloud/src/compute-session-index.ts
+++ b/apps/notebook-cloud/src/compute-session-index.ts
@@ -2,7 +2,14 @@ import type { DurableObjectNamespace, DurableObjectState, Env } from "./cloudfla
 import { isNotebookComputeSessionSummary, type NotebookComputeSessionSummary } from "runtimed";
 import { json } from "./http-responses.ts";
 import { cloudLog, errorMessage } from "./observability.ts";
-import { workstationEventsObjectName } from "./workstation-events.ts";
+import {
+  failActiveWorkstationAttachJobsForWorkstation,
+  type WorkstationAttachJobRow,
+} from "./storage.ts";
+import {
+  workstationEventsObjectName,
+  type WorkstationAttachJobNotification,
+} from "./workstation-events.ts";
 
 const COMPUTE_SESSION_KEY_PREFIX = "session:";
 const WORKSTATION_LEASE_KEY_PREFIX = "lease:";
@@ -16,6 +23,9 @@ export const WORKSTATION_LEASE_GC_MS = 24 * 60 * 60_000;
 // Hard ceiling on a single went_offline delivery so a slow or dead events DO
 // cannot keep the background notify (and thus the registry DO) alive.
 const WORKSTATION_WENT_OFFLINE_NOTIFY_TIMEOUT_MS = 5_000;
+const WORKSTATION_LEASE_EXPIRED_ATTACH_JOB_MESSAGE =
+  "workstation lease expired: no heartbeat within the lease window";
+const WORKSTATION_LEASE_RUNTIME_REPAIR_TIMEOUT_MS = 5_000;
 
 /**
  * A workstation liveness lease held in the owner-scoped registry DO. The
@@ -196,6 +206,7 @@ export class OwnerComputeIndex {
   private async sweepExpiredLeases(now: number): Promise<number> {
     const leases = await this.readLeases();
     const wentOffline: WorkstationLeaseRecord[] = [];
+    const failedAttachJobs: WorkstationAttachJobRow[] = [];
     for (const [key, lease] of leases) {
       // Drop leases that have stayed offline well past their window so the DO
       // does not accumulate dead records.
@@ -204,12 +215,14 @@ export class OwnerComputeIndex {
         continue;
       }
       if (lease.online && lease.lease_expires_at <= now) {
+        const failedJobs = await this.failActiveAttachJobsForExpiredLease(lease);
         const offline: WorkstationLeaseRecord = {
           ...lease,
           online: false,
           offline_reason: "lease expired: no heartbeat within the lease window",
         };
         await this.state.storage.put(key, offline);
+        failedAttachJobs.push(...failedJobs);
         wentOffline.push(offline);
         cloudLog("info", "fleet_registry.lease_expired", {
           owner_principal: lease.owner_principal,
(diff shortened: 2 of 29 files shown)


quillaid:
Deploy smoke note from preview:

- Deployed `90cf74cfaa1301cb848bbafd769b211109317891` to `preview.runt.run`; health now serves that SHA and the runtime WASM manifest points at `runtimed_wasm.75e005eccf26efb8.js` / `runtimed_wasm_bg.7b3886bae700b7cf.wasm`.
- Initial workstation routes were failing with Cloudflare 1101. `wrangler tail` showed `initializeCatalogSchema` failing on `CREATE UNIQUE INDEX ... workstation_attach_jobs(notebook_id, owner_principal) WHERE status IN (...)` with `UNIQUE constraint failed: workstation_attach_jobs.notebook_id, workstation_attach_jobs.owner_principal`.
- Preview D1 had `duplicate_groups=1`, `extra_active_rows=1` for active attach jobs. Applying `apps/notebook-cloud/migrations/0008_workstation_attach_active_owner_unique.sql` fixed the data/index state; after that `/api/workstations` returned 200 and the duplicate aggregate was zero.

The migration file has the right cleanup, but the runtime lazy schema path currently creates the new owner-level unique index in `SCHEMA_STATEMENTS` before `runCatalogMigrations()` runs, so an existing DB with duplicate active rows can fail before it reaches the cleanup. Worth hardening that path before merge, unless production deploys always apply this D1 migration out of band before any Worker request can hit the lazy schema initializer.

I also deleted the current owner's preview workstation rows after validation and stopped the local `lab2` workstation-agent so the delete would stick. Final owner-scoped `/api/workstations` check is 200 with `count=0`.

Remaining smoke gap: the hosted toolbar browser smoke is blocked locally because `~/token.preview.json` is expired and the app-session exchange only accepts OIDC identities. API-key health is still OK.

PR #3978, opened 2026-07-09
feat(cloud): peer loss reads as calm Disconnected and resumes on Run

### LINKED ISSUE #3973: Idle attachment relabeled to error when Run hits an offline workstation, killing lazy resume
When a notebook's workstation attachment is `idle` and the owner hits Run while the workstation itself is offline or stale, `ensureRuntimeForHostedExecution` returns false and `reconcileMissingRuntimePeer` (apps/notebook-cloud/src/notebook-room.ts:1907) relabels the attachment `idle` → `error` via `runtime_peer_gone_workstation_attachment` (crates/runtimed-wasm/src/lib.rs:1976, unconditional `status="error"`). Since `error` is not in `runtimeAttachmentCanResumeForExecution`, lazy resume is dead until manual re-attach.

Not a regression: main relabeled every no-peer execute the same way. But the idle-TTL arc made resume the happy path, so the offline case now degrades a recoverable state into a terminal one.

Fix options, either:
- skip the relabel when the current status is `idle` (keep the sticky attachment; surface the offline workstation through the rail's registry status instead), or
- admit `error` to `runtimeAttachmentCanResumeForExecution` so a later Run can still mint a resume job once the box is back.

Found by the final invariant sweep on the idle-TTL PR; invariant 3 ("peer-gone never relabels an idle attachment") over-promised relative to shipped behavior.

rgbkrk:
When a cloud runtime peer drops out (grace-window expiry or missing runtime peer at reconcile), the notebook now reads as calmly disconnected instead of broken. Reconcile writes attachment status `disconnected` with a `compute disconnected: <reason>` message rather than `error`, the viewer renders a quiet "Compute disconnected" notice with retry/run-to-wake guidance instead of the red kernel-failure banner, and the next Run admits hosted execution and creates a resume attach job. Genuine launch failures still surface the loud error banner.

## Design decision

Peer loss and launch failure are different states and now carry different vocabulary. `error` stays reserved for real failures, which keeps the resume gate (`runtimeAttachmentCanResumeForExecution` still excludes `error`) and the monotonic publish guard untouched. The viewer discriminates on the `"runtime peer disconnected:"` details prefix, the same discriminator `clear_stale_runtime_peer_gone_error` already keys on, so no new state field was added. `runtime-doc` docs were sharpened to pin down what `disconnected`, `idle`, and `error` each mean.

## Review protocol

Codex implemented; an opposing Claude model reviewed. All blocking findings from that review were fixed and re-verified: WASM tests, notebook-cloud typecheck and tests, full vp suite, notebook build, and lint are green.

Closes #3973.

### DIFF
--- a/apps/notebook-cloud/src/notebook-room.ts
+++ b/apps/notebook-cloud/src/notebook-room.ts
@@ -1549,6 +1549,11 @@ export class NotebookRoom {
     }
   }
 
+  /**
+   * A disconnected selected runtime session is recoverable room state: the
+   * same runtime peer may rejoin with its preserved workstation/session IDs
+   * and publish a ready attachment. Idle remains terminal execution teardown.
+   */
   private async runtimePeerAuthorityError(
     notebookId: string,
     workstation: RuntimePeerWorkstationMetadata | null,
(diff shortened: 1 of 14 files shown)


rgbkrk:
CI has a failing check; holding this PR in draft. Details in the checks tab — will investigate.

PR #3980, opened 2026-07-09
refactor(runtimed): KernelSession struct and typed launch spec

rgbkrk:
Two behavior-neutral parameter-shape refactors in `crates/runtimed`, one per commit. `apply_initial_launch` threaded six parallel kernel-session slots (`kernel`, `kernel_state`, `seen_execution_ids`, `lifecycle_rx`, `work_rx`, `interrupt_handle`) that the eager launch block and the LaunchKernel RPC arm threaded too; a `KernelSession` struct now owns them and the launch paths take `&mut KernelSession`. `allocate_current_python_runtime` took 8 positional config args; `CurrentPythonLaunchSpec` names them and the `main.rs` caller builds the spec with named fields. Both `#[allow(clippy::too_many_arguments)]` markers from #3972 are gone and `cargo clippy -p runtimed --all-targets -- -D warnings` is clean without them.

## Design decision

`seen_execution_ids` lives inside `KernelSession` rather than staying loop-local. The call-site inventory decided it: every consumer (the RPC dispatcher, sync-driven queueing on RuntimeStateSync, queue release on launch) already threads it together with `kernel` and `kernel_state`, so keeping it outside would preserve exactly the parallel parameter list this PR removes. It does outlive any single kernel process, surviving transport reconnects and kernel death and clearing only in the RestartKernel arm, same as `kernel_state`; the struct doc comment carries that lifetime note so membership doesn't imply it dies with the kernel.

Channel installation goes through one method, `KernelSession::install_kernel_channels`: the `QueueCommandReceivers` pair installs indivisibly and the interrupt handle re-derives from the kernel slot in the same call. The s5 review had to hand-verify that every launch path installed all channels; the compiler enforces it now, since a call site can no longer install a subset.

The checkpoint-ordering invariant (launch unreachable before `runtime_state_doc_seen`) and the one-shot `pending_initial_launch` `take()` semantics are structurally unchanged, and `plan_current_python_allocation` stays positional because it is the pure, unit-tested planner whose tests exercise the argument mapping directly.

### DIFF
--- a/crates/runtimed/src/main.rs
+++ b/crates/runtimed/src/main.rs
@@ -621,14 +621,16 @@ async fn main() -> anyhow::Result<()> {
                         _ => unreachable!("clap validates launch-mode"),
                     };
                     runtimed::workstation::allocate_current_python_runtime(
-                        target,
-                        config.auth,
-                        python_path,
-                        notebook_path,
-                        launch_working_dir,
-                        std::collections::HashMap::new(),
-                        blob_root,
-                        launch_trigger,
+                        runtimed::workstation::CurrentPythonLaunchSpec {
+                            target,
+                            auth: config.auth,
+                            python_path,
+                            notebook_path,
+                            working_dir: launch_working_dir,
+                            env_vars: std::collections::HashMap::new(),
+                            blob_root,
+                            launch_trigger,
+                        },
                     )
                     .await
                 }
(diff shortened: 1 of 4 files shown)


rgbkrk:
CI has a failing check; holding this PR in draft. Details in the checks tab — will investigate.

rgbkrk:
CI has a failing check; holding this PR in draft. Details in the checks tab — will investigate.

PR #3997, opened 2026-07-10
feat(runtimed): make notebook room lifecycle durable

rgbkrk:
## Summary

- give each notebook room an explicit, generation-bearing source lifecycle and a separate availability/capability lifecycle
- prepare immutable Automerge imports outside the live document lock, journal their exact changes, and publish them in bounded batches under a room-owned task lease
- add append-only recovery journals, causal file checkpoints, restart recovery, file-source conflict detection, and explicit reconciliation operations
- preserve acknowledged peer edits through loading, failure, reaping, and daemon restart while keeping UI and MCP mutations gated until the room is interactive
- return `connect_notebook` from a retained, heads-qualified projection and continue local document/runtime convergence in the background
- add MCP activation generations, same-target coalescing, path/UUID canonicalization, structured readiness errors, and required-head execution guards
- migrate desktop dirty-state and save reporting from timestamps to causal exported heads
- extend the raw stdio MCP harness across cold, warm, parallel, stalled-peer, target-switch, degraded-source, and runtime-not-ready scenarios

Related to #3907.

## Delivery state

This remains a draft integration/reference branch and should not be merged as one undifferentiated slice. It preserves the useful room-loader and projection work already developed on this branch, then adds three reviewable boundaries on top:

1. architecture decisions and Automerge fork baseline;
2. durable room recovery and causal file checkpoints;
3. progressive desktop and MCP clients.

The final merge sequence should land those boundaries in dependency order, each rebased onto the latest `origin/main` with its own fault-injection review. If the staged PRs absorb the whole diff, this draft can close as superseded.

## Architecture and behavior

The accepted room lifecycle ADR makes these invariants explicit:

- `.ipynb` remains the user-visible interchange format, the live Automerge document is active session truth, and the recovery journal is crash/restart durability.
- `RoomSourceState` (`Preparing`, `Publishing`, `Ready`, or `Failed`) is independent from `RoomAvailability` (`Attached`, `ProjectionReady`, `Interactive`, or `Degraded`).
- a room is not published in a non-terminal source state until its task lease is owned; cancellation or panic terminalizes the generation.
- source and peer changes are acknowledged or broadcast only after the corresponding journal record is durable.
- a save is successful only after temporary-file flush, atomic replacement, and causal checkpoint commit; stale completions cannot regress exported heads.
- disk/journal divergence preserves both histories and becomes `source_conflict` until the user explicitly reconciles it.
[...]

Issue #3907, opened 2026-07-05
design(mcp): decouple connect_notebook initial projection from full peer convergence


quillaid:
> *This was generated by AI during triage.*

## Current triage (2026-07-10)

- **Verdict:** Keep open as a measured protocol and latency design issue.
- **Current code:** `connect_notebook` can wait on full convergence and shares a long timeout path, but the reported MissingOps case lacks a stable diagnostic id or reproducible trace.
- **Next slice:** Instrument time to authoritative initial projection separately from full peer convergence, capture a reproducible failure, and then define which readiness contract MCP callers actually require.
- **Boundary:** Do not weaken correctness or hide missing operations merely to return earlier.

## Summary

`connect_notebook` currently acts as both:

1. the MCP session attach operation, and
2. the initial notebook read that gives agents stable cell IDs to use in later tool calls.

That one-call agent contract is important and should probably stay. Agents should not need to call `connect_notebook`, then separately discover initial cells before they can operate.

The concern is that the implementation appears to couple that initial read to full MCP-peer Automerge convergence. In at least one diagnostic case, `connect_notebook` spent the full 120s readiness timeout even though this should usually be milliseconds of work.

## Observed symptom

A submitted diagnostic for `nteract_connect_notebook` showed:

- `list_active_notebooks` worked.
- `connect_notebook` hung/timed out.
- The daemon was otherwise healthy.
- Runtime state reached `Ready`.
- Notebook document sync stayed `Syncing`.
- The MCP logs showed an Automerge warning like `InvalidChanges(MissingOps)` / partial load before the session readiness timeout.

This suggests the expensive part was not kernel startup or environment preparation. It was waiting for the MCP peer's notebook Automerge replica to converge enough for `session_ready()`.

## Current coupling

Today the MCP connect path waits for full session readiness before returning the initial cells/context. That effectively makes the initial read depend on the MCP process becoming a fully converged Automerge peer.

The user-facing contract we want is more like:

- resolve/open the notebook room,
- establish the MCP session,
- return initial cell IDs and context in the same tool call,
- continue live sync for future updates/writes.

The open question is whether the initial cell list really needs to be produced by the MCP peer's local Automerge projection, or whether it can come from a daemon-owned room projection/snapshot.

## Constraints to preserve

- `connect_notebook` should remain the one tool call that both establishes the session and gives the agent initial cell IDs.
- Returned cell IDs must be stable real notebook cell IDs, not temporary handles.
[...]

PR #4011, opened 2026-07-13
feat(runtimed): make saves causal and reconciliation explicit

rgbkrk:
## Summary

PR 4 of 6 in the room lifecycle stack (carved from #3997, related to #3907). Second half of the durable room recovery slice, on top of the source lifecycle in the previous PR:

- the watcher/save flow moves onto the causal file checkpoint: temp-file flush, atomic replacement, monotonic save sequences, and typed `Saved`/`AlreadyCurrent`/`Blocked` outcomes; exported heads advance only after replacement, and stale completions cannot regress the checkpoint
- disk/journal divergence preserves both histories and becomes `source_conflict`; `reconcile_notebook_source` provides the explicit save-recovered-as / keep-recovered / archive-and-reload operations
- the watcher skips events whose bytes match the recorded disk baseline. This is load-bearing on Linux: inotify reports reads (IN_ACCESS), so a poller merely reading the `.ipynb` would otherwise re-run the merge every debounce window, churning the journal and resetting the autosave debounce forever (the Linux-lane `test_promote_untitled_starts_autosave` failure, root-caused via a throwaway debug branch). The binding save of an untitled room now seeds that baseline. The underlying non-idempotent metadata apply is #4015
- protocol grows `NotebookSaved` `exported_heads`/`save_sequence`, typed `notebook_save_blocked` and reconciliation results; generated TS request types and the contract test update with the Rust source; runt-mcp adapts mechanically until the progressive-connect slice
- re-gates the legacy apply/parse wrappers to test-only and carries the replacement coverage for the retired stale-persisted-doc deletion tests

## Stack

1. #4009 room-loader foundations and ADRs
2. #4010 durability primitives
3. room source lifecycle and staged imports (base of this PR)
4. causal saves and explicit reconciliation (this PR)
5. #4012 client causal dirty-state
6. #4013 progressive MCP connect

## Validation

- `cargo test -p runtimed --lib`: 1190 passed, 2 ignored (tree at this tip is byte-identical to the previously validated slice tip, including the Linux-lane fix verified green 3x on the debug branch)
- full earlier-slice validation carries: workspace check, CI-scope clippy, protocol/sync/doc suites, TS typecheck + contract tests, app build, `cargo xtask lint --fix`


rgbkrk:
Superseded: content merges to main via #4013. Review discussion here remains the record for the causal-saves slice.

PR #4014, opened 2026-07-14
debug: isolate Linux-only promote-autosave failure (DO NOT MERGE)

rgbkrk:
Throwaway CI-debug branch for the Linux-only `test_promote_untitled_starts_autosave` timeout seen on #4011/#4013. Runs only that test 3x with tracing enabled. Will be closed and deleted once the root cause is captured.


rgbkrk:
Root cause confirmed and fixed on #4011 (d09e8d5e1): Linux inotify IN_ACCESS events re-triggered the watcher merge loop, starving autosave. This debug branch's isolated 3x run is now green on the Linux lane.

PR #4056, opened 2026-07-15
perf(cloud): session diet, GET-first bootstrap with Server-Timing

rgbkrk:
A production trace of one logout, login, and notebook open shows nine /api/auth/session calls costing ~7.2s, because page mounts POST and the POST handler validates upstream on every call (docs/audits/cloud-login-flow-2026-07-15.md). This puts the session flow on a diet with three invariants. A page mount with a live cookie session costs at most one status GET and zero establish POSTs: the auth store adopts the token when the fetched session is fresh, so OIDC token rotation alone never re-validates upstream, and a fresh bootstrap session delivered with the page costs no session fetch at all. The establish POST happens in exactly two places: the OIDC callback exchange, and the store's fallback when the status GET reports no session while the client holds a valid OIDC token. A present-but-expiring session renews through the status GET, which already slides the cookie server-side, instead of re-validating upstream.

The store's establish completion and the OIDC refresh driver's post-await writes now capture the activation epoch at issue and drop their writes when it no longer matches, so work from a torn-down activation cannot mutate the module singleton. The manual notebook-list refresh drops its establish POST and reads status through the store; pending invites still resolve because the cookie-authenticated list fetch runs syncStoredAppSessionProfile itself. The OIDC refresh cadence is unchanged.

routeAppSession responses now carry Server-Timing with stable metric names so session latency explains itself in any future trace: session_read, host_bootstrap, and renew_cookie on GET; auth_validate, profile_sync, and cookie_create on POST; total on every method. Cloudflare advances Date.now() only across I/O, so a near-zero total against a slow trace points at scheduling upstream of the handler, which is where the 2.8s DELETE for a three-line cookie-clear has to live. Wire shapes, cookie semantics, and OIDC token handling are unchanged.

### DIFF
--- a/apps/notebook-cloud/src/index.ts
+++ b/apps/notebook-cloud/src/index.ts
@@ -1046,30 +1046,68 @@ function withNoStore(response: Response): Response {
   return response;
 }
 
+/**
+ * Server-Timing phase timers for the session endpoint. Cloudflare advances
+ * `Date.now()` only across I/O, so each phase reports the awaited upstream or
+ * database time it wrapped; a near-zero `total` against a slow client-side
+ * trace points at scheduling upstream of the handler. Metric names are stable
+ * so traces stay comparable: `session_read`, `host_bootstrap`, `renew_cookie`
+ * (GET), `auth_validate`, `profile_sync`, `cookie_create` (POST), and `total`
+ * on every method.
+ */
+function appSessionServerTiming(): {
+  time<T>(name: string, op: () => Promise<T>): Promise<T>;
+  apply(response: Response): Response;
+} {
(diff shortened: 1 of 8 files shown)


rgbkrk:
Bedrock review (GLM-5) verdict: two nits, no blockers.

1. Test hygiene: second activation after dispose reuses the scheduler instance in one cloud-auth-store test.
2. The fallback-establish failure path logs a warning but does not surface an error view state; worth confirming whether that visibility gap predates this PR before deciding to extend the session view.

Both deferred for the auth-owner read; neither affects the GET-first semantics. Staging validation on preview.runt.run follows this comment.

PR #4064, opened 2026-07-16
fix(runtimed): surface conda warm-env failures and raise the Windows timeout

### LINKED ISSUE #4017: In-App Feedback
how to install Conda to support this nteract 2.0 in Window? look like no way the Checking Conda runtime is working?

---
**System Info**
- App: 2.6.2 (1f0d54c)
- OS: windows Microsoft Windows [Version 10.0.19045.6466] (x86_64)
- Released: 2026-06-29

markmiro:
# Fix conda warm-env: surface failures and raise the Windows timeout

Fixes #4017 — onboarding's "Checking Conda runtime" spins forever on Windows, with no clue why.

## Background: three processes

The daemon keeps a **pool of pre-built conda environments** (default size 1). Onboarding just polls until the pool has `available > 0`.

| Process | Role |
|---|---|
| **Onboarding UI** | React; polls pool status ~1/sec |
| **`runtimed` daemon** | Long-running background process; owns the pool |
| **`warm-env` child** | Short-lived subprocess the daemon spawns to build one env (download ~120 conda packages, then launch Python to validate) |

```mermaid
flowchart TD
    subgraph Desktop["Desktop app (Tauri + React)"]
        UI["Onboarding UI<br/>shows 'Checking Conda runtime'"]
    end
    subgraph Daemon["runtimed daemon"]
        Loop["warming loop<br/>keeps the pool full"]
        Spawn["spawn_warm_env()<br/>the PARENT side"]
        PoolDoc["Pool status<br/>available / error / error_kind ..."]
    end
    subgraph Child["warm-env CHILD process"]
        Work["solve + download + install<br/>conda packages, then launch python"]
    end

    UI -->|"polls pool status ~1/sec"| PoolDoc
    Loop -->|"pool empty → build one"| Spawn
    Spawn -->|"spawns subprocess"| Work
    Work -->|"stdout: JSON progress + result"| Spawn
    Work -.->|"stderr: the REAL error detail"| Spawn
    Spawn -->|"writes outcome"| PoolDoc
    PoolDoc -->|"available > 0 ⇒ unblock UI"| UI

    classDef bug fill:#ffe0e0,stroke:#d33,stroke-width:2px,color:#000
    class Work,Spawn bug
    linkStyle 4 stroke:#d33,stroke-width:2px
```
*Red = where the bug lives: the parent that spawns the child, and the child's stderr channel (dotted).*

## The bug

Two problems compounded into an eternal spinner with zero breadcrumbs:

- **Timeout too short.** The parent gave the child a hard **600s** budget. On Windows the conda download alone can exceed that (slow/proxied networks push it further). On timeout the child is killed, the pool never fills, and the UI polls `available = 0` forever.
- **Error invisible.** The child's stderr was **inherited** by the daemon. But the Windows daemon is detached (no console), so inherited stderr goes to `/dev/null` — the detailed failure reason was written and instantly discarded.

```mermaid
sequenceDiagram
    participant UI as Onboarding UI
    participant D as Daemon (parent)
    participant C as warm-env child

    D->>C: spawn — stderr = INHERIT ❌
    C->>C: download conda packages (slow on Windows)
    Note over C: download takes > 600s
    D->>D: 600s timeout fires ❌ too short
    D-->>C: kill the child
    Note over D: ❌ child stderr was inherited →<br/>went nowhere, daemon detached
    loop every ~1s, forever
        UI->>D: get pool status
        D-->>UI: available = 0
    end
    Note over UI: spins forever
```

## The fix

```mermaid
sequenceDiagram
    participant UI as Onboarding UI
    participant D as Daemon (parent)
    participant C as warm-env child
    participant Buf as Daemon log + 2KB tail buffer

    D->>C: spawn — stderr = PIPED ✅
    C-->>Buf: ✅ every stderr line captured live
    C->>C: download conda packages
    alt finishes within 900s (was 600s)
        C->>D: stdout result → success
        D->>UI: available = 1 ✅ spinner clears
    else exceeds 900s
        D->>D: 900s timeout fires, now configurable
        D-->>C: kill the child
        D->>Buf: read captured stderr tail
        D->>UI: ✅ error = timed out after 900s + captured stderr tail<br/>error_kind = timeout
        Note over UI: shows a real message,<br/>stops spinning, retries in background
    end
```

- **Capture stderr (headline fix).** Switched the child's stderr from *inherit* to *piped*. The parent now reads it line by line into the daemon log and keeps the last ~2KB, which backfills the error message when the child dies without a reason. The failure is now visible in both logs and pool status.
- **Raise + unlock the timeout.** 600s → **900s**, now overridable via `RUNTIMED_WARM_ENV_TIMEOUT_SECS` so a slow network can extend it without a rebuild. This lets a genuinely-slow-but-fine install finish.
- **Classify timeouts.** Timeouts were tagged as the generic `"setup_failed"`; now errors containing "timed out" get `error_kind: "timeout"`, so the frontend can show the correct "slow network, retrying" message.
- **Frontend reacts.** The UI previously treated "not ready" as "still warming." It now checks for a recorded failure and, if present, stops spinning and shows the message while the daemon keeps retrying with backoff.

## Verification

Verified **end-to-end on Windows through the real `runtimed` daemon** — not the old repro that called the child's conda logic directly and bypassed the parent (where all three fixes live).

- **Scenario 1 (forced short timeout):** pool status shows `error_kind: "timeout"` and the real rattler download logs captured in `error` — the stderr Windows used to drop. ✅
- **Scenario 2 (default 900s):** the same slow install completes; `available: 1`. ✅
- Rust unit tests (timeout override, stderr drain) + frontend tests green; `cargo xtask lint --fix` clean.

### DIFF
--- a/apps/notebook/onboarding/App.tsx
+++ b/apps/notebook/onboarding/App.tsx
@@ -253,6 +253,7 @@ export default function App() {
     const envLabel = PYTHON_ENV_LABELS[pythonEnv];
     setSelectedPoolReady(false);
     setPoolWaitTimedOut(false);
+    setErrorMessage(null);
     setSteps((prev) =>
       prev.map((s) =>
         s.id === "tools"
@@ -372,6 +373,7 @@ export default function App() {
     setPythonEnv(selected);
     setSelectedPoolReady(false);
     setPoolWaitTimedOut(false);
+    setErrorMessage(null);
     setSteps((prev) =>
       prev.map((s) =>
         s.id === "tools"
--- a/apps/notebook/onboarding/App.tsx
+++ b/apps/notebook/onboarding/App.tsx
@@ -6,7 +6,13 @@ import { Button } from "@/components/ui/button";
 import { Progress } from "@/components/ui/progress";
 import { cn } from "@/lib/utils";
 import { CondaIcon, DenoIcon, PixiIcon, PythonIcon, UvIcon } from "@/components/environment";
-import { isOnboardingPoolReady, type OnboardingPoolState, type PythonEnv } from "./pool-readiness";
+import {
+  hasOnboardingPoolError,
+  isOnboardingPoolReady,
+  onboardingPoolErrorMessage,
+  type OnboardingPoolState,
+  type PythonEnv,
+} from "./pool-readiness";
 import type { DaemonStatus } from "./types";
 
 type Runtime = "python" | "deno";
@@ -268,10 +274,28 @@ export default function App() {
 
           if (isOnboardingPoolReady(pythonEnv, state)) {
             setSelectedPoolReady(true);
+            setErrorMessage(null);
             setSteps((prev) =>
               prev.map((s) =>
                 s.id === "tools"
                   ? { ...s, label: `${envLabel} runtime ready`, status: "completed" }
+                  : s,
+              ),
+            );
+            return;
+          }
+
+          // A recorded warm-env failure means the daemon hit a real problem
+          // (bad package, import error, timeout, missing runtime). Surface it
+          // instead of spinning on "warming" until the poll cap — the daemon
+          // keeps retrying with backoff, so the user can still continue.
+          if (hasOnboardingPoolError(pythonEnv, state)) {
+            setPoolWaitTimedOut(true);
+            setErrorMessage(onboardingPoolErrorMessage(envLabel, selected));
+            setSteps((prev) =>
+              prev.map((s) =>
+                s.id === "tools"
+                  ? { ...s, label: `${envLabel} runtime setup failed`, status: "failed" }
                   : s,
               ),
             );
--- a/apps/notebook/onboarding/pool-readiness.ts
+++ b/apps/notebook/onboarding/pool-readiness.ts
@@ -4,6 +4,16 @@ export type OnboardingPoolRuntimeState = {
   available?: number;
   warming?: number;
   pool_size?: number;
+  /** Human-readable error message from the daemon (absent if healthy). */
+  error?: string;
+  /** Package that failed to install, when the daemon could identify one. */
+  failed_package?: string;
+  /** Error classification: "timeout" | "invalid_package" | "import_error" | "setup_failed". */
+  error_kind?: string;
+  /** Consecutive warm-env failures (0 if healthy). */
+  consecutive_failures?: number;
+  /** Seconds until the daemon retries warming (0 if imminent or healthy). */
+  retry_in_secs?: number;
 };
 
 export type OnboardingPoolState = Partial<Record<PythonEnv, OnboardingPoolRuntimeState>>;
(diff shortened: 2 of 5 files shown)


apps/notebook/onboarding/App.tsx
```diff
@@ -278,6 +286,23 @@ export default function App() {
             return;
           }
 
+          // A recorded warm-env failure means the daemon hit a real problem
+          // (bad package, import error, timeout, missing runtime). Surface it
+          // instead of spinning on "warming" until the poll cap — the daemon
+          // keeps retrying with backoff, so the user can still continue.
+          if (hasOnboardingPoolError(pythonEnv, state)) {
```
rgbkrk, line 293: [P1] There is one timing mismatch here that keeps the new timeout message from reaching onboarding. The UI stops polling after `WARMING_POOL_POLL_ATTEMPTS = 180` (about three minutes), while the daemon now gives `warm-env` 900 seconds by default. In that default path, we stop watching roughly twelve minutes before the daemon records the timeout. Could we make the 180-second mark a soft threshold that lets the user continue, while leaving the poll alive for the eventual success or failure?

apps/notebook/onboarding/App.tsx
```diff
@@ -278,6 +286,23 @@ export default function App() {
...
+            setPoolWaitTimedOut(true);
```
rgbkrk, line 294: [P1] There is a small but real selection race here. `cancelled` is checked before the awaited `invoke`, but not after it. If a Conda request comes back after the user has switched to UV, this branch can still set `poolWaitTimedOut`, show the Conda error, and make the UV selection eligible to continue before UV is ready. Could we re-check the captured selection or cancellation identity immediately after the await, before applying any of this state?

apps/notebook/onboarding/App.tsx
```diff
@@ -278,6 +286,23 @@ export default function App() {
...
+            setErrorMessage(onboardingPoolErrorMessage(envLabel, selected));
+            setSteps((prev) =>
+              prev.map((s) =>
+                s.id === "tools"
+                  ? { ...s, label: `${envLabel} runtime setup failed`, status: "failed" }
+                  : s,
+              ),
+            );
+            return;
```
rgbkrk, line 303: [P2] Since the daemon deliberately backs off and retries, I do not think the first recorded failure should end observation. `Pool::add` clears the failure once an environment becomes available, but this return leaves onboarding stuck on the old failure forever. Could we let the error unlock "continue anyway" while keeping the poll alive? That lets the UI naturally move from failed, to retrying, to ready when the daemon recovers.

rgbkrk (dismissed):
Mark, thanks for taking this all the way through the real daemon path on Windows. That verification matters here, and the daemon-side shape makes sense to me: capture the detached child's stderr, keep a bounded diagnostic tail, and put a classified failure into pool state instead of letting it disappear.

The bit I'd like us to tighten before merge is the onboarding lifecycle. Right now the UI stops watching too early, can accept a late response from a previously selected environment, and stays failed even if the daemon's background retry succeeds. I left three focused comments on that path.

For live progress, I like the terminal-view direction, especially since nteract already has terminal rendering components we can reuse. `warm-env` is already emitting structured `Progress { phase, detail }` events for resolving, installing, vendoring, validating, and warming. Publishing the latest phase and detail, plus an update timestamp and any bounded log data the terminal surface needs, through `RuntimePoolState`/`PoolDoc` would give onboarding something useful to show immediately. It could render that as calm status steps, an existing terminal surface, or both. The important part is reusing what we have and keeping the daemon and `PoolDoc` as the source of truth instead of inventing a second frontend state model.

If we extract the frontend polling, this is a good fit for a small onboarding-scoped RxJS controller: `switchMap` on the selected environment, one non-overlapping poll, and one discriminated pool-gate state. I don't think that full refactor needs to block this fix. The minimum here is to guard stale completions and keep observing after the soft timeout or a retryable failure. When progress fields land, we should also reuse the shared `PoolState` type from `runtimed` rather than extending the parallel optional onboarding shape.

PR #4081, opened 2026-07-20
feat(cloud): full-stage sign-in gate for signed-out private notebooks

markmiro:
## What & why

The signed-out private-notebook surface used to be a warning banner floating over an empty notebook stage — it read as broken or still-loading, with 3–4 competing weak sign-in affordances and a leaked/humanized URL slug as the title. This replaces it with a **centered full-stage gate** so a gated route reads as *intentionally gated*.

- Add a shared **`NotebookAccessGate`** primitive (centered icon badge, title, one line of detail, one primary action; `neutral`/`info`/`attention` tones), exported from `src/components/notebook` and cataloged in Elements `unhappy-states`.
- Route `signedOutNotebookSignInRequired` to the gate on the notebook stage: suppress the empty `NotebookView`, hide the rail, and quiet header presence/edit/identity chrome behind the wall (`notebookStageGated`).
- Reuse **`CloudNotebookSignInButton`** as the single sign-in source of truth so the gate's CTA copy and OIDC-vs-localDev priority cannot drift.
- Withhold the redundant inline sign-in banner when the gate owns the surface (`signInRequiredOwnedByStage`) while still suppressing the stale loading notice.
- Show a clean **"Private notebook"** title instead of the humanized URL slug fragment (the "Ob…" bug) until access resolves.

## Before → After

Both captured from the **real cloud viewer** (`wrangler dev` + local OIDC) in the identical signed-out private-notebook state (`signedOutNotebookSignInRequired`: transport reconnecting + anonymous + no snapshot). The only variable is this diff.

| | Before | After |
|---|---|---|
| **Title** | "Quarterly Forecast" — leaked/humanized URL slug (the "Ob…" bug class) | "Private notebook" — clean intentional label |
| **Sign-in affordances** | 3 competing: top-bar button + banner text + weak "Sign in" link | 1 centered high-emphasis button |
| **Primary area** | Huge empty void — reads as broken or still loading | Centered gate panel: icon badge, title, one line of context, the CTA |
| **Left rail** | Live-looking dead icons (outline, packages, comments) | Hidden entirely |
| **Banner** | "Sign in required. Sign in to open the live notebook room." (redundant) | Gone — the gate owns the surface |

### After (light / dark)

 

> The sign-in button reads "Local dev OIDC" in the screenshots only because that's the dev issuer label. Production renders **"Sign in with Anaconda"** through the identical `CloudNotebookSignInButton` code path — the provider label is the only difference, server-derived from `NOTEBOOK_CLOUD_OIDC_PROVIDER_LABEL`. The gate is presentation-only over the existing shared sign-in button; nothing in this diff touches the Anaconda auth mechanics.

## Verification

- `cargo xtask lint --fix` — clean (1201 JS/TS files, ruff + ty pass)
- notebook-cloud `tsc --noEmit` — exit 0
- Modified viewer tests pass their new assertions (`viewer-shared-cell-surface`, `notebookStageGated` regex)
- Elements catalog compiles and renders the gate in both themes (all three tone variants)
- Verified end-to-end in the real `wrangler dev` cloud viewer — the gate fires on the genuine `signedOutNotebookSignInRequired` state, not a fixture
- Caught + fixed along the way: `unhappy-states-example.tsx` needed a `"use client"` directive (an RSC boundary violation that `tsc`/lint missed but broke every Elements route)

### DIFF
--- a/apps/elements/components/unhappy-states-example.tsx
+++ b/apps/elements/components/unhappy-states-example.tsx
@@ -1,3 +1,5 @@
+"use client";
+
 import {
   BookOpen,
   Check,
@@ -15,6 +17,7 @@ import {
   type LucideIcon,
 } from "lucide-react";
 import type { CSSProperties } from "react";
+import { NotebookAccessGate, type NotebookAccessGateTone } from "@/components/notebook";
 import { Button } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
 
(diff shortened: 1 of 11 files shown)


rgbkrk:
Nice, fix up the test and I think you're good to go!

Issue #4018, opened 2026-07-14
room-stack review sweep: deferred findings from Opus/Sonnet/GLM cross-model reviews


rgbkrk:
Cross-model review sweep of the room lifecycle stack (#4009 #4010 #4016 #4011 #4012 #4013) using Bedrock Opus 4.8 on the three hardest slices, Sonnet 5 and GLM elsewhere. No architectural findings. Two review findings became fixes on the stack (`runtime_not_ready` gate code on #4013, watcher manifest-fingerprint skip on #4011). The rest triaged as deferred polish, kept here so they survive the merge.

**Resolved since filing** (room-train, 2026-07-14):
- ~~#4012 `SaveNotebookError` dead code~~ - deleted with the legacy save error wire surface in #4024.
- ~~Dead `status == "error"` check in `execution.rs` `run_all_and_wait`~~ - the wait was rewritten as a batch terminal wait in #4038; the hand-rolled poll loop and its dead check are gone.
- The `run_all_and_wait` per-cell grace concern is also superseded by #4038 (one shared deadline, single trailing grace pass).

**Cross-slice artifacts (non-issues at stack tip, noted for reviewers):**
- #4011 "TS client ignores notebook_already_current/notebook_save_blocked" - handled at `notebook-client.ts:423/430` in #4012.
- #4016 "DocHandle head-containment methods unused" - consumed by #4013 execution guards.
- #4009 "failed initial load has no retry path" / "ADR describes unimplemented machinery" - the retry/reconciliation machinery is #4016/#4011; the ADR intentionally ratifies the full decision ahead of the slices.

**Still open (real, small, non-blocking):**
- ~~`sync_task.rs` fault-injection env var compiles into production builds~~ - gated `cfg(any(test, debug_assertions))` in #4046.
- `daemon.rs` attached-availability projection returns untyped string errors (#4016).
- `kernel.rs` restart retry collapses `session_superseded` into `Disconnected` (#4013).
- Hosted resume/connect no longer polls kernel-status settle (#4013) - verify hosted UX before enabling progressive connect for hosted rooms. This gates the progressive-connect rollout.
- Harness self-test count label hardcodes 22 (#4013).
- `commit_external_source_revision`/`commit_staged_source` could mirror `commit_snapshot`'s durable-heads validation (GLM medium, #4016; inputs are internally generated and `AutoCommit::load` validates structure, so risk is low).
[...]

PR #4082, opened 2026-07-21
fix(mcp): count connect harness self-tests

rgbkrk:
## Summary

- replace the connect harness's raw `assert` statements with explicit counted checks
- report the number of checks actually executed instead of a stale hardcoded value
- keep self-test failures active when Python runs with `-O`

## Root cause

The harness printed `self_tests: 22` independently of the checks in `run_self_tests`. The suite now contains 19 checks, so the diagnostic label had drifted and could drift again.

This addresses the harness self-test count finding tracked in #4018. It does not close that issue because other deferred findings remain.

## Verification

- `python3 scripts/mcp-connect-harness.py --self-test`
- `python3 -O scripts/mcp-connect-harness.py --self-test`
- forced a failing first check under `python3 -O` and verified `AssertionError: self-test 1 failed`
- in-memory Python compilation of `scripts/mcp-connect-harness.py`
- `cargo xtask lint --fix`

### DIFF
--- a/scripts/mcp-connect-harness.py
+++ b/scripts/mcp-connect-harness.py
@@ -1755,6 +1755,14 @@ def parse_args() -> argparse.Namespace:
 
 
 def run_self_tests() -> int:
+    self_test_count = 0
+
+    def check(condition: bool) -> None:
+        nonlocal self_test_count
+        self_test_count += 1
+        if not condition:
+            raise AssertionError(f"self-test {self_test_count} failed")
+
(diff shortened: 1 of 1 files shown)


quillaid:
Bedrock Claude Opus 4.6 final review: clear, with no findings. The reviewer specifically verified that the counted `check()` helper remains active under Python optimization, increments before a failure for accurate numbering, and removes the stale hardcoded count without introducing authority or coverage gaps.

Issue #4069, opened 2026-07-17
getpass not implemented


intentionally-left-nil:
```
import getpass, os

os.environ["MY_SECRET"] = getpass.getpass("Enter in your secret ")
```

fails with
```
Traceback (most recent call last):
  Line 3 in Current Cell (cell_id=cell-09eb4dd1-9e38-4832-bfde-4d152b469e10, execution_id=1c74f13a-89ce-43fa-870f-20970ded2c30, source_hash=sha256:007e5022121eba0d85d034bb5a826b28f352bc5c99c438b6505190f9119f855c)
                raise StdinNotImplementedError(msg)
StdinNotImplementedError: getpass was called, but this frontend does not support input requests.
```

rgbkrk:
Thanks for the clean repro. Confirmed: the kernel raises `StdinNotImplementedError` because the runtime agent does not implement the Jupyter stdin channel, so `allow_stdin` is false for every execute request and anything that prompts (`getpass`, `input()`) fails.

Supporting it properly means carrying an `input_request`/`input_reply` round trip from the kernel through the daemon to whichever client ran the cell, plus a masked-input affordance in the UI for the `password=True` case. That is on the table but not scheduled yet.

Until then, the working patterns are environment variables (`os.environ`), an `.env` file loaded with `python-dotenv`, or your OS keychain via `keyring`, all of which avoid putting the secret in notebook source or outputs.

Marking this as a feature request for the stdin channel.

PR #4084, opened 2026-07-21
docs(runtime): propose kernel stdin control protocol

quillaid:
## What changed

- adds a proposed architecture memo for Jupyter `input()` / `getpass()` support
- defines stdin as an explicit, default-off execution capability
- proposes a reliable actor-targeted prompt frame and correlated `ReplyInput` request
- specifies stale-reply rejection, interrupt/restart/disconnect cleanup, and password-handling invariants
- breaks implementation into independently safe, default-off slices
- links the proposal from the documentation front door

## Why

The current kernel path constructs `execute_request` with `allow_stdin = false` and has no stdin DEALER reader. Simply enabling stdin would block executions indefinitely and a room-wide broadcast would be both lossy and visible to unrelated peers.

This proposal keeps prompt traffic ephemeral and targeted while preserving RuntimeStateDoc as the durable execution record. It also records a concrete limitation discovered during review: the current daemon-mediated hosted bridge does not preserve the initiating local actor on forwarded requests, so stdin must remain disabled on that path until delegated end-actor routing exists.

## Impact

This is a design-only draft. It does not enable stdin or change the wire protocol yet. It provides reviewable boundaries for local rooms and direct cloud sessions, plus an explicit safe staging order.

Closes no issue; informs #4069.

## Validation

- `cargo xtask lint --fix`
- `git diff --check`
- verified every referenced source path exists
- two US Bedrock Claude Opus 4.6 architecture passes
  - first pass identified the daemon-mediated bridge attribution gap and several clarification nits; all were addressed or dispositioned
  - second pass returned `clear` with no findings


quillaid:
Rebased safely onto current main at a246105a. New head: 5de1ed89. Git range-diff confirms the proposal patch is identical to its pre-rebase form. cargo xtask lint --fix, git diff --check, and Docs & Agent Guidance all pass.

rgbkrk:
Closing the stdin design draft during release cleanup. #4084 and #4089 intentionally propose mutually exclusive protocol contracts; neither is scheduled for 2.7.0, and both now conflict with `main`. The reviewed targeted-frame design remains in git history and can be revived from issue #4069 when the protocol decision is active.

PR #4085, opened 2026-07-21
fix(runtimed): stage full-coverage save baselines

### LINKED ISSUE #4065: fix(runtimed): created-at-path notebook journal restores as permanently degraded room (reconnect + auto-launch storm)
A notebook created at a path (never imported from disk) produces a recovery journal that restores as a permanently degraded room on the next daemon start. The room then rejects every client connection, the client reconnect loop has no backoff, and each early connect fires an unguarded kernel auto-launch. Observed on nightly 2.6.3-nightly.202607160939: 2,201 rejected connections over 2.5 hours, 121 runtime-agent spawns with 118 SIGKILL reaps, and the daemon hit `Too many open files (os error 24)` on its accept loop, which is daemon-wide blast radius. In the app this reads as a flickering "Runtime unavailable / Runtime disconnected. Attempting to reconnect..." banner over an empty notebook stuck "Initializing."

## Root cause chain

Three bugs stack:

### 1. Ordinary saves never establish a source baseline

`commit_file_checkpoint` (`crates/runtimed/src/notebook_sync_server/durability.rs:783`) always passes `source_generation: None`; only explicit reconciliation (`commit_reconciled_file_checkpoint`) stages a generation. `apply_mutation_to_manifest` advances `source_phase` to `DurablyStaged` only when the checkpoint carries a generation (`durability.rs:1234`). A notebook born at a path has no file import, so its manifest stays `source_phase: pending, source_generation: 0` forever, no matter how many clean saves it makes.

### 2. Restore condemns a complete-checkpoint journal as degraded

The startup restore branch (`crates/runtimed/src/notebook_sync_server/room.rs:1519`) sees `Pending|Failed` + non-empty `peer_change_hashes` and declares `source_degraded: recovery contains N peer changes but no durably staged source generation`, failing initial materialization. But the actual manifest in this incident proves nothing is missing:

- `durable_heads == exported_heads` (the single peer change is fully exported)
- `file_save_sequence: 1` with `canonical_path` set
- disk file sha256 exactly matches `source_fingerprint`

The export is the baseline. The journal is durable recovery evidence, not danger.

The retry gate makes this terminal by construction: `retry_failed_room_initial_load_if_safe` (`load.rs:1446`) requires a pristine doc, but restoring the journal applies the peer change, so the doc is never pristine. The one thing that degrades the room is the same thing that disqualifies its repair.

### 3. Failed materialization becomes an unthrottled reconnect and spawn storm

`peer_session.rs:448` returns a connection error on `Failed` initial load, tearing down the socket after sending `InitialLoadPhaseWire::Failed`. The client treats the drop as a transient runtime disconnect and reconnects immediately, several times per second, indefinitely. Before `has_kernel` flipped true, each connect also fired auto-launch with no per-room single-flight guard: every new spawn superseded the previous agent, the manifest reaper SIGKILLed it mid-launch, the launch failure triggered another spawn. That storm is what exhausted file descriptors.

## Proposed fixes

1. **Restore-side heal (fixes existing journals):** in the `Pending|Failed` + peer-changes branch, a manifest with `exported_heads == durable_heads` and a recorded file checkpoint is a complete export baseline; restore it like the healthy case. The existing watcher/source-conflict machinery still catches a foreign disk write via fingerprint mismatch, so this does not weaken conflict detection.
2. **Save-side baseline (fixes new journals):** when a save fully exports the durable heads and no source generation is staged, claim and stage one in the same checkpoint commit. The export is a source generation by construction.
3. **Terminal failure UX:** a `Failed` initial load should surface as a terminal "couldn't load this notebook" state in the client, not a reconnect loop. Better: keep the session open in last-known-good browse mode per the app-shell doctrine, gating actions rather than presence.
4. **Single-flight auto-launch per room** so no connect pattern can become an agent spawn storm.

## Evidence

Nightly log 2026-07-16 12:51-15:38 (log rotated; quarantined copies retained). Poisoned journal triple (`.recovery`, `.recovery.manifest.json`, `.automerge`) plus the matching `.ipynb` captured before mitigation for use as a regression fixture. Mitigation applied: quarantine the journal triple, restart the daemon, room imports fresh from disk.

## Related

Third head of the same durability hydra: #4053 (cross-channel split-brain at open), #4062 (healthy-journal source conflict blocks daemon shutdown). All three share the shape "lifecycle gate treats journal-fully-represented-on-disk as danger instead of health."

Also observed during mitigation, minor: `runt-nightly daemon start` prints "Service started" but launchd does not spawn the process until `launchctl kickstart`; `daemon status` shows "Daemon running: no" until then.

quillaid:
Fixes #4065.

## Summary

- promote an ordinary file checkpoint to `DurablyStaged` in the same journal record when it exports every durable head
- let an `AlreadyCurrent` ordinary save append a missing promotion once, while keeping later no-op saves journal-idempotent
- preserve the existing source generation, including generation 0 for created-at-path rooms, matching the landed restore/resolver normalization instead of introducing a second async lifecycle publication transaction
- keep a staged baseline valid when later durable peer changes form a replayable tail

## Verification

- `cargo test -p runtimed full_coverage_checkpoint --lib` (4 passed)
- `cargo test -p runtimed recovered_checkpoint_with_unexported_tail --lib` (1 passed)
- `cargo test -p runtimed already_current_save_does_not_advance_checkpoint_or_saved_timestamp --lib` (1 passed)
- `cargo test -p runtimed notebook_sync_server::durability::property_tests --lib` (7 passed)
- `cargo xtask lint --fix`
- `git diff --check`

## Review

Bedrock Claude Opus 4.6 architecture review is clear after one remediation pass. The final review specifically challenged generation-0 preservation, crash/idempotence, degraded and `AlreadyCurrent` paths, and peer-tail replay; it found no correctness blockers.

This remains draft so the narrower generation-preserving choice is easy to review against the larger generation-claim alternative described in the issue.

### DIFF
--- a/crates/runtimed/src/notebook_sync_server/durability.rs
+++ b/crates/runtimed/src/notebook_sync_server/durability.rs
@@ -924,18 +924,28 @@ impl RoomDurability {
                 )));
             }
         }
-        if state.has_durable_record && mutation_is_already_reflected(&state.manifest, &mutation) {
+        let already_reflected =
+            state.has_durable_record && mutation_is_already_reflected(&state.manifest, &mutation);
+        let mut manifest = state.manifest.clone();
+        if !already_reflected {
+            apply_mutation_to_manifest(&mut manifest, mutation);
+        }
+        // A successful ordinary save that exports every durable head is the
+        // same complete source baseline that restore and intent resolution
+        // already recognize. Stage it in the save's commit record so a room
+        // created at a path is joinable before it has to restart once.
+        if state.degraded.is_none() {
+            promote_full_coverage_checkpoint_baseline(&mut manifest);
+        }
+        if already_reflected && manifest == state.manifest {
             return Ok(DurableCommitOutcome::AlreadyDurable(status_from_state(
                 &state,
             )));
         }
-
-        let mut manifest = state.manifest.clone();
         manifest.sequence = manifest
             .sequence
             .checked_add(1)
             .ok_or(RoomDurabilityError::SequenceExhausted)?;
-        apply_mutation_to_manifest(&mut manifest, mutation);
 
         if let Some(journal) = self.journal() {
             if let Err(error) = journal.append(&manifest, &state.durable_snapshot) {
@@ -1295,6 +1305,10 @@ fn status_from_state(state: &DurabilityState) -> RoomDurabilityStatus {
 /// dedicated resolution path owns the transition) or when any durable head is
 /// missing from the export. Callers gate on their degraded state: a journal
 /// with preserved corrupt data must not be promoted.
+///
+/// This normalization changes only `source_phase`. In particular it preserves
+/// both `source_generation` and journal `sequence`; the caller owns the append
+/// sequence for the record carrying the transition.
 fn promote_full_coverage_checkpoint_baseline(manifest: &mut RecoveryManifest) {
     if matches!(
         manifest.source_phase,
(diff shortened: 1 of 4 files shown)


quillaid:
Bedrock Opus 4.6 review disposition:

- Replaced the artificial in-memory legacy-phase test with a real degraded checkpoint, degradation clear, and same-checkpoint recommit; it proves the missing promotion appends once and later commits are idempotent.
- Documented that baseline normalization changes only source phase and preserves generation and journal sequence.
- Added an `AlreadyCurrent` integration assertion that no extra journal record is written.
- Added evidence that an unexported peer tail remains present in the recovered durable Automerge snapshot.
- Updated the shadow model comments and fixed sequence to mirror the production promotion guard.

A second architecture review explicitly challenged generation-0 preservation and the peer-tail classification and returned clear with no findings.

PR #4086, opened 2026-07-21
fix(runtimed): rebuild captured env after launch failure

### LINKED ISSUE #1969: runtimed: auto-detect and rebuild corrupted captured envs
**Priority:** P2 (ship before stable)
**Related:** #1954 (unified env resolution design), #1960 (captured envs), #1963 (preserve on eviction)

## Problem

When a captured env on disk becomes unusable (partial install, disk-level corruption, missing interpreter, interrupted install that left a broken venv), the next reopen still cache-hits via `unified_env_on_disk` and tries to launch Python from the broken dir. The kernel launch fails, and the user sees the failure with no path to recovery beyond manually editing the `.ipynb` to clear `metadata.runt.env_id`.

Since #1963 preserves captured envs aggressively across daemon restart, a single bad state can persist indefinitely. Manual recovery requires knowledge of internal metadata fields.

## Detection signals

Two viable approaches:

### 1. Kernel-handshake failure on the captured path

`jupyter_kernel.rs` spawns Python and waits for `kernel_info_reply`. Today, handshake failure bubbles up as `KernelLaunched { error }` and the user sees it. Distinguish "infrastructure failure on the captured path" from "user code failure":

- If `env_source` is `uv:prewarmed` or `conda:prewarmed` AND `unified_env_on_disk` was the source of the venv path AND the handshake fails, treat the env as corrupted.
- Out-of-band signals that are always infrastructure: process exits before `kernel_info_reply`, interpreter binary fails to execute, missing `ipykernel` module on stdout/stderr.

### 2. Sidecar `.captured-ok` marker

Write `{cache}/{unified_hash}/.captured-ok` after the first successful kernel handshake. Cache-hit check in `unified_env_on_disk` gains a second step: if the sidecar is missing, treat as corrupted.

The sidecar catches "install was interrupted, kernel has never actually worked in this env" which is the most common corruption pattern. The handshake approach catches "env was fine, then got broken later" (harder to simulate but does happen).

Ship both. They're cheap and orthogonal.

## Recovery flow

On corruption detection, before returning failure to the user:

1. Log: `[notebook-sync] Captured env for {notebook_id} appears corrupted ({reason}), rebuilding from captured deps`.
2. Delete `{cache}/{unified_hash}/`.
3. Fall through to the inline-deps rebuild path: `prepare_{uv,conda}_inline_env(captured_deps, env_id)`.
4. Re-attempt kernel launch once against the rebuilt env.
5. If the rebuild + relaunch still fails, surface the original error to the user (don't mask the real problem).

## Interaction with #1968 (user-facing rebuild button)

Auto-recovery makes the manual "rebuild env" action a pure escape hatch rather than the primary recovery path. Ship order: this first (correctness), #1968 after (convenience).

## Out of scope

- Retry on user-code errors. Recovery is strictly for handshake-phase failures.
- Proactive env health checks on every launch. That's a perf regression for a corner case.

## Acceptance

- [ ] Write `.captured-ok` sidecar after first successful handshake.
- [ ] `unified_env_on_disk` treats missing sidecar as a cache miss.
- [ ] Handshake failure on a captured env triggers one rebuild attempt.
- [ ] Test: tmpdir-backed cache with a captured env whose Python binary has been removed — daemon detects, rebuilds, relaunches.
- [ ] Test: sidecar missing triggers rebuild even on an otherwise-complete env dir.
- [ ] Second consecutive failure surfaces the underlying kernel error.

quillaid:
## Summary

- add backend-owned force rebuilds for captured UV and Conda environments
- rebuild and relaunch exactly once after captured-environment startup infrastructure failures
- apply the policy to auto-launch, manual fresh launch, and manual restart
- preserve the original launch error and avoid publishing an intermediate terminal lifecycle

Fixes #1969.

## Root cause

Captured environments were considered usable when their Python executable existed. If packages or `ipykernel` later became corrupt, the runtime agent reported a typed handshake/startup failure, but the coordinator immediately surfaced it without rebuilding the environment from the notebook's captured declaration.

## Implementation

The coordinator retains the `CapturedEnv` resolved for the original launch. For `ProcessExited`, `StartupTimeout`, or `ToolBootstrap`, it force-rebuilds that exact UV/Conda unified identity and sends one more launch or restart request. Each launch attempt retains the existing bounded fresh-port retry policy.

The runtime agent defers terminal lifecycle publication for failure kinds owned by coordinator retry policy. The outer launch handlers publish the final error after the retry decision, so RuntimeStateDoc does not expose `Error` between the first failure and repair.

Conda force rebuilds preserve the existing matching-lock path. UV intentionally recreates from the captured declaration because it has no equivalent environment lock sidecar.

## Validation

- `cargo test -p runtimed notebook_sync_server::runtime_bridge::tests --lib` — 13 passed
- `cargo test -p runtimed kernel_launch_failure::tests` — 8 passed
- `cargo test -p runtimed captured_env_retry_transport_failure_response_clears_launching --lib` — passed
- `cargo test -p kernel-env` — 94 unit tests and launcher E2E passed; 2 doc tests ignored
- `cargo xtask lint --fix` — passed
- US Bedrock Opus 4.6 structured architecture review; its actionable port-retry boundary test was added
- final US Bedrock Opus 4.6 PR review (`ses_07974da64ffe7U8LdD3RJurixa`) — clear, no findings

`cargo test -p runtimed` was also attempted. The host exhausted disk space while rustc linked all examples and integration targets; no test executed or failed in that attempt. Focused runtimed targets compiled and passed before and after the final coverage change.

## Decisions and residual risks

- Rebuild is deliberately limited to captured UV/Conda environments and three infrastructure-oriented failure kinds. Port/transport failures remain owned by the existing fresh-port loop; unsupported, misconfigured, and unknown failures do not trigger destructive repair.
- The dependency declaration and `env_id` are frozen from the initial resolution. Metadata changes during repair apply to a later launch cycle.
- The failed kernel child uses `kill_on_drop(true)` and its watcher is aborted before the failure returns, so the coordinator does not intentionally rebuild under a live kernel process.
- This PR tests retry orchestration with injected send/repair futures and exercises the full backend suites. It does not add a slow end-to-end Conda corruption fixture; lock-assisted and full-solve behavior remain covered at the backend level rather than through a spawned daemon.

## Reviewer dispositions

The final structured review is clear. Earlier findings were fixed or dispositioned with code evidence in PR comments: the retry publishes `PreparingEnv -> Launching`, then all three outer handlers own the terminal outcome; retry transport failures normalize back into `KernelLaunchFailed`, and a direct state test proves `Launching` cannot stick. Cancellation can leave a partial backend directory, which is an existing typed state repaired on the next launch; atomic environment replacement remains outside this issue.

### DIFF
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -386,7 +386,9 @@ async fn prepare_environment_unified_inner(
 
     tokio::fs::create_dir_all(cache_dir).await?;
 
-    // Try lock-based rebuild before full re-creation
+    // Try lock-based rebuild before full re-creation. A forced rebuild enters
+    // this path even when Python exists because the launch handshake already
+    // proved the otherwise-usable environment cannot start a kernel.
     if env_path.exists() && (force_rebuild || !python_path.exists()) {
         if let Some(lock) = crate::lock::LockFile::read_from(&env_path).await {
             let expected_specs = build_spec_strings(deps);
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -1539,6 +1539,8 @@ mod tests {
         assert!(unified_cache_is_reusable(false, true, true));
         assert!(!unified_cache_is_reusable(true, true, true));
         assert!(unified_lock_rebuild_is_applicable(true, true, true));
+        assert!(!unified_lock_rebuild_is_applicable(true, false, true));
+        assert!(!unified_lock_rebuild_is_applicable(true, false, false));
     }
 
     #[test]
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -334,6 +334,18 @@ pub async fn rebuild_environment_unified(
     prepare_environment_unified_inner(deps, env_id, cache_dir, handler, true).await
 }
 
+fn unified_cache_is_reusable(force_rebuild: bool, env_exists: bool, python_exists: bool) -> bool {
+    !force_rebuild && env_exists && python_exists
+}
+
+fn unified_lock_rebuild_is_applicable(
+    force_rebuild: bool,
+    env_exists: bool,
+    python_exists: bool,
+) -> bool {
+    env_exists && (force_rebuild || !python_exists)
+}
+
 async fn prepare_environment_unified_inner(
     deps: &CondaDependencies,
     env_id: &str,
@@ -357,7 +369,7 @@ async fn prepare_environment_unified_inner(
     let python_path = env_path.join("bin").join("python");
 
     // Cache hit
-    if !force_rebuild && env_path.exists() && python_path.exists() {
+    if unified_cache_is_reusable(force_rebuild, env_path.exists(), python_path.exists()) {
         info!("Using cached unified conda env at {:?}", env_path);
         crate::gc::touch_last_used(&env_path).await;
         crate::launcher::vendor_into_venv(&python_path)
@@ -389,7 +401,7 @@ async fn prepare_environment_unified_inner(
     // Try lock-based rebuild before full re-creation. A forced rebuild enters
     // this path even when Python exists because the launch handshake already
     // proved the otherwise-usable environment cannot start a kernel.
-    if env_path.exists() && (force_rebuild || !python_path.exists()) {
+    if unified_lock_rebuild_is_applicable(force_rebuild, env_path.exists(), python_path.exists()) {
         if let Some(lock) = crate::lock::LockFile::read_from(&env_path).await {
             let expected_specs = build_spec_strings(deps);
             let expected_channels = if deps.channels.is_empty() {
@@ -1520,6 +1532,13 @@ mod tests {
         let h1 = compute_unified_env_hash(&deps, "notebook-1");
         let h2 = compute_unified_env_hash(&deps, "notebook-2");
         assert_ne!(h1, h2);
+    }
+
+    #[test]
+    fn forced_unified_rebuild_bypasses_cache_and_uses_lock_path_when_available() {
+        assert!(unified_cache_is_reusable(false, true, true));
+        assert!(!unified_cache_is_reusable(true, true, true));
+        assert!(unified_lock_rebuild_is_applicable(true, true, true));
     }
 
     #[test]
--- a/crates/kernel-env/src/conda.rs
+++ b/crates/kernel-env/src/conda.rs
@@ -317,6 +317,30 @@ pub async fn prepare_environment_unified(
     cache_dir: &Path,
     handler: Arc<dyn ProgressHandler>,
 ) -> Result<CondaEnvironment> {
+    prepare_environment_unified_inner(deps, env_id, cache_dir, handler, false).await
+}
+
+/// Force-rebuild a notebook-captured Conda environment at its unified hash.
+///
+/// A matching lock sidecar is consumed before the existing environment is
+/// removed, preserving the normal lock-assisted rebuild path. The target path
+/// remains derived exclusively from the captured declaration and `env_id`.
+pub async fn rebuild_environment_unified(
+    deps: &CondaDependencies,
+    env_id: &str,
+    cache_dir: &Path,
+    handler: Arc<dyn ProgressHandler>,
+) -> Result<CondaEnvironment> {
+    prepare_environment_unified_inner(deps, env_id, cache_dir, handler, true).await
+}
+
+async fn prepare_environment_unified_inner(
+    deps: &CondaDependencies,
+    env_id: &str,
+    cache_dir: &Path,
+    handler: Arc<dyn ProgressHandler>,
+    force_rebuild: bool,
+) -> Result<CondaEnvironment> {
     let hash = compute_unified_env_hash(deps, env_id);
     let env_path = cache_dir.join(&hash);
 
@@ -333,7 +357,7 @@ pub async fn prepare_environment_unified(
     let python_path = env_path.join("bin").join("python");
 
     // Cache hit
-    if env_path.exists() && python_path.exists() {
+    if !force_rebuild && env_path.exists() && python_path.exists() {
         info!("Using cached unified conda env at {:?}", env_path);
         crate::gc::touch_last_used(&env_path).await;
         crate::launcher::vendor_into_venv(&python_path)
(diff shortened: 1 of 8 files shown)


quillaid:
Opus 4.6 review disposition (session `ses_0798d7e80ffeItgOQmomtO8LLk`):

1. **`CapturedEnv` import — not actionable.** `runtime_bridge.rs` imports `super::*`; `notebook_sync_server/mod.rs` re-exports `metadata::*`, so `CapturedEnv` is in scope. The focused `runtimed` tests compile and pass.
2. **Lifecycle around the rebuild retry — fixed in `525f7a9a`.** A successful rebuild now publishes `Launching` before the retry request is handed to the runtime agent, making the transition `PreparingEnv -> Launching` explicit. A retry transport failure is normalized to the original `KernelLaunchFailed`, so all three callers execute their existing terminal-state reset/error path. Added `retry_transport_failure_preserves_original_error_without_a_third_send` to cover the closed-channel branch.
3. **Conda forced-rebuild comment — fixed in `525f7a9a`.** The comment now explains why a forced rebuild intentionally enters the lock-assisted path even when Python exists.
4. **`FnMut` compatibility — fixed in `525f7a9a`.** Restored the lower-level `send_runtime_agent_request_with_kernel_ports` callback contract to `FnMut`. Only the new outer repair wrapper requires `Fn`, because it must be able to construct requests for two bounded launch rounds.
5. **Cancellation/concurrent rebuild test — deferred, non-blocking nit.** Launch admission is already serialized through runtime lifecycle/auto-launch gates, and interrupted environment creation already follows the existing partial-cache repair behavior. Making cache replacement atomic would be a broader environment-installer change rather than part of the single launch-handshake repair in #1969.
6. **`captured.cloned()` — retained intentionally.** The owned clone freezes the exact captured environment identity across both async launch rounds while callers still borrow the same resolved configuration to build requests. The payload is small, and ownership here makes the retry boundary explicit.

Post-remediation verification:

- `cargo test -p runtimed notebook_sync_server::runtime_bridge::tests --lib` — 13 passed
- `cargo test -p runtimed kernel_launch_failure::tests --lib` — 8 passed
- `cargo test -p kernel-env` — 92 unit tests + launcher E2E passed; 2 doc tests ignored
- `cargo xtask lint --fix` — passed

quillaid:
Follow-up for the second Opus review (session `ses_0798476c5ffe5M9DTsqMOg7QMp`), addressed in `e8796bea`:

- Added backend-level cache-policy regression tests for both UV and Conda. They assert that `force_rebuild=true` bypasses an otherwise reusable environment; the Conda test also asserts the matching-lock rebuild path remains applicable. The production unified-env paths now use those tested decisions directly.
- Added `captured_env_retry_transport_failure_response_clears_launching`, which composes the exact normalized `KernelLaunchFailed` response with the production terminal reset and directly asserts `Launching -> Error`, preserved error details, and cleared environment progress.
- Clarified the repair wrapper's contract: retry transport errors are normalized specifically so each of its three production callers enters its existing contextual terminal-state arm.

The other second-pass entries explicitly withdrew themselves or described intentional/no-bug behavior (the owned `CapturedEnv`, UV full recreation, existing `env_id()` method, fresh-port allocation per launch round, and generic helper used for testability).

Verification after this follow-up:

- `cargo test -p kernel-env` — 94 unit tests + launcher E2E passed; 2 doc tests ignored
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p runtimed captured_env_retry_transport_failure_response_clears_launching --lib` — passed
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p runtimed notebook_sync_server::runtime_bridge::tests --lib` — 13 passed
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p runtimed kernel_launch_failure::tests --lib` — 8 passed
- `cargo xtask lint --fix` — passed

quillaid:
Disposition for malformed Opus review session `ses_07979af1fffeY6RKD355cLIcz9` (the report was `infra_uncertain` because one finding used a string line number):

- **Concurrent restart / stale retry — not actionable.** Manual launch/restart admission atomically treats `Resolving`, `PreparingEnv`, `Launching`, and `Connecting` as an in-progress launch and waits for a terminal lifecycle before it can continue (`requests/launch_kernel.rs`). Auto-launch is single-flight via `AutoLaunchAttempt`; once this retry path is reachable, the existing runtime-agent handle also causes any racing auto-launch to return. The request channel itself is processed sequentially. There is no second launch that can interleave in the described rebuild-to-retry window.
- **Conda prefix allegedly not removed — factually incorrect.** In the matching-lock branch, the implementation reads the lock first, then calls `tokio::fs::remove_dir_all(&env_path)` and `tokio::fs::create_dir_all(&env_path)` immediately before `install_from_lock`. The suggested fix is already the code under review.
- **Captured snapshot divergence — intentional invariant.** The retry deliberately owns the originally resolved `CapturedEnv` across its await boundary; re-reading mutable notebook metadata midway would let a retry switch environment identity. A later metadata edit is handled as subsequent configuration drift, not by mutating an in-flight launch.
- **UV lock-path comment — non-actionable nit.** UV has no lock-sidecar restore path; force mode bypasses cache and falls through to remove/recreate. The backend cache-policy test now locks that decision down.

No code change was warranted from this malformed pass.

quillaid:
Disposition for Opus session `ses_079781e43ffe1G7dPrMMLGlbW6`:

- The reported lifecycle item explicitly concedes this is not a stuck-state bug and asks only that the visible `Launching -> Error` transition be documented. Commit `4bdecc19` now documents that sequence and the caller-owned terminal reset directly on `rebuild_captured_environment`.
- The single-flight precondition and the exact one-plus-optional-one sequential `FnMut` call contract are now documented in `4bdecc19`.
- Added the suggested missing-directory assertions to the Conda cache-policy test and documented UV's intentional remove/recreate fallthrough.
- `CapturedEnv` already has `#[derive(Debug, Clone)]`, and `CapturedEnv::env_id()` is defined immediately below it in `metadata.rs`; both compilation concerns are disproven by the passing tests.

Verification: both focused backend policy tests pass, and `cargo xtask lint --fix` passes.

quillaid:
Final Bedrock Opus 4.6 review is **clear** (`review_complete`), session `ses_07974da64ffe7U8LdD3RJurixa`.

> The PR correctly introduces a single-shot captured-environment rebuild after qualifying kernel launch failures. The retry logic is well-bounded (at most one rebuild, one relaunch), fresh-port failures are excluded from the rebuild layer, terminal lifecycle transitions are preserved by delegating to existing reset arms, and the runtime agent's premature terminal-state publish is suppressed via `coordinator_may_retry`. Helper predicates are unit-tested, the async retry composition is thoroughly covered with mock closures, and the integration test confirms the `Launching` state cannot stick after a transport failure on retry. No concrete bugs, authority bypasses, ordering hazards, or missing test coverage found.

quillaid:
CI follow-up: Linux Rust Clippy flagged the retry callback as a redundant closure. Commit c0b0fccc applies Clippy’s semantics-preserving suggestion and passes `cargo clippy -p runtimed --lib -- -D warnings`, `cargo xtask lint --fix`, and `git diff --check`. The behavioral suite was already green on the preceding commit; I stopped a redundant fresh test-profile rebuild because the runner had under 2 GiB free.

Issue #4000, opened 2026-07-13
Feature Request: Introduce "Continue on SyntaxError" Mode for Batch Execution


NotesOfReality:
## Description
I would like to propose a "Continue on SyntaxError" toggle for batch notebook execution in `nteract`. While it is certainly possible to execute notebook cells manually one by one, batch execution currently halts at the first `SyntaxError`.
This optional "QoL" feature would allow users to demonstrate both functional code and intentionally incorrect syntax within a single batch run, without the execution engine force-stopping.

## Requirements

*   **Non-blocking Parsing:** The batch runner should optionally catch `SyntaxError` exceptions, allowing it to gracefully skip the malformed block and continue processing subsequent cells instead of terminating the sequence.
*   **Output Capture:** Rather than terminating, the notebook should capture the `SyntaxError` traceback and display it directly within the affected cell’s output area, keeping the kernel environment active.
*   **Visual Status Indicators:** To improve readability, I propose adding distinct status badges:
    *   **Syntax Error Badge:** For the specific cell that failed to parse.
    *   **Dependency Error Badge:** For downstream cells automatically skipped. This visual distinction helps users immediately identify the root cause versus secondary side effects.
*   **Smart Dependency Handling:** If a cell fails, the system should prevent the execution of downstream cells that rely on the failed logic. When a dependent cell is skipped, it should display: *"Execution skipped: requires [variable] from [Cell ID/Index], whose execution failed."* 

## Educational Rationale

*   **Decoupling:** Separating syntax validation from the kernel’s execution lifecycle creates a smoother experience for teaching and testing.
*   **Instructional Clarity:** This allows instructors to present broken code snippets as part of a lesson while ensuring the rest of the notebook remains functional, making it easier to show how valid code should be structured in comparison.
*   **Dependency Handling:** This approach prevents the confusion caused by cascading `NameError` exceptions, clarifying the distinction between a fundamental syntax failure and a subsequent logic error.

## Use Case
This feature is a "nifty" enhancement for educational content, such as tutorials, workshops, or course materials, where presenting both correct and incorrect Python syntax in a single batch execution stream is beneficial, but not strictly required.

---
**System Info**
- App: 2.6.3 (bde4a67)
- OS: windows Microsoft Windows [Version 10.0.26200.8655] (x86_64)
- Released: 2026-07-10

PR #4087, opened 2026-07-21
docs(execution): frame batch failure policy

rgbkrk:
## Summary

- map the current Run All path from protocol through local/hosted coordinators, RuntimeStateDoc, the runtime-agent FIFO, and shared cell UI
- propose a fail-closed, request-scoped `continue_on_syntax_error` contract while preserving stop-on-error as the default
- define coordinator-owned plan membership, adapter-owned syntax classification, aggregate result semantics, and shared desktop/cloud presentation
- separate sequential syntax-error continuation from future dependency-engine `blocked` semantics
- link the memo from the documentation front door

## Why a memo

The current runtime already captures syntax tracebacks and keeps the kernel alive, but `CellError` loses error classification and clears the entire FIFO. There is no durable batch identity, accepted failure policy, cancellation cause, or dependency graph. A one-line queue change would continue unrelated work, provide no fail-closed behavior across version skew, and could not support the requested dependency explanations honestly.

This memo makes the smallest coherent implementation boundary concrete before those protocol and RuntimeStateDoc fields become compatibility commitments. It addresses the design/probe work for #4000; it does not close the feature request.

## Proposed first slice

- keep existing Run All behavior unchanged by default
- use a distinct opt-in request discriminant so older daemons reject before queue mutation
- stamp plan identity/index/policy on coordinator-accepted member executions
- carry adapter-classified syntax errors through the reliable lifecycle path after durable output commit
- continue only syntax-family errors; stop on ordinary or unknown errors
- wire one shared secondary Run All action across desktop and hosted shells

Dependency-aware skipping remains a later execution-engine capability; the sequential engine must not infer Python dependencies from visual order or traceback text.

## Verification

- validated every local Markdown link in the changed docs
- `git diff --check`
- `cargo xtask lint --fix`


quillaid:
Bedrock Opus 4.6 review disposition:

- Added a source-audit table and verified the current `RunAllCells`, runtime-agent FIFO, `LifecycleSignal::CellError`, user-error normalization, and shared UI symbols/paths against this branch.
- Made version-skew behavior exact: the new request must use a distinct enum discriminant so older daemons fail closed before queue mutation.
- Marked all illustrated schema names explicitly non-canonical and reordered the docs index behind the broader execution-engine memo.
- Corrected the programmatic-surface wording and exact error-normalization citation after the second pass.

A third Opus 4.6 pass returned clear with no findings. Docs & Agent Guidance CI is green.

quillaid:
Rebased safely onto current main at a246105a. New patch commits: d729a2ea, 0cda8bc7, 54ae6665. Git range-diff confirms all three patches are unchanged. cargo xtask lint --fix, git diff --check, and Docs & Agent Guidance all pass.

rgbkrk:
Closing to keep the upcoming stable release focused; this design memo can return when batch failure policy becomes active implementation work.

PR #4088, opened 2026-07-21
fix(desktop-cloud): surface hosted bridge health

### LINKED ISSUE #3599: feat(desktop-cloud): show hosted bridge link health in desktop status
> *This was generated by AI during triage.*

## Current triage (2026-07-10)

- **Verdict:** Keep open as a Desktop-specific implementation slice.
- **Current code:** The shared cloud-room link surface has landed, but the Desktop hosted bridge does not expose a reactive health signal beyond ordinary daemon status.
- **Next slice:** Project connected, reconnecting, offline, authentication-failed, and terminal-error bridge states into Desktop status with actionable recovery.
- **Boundary:** Do not duplicate the already shipped cloud-shell room-link indicator.

ADR-recorded future work from #3570: the desktop connectivity dot renders for remote contexts (`runtime_peer` / cloud-attached) but measures only the first hop — frontend↔daemon lifecycle (`ready`/`disconnected`/`unavailable`). The link whose health the remote context actually depends on is **daemon↔cloud room**, which the dot cannot see; the copy is scoped honestly ("Daemon connection: …") as a stopgap.

**Fix shape:** the daemon's cloud runtime agent already runs the reference reconnect loop (`crates/runtimed/src/runtime_agent.rs`, `reconnect_with_backoff`) — surface its connection state to the frontend (daemon event or RuntimeStateDoc-adjacent control signal, NOT output transport per the control-plane invariant), compose it into the desktop status source (`apps/notebook/src/lib/desktop-connection-status.ts`), and widen the scoped copy. The shared component already renders all four ConnectionStatus values — zero UI changes.

Size: M (daemon signal + plumbing + tests). Cross-layer but well-bounded.

quillaid:
## What changed

- publish daemon-to-hosted-room bridge lifecycle through the reliable session-control path
- expose that typed status through the WASM sync engine and compose it with desktop daemon health
- keep authentication and terminal failures visible during backoff without stranding the retained bridge handle
- preserve local-room behavior and scope the hosted desktop connection copy to the complete notebook path

## Why

Desktop previously reported only app-to-daemon health. A hosted notebook could therefore show Connected while the daemon's outbound room bridge was connecting, reconnecting, or failing authentication.

The bridge is the daemon's observed source of truth for that second hop. Publishing its status as connection-local control state keeps lifecycle signals separate from output transport and avoids duplicating state into notebook documents.

Fixes #3599.

## Validation

- `cargo xtask lint --fix`
- `cargo test -p runtimed hosted_bridge` (13 passed)
- `cargo test -p notebook-wire -p notebook-protocol` (91 passed)
- `cargo test -p notebook-cloud-transport` (33 passed)
- `pnpm test:run` (2,784 passed, 6 skipped)
- `pnpm --dir packages/runtimed typecheck`
- `pnpm --dir apps/notebook exec tsc -b --pretty false`

### DIFF
--- a/apps/notebook/src/App.tsx
+++ b/apps/notebook/src/App.tsx
@@ -379,6 +379,7 @@ function AppContent() {
     getHandle,
     getEngine,
     sessionStatus$,
+    hostedBridgeStatus$,
     notebookDocChanged$,
     triggerSync,
     localActor,
@@ -1091,8 +1092,13 @@ function AppContent() {
   // daemon restarts, and must not claim "reconnecting" while the governor
   // is latched terminal and nothing is redialing).
   const desktopConnectionStatus = useMemo(
-    () => createDesktopConnectionStatusSource(host.daemonEvents, host.daemon.autoReconnect),
-    [host],
+    () =>
+      createDesktopConnectionStatusSource(
+        host.daemonEvents,
+        host.daemon.autoReconnect,
+        hostedBridgeStatus$,
+      ),
+    [host, hostedBridgeStatus$],
   );
   useEffect(() => () => desktopConnectionStatus.dispose(), [desktopConnectionStatus]);
 
@@ -2028,12 +2034,13 @@ function AppContent() {
             // Connection/identity slot: renders nothing for a purely local
             // session (isRemoteNotebookContext) — conditionality is the
             // point. The source derives from daemon lifecycle events (the
-            // IPC transport's status is constant in practice), and the
-            // copy is scoped to the link it measures.
+            // IPC transport's status is constant in practice). Hosted rooms
+            // compose daemon and bridge health, so their copy names the whole
+            // notebook connection rather than only the first hop.
             <NotebookConnectionIdentity
               capabilities={shellCapabilities}
               connectionStatus$={desktopConnectionStatus}
-              connectionLabel="Daemon connection"
+              connectionLabel={hostedNotebookUrl ? "Notebook connection" : "Daemon connection"}
             />
           }
         />
--- a/apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx
+++ b/apps/notebook/src/components/__tests__/notebook-toolbar.test.tsx
@@ -694,10 +694,10 @@ describe("connection/identity slot wiring", () => {
     // vp test runs from the repo root.
     const appSource = readFileSync(resolve(process.cwd(), "apps/notebook/src/App.tsx"), "utf8");
     expect(appSource).toMatch(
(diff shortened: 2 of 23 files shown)


quillaid:
Bedrock Opus 4.6 review dispositions (session `ses_07974869bffemaZFGyQguMXshl`):

- **Confirmed — fixed:** initialize the app-level bridge status with `not_applicable`, matching the daemon's local-room default and avoiding a stale local Connecting state.
- **Confirmed — fixed:** clarify the governor-latch test and assert the bridge-supplied Reconnecting state before applying the unrelated daemon latch.
- **Confirmed — fixed:** document that `hosted_bridge_status` is a protocol-v4 session-control addition at both send gates.
- **Confirmed — fixed:** attach the app-level bridge subscription before `engine.start()` so a synchronous transport replay cannot be missed.
- **Non-issue, clarified:** a successful cloud connect without a principal violates `CloudWsFrameTransport`'s contract, so `terminal_error` is intentional during backoff. The bridge still retries to avoid retaining a dead handle; an inline comment now records both decisions.
- **Confirmed — fixed where introduced:** new session-control test deliveries now use `FrameType.SESSION_CONTROL` rather than the raw `0x07` byte.

Remediation is in `5ec36e8e`. Post-change checks: 13 focused Rust tests, 171 focused TypeScript tests, and `cargo xtask lint --fix` all pass.

quillaid:
Final Bedrock Opus 4.6 review dispositions (session `ses_079705e23ffe2PNY9anT1M0i7e`):

- **Non-issue — replay race:** the app-level subscription is installed once before `engine.start()` and remains live across relay bootstraps. `ReplaySubject(1)` always replays its latest value, and every v4 peer attach explicitly sends the current room watch snapshot with `borrow_and_update()`, even when the bridge itself did not transition.
- **Non-issue — effect dependency:** `hostedBridgeStatusSubject` is intentionally stable via `useMemo([])` and is correctly listed because the effect closes over it. It cannot orphan a subscription without a corresponding effect cleanup/re-run.
- **Confirmed-defer — source-regex test:** this is the established static boundary-test style in that file and is a test-maintainability nit, not a product defect. The updated assertion passed both focused and full frontend suites; replacing the surrounding test architecture is outside this PR.
- **Non-issue — protocol negotiation:** the shared client/server constant is already `PROTOCOL_VERSION: u8 = 4` in `crates/notebook-protocol/src/connection/handshake.rs`; the desktop relay also sends version 4. The new server gates use that already-negotiated version.
- **Non-issue — clean EOF terminal branch:** `run_bridge` takes the concrete `CloudWsFrameTransport`, whose `clean_eof_is_recoverable()` unconditionally returns `true`. Cloud clean EOF therefore takes the Reconnecting branch; the false branch cannot classify a cloud rolling restart as terminal.
- **Non-issue — event payload name:** `status` is already the flat `SessionStatus` field on `FrameEvent`. `hosted_bridge_status` keeps the new discriminated event unambiguous without widening `status` to an unrelated union.

No further code changes are warranted from the final pass. The PR remains draft.

PR #4089, opened 2026-07-22
docs(runtime): propose CRDT-backed kernel stdin slot

quillaid:
## What changed

- adds an alternative architecture proposal for issue #4069
- models a waiting stdin prompt as a single-assignment RuntimeStateDoc execution slot
- keeps the submitted value on one authenticated, correlated ReplyInput request rather than in CRDT history
- documents prompt-visibility, exact-actor authority, hosted-bridge, cancellation, and terminal-ordering tradeoffs
- compares the simplification directly with draft PR #4084

## Why

The targeted-frame design in #4084 is secure and explicit, but it introduces a new server frame, targeted delivery path, frontend ephemeral store, and reconciliation with execution state. RuntimeStateDoc already provides reliable execution-scoped projection.

This proposal tests whether nteract can use that existing document for prompt lifecycle while preserving a narrow ephemeral path for the secret-bearing reply. The two proposals are alternatives and should not both become protocol contracts.

## Impact

Design-only. It does not enable kernel stdin or change the protocol. The proposal keeps stdin default-off and calls out the decision test that can reject this shape: if prompt text itself must be actor-confidential, #4084's targeted-frame design is the correct boundary.

## Validation

- cargo xtask lint --fix
- git diff --check
- source-path and documentation-index review

## Review plan

A Bedrock Claude architecture review will be posted after the draft PR exists.


quillaid:
Opus 4.8 review loop complete.

Initial findings and dispositions:
- Prompt-text visibility and permanent CRDT retention needed explicit product/security sign-off: confirmed-fix in bba9d34f.
- Input-before-terminal ordering was phrased too much like an inherited guarantee: confirmed-fix; it is now an implementation requirement and decision test.
- Consume-once relied on an unstated single-runtime-peer assumption: confirmed-fix; ownership transfer/restart cancels rather than adopts a waiting slot.
- README could make a competing proposal look adopted: confirmed-fix; the entry is marked proposal.
- Referenced source paths: locally verified.
- 2026-07-22 date: non-issue; it is the actual authoring date.

Final US Bedrock Claude Opus 4.8 review: clear, no actionable findings.
Session: ses_07811752affeoxyGEFEX3ieevH

rgbkrk:
Closing the alternative stdin design draft during release cleanup. #4084 and #4089 intentionally propose mutually exclusive protocol contracts; keeping both open without an active decision obscures the actionable queue. The design remains in git history and can be revived from issue #4069 when stdin work is scheduled.

PR #4094, opened 2026-07-28
feat(notebook): redesign workstations panel layout and simplify pairing steps

ecyee3:
- Reworks the workstations panel into a selectable list (icon, name, status) with a separate collapsible "Workstation details" section below, instead of one dense always-expanded target section.
- Moves "Add workstation" into the rail header, next to the panel title.
- Cuts the "Connect a machine" card down to the two steps actually required to connect (install, pair). PATH export and platform-specific persistence commands move into "Additional setup options."
OLD:

image
NEW:

image
image
image

### DIFF
--- a/apps/notebook-cloud/scripts/hosted-workstation-toolbar-smoke.mjs
+++ b/apps/notebook-cloud/scripts/hosted-workstation-toolbar-smoke.mjs
@@ -456,7 +456,7 @@ async function assertOwnerBlockedWorkstationStates({
   const offlineDefault = await assertOwnerToolbarActionWithMockedWorkstations({
     browser,
     expectedLabel: "Review compute",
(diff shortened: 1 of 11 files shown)


apps/notebook-cloud/scripts/hosted-workstation-toolbar-smoke.mjs
```diff
@@ -456,7 +456,7 @@ async function assertOwnerBlockedWorkstationStates({
   const offlineDefault = await assertOwnerToolbarActionWithMockedWorkstations({
     browser,
     expectedLabel: "Review compute",
-    expectedPanelText: "No heartbeat from this workstation recently.",
+    expectedPanelText: "Offline workstation",
```
rgbkrk, line 459: Thank you for replacing my totally generated text here!

PR #4095, opened 2026-07-28
feat(e2e): opt-in UI recording rig for agent-inspectable video and frames

markmiro:
## Summary

- Adds an opt-in recording rig behind `NTERACT_E2E_RECORD=1`: a 1200x800 viewport with 2x CSS zoom (Playwright's recorder ignores `deviceScaleFactor`, so zoom is the only route to retina density), plus `pauseForVideo()`/`screenshot()` helpers that are no-ops otherwise — so a normal suite run produces no video, no PNGs, and no pacing delay.
- Adds `show-video.mjs` (trim the load prefix via a `video-meta.json` sidecar, re-encode, archive to `recordings/<spec>/<test>-<timestamp>.webm`), `show-frames.mjs`, `test-results.mjs`, an HTML viewer, and the `record-ui` skill documenting the workflow.
- Records `add-cell.spec.ts` adding and running a code cell as the first flow using the rig.
- Fixes `src/components/ui/avatar.tsx`: the root's `overflow-hidden` clipped the absolutely-positioned `AvatarBadge` into a quarter wedge and ate its ring; the circle crop now lives on `AvatarImage`.

## Test plan

- [x] `pnpm test:run` — 2785 passed, 6 skipped
- [x] Avatar fix verified visually in the Elements docs app against a worktree without the fix (wedge before, full ringed circle after); photo avatars still crop circular
- [ ] CI green


markmiro:
Superseded by #4096, which carries only the avatar badge fix. The e2e recording rig work is still on the `markmiro/checkout-run-video-commit` branch and can get its own PR.

PR #4096, opened 2026-07-28
fix(ui): stop clipping the avatar presence badge

markmiro:
`Avatar`'s root had `overflow-hidden` to circle-crop photos, which clipped the absolutely-positioned `AvatarBadge` into a wedge and ate its ring. Moved the crop onto `AvatarImage` via `rounded-full`.

before and after

CI green; photo avatars still crop circular.


rgbkrk:
Nice

PR #4098, opened 2026-07-28
feat(e2e): opt-in UI recording rig for agent-inspectable video and frames

markmiro:
## Summary

A UI change is hard to verify from a test assertion alone, and hard to show in a PR. This adds a recording rig that gets both artifacts from a single spec run — a `.webm` for humans and PR attachments, and `frame-NN-<label>.png` screenshots that an agent can actually `Read` to check its own work.

Everything is gated on `NTERACT_E2E_RECORD=1`. Without it, `pauseForVideo()` and `screenshot()` are no-ops and no video viewport is configured, so a spec carrying recording marks runs as a plain test in CI: no video, no PNGs, no pacing delay. Specs stay free of viewport/video config — that lives in `playwright.config.ts` and `helpers.ts`.

Two non-obvious details drove the design:

- **Retina.** Playwright's recorder captures at CSS-pixel size and ignores `deviceScaleFactor`, so a compact window yields a blurry low-res video. The rig records at 1200x800 and CSS-zooms the page 2x, so the framing reads as 600x400 at full density.
- **Load prefix.** Recording starts at context creation with no pause/resume. `markKernelReady()` waits for `data-kernel-status="idle"`, then stamps the elapsed time into a `video-meta.json` sidecar; `show-video.mjs` seeks past that offset so the clip opens on the first real interaction instead of ~30s of kernel boot.

Helper scripts: `show-video.mjs` (trim, re-encode at 2x, archive to a gitignored `e2e/recordings/<spec>/<test>-<timestamp>.webm` so before/after clips survive Playwright wiping its output dir), `show-frames.mjs` (print the PNG paths for an agent to read), `test-results.mjs` (shared artifact lookup), and an HTML viewer that scales the 2x video down by `devicePixelRatio` — QuickTime shows the raw pixels oversized.

`add-cell.spec.ts` is the first flow on the rig: add a code cell from the toolbar, type into it, run it, wait for output. The `record-ui` skill documents the loop.

Final frame from a recorded run:

add-cell output rendered

## Test plan

- [x] `NTERACT_E2E_RECORD=1` run of `add-cell.spec.ts` produces `video.webm` + 3 labeled frames; `show-video.mjs` trims and archives; viewer opens at correct retina scale
- [x] Default (non-recording) mode verified in CI: the browser E2E job ran all 32 chromium tests — `add-cell.spec.ts:11 › via toolbar` among them — and passed with no video or PNG output
- [x] CI green


rgbkrk (commented):
The shared `@nteract/e2e-record` extraction is coherent and the opt-in behavior keeps normal CI cheap. One reproducibility gap keeps this draft-worthy: `run-cloud-e2e.mjs` reuses any healthy worker already listening at the derived URL, but only a newly spawned worker is guaranteed to have `--no-local-oidc`. The documented local-auth specs depend on that mode, so the same command can pass in a clean environment and fail when a default dev worker is already running.

Please make the runner verify the active worker mode or own/restart an isolated worker for this suite. After that, this looks like a good candidate to promote and merge.

rgbkrk:
Closing during the 2.7.0 release cleanup. The recording rig is useful but large, non-release-critical, and still has the unresolved cloud-worker mode reproducibility issue from review. Please revive the shared recording package after the release with that boundary fixed.

PR #4100, opened 2026-07-28
feat(cloud): raise a connect-compute notice on the notebook stage

markmiro:
A fresh cloud notebook has nowhere to run but only said so inside the collapsed workstations rail. This raises that state onto the stage as an inline notice; its action opens the workstations rail so the reader can register a machine.

notice


rgbkrk:
For "lacking compute" we need to make sure our UI is minimal in the case of "just reading" the notebook. Also if the user doesn't have ownership access then they shouldn't see this banner (because they can't launch compute).

rgbkrk (commented):
The earlier owner/viewer concern is addressed in the current diff: the notice requires `canRegisterWorkstation`, which comes from the owner-only workstation capability, and the pure test proves the notice stays quiet when that capability is false. The state gate is also appropriately narrow (`needs_registration` only).

This now looks like a rebase/refresh candidate rather than a product blocker. Please resolve the current `main` conflict and rerun the focused UI tests/CI; if the refreshed diff remains this small, I would be comfortable promoting it.

rgbkrk:
Taking this over for the release push. I flattened the historical dialog-introduction/revert sequence into the reviewed six-file net change, rebased it onto current `main`, preserved the newer pairing-dialog wiring, and pushed the result at `bf661983`. Focused component tests, cloud typecheck, and the required lint gate pass; fresh CI is running.

PR #4105, opened 2026-07-29
docs(outputs): frame the large cell payload problem

rgbkrk:
Frames the problem behind #4103 and records what is still open. Companion to `memos/arrow-bounded-head-and-progressive-fetch.md`, which bounded how many rows cross the wire and said nothing about one column being nearly all of the payload.

A table cell can be larger than most tables. `moonshotai/PerceptionBench` declares `image` as `List(Value('string'))`, so every cell is a 43 KB `data:image/jpeg;base64,...` string. That rendered as text and hung the app.

## What the memo records

**The diagnosis is a detection gap, not a regression.** `is_image_like_data_type` matched only `Struct{bytes,path}` or a list of that struct, so data-URI strings were never handled. A dataset using the proper `Image` feature always worked.

**Rendering was already lazy; transport was not.** Image bytes are read per visible cell, deliberately bypassing the viewport cache. But they sit inline in the Arrow stream, so every byte crosses kernel to CAS to browser whether or not anyone scrolls there. That split is the whole design space, and each candidate layer fixes a different half.

**The manifest already has the slot for the transport half.** `blobs[]` sits beside `chunks[]` and `coalesced`, and the publish path, hosted snapshot collector, and cloud viewer ref collector all walk it. Per-cell refs would fill an existing seam rather than extend the format. It is undocumented in `adr/blob-ref-and-chunk-manifest-protocol.md`, which is worth fixing on its own.

**No new Python-to-daemon channel is required.** `pending_buffers()` plus the buffer hook already move arbitrary bytes into CAS, `preflight_ref_buffers` hash-verifies and commits them, and ADR Decision 2 already permits a multi-ref envelope. The `nteract.dx.blob` comm target stays reserved and its comment is accurate: the target is unused because bytes ride IOPub buffers instead. A direct channel is still wanted for pull, where the frontend asks the kernel for more rows, but push-side extraction does not block on it.

**Range requests are ruled out.** Neither the desktop daemon nor the cloud blob path serves them, which leaves packed side-car blobs with an offset index as the coalescing shape to measure.

**Text needs a higher threshold than binary.** Sorting and filtering cannot operate on a hash or a preview, so a blobbed text column needs explicit materialization semantics. Roughly 64 KiB for binary against 256 KiB for text.

## Priorities recorded

What the reader is owed is a table they can work with, and once images render they will expect to see them. That ordering makes truncation and substitution correct rather than a compromise, including for an agent reading the text projection, where a marker beats 43 KB of base64 nobody will read.

## Open questions

Fetch amplification gates the entire transport half and wants measurement rather than reasoning: the table is already batched into a handful of chunk blobs, so per-cell refs could trade a payload problem for a request problem.

Also open: where the thresholds sit, materialization semantics for sorting blobbed text, whether extraction is automatic or opt-in, GC reachability for per-cell blobs, and whether Sift's WASM work belongs in a worker inside the isolated frame. On that last one, the frame omits `allow-same-origin` for security, which says nothing about scheduling, so a large table can still block the thread that renders it.

A longer feasibility study on the blob-ref design lives in `.context/` on the `feat/large-cell-blob-refs` branch. It stays out of the repo per the documentation taxonomy; its conclusions are folded in here.


rgbkrk:
Closing to keep the upcoming stable release focused; the large-cell transport research remains recoverable from this branch for a later output-focused effort.

PR #4111, opened 2026-07-30
feat(notebook-cloud): add settings drawer for theme and mode

ecyee3:
Adds a settings drawer to the cloud notebook viewer so users can switch light/dark/system mode and color theme (classic/cream).
- New NotebookSettingsDrawer component, wired into the account menu
- New useColorThemePreference hook to persist the color theme choice
- Cloud viewer bootstraps the saved theme on load (viewer-theme-bootstrap.ts)
- Elements docs/example page for the new drawer
- Tests for the drawer and viewer theme bootstrap

https://github.com/user-attachments/assets/889da6aa-1a67-43f8-a79d-648e7771b215

### DIFF
--- a/apps/elements/app/page.tsx
+++ b/apps/elements/app/page.tsx
@@ -20,6 +20,7 @@ import {
   Rows3,
   Search,
   Server,
+  Settings,
   ShieldCheck,
   SlidersHorizontal,
   SquareCode,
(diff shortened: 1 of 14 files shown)


markmiro:
I don't think we need Account and sign out there, but otherwise it's good!

rgbkrk:
I love it. I do actually want sign out somewhere in there.

PR #4115, opened 2026-07-31
feat(runt): warn at workstation startup when ipykernel is missing

markmiro:
## The problem

Screenshot 2026-07-31 at 7 29 19 AM

Started from this screenshot. End of desc explains more about why it doesn't fix the banner (would be a bigger change), but does help.

---

The workstation `current_python` policy (`crates/runtimed/src/workstation/launch_on_attach.rs`) launches kernels against the machine's interpreter **as-is**, with package management disabled. There is no env-prepare step that could add `ipykernel`.

So if that interpreter lacks `ipykernel`, every launch dies with a raw `runpy.py` / `ModuleNotFoundError` traceback — and it surfaces in a **browser**, potentially on another continent from the operator who could fix it with one `pip install`. Worse, it is not classified as a typed `missing_ipykernel` reason: `kernel_env::diagnose_ipykernel` only runs for the `uv:inline` / `conda:inline` / `uv:pep723` arms, not `uv:current_python`.

Before this change `runt workstation run` printed the interpreter path only when you passed `--python-path`, and never checked anything.

## The change

Probe at `runt workstation run` startup — the one moment an operator is sitting at a terminal — and print the interpreter plus the exact remediation:

```
Python interpreter: /usr/bin/python3
Warning: ipykernel is not installed in /usr/bin/python3; kernels cannot launch.
         Install it with: /usr/bin/python3 -m pip install ipykernel
```

**It warns; it does not refuse.** Exiting non-zero would keep the workstation from ever appearing in the panel, trading a fixable error message for a mystery absence.

### The one structural change, and why it was forced

The preflight must probe the interpreter the daemon will *actually* use, or it is worse than nothing — a check that reports "fine" while kernels keep failing.

`workstation_agent_args` omits `--python-path` when the operator didn't pass one, and the daemon then falls back to `resolve_python_on_path` (`crates/runtimed/src/main.rs`). That function lived inside the `runtimed` crate, and `runt` does not depend on `runtimed` (only `runtimed-client` / `runtimed-service`), so the CLI could not call it.

Both crates already depend on `kernel-env`, so the resolver moved there:

```mermaid
flowchart LR
  subgraph runtcrate[crate runt - CLI binary]
    RUN[workstation_cli.rs<br/>NEW preflight]
  end
  subgraph runtimedcrate[crate runtimed - daemon binary]
    MAIN[main.rs<br/>workstation-agent]
    MOD[workstation/mod.rs<br/>re-export retargeted]
  end
  subgraph shared[crate kernel-env - shared lib]
    RESOLVE[resolve_python_on_path<br/>MOVED here]
    DIAG[diagnose_ipykernel<br/>already existed]
  end
  RUN -->|spawns as subprocess| MAIN
  MAIN --> MOD
  MOD --> RESOLVE
  RUN -->|same resolver| RESOLVE
  RUN -->|same probe| DIAG
  MAIN --> DIAG
```

That makes "CLI and daemon pick the same Python" true by construction rather than by coincidence. The alternative — duplicating the ~25-line PATH search in `runt` — gets a one-file diff at the cost of two copies that drift into exactly the silent failure above.

## Reading the diff

| File | Δ | Role |
|---|---|---|
| `crates/runt/src/workstation_cli.rs` | +130/−2 | All new logic. ~60 lines of that is two tests. |
| `crates/kernel-env/src/lib.rs` | +67 | Destination of the move (function + `is_executable_file` + its test). |
| `crates/runtimed/src/workstation/agent_loop.rs` | −59 | Source of the move. Deletion only. |
| `crates/runtimed/src/workstation/mod.rs` | +2/−2 | `pub use` retargeted at `kernel_env`. |
| `docs/runbooks/remote-workstation.md` | +9 | The runbook never stated the `ipykernel` requirement. |

Net new code across the three refactor files is **+7 lines** (a doc comment and `std::path::` qualifiers); the function bodies diff clean.

`mod.rs` is the only file with blast radius, and it is a 4-line diff: retargeting the re-export is why `crates/runtimed/src/main.rs` still says `runtimed::workstation::resolve_python_on_path(...)` and compiles untouched.

`ipykernel_preflight_warning` is a pure function from diagnostic to `Option<String>` — no I/O, so all four arms are unit-testable: present, missing, site-packages mismatch (ABI-mismatched sibling), and probe failure. `first_line_truncated` exists because pointing `--python-path` at a non-Python executable makes it echo the entire probe script to stderr.

## Test plan

- [x] `cargo test -p kernel-env --lib` — 95 pass, including the relocated resolver test
- [x] `cargo test -p runt --bins` — 30 pass, including the two new preflight tests
- [x] `cargo check -p runtimed` — proves every existing caller still resolves through the re-export
- [x] Real binary driven through each path: venv without `ipykernel` → warns and continues; after `pip install ipykernel` → silent (no false positive); no `--python-path` → resolves and reports the PATH interpreter; `--python-path /bin/ls` → one-line probe-failure warning; empty `PATH` → no-interpreter warning

## Known limits

This narrows the window; it does not close it. The probe runs **once at startup**, so a later `pip uninstall`, a rebuilt venv, or a system update changing which `python3` wins on `PATH` all leave the agent running past a check that is no longer true. And the runbook's recommended happy path (`workstation service install --start`) sends this warning to the systemd journal, where nobody looks.

The complete fix is a daemon-side check in the `uv:current_python` launch arm emitting a typed reason the browser can render with remediation. That is deliberately **not** in this PR — it is a separate, larger change to the launch path. The screenshot that started this investigation would still look the same after this change; only the operator's terminal improves.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/crates/kernel-env/src/lib.rs
+++ b/crates/kernel-env/src/lib.rs
@@ -68,6 +68,46 @@ pub fn strip_base(installed: &[String], base: &[&str]) -> Vec<String> {
         .collect()
 }
 
+/// Resolve the default Python interpreter: first `python3`, then `python`,
+/// searched across `path_var` (the `PATH` environment value).
+///
+/// Shared so the workstation CLI can probe the same interpreter the
+/// `workstation-agent` will pick when `--python-path` is omitted.
+pub fn resolve_python_on_path(path_var: Option<&str>) -> Option<std::path::PathBuf> {
+    let path_var = path_var?;
+    let names: &[&str] = if cfg!(windows) {
+        &["python3.exe", "python.exe", "python3", "python"]
+    } else {
+        &["python3", "python"]
+    };
+    for name in names {
+        for dir in std::env::split_paths(path_var) {
+            if dir.as_os_str().is_empty() {
+                continue;
+            }
+            let candidate = dir.join(name);
+            if is_executable_file(&candidate) {
+                return Some(candidate);
+            }
+        }
+    }
+    None
+}
+
+#[cfg(unix)]
+fn is_executable_file(path: &std::path::Path) -> bool {
+    use std::os::unix::fs::PermissionsExt;
+    match std::fs::metadata(path) {
+        Ok(meta) => meta.is_file() && meta.permissions().mode() & 0o111 != 0,
+        Err(_) => false,
+    }
+}
+
+#[cfg(not(unix))]
+fn is_executable_file(path: &std::path::Path) -> bool {
+    path.is_file()
+}
+
 /// Diagnostic result for checking whether a prepared Python environment can
 /// import `ipykernel`.
 #[cfg(feature = "runtime")]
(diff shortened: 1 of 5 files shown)


markmiro:
The biggest change is move of the resolver to `kernel-env` so `runt`  and `runtimed` both have access to it. I think that's safe but not sure if it makes sense or not

rgbkrk (approved):
Reviewed the exact live head. The resolver belongs in kernel-env so the CLI and daemon share one Python selection rule. Current CI is green; exact-head local verification passed: cargo test -p kernel-env --lib (95) and cargo test -p runt --bins (30).

PR #4116, opened 2026-07-31
chore(scripts): add a Linux container that pairs with the local cloud worker

markmiro:
> **Stacked on #4115.** Base is `markmiro/ipykernel-fix-simple`, so the diff here is one new file. Retarget to `main` once #4115 lands.

## Why

Iterating on the *populated* Workstations panel — registered workstation rows, pairing cards, live status — needs a real agent registered against local D1. Elements fixtures only fake the appearance. This provisions a throwaway Linux container that can run the panel's "Connect a machine" commands **verbatim**, which is also how the `ipykernel` bug in #4115 was originally reproduced.

## Usage

```bash
pnpm --dir apps/notebook-cloud dev            # must be running first
scripts/local-workstation-container.sh up     # ~3 min cold, idempotent
scripts/local-workstation-container.sh shell
# then paste the panel's connect + run commands as-is
```

Also `status`, `logs`, `down`. Container name and bridge port are derived from the worktree's wrangler port offset, so parallel worktrees don't collide.

## What the script encodes

Most of its length is working around constraints that each cost real debugging time:

- **The worker binds loopback only**, so a container cannot reach `127.0.0.1:<wrangler>` at all. Two `socat` hops: host bridge on `0.0.0.0:47000+offset` → host loopback, then an in-container listener on *the same port the panel prints* → `host.docker.internal`. That second hop is the point — it's what lets copied commands run unedited instead of needing a hand-rewritten host:port.
- **Releases are linux-x64 only** (there is no `runt-linux-arm64` asset), so Apple Silicon needs `--platform linux/amd64`, which needs colima with `--vm-type vz --vz-rosetta`. The x64 `runt`/`runtimed` binaries do run under Rosetta.
- **No init system**, so `runt daemon start` and `runt workstation service install` both fail. The script runs `runtimed run` detached instead and tells you to use foreground `runt workstation run`.
- **`ubuntu:22.04` ships no `python3`** — and installing it isn't enough, `ipykernel` has to go into that same interpreter. `ensure_ipykernel` does it, for the same reason #4115 exists: the `current_python` policy launches against the interpreter as-is with package management disabled.
- **`PATH` goes in `/etc/profile.d/`, not `~/.bashrc`** — `bash -l` sources the former only, so `docker exec -it … bash -l` otherwise gets `runt: not found`.
- **Two probe traps**, both commented at the call site: `pgrep -f 'TCP-LISTEN:…'` matches the `bash -lc` wrapper doing the probing (so it probes with `curl`), and `runt daemon status | grep -q` exits early, SIGPIPEs `runt`, and reads as a failure under `set -o pipefail` (so it captures to a variable and pattern-matches).
- The host `socat` is fully detached with redirected stdio, because an inherited stdout keeps `… | tail` pipelines from ever closing.

## Test plan

- [x] `up` from cold on Apple Silicon under colima: bridge, container, apt deps, headless install, daemon
- [x] `up` again while running — idempotent, no duplicate bridges or containers
- [x] `runt workstation connect` + `runt workstation run` pasted from the panel unedited; workstation appears registered in the panel
- [x] `status` / `down` / re-`up` cycle
- [ ] Not exercised on Intel macOS or Linux hosts, or with Docker Desktop instead of colima

## Scope

Dev tooling only — one new file under `scripts/`, nothing shipped, no CI hookup, not referenced by any build path. The `Ready.` banner at the end of `up` restates the container-specific deviations (skip the installer step, no systemd) so the panel's instructions aren't followed off a cliff.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/scripts/local-workstation-container.sh
+++ b/scripts/local-workstation-container.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+# Linux container that can pair with this worktree's local notebook-cloud worker,
+# so the "Connect a machine" commands from the Workstations panel can be pasted
+# in verbatim.
+#
+#   scripts/local-workstation-container.sh up      # bridge + container, then exit
+#   scripts/local-workstation-container.sh shell   # interactive shell in it
+#   scripts/local-workstation-container.sh status
+#   scripts/local-workstation-container.sh down
+#
+# Requires `pnpm --dir apps/notebook-cloud dev` already running.
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+IMAGE="${NTERACT_WORKSTATION_IMAGE:-ubuntu:22.04}"
+PLATFORM="${NTERACT_WORKSTATION_PLATFORM:-linux/amd64}"
+
+read -r WRANGLER_PORT PORT_OFFSET <<<"$(
+  cd "$REPO_ROOT" && node -e '
+import("./apps/notebook-cloud/scripts/local-dev.mjs").then((m) => {
+  const ports = m.notebookCloudDevPorts();
+  process.stdout.write(
+    `${ports.port} ${ports.port - m.WRANGLER_HTTP_PORT_BASE}\n`,
+  );
+});
+'
+)"
+# Above the wrangler http (45000+) and inspector (46000+) ranges.
+BRIDGE_PORT=$((47000 + PORT_OFFSET))
+CONTAINER="nteract-workstation-${PORT_OFFSET}"
+BRIDGE_PIDFILE="${TMPDIR:-/tmp}/${CONTAINER}-bridge.pid"
+
+log() { printf '\033[36m==>\033[0m %s\n' "$*"; }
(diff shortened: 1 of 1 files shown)


rgbkrk (commented):
The one-file dev-tool scope is attractive, and syntax/ShellCheck found no substantive shell error. Two behavior/safety details should be fixed before the draft is promoted:

1. `start_daemon` only logs when its 30-second readiness loop expires, then returns success and the script prints `Ready.`. A failed daemon start must exit nonzero so the final banner cannot overclaim readiness.
2. The host bridge binds `0.0.0.0`, exposing the local cloud worker on every host interface, not only to the container VM. Please either bind to the narrow VM-reachable interface or make the exposure explicit and guard it. While touching this, validate that a stale PID file still belongs to the expected `socat` bridge before treating it as live or killing it.

rgbkrk:
Closing during the 2.7.0 release cleanup. This is dev-only tooling and still has unresolved readiness-reporting, host-interface exposure, and stale-PID safety concerns. The approach can return after the release as a fresh, narrowly reviewed script.

PR #4118, opened 2026-07-31
docs(runtimed-node): record the event model and what it actually emits

rgbkrk:
`@runtimed/node` documents `runCell` as its front door: submit a cell, await a `CellResult`. That fits running a few cells from a script. It is the wrong shape for an editor, which wants to stream outputs into a cell as they arrive rather than await a terminal result.

The bindings already carry the frontend-shaped surface for that. `executionTransitions$`, `executionViewChanges$`, `sessionStatus$`, and `getExecutionView()` are all on `Session` today, as RxJS observables, with `ExecutionView` as manifested state and `ExecutionViewChangeset` as the delta. None of it is documented.

Measured against a live dev daemon, two thirds of it does not behave as typed.

## What was measured

**`executionTransitions$` delivers, in the wrong envelope.** Both transitions arrive in order with the right execution id and count:

```
["{\"execution_id\":\"85e2e83b-...\",\"kind\":\"started\",\"execution_count\":1}"]
["{\"execution_id\":\"85e2e83b-...\",\"kind\":\"done\",\"execution_count\":1}"]
```

Each emission is a single-element array holding a JSON string, not the declared `ExecutionTransition`. So a consumer following `index.d.ts` writes `if (t.kind === "done")` and it is `undefined`, silently, forever. The memo records the working form.

`sessionStatus$` shares the envelope and additionally reports a `connection` field the type omits, with PascalCase values where the type declares lowercase.

**`executionViewChanges$` fires empty.** Changesets arrive throughout an execution with `execution_upserts` empty every time and `queue` never populated. The stream says something changed without saying what.

**`getExecutionView()` was empty throughout.** Sampled before submitting, mid-execution, immediately after the promise resolved, and a second later. Empty at all four points for an execution that completed successfully. Recorded as observed rather than diagnosed, since a precondition may be unmet.

## Why it matters together

The two halves fail in opposite directions. The promise API is documented and fabricates an empty result on timeout:

```rust
fn timeout_cell_result(cell_id: String, execution_id: String) -> CellResult {
    CellResult { cell_id, execution_id, execution_count: None,
                 status: "timeout".to_string(), success: false, outputs: vec![] }
}
```

`outputs: vec![]` is hardcoded, and the default timer is 120 seconds. Every other path builds from an `ExecutionState` read out of the document; this one does not. So a caller cannot distinguish "produced nothing" from "produced output but I stopped waiting" from "kernel never started."

The event API is honest, undocumented, and mistyped.

## Scope

Documentation only. No code changes, so nothing here fixes the envelope, the empty projection, or the fabricated timeout result. The memo is written so an integrator can use the event API today with the caveat spelled out, and so the three defects are recorded with what was actually observed rather than inferred.

The timeout fix is the one with a clear answer and no dependencies: read the document's current state and mark it timed out, instead of fabricating an empty result. Worth doing on its own.


rgbkrk:
Closing as stale documentation during the 2.7.0 release cleanup. This memo records broken `@runtimed/node` behavior from before the current repair work; #4138 now fixes the event envelope/session status and durable output read path. Any remaining execution-view gap should be documented with the fresh main-based follow-up rather than preserving an obsolete integration guide.

PR #4122, opened 2026-08-04
feat(notebook-cloud): rebuild the sharing popover on shared primitives with aligned rows

markmiro:
Rebuilds the cloud sharing popover on the shared shadcn primitives (`Button`, `Input`, `Select`, `Label`, `Separator`) and deletes the ~320 lines of hand-rolled `viewer/index.css` that used to style it, then extracts `CloudSharingPanel` so the new Elements `sharing-sheet` fixture renders the real markup instead of a copy. Every access section — current access, edit requests, and compute access — now renders through one `CloudShareRow` that owns row geometry and reserves the trailing action column, so buttons, badges, and plain text land on the same right inset rather than each row positioning its own. Icon sizing and `aria-hidden` come from the `Button` variant and lucide defaults, and the Enable/Disable and Copy link controls became outline buttons so they read as buttons instead of ghost text; `viewer-render-props.test.ts` asserts these invariants against the source so they can't silently drift back.

Before/after on the default owner state — the earlier pass left oversized icons, loose header padding, and section headers in a different type style:

Before and after

Nine fixture states are reviewable in Elements at `/docs/sharing-sheet`:

Elements sharing-sheet fixture states

`cargo xtask lint --fix` and the `notebook-cloud` typecheck are clean, and CI is green. One test in the file this PR touches, `cloud callback keeps sign-in handoff in the entry surface language`, still fails locally — it also fails on `origin/main`, asserting OIDC callback copy that changed upstream, unrelated to sharing.


rgbkrk (commented):
The extraction to a production-backed `CloudSharingPanel` is a good direction, and I approved the small stacked reset in #4123. Two concrete items keep this parent draft-worthy:

1. `sharing-panel.tsx` adds notice-style `border-l-2` treatments plus raw `emerald-*`/`amber-*` palette classes. That conflicts with the shared UI law: left rules are for code lines, notices use a full semantic border/tint, and component chroma comes from the repository meaning tokens.
2. The Elements page renders nine `CloudSharingPanel` instances, but every instance emits the same `cloud-share-invite-email` and `cloud-share-invite-scope` IDs. Labels in later fixtures therefore target the first matching control and the catalog has duplicate IDs. Give the panel an ID prefix or use a per-instance `useId()`-derived pair.

Both are contained follow-ups; the underlying sharing behavior and #4123 reset look sound.

PR #4123, opened 2026-08-04
fix(cloud): reset copy-link state on share popover close

markmiro:
## Summary
- Stacked on #4122. `CloudSharingControls` kept `copyState` in React state but never reset it on popover close, so reopening the Share popover showed a stale "Copied link"/"Copy failed" label from the previous session.
- Added an `onOpenChange` handler that resets `copyState` to `"idle"` whenever the popover closes.

## Test plan
- [x] Verified locally via Playwright against the notebook-cloud dev server (rebuilt with #4122's reworked sharing panel): clicked Copy link (label became "Copied link" / "Link copied."), closed the popover, reopened it, and confirmed it now reads "Copy link" with no stale message.

Share popover after reopen

### DIFF
--- a/apps/notebook-cloud/viewer/sharing-controls.tsx
+++ b/apps/notebook-cloud/viewer/sharing-controls.tsx
@@ -332,8 +332,15 @@ export function CloudSharingControls({
     }
   };
 
+  const handleOpenChange = (nextOpen: boolean) => {
+    setOpen(nextOpen);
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
Small, focused fix: closing the controlled popover resets the transient copy result while preserving the existing open/load lifecycle. Current CI is green.

PR #4128, opened 2026-08-06
docs(cloud): explore self-hosted celld deployment

rgbkrk:
## Summary

- capture an exploratory self-hosted nteract topology using celld for live room ownership
- keep kernels on customer-managed BYOC workstations connected as scoped `runtime_peer`s
- compare native celld D1, Postgres behind a private control service, and a disposable all-celld catalog proof
- define company ingress, identity, object storage, backup/restore, deployment profiles, and go/no-go gates

## Why

celld maps closely to the current Worker and Durable Object room model, but its current compatibility surface does not include D1 or R2. This memo records the smallest credible on-prem shape without treating an alpha substrate or an invented `CatalogDb` bridge as a production decision.

## Current recommendation

Use celld for live room objects, Postgres for relational control-plane state, and separate S3-compatible buckets for application artifacts and celld fleet state. Do not add Redis by default. Recheck native celld D1 immediately before implementing the catalog port.

## Evidence

- reviewed the current notebook-cloud D1/R2/DO bindings and storage schema
- compared the current Worker surface with celld v0.1.0 compatibility and security documentation
- ran a local celld bundle dry run; the stock command failed on the WASM import, while an esbuild `.wasm` loader override bundled the Worker and its three existing Durable Object classes
- ran `cargo xtask lint --fix`
- verified local documentation links and `git diff --check`

## Review focus

- whether Postgres plus a private domain API is the right on-prem default while celld D1 is only planned
- whether the state and operational ownership boundaries are complete enough for a customer deployment
- whether the all-celld variant should remain an evaluation-only path

## Follow-up

Stacked draft PR #4129 adds the smallest celld packaging harness and records what bundles, boots, or fails without implementing the Postgres control service yet.


rgbkrk:
Closing this self-hosted celld exploration because it is not part of the upcoming stable release scope; the research remains recoverable from the branch.

PR #4129, opened 2026-08-06
build(cloud): add celld packaging spike

rgbkrk:
## Stack

- Base: #4128
- This PR: reproducible celld packaging and bundle compatibility spike

## Summary

- add a celld-specific Wrangler JSONC manifest for the current Worker, static assets, and three active Durable Object classes
- add a dry-run command that builds the real viewer/WASM artifacts and invokes celld without writing a deployment
- add an explicit esbuild dependency and `.wasm` loader wrapper required by the current celld bundler
- test that the manifest excludes unsupported D1/R2 bindings and does not resurrect deleted Durable Object classes

## Measured result

Using celld v0.1.0 on Apple Silicon, the checked-in command successfully bundled:

- the current notebook-cloud Worker
- 61 static assets with 52 unique bodies
- `NotebookRoom`, `WorkstationEvents`, and `OwnerComputeIndex`
- the `runtimed-wasm` import through the explicit `.wasm` loader

celld reported 6,430.14 KiB raw / 1,870.26 KiB gzip and produced dry-run version `501a778eb6e3ca41`. Nothing was written to object storage.

The first reproducibility run failed because the loader wrapper inherited no directly available `esbuild` binary. This PR now declares `esbuild` explicitly in the cloud app rather than relying on a transitive absolute path.

## Deliberate limits

- no D1 catalog implementation
- no R2/artifact replacement
- no Postgres control service
- no production ingress or identity configuration
- no claim that the Worker has booted on a live celld node yet

Routes requiring D1 or R2 remain unavailable by design. This PR establishes the packaging floor before introducing either backend.

## Local live-node result

A disposable local probe used celld v0.1.0 with MinIO `RELEASE.2025-09-06T17-38-46Z`:

- celld uploaded the immutable Worker version, manifest, asset index, and 52 content-addressed asset bodies
- publishing `deploy/.../current.json` failed twice because celld's create-if-absent conditional write treated MinIO's missing-key response as an ambiguous CAS failure
- after manually seeding the deployment pointer and fleet peer secret for diagnosis only, node startup failed on the same CAS behavior while creating its node lease
- the process exited with `node authority was not established before wake scan` before binding its HTTP listener

No workaround is committed. This result blocks the MinIO-backed live profile at celld's fleet storage layer, before nteract Worker JavaScript executes. The next live probe should use an object store with celld-verified conditional-write semantics or a celld/MinIO compatibility fix.

## Verification

- `node --test apps/notebook-cloud/test/celld-config.test.mjs`
- `CELLD_BIN=/absolute/path/to/celld pnpm --dir apps/notebook-cloud celld:dry-run`
- `cargo xtask lint --fix`
- `git diff --check`

The full `pnpm --dir apps/notebook-cloud test` suite was also run. The new celld tests pass, but the existing suite remains red on failures outside this PR's diff, including stale viewer source-text expectations, workstation attachment fixtures that omit `accelerators: null`, renderer artifact drift, and other current room/sharing assertions. The stacked diff does not modify the corresponding runtime, viewer, or pre-existing test files, so those failures are recorded rather than folded into this packaging spike.

## Next experiment

Repeat the live probe against an object store with celld-verified conditional-write semantics, then probe `/api/health`, static assets, and Durable Object construction. Record each failure before adding the Postgres or artifact adapters.

### DIFF
--- a/apps/notebook-cloud/README.md
+++ b/apps/notebook-cloud/README.md
@@ -38,6 +38,26 @@ the gitignored `runtimed-wasm` bindings before TypeScript reads them. Use
 `pnpm --dir apps/notebook-cloud wasm` when you explicitly want a full
 `runtimed-wasm` rebuild.
 
+### celld packaging spike
+
+The celld manifest is deliberately separate from `wrangler.toml`. It declares
+the current Worker, static assets, and three active Durable Object classes, but
+omits D1 and R2 because celld does not implement those bindings yet. This is a
+bundle compatibility check, not a functional self-hosted deployment.
+
+Install celld, then run:
+
+```bash
+pnpm --dir apps/notebook-cloud celld:dry-run
+```
+
+Set `CELLD_BIN=/absolute/path/to/celld` to test a downloaded binary that is not
+on `PATH`. The command builds the current viewer and WASM artifacts, adds the
+required esbuild `.wasm` loader, and invokes `celld deploy --dry-run`; it does
+not write a deployment to object storage. A live proof still needs an
+S3-compatible fleet bucket, a celld node, and replacements for the D1 catalog
+and R2 artifact boundaries.
+
 With Wrangler running:
 
 ```bash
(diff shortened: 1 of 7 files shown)


rgbkrk:
Closing this celld packaging spike with its parent exploration because it is outside the upcoming stable release scope.

PR #4130, opened 2026-08-07
feat(ui): Updates to page layout

ecyee3:
Created a side rail with the home button and avatar on it. Moved the header to show above the notebook area. Moved the notebook utilities above the notebook to the right of the notebook panel. 

OLD:
image

NEW:
image

### DIFF
--- a/apps/notebook-cloud/viewer/cloud-notebook-title.tsx
+++ b/apps/notebook-cloud/viewer/cloud-notebook-title.tsx
@@ -1,4 +1,3 @@
-import { House } from "lucide-react";
 import { DocumentTitle } from "@/components/notebook/DocumentTitle";
 import {
   cloudNotebookRouteTitleFromPathname,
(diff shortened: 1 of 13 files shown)


rgbkrk (commented):
The layout direction is coherent and current CI is green. I found two follow-up items before I would call the result architecture-clean:

1. `docs/adr/notebook-host-shell-convergence.md` explicitly keeps application-level notebook catalogs/home navigation outside `NotebookDocumentShell` and its document rail. This PR puts the `/n` home action and account/connection identity into `NotebookDocumentRail` through `leadingSlot`/`trailingSlot`. Either keep that chrome in a host/app-shell layer or make the decision change explicit; otherwise the shared document rail silently becomes application chrome.
2. The new `NotebookRailHomeButton` and changed `RailButton` use `rounded-xl`; the current shared-UI design law requires `rounded-md` for controls. The new home button also extends the existing noncanonical `ring-2 ring-primary/*` focus treatment instead of `ring-1 ring-ring`.

The radius/focus cleanup is easy to carry as a follow-up. The app-chrome/document-rail boundary needs an intentional call rather than being accidental.

rgbkrk (approved):
Approved with the documented architecture and design-law cleanup carried as follow-up work; the current head is clean, mergeable, and green.

PR #4132, opened 2026-08-10
feat(build): publish host-neutral notebook web artifact

rgbkrk:
## Summary

- package the existing production notebook Vite app as a host-neutral `nteract-notebook-web-<version>.tar.gz` release asset
- include runtime WASM, content-hashed frontend assets, a revision/compatibility manifest, payload checksums, and an adjacent archive checksum
- add `cargo xtask notebook-web`, package-level tests, build artifact upload, and release workflow wiring
- document the boundary between the host-owned app origin/relay and daemon-owned kernels plus isolated renderer plugins

## Why

External desktop hosts need a supported way to embed the notebook workspace without copying `apps/notebook/dist` by hand or coupling the UI to Electron. A separately versioned archive preserves the existing browser transport, lets hosts supply their own origin and CSP, and gives them provenance and integrity data with which to reject daemon/UI source-revision skew.

This intentionally keeps the complete notebook app out of the `runtimed` binary. `runtimed` continues to own execution and its embedded isolated-output renderers; the embedding host owns delivery of the larger workspace UI.

## Validation

- `cargo xtask lint --fix`
- `cargo test -p xtask` (45 passed)
- `node --test scripts/package-notebook-web.test.ts` (2 passed)
- `cargo xtask notebook-web` from a clean build (142 files, runtime WASM included)
- `cargo xtask notebook-web --skip-build`
- `actionlint .github/workflows/build.yml .github/workflows/release-common.yml`
- packaged-host integration smoke with the matching `runtimed 2.6.3+2d943d5`: edited and executed a synced cell and observed `5050`

## Release scope

This PR wires the artifact into existing release workflows but does not itself publish a release, sign an archive, or decide installer ownership. Those remain explicit operational gates for downstream adopters.

PR #4135, opened 2026-08-11
feat(notebook): use daemon output document in browser host

rgbkrk:
## Summary

- expose an optional live `NotebookHost.outputDocumentUrl` from the browser host
- derive the URL from the relay's current daemon blob port
- pass the host value through `NotebookView.outputHostContext`
- retain the existing `srcDoc` fallback until a blob port is available

## Motivation

Browser embedding hosts may serve the notebook application under a restrictive parent Content Security Policy. The `srcdoc` output-frame fallback inherits that policy, which can block the isolated frame's inline bootstrap and dynamic renderer-plugin installation.

runtimed already serves the canonical sandboxed output shell at `http://127.0.0.1:<blob_port>/output-frame` with output-specific CSP headers. Reusing that document keeps renderer execution off the notebook application origin and avoids packaging a duplicate shell.

The iframe sandbox remains unchanged and continues to omit `allow-same-origin`. Embedding hosts remain responsible for allowing the daemon output-frame URL in their `frame-src` policy.

## Verification

- focused notebook-host and output-context tests
- notebook and notebook-cloud TypeScript checks
- runtimed `/output-frame` route test
- `cargo xtask lint --fix`

## Stack

This draft is stacked on #4132 so the host-neutral artifact memo documents the complete browser embedding contract. It can be rebased onto `main` after that PR lands.

### DIFF
--- a/apps/notebook/src/AGENTS.md
+++ b/apps/notebook/src/AGENTS.md
@@ -31,6 +31,7 @@ Host-platform side effects (Tauri IPC, plugin calls, window chrome) flow through
 | `host.daemon` | `isConnected`, `reconnect`, `getInfo`, `getReadyInfo`, optional `autoReconnect` (backoff + terminal failed-load latch) |
 | `host.daemonEvents` | `onReady` / `onProgress` / `onDisconnected` / `onUnavailable` |
 | `host.relay` | `notifySyncReady()` outbound signal |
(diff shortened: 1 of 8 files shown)


rgbkrk:
Closing with its parent stack, #4132, to keep the 2.7.0 stable-release scope focused. The daemon-served output document remains a useful browser-host integration direction; it should return as a fresh PR after the host-neutral artifact boundary is rebased and agreed.

PR #4136, opened 2026-08-11
feat(node): publish native relay release assets

rgbkrk:
## Summary

- publish deterministic `@runtimed/node` wrapper and native package archives with each nteract release
- cover macOS arm64, macOS x64, Linux x64 glibc, and Windows x64 MSVC
- add a machine-readable asset manifest carrying release provenance, npm package version, Node-API minimum, and target/package mapping
- add the missing macOS x64 npm package surface and release/publish matrix coverage
- load every platform archive on its native runner in both PR and release workflows

## Motivation

Embedding hosts often acquire the notebook web bundle and daemon from one immutable nteract release rather than installing npm packages dynamically at runtime. The native relay should be available through the same release boundary so a host can select and verify a wrapper/native asset set without compiling Rust or reaching a package registry during packaging.

## Release contract

Each release adds:

- `runtimed-node-wrapper-${releaseVersion}.tgz`
- `runtimed-node-darwin-arm64-${releaseVersion}.tgz`
- `runtimed-node-darwin-x64-${releaseVersion}.tgz`
- `runtimed-node-linux-x64-gnu-${releaseVersion}.tgz`
- `runtimed-node-win32-x64-msvc-${releaseVersion}.tgz`
- `runtimed-node-assets-${releaseVersion}.json`

The GitHub Releases API `sha256:` digest remains the archive authenticity source. The manifest describes provenance, Node-API compatibility, and package placement; it intentionally does not duplicate archive checksums. A host installs the wrapper plus exactly one platform package and verifies Node-API support before loading.

`source_revision` records the release commit for audit and reproduction. It is not a runtime admission check and hosts must not require it to equal the daemon or notebook web revision. Runtime compatibility is negotiated by protocol number and versioned semantic capabilities on the relay connection.

Directly extracting the wrapper and native archive self-contains the `@runtimed/node/relay` surface. The higher-level package root still requires its declared runtime dependencies, including `rxjs`, supplied by a package manager or embedding host.

The workflows assert the actual Node architecture and Rust host tuple before building each native target, so runner-label drift fails explicitly rather than publishing a mislabeled archive. Each extracted archive pair must load the public relay API and exercise native socket discovery before release. The PR matrix is conservatively triggered by any local Rust crate change, and release build jobs run with read-only repository permissions.

## Verification

- `cargo xtask lint --fix`
- 45 xtask tests
- 9 release-asset contract/reproducibility tests
- frozen lockfile install
- deterministic wrapper packing
- isolated wrapper + native `/relay` extraction and load on macOS arm64
- real wrapper and arm64 platform package dry-runs
- deterministic archive rename and manifest generation checks
- workflow lint

The four-platform native package/load matrix is a required check on this draft's pushed head; signed/notarized application smoke remains outside this package PR.

## Stack

- base: #4134

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,7 @@ jobs:
       wasm_changed: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || steps.filter.outputs.wasm == 'true' }}
       python_changed: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || steps.filter.outputs.python == 'true' }}
       node_changed: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || steps.filter.outputs.node == 'true' }}
+      node_release_changed: ${{ steps.filter.outputs.node_release == 'true' }}
       windows_changed: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || steps.filter.outputs.windows == 'true' }}
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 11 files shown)


rgbkrk:
Closing because the native Node packages already ship through npm; a separate GitHub Release archive channel is not required for the current relay integration.

PR #4139, opened 2026-08-11
build(web): add artifact compliance metadata

rgbkrk:
## Summary

- package the root BSD license, deterministic third-party notices, and an SPDX 2.3 document with the Notebook Web archive
- inventory a deterministic union of the `notebook-ui` runtime closure, shared root-source dependencies, embedded-renderer dependencies, recursively reached workspace packages, explicit CSS build inputs, and the non-dev `runtimed-wasm` Cargo closure
- cover every regular payload file in SPDX, including HTML, CSS, fonts, JavaScript, WASM, notices, and Vite's renderer provenance metadata; include KaTeX font OFL-1.1 terms and Tailwind CSS build provenance
- emit a deterministic Vite build record that binds each opaque renderer source hash to its content-hashed output chunk, then fail packaging when either side is stale
- parse license expressions with the SPDX grammar and fail packaging when syntax, license metadata, license text, required component coverage, renderer provenance, shipped-file provenance, or regular-file constraints are missing or unknown; reject symlinks before copying the release payload
- invoke pnpm through the platform command shim so the same inventory commands work on Windows
- exclude the bundle visualizer's `stats.html` diagnostic from the release payload because it is not used at runtime and contains local build paths

This is stacked on #4132.

## Verification

- `cargo xtask wasm`
- `cargo xtask renderer-plugins`
- `pnpm --dir apps/notebook build`
- `pnpm notebook-web:test`
- `cargo xtask notebook-web --skip-build`
- `cargo xtask lint --fix`
- `uvx --from spdx-tools pyspdxtools --infile target/notebook-web/notebook-web.spdx.json`
- packaged-artifact check: 483 package records, all 144 non-SBOM payload files represented with matching SPDX `CONTAINS` relationships and verification code, all three compliance documents present, `stats.html` absent, no non-regular entries, and no absolute workspace paths in generated metadata

## Follow-up

- Add exact per-chunk npm component attribution emitted by the bundler. The current deterministic union is intentionally conservative for shared root sources and embedded-renderer dependency closures, while workspace packages, renderer source-to-output mappings, and payload file hashes are exact.

### DIFF
--- a/apps/notebook/vite-plugin-isolated-renderer.ts
+++ b/apps/notebook/vite-plugin-isolated-renderer.ts
@@ -17,6 +17,7 @@
  */
 
 import fs from "node:fs";
+import { createHash } from "node:crypto";
 import path from "node:path";
 import tailwindcss from "@tailwindcss/vite";
 import { build, type Plugin } from "vite-plus";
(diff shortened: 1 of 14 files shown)


rgbkrk:
Closing this broad compliance follow-up for release focus; the host-neutral web artifact should carry only the minimum required notices now, with deeper SPDX provenance revisited separately.

PR #4140, opened 2026-08-11
[WIP] feat(node): queue existing synced cells

rgbkrk:
> [!IMPORTANT]
> **WIP programmatic-client exploration.** This convenience API is not required for embedding the notebook frontend. Embedded hosts should pump opaque frames through `@runtimed/node/relay` and let the shared JS/WASM `NotebookClient` own document and execution semantics. This draft remains open while we decide whether its higher-level Node surface should move into or delegate to that shared client.

## Summary

- add `Session.queueExistingCell(cellId, { executionId? })` to `@runtimed/node`
- queue only source already present in the synced NotebookDoc and return the accepted execution handle without waiting for terminal output
- let callers supply a UUID as an idempotency key so a lost response can recover the same active or terminal execution
- reject reuse of an execution UUID for a different cell in both the daemon and hosted WASM room host
- preserve edit cursor/selection presence when execution is queued

## Motivation

Programmatic Node clients that separate execution admission from observation need the execution ID before they can call `waitForExecution`, correlate progress, or make safe ownership decisions. The existing `executeCell` convenience method returns only after terminal output, while sending source again would bypass the live synced notebook document.

The daemon-side operation keeps NotebookDoc as the source of truth. It captures current heads, sends only the cell ID, and uses the daemon's required-head gate before execution is accepted.

Supplying `executionId` also gives callers retry safety:

- same UUID + same cell returns the existing active or terminal execution
- same UUID + different cell is rejected
- a new requested UUID cannot bypass other active work on the cell

The call returns after kernel readiness and daemon acceptance. Cold kernel startup can therefore delay admission; terminal waiting remains a separate operation.

## Compatibility

The wire operation is additive to protocol v4 but must be gated by an advertised semantic capability before independently released clients rely on it. Daemon version/source revision is diagnostic metadata, not the compatibility boundary.

## Verification

- `cargo test -p runtimed-node`: 25 passed
- focused daemon retry/cross-cell tests: passed
- `cargo test -p runtimed-wasm`: 87 passed
- exported WASM Deno tests: 149 passed, 5 ignored
- `@runtimed/node` wrapper tests: 9 passed
- real Node -> native addon -> daemon -> kernel integration: passed
- Tokio mutex lint: passed
- `cargo xtask lint --fix`: passed

## Stack

- base: #4138

### DIFF
--- a/crates/runtimed-node/src/session.rs
+++ b/crates/runtimed-node/src/session.rs
@@ -289,6 +289,14 @@ pub struct QueueCellOptions {
     pub cell_type: Option<String>,
 }
 
+/// Options for `Session.queueExistingCell()`.
+#[napi(object)]
+#[derive(Default)]
+pub struct QueueExistingCellOptions {
+    /// Caller-owned UUID used as an idempotency key for this cell execution.
+    pub execution_id: Option<String>,
+}
+
 /// Options for `Session.waitForExecution()`.
 #[napi(object)]
 #[derive(Default)]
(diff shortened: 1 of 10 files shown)


rgbkrk:
Closing this programmatic Node authoring experiment because embedded hosts should use the shared JS/WASM notebook client over the raw relay.

PR #4141, opened 2026-08-11
[WIP] fix(node): make notebook edit acknowledgements durable

rgbkrk:
> [!IMPORTANT]
> **WIP programmatic-client exploration.** The durable confirmation primitive may be broadly useful, but the higher-level Node authoring API is not part of the notebook embedding path. Embedded hosts should relay bytes and use the shared JS/WASM client. This draft remains open while that ownership boundary and capability negotiation are settled.

## Summary

- make native notebook edit promises resolve only after a correlated daemon response confirms the captured Automerge heads are durable
- add an explicit `ConfirmNotebookHeads` protocol barrier backed by the room recovery journal
- make caller-keyed cell creation atomic and retry-safe, including concurrent conflicting attempts
- preserve the existing best-effort sync confirmation behavior for older call sites

## Motivation

Programmatic clients need to distinguish a completed notebook edit from an edit whose outcome is still indeterminate after a timeout or disconnect. Observing another sync frame is not sufficient: the daemon must confirm the exact captured document heads after its durability boundary.

This keeps retry behavior safe for clients that provide stable cell IDs. Repeating the same cell ID, type, and source is idempotent; trying to reuse the ID for different content is rejected.

## Compatibility

The durable barrier is an additive protocol-v4 operation but must be gated by an advertised semantic capability before independently released clients rely on it. Daemon version/source revision is diagnostic metadata, not the compatibility boundary.

## Verification

- `cargo xtask lint --fix`
- `cargo test -p notebook-protocol -p notebook-sync -p runtimed-node`
- `cargo test -p runtimed --lib` (1,262 passed; 2 ignored)
- `vp test run packages/runtimed-node/tests` (19 passed; 4 skipped)
- rebuilt the native addon and ran the native integration suite
- journal-failure regression proves failed peer changes cannot satisfy the barrier, then verifies replayed changes survive journal recovery

## Stack

- base: #4140


rgbkrk:
Closing this programmatic Node authoring experiment because the high-level mutation surface is not part of the current embedding or stable release boundary.

PR #4142, opened 2026-08-11
[WIP] feat(node): cancel exact executions

rgbkrk:
> [!IMPORTANT]
> **WIP programmatic-client exploration.** Exact cancellation is a daemon-wide semantic, but the convenience method on the high-level Node `Session` is not required for notebook embedding. Embedded hosts should relay bytes and let the shared JS/WASM client own execution state. This draft remains open while that API ownership and capability negotiation are settled.

## Summary

- add exact execution cancellation to the daemon and `Session.cancelExecution(executionId)` to the native `@runtimed/node` authoring API
- cancel queued work without interrupting the kernel, or interrupt only the matching running execution
- preserve unrelated queued work and release the interrupt gate only for the exact execution that was interrupted
- make terminal retries explicit with `already_terminal` and unknown IDs explicit with `not_found`
- durably retire queued work before runtime-agent dispatch so a lagging or restarted agent cannot admit it later
- preserve a completion that races cancellation

## Motivation

Long-lived programmatic clients need to cancel work they own without using the existing notebook-wide interrupt command. A generic interrupt can terminate another caller's running execution and clear unrelated queued work. Caller-owned execution UUIDs provide the identity needed for an exact operation.

## Behavior

`cancelExecution()` returns one of:

- `interrupted` — the named execution was running and its kernel request was interrupted
- `cancelled_queued` — the named execution was removed before it ran
- `already_terminal` — the execution had already completed, errored, or been cancelled
- `not_found` — the notebook has no execution with that ID

Running cancellation preserves successors behind an execution-scoped interrupt gate. Queued cancellation updates RuntimeStateDoc before querying the attached runtime agent. Concurrent terminal outcomes are normalized from the full execution record so cancellation cannot overwrite a real completion.

## Compatibility

The operation is additive to protocol v4 but must be gated by an advertised semantic capability before independently released clients rely on it. Daemon version/source revision is diagnostic metadata, not the compatibility boundary.

## Verification

- `cargo xtask lint --fix`
- notebook protocol suite: 93 tests
- runtimed library suite: 1,271 passed, 2 ignored
- runtimed integration and lint suites
- `runtimed-node` Rust suite: 25 tests
- Node wrapper suite: 10 tests
- native daemon/kernel integration, including running and queued cancellation and successor survival
- concurrent completion-versus-cancellation race coverage

## Stack

- base: #4141


rgbkrk:
Closing this programmatic Node authoring experiment because exact cancellation and its high-level Node API are outside the current embedding and stable release boundary.

PR #4143, opened 2026-08-11
[WIP] feat(runtimed): explore opt-in host-owned daemon instances

rgbkrk:
> [!IMPORTANT]
> **WIP design exploration.** This is not proposed as the default daemon model or as a
> release dependency. The preferred default remains one user-level singleton so nteract,
> embedding hosts, users, and multiple agents can join the same live notebook through one
> daemon. Client compatibility should be governed by negotiated protocol versions and
> capabilities, not by creating a second daemon for each host revision.

## Current direction

- keep the normal stable/nightly singleton as the product default and collaboration authority
- let independently released clients attach when their protocol version and required capabilities are compatible
- retain source revisions as diagnostic provenance rather than the compatibility boundary
- consider an instance namespace only as an explicit opt-in for tests or deployments that intentionally require isolation

The current implementation remains open as research while that narrower opt-in use case is
evaluated. Its shared file claim prevents two daemons from writing one file-backed notebook,
but that safety property is not equivalent to collaboration: the second daemon is refused
instead of joining the existing live document.

## Experimental implementation

- add an opt-in `RUNTIMED_INSTANCE_ID` namespace for desktop and Electron hosts that bundle an independently versioned daemon
- isolate the singleton, endpoint, settings, documents, execution records, blobs, trust state, kernel IPC, and mutable environment/tool roots while preserving the legacy stable/nightly paths when the variable is absent
- keep the cross-daemon file-claim registry shared so separate instances still refuse concurrent ownership of the same file-backed notebook
- expose deterministic `socket_path_for_instance` / `socketPathForInstance` helpers and additive daemon diagnostics

## Original motivation

A custom socket alone does not isolate a daemon. The channel-wide singleton and other mutable state still resolve through the normal stable or nightly root, so an embedding host cannot safely run a bundled revision beside the user's ordinary daemon.

This experiment gives a host one stable opaque instance id. The daemon hashes it into a short path-safe key, uses a whole-root namespace, and reports the live instance identity. Hosts can derive and probe the endpoint before spawning, set the id on the daemon child only, and pass the derived endpoint explicitly to every client session.

Materialized environments are deliberately instance-local because each daemon's garbage collector only knows about its own live kernels. Shared package-manager download caches remain available through their existing upstream mechanisms.

## Compatibility and security

- no instance id means the exact existing channel/worktree paths and endpoints
- the instance id is collision isolation within one OS user, not authentication
- endpoint and dev-worktree overrides should be scrubbed from a production child environment
- explicit CLI/programmatic path overrides still take precedence for the resources they configure
- the shared file-claim lease remains the cross-instance guard for file-backed notebooks

## Verification

- `cargo xtask lint --fix`
- `cargo clippy -p runt-workspace -p runtimed-client -p runtimed -p runtimed-node --all-targets -- -D warnings`
- `cargo test -p runt-workspace -p runtimed-client --lib`
- `cargo test -p runtimed --test instance_namespace -- --nocapture`
- `cargo test -p runtimed singleton`
- `cargo test -p runt-mcp daemon_watch`
- `pnpm --dir packages/runtimed-node build`
- `RUNTIMED_NODE_NATIVE_INTEGRATION=1 pnpm exec vp test run packages/runtimed-node/tests/arrow-ipc.test.ts`
- `cargo check -p runt-workspace --target x86_64-pc-windows-msvc`

Manual process smoke also kept the ordinary nightly daemon running while two isolated instances started with different endpoints, locks, settings, PIDs, and blob ports. A duplicate of the same instance id exited without disturbing the owner; a second instance was refused when it tried to open the same file-backed notebook; execution worked through the native Node client; and reopening after an instance restart recovered notebook source and the durable execution result.

### DIFF
--- a/crates/kernel-env/AGENTS.md
+++ b/crates/kernel-env/AGENTS.md
@@ -89,7 +89,7 @@ Every notebook's env is isolated per notebook via `env_id` — no cross-notebook
 | UV | `kernel_env::uv::compute_unified_env_hash` | `~/.cache/runt/envs/{hash}/` |
 | Conda | `kernel_env::conda::compute_unified_env_hash` | `~/.cache/runt/conda-envs/{hash}/` |
 
-Cache paths are channel-aware: stable → `runt/`, nightly → `runt-nightly/`, dev worktree → `runt/worktrees/{hash}/`.
+Cache paths are daemon-context-aware: stable → `runt/`, nightly → `runt-nightly/`, dev worktree → `runt/worktrees/{hash}/`, and a host-owned daemon instance → `<channel>/instances/{hash}/`. Materialized environments must not be shared across daemon instances because each daemon's GC only knows its own live kernels.
 
 Cache hit check: verify `{hash}/bin/python` (Unix) or `{hash}/Scripts/python.exe` (Windows).
 
--- a/crates/runt-mcp/src/daemon_watch.rs
+++ b/crates/runt-mcp/src/daemon_watch.rs
@@ -702,6 +702,8 @@ mod tests {
             execution_store_dir: None,
             worktree_path: None,
             workspace_description: None,
+            instance_id: None,
+            daemon_base_dir: None,
         }
     }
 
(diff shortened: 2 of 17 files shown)


rgbkrk:
Closing this exploration because the product direction remains one user-level singleton daemon with compatible clients joining the same live notebook.

PR #4145, opened 2026-08-11
feat(protocol): negotiate semantic capabilities

rgbkrk:
## Summary

- keep protocol v4 as the framing and serialization compatibility boundary
- add a versioned `features` map to `ProtocolCapabilities` for additive semantics within v4
- advertise caller-owned execution IDs, durable notebook-head confirmation, and exact execution cancellation at feature version 1
- carry negotiated features through `notebook-sync`, the native relay, and the high-level Node session
- reject feature-dependent Node operations before kernel launch, request delivery, or local document mutation when the daemon did not advertise the required version
- remove exact source-revision admission from the Node relay; artifact versions remain diagnostic metadata

## Contract

The handshake now distinguishes two independent compatibility axes:

- `protocol_version` governs framing and wire serialization
- `features: Record<string, number>` governs additive operation semantics

A missing feature is unsupported. Unknown features are ignored. A caller may use a feature only when the advertised integer is at least the version it requires.

The local runtimed endpoint currently advertises:

```json
{
  "caller_owned_execution_ids": 1,
  "confirm_notebook_heads": 1,
  "exact_execution_cancellation": 1
}
```

Pre-authenticated frame transports that do not forward handshake capabilities default to an empty feature set. This prevents a cloud or custom transport from accidentally inheriting local-daemon semantics.

## Node behavior

- `RelaySession.info.features` exposes the generic version map to embedding hosts
- `Session.protocolFeatures` exposes the same map to programmatic clients
- caller-supplied execution IDs are checked before lazy kernel launch
- exact cancellation is checked before its request is sent
- durable cell create/set/delete/move APIs are checked before applying a local Automerge mutation
- malformed JavaScript feature metadata is treated as unsupported

## Verification

- `cargo xtask lint --fix`
- `cargo test -p notebook-protocol -p notebook-sync -p runtimed-node`
- `cargo check -p runtimed`
- focused Node relay/session wrapper tests: 21 passed
- rebuilt `@runtimed/node` native addon
- live isolated dev-daemon relay smoke passed
- live handshake observed protocol 4 and all three feature flags at version 1

## Stack

- base: #4142

This remains draft while the programmatic Node authoring stack is evaluated against the shared JS/WASM client ownership boundary.

### DIFF
--- a/crates/notebook-protocol/src/connection.rs
+++ b/crates/notebook-protocol/src/connection.rs
@@ -20,7 +20,8 @@ pub use framing::{
 pub use handshake::{
     recv_typed_bootstrap_frame, send_typed_bootstrap_frame, ConnectionBootstrap,
     CreateNotebookRequest, Handshake, NotebookConnectionInfo, ProtocolCapabilities,
-    PutBlobCapability, PROTOCOL_V4, PROTOCOL_VERSION,
+    PutBlobCapability, FEATURE_CALLER_OWNED_EXECUTION_IDS, FEATURE_CONFIRM_NOTEBOOK_HEADS,
+    FEATURE_EXACT_EXECUTION_CANCELLATION, FEATURE_VERSION_1, PROTOCOL_V4, PROTOCOL_VERSION,
 };
 
 #[cfg(windows)]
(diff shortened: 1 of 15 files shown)


rgbkrk:
Closing this stacked capability experiment for now; semantic capability negotiation can return as a focused main-based change when a concrete optional protocol semantic requires it.

PR #4157, opened 2026-08-18
Layout updates

ecyee3:
Updates to layout to make panel slide out under the utility bar but keeping the sidebar in place.
image
image


rgbkrk:
Superseded by #4158, which is now merged. Erika’s panel-under-toolbar direction and follow-up commit `3997f7c1f` were carried into #4158 with the original authorship preserved, then refined into the shared desktop/cloud rail-drawer grid. Thank you, Erika, for establishing the layout direction we built on.

PR #4167, opened 2026-08-20
fix(mcp): preserve sessions across daemon replacement

rgbkrk:
## Summary

- make completed same-target `connect_notebook` calls truly idempotent so they keep the live/converging replica instead of installing another retained startup projection
- clear stale local sessions before rejoining after a same-version daemon process replacement, while preserving the durable target for bounded retries
- promote a created notebook's restart target from UUID to canonical path after save, and let the daemon reverse-resolve a saved UUID through `notebook-registry.sqlite`
- materialize the saved `.ipynb` when UUID recovery has only the durable path registry and no Automerge recovery journal
- report that a reconnect was requested instead of claiming it already succeeded

## Why this is patch-release work

The 2.6.2-to-2.7.0 audit found a version-specific regression in #4013 (`c805f3a6f`). That lifecycle change made any active MCP session authoritative during publication, including a stale same-UUID session whose `DocHandle` belonged to the replaced daemon. The fresh rejoin was therefore dropped. The same progressive-connect work also allowed a completed same-target call to start another generation, which could repeatedly replace a converging replica with the bounded initial projection.

The UUID-to-path handoff is an older latent gap rather than a 2.7 regression: an untitled notebook keeps its UUID when saved, but the supervisor previously retained only that daemon-scoped UUID. This patch closes both sides of that recovery contract without changing notebook identity.

Independent review found one additional data-loss risk in that recovery path: reverse registry lookup could recover the correct saved path but construct a fresh notebook whenever the auxiliary Automerge journal was missing. UUID recovery now schedules the same `.ipynb` import used by open-by-path before the session becomes ready.

The submitted diagnostic archive is consistent with this chain: the upgrade child received a UUID the replacement daemon could not find; a same-version restart later logged a fresh same-UUID rejoin being dropped as superseded; and repeated successful connects exposed a frozen projection timestamp and one-cell preview while the file contained 19 populated cells. This is strong causal evidence, though the archive cannot prove which link Anil encountered first.

## Verification

- `cargo test -p runt-mcp-proxy` (108 passed)
- `cargo test -p runt-mcp` (212 passed)
- `cargo test -p runtimed notebook_registry` (9 passed)
- `cargo test -p runtimed --test integration test_notebook_sync_uuid_ -- --nocapture` (2 passed, including registry-only recovery without a journal)
- `cargo xtask lint --fix` (passed)


rgbkrk:
Superseded by #4169, which incorporates the broader daemon-incarnation and durable notebook-identity recovery design plus the completed Fable/Opus review findings. #4169 is ready for review and CI is green.

